PR #15, opened 2018-11-26
Initial selenium test module commit

Zejnilovic:
Adds selenium-test to modules in the parent pom
Adds selenium-test module with models and some smoke tests


selenium-tests/src/test/java/za/co/absa/SchemaTest.java
```diff
@@ -0,0 +1,95 @@
...
+    By newSchemaName = By.id("newSchemaName-inner");
+    By newSchemaDescription = By.id("newSchemaDescription-inner");
+    By newSchemaAddButton = By.id("newSchemaAddButton");
+    By newSchemaCancelButton = By.id("newSchemaCancelButton");
+
+    chooseTab(schemasTab);
+    hoverClick(addSchemaButton);
+    TimeUnit.SECONDS.sleep(1);
+    WebElement eName = driver.findElement(newSchemaName);
+    WebElement eDescription = driver.findElement(newSchemaDescription);
```
GeorgiChochov: I'd go for `elementName` and `elementDescription` to be more descriptive.

selenium-tests/src/test/java/za/co/absa/BaseTest.java
```diff
@@ -0,0 +1,192 @@
...
+  private DateTimeFormatter dtf = DateTimeFormatter.ofPattern("yyyy_MM_dd-HH_mm_ss");
+  String postfix = dtf.format(LocalDateTime.now());
+
+  By schemasView = By.id("__navigation0---schemaMainView");
+  By datasetsView = By.id("__navigation0---datasetMainView");
+  By mappingTablesView = By.id("__navigation0---mappingTableMainView");
+  private By nameField = By.name("username");
+  private By passwordField = By.name("password");
+  String username = "user";
+  private String password = "changeme";
```
GeorgiChochov: 1. Generally not a good style to have fields with different visibility mixed. First public, then protected, then default, finally private.
2. These seem like constants that duplicate the ones in `Base`. I'd extract them in a Constants class and make them all `public static final`.

GeorgiChochov:
One more thing, add the License to the top of each file. (Intellij has an option to do this automatically for you)

selenium/src/test/java/za/co/absa/enceladus/selenium/AppTest.java
```diff
@@ -0,0 +1,91 @@
+/*
+ * <!--
+ *   ~ Copyright 2018 ABSA Group Limited
```
GeorgiChochov: The Copyright should only have text in it, not the xml formatted comment prefix per line. (Your IntelliJ settings might be a bit off)

GeorgiChochov:
The problem is that to have these tests run, you need a properly set up and running environment, with data fixtures in place. Just running `mvn test` fails, which means if we have the selenium tests as a module of Enceladus, a normal build of the project always fails.
@Zejnilovic wants to separate it out to its own repo, rather than have the set up automated within the tests, as that could really slow down local build processes and overload the functionality of the selenium module, which should just run tests.

benedeki:
> 
> 
> The problem is that to have these tests run, you need a properly set up and running environment, with data fixtures in place. Just running `mvn test` fails, which means if we have the selenium tests as a module of Enceladus, a normal build of the project always fails.
> @Zejnilovic wants to separate it out to its own repo, rather than have the set up automated within the tests, as that could really slow down local build processes and overload the functionality of the selenium module, which should just run tests.

Having env built scripts in a different repo doesn't make much sense IMO, at least I don't see the benefits.
Also correct me, but doesn't most more elaborate tests require env setup anyway? So it's not just for the selenium ones...

Zejnilovic:
> > The problem is that to have these tests run, you need a properly set up and running environment, with data fixtures in place. Just running `mvn test` fails, which means if we have the selenium tests as a module of Enceladus, a normal build of the project always fails.
> > @Zejnilovic wants to separate it out to its own repo, rather than have the set up automated within the tests, as that could really slow down local build processes and overload the functionality of the selenium module, which should just run tests.
> 
> Having env built scripts in a different repo doesn't make much sense IMO, at least I don't see the benefits.
> Also correct me, but doesn't most more elaborate tests require env setup anyway? So it's not just for the selenium ones...

For having another repo:
- CI pipeline separation - more readable and stable pipeline with separate jobs/parts
- env setup scripts are not always Java (in most case they aren't) and the repo is a maven repo through and through. How structured/readable will it be when we start getting shell, mongo, etc. scripts into it or want to run those scripts.
- running regression tests - we have app version 0.2.0 and test 0.2.0 - but we might want to see if the previous tests are still working or we want to test previous app version but keep test versions to see when was the bug introduced into the code.

GeorgiChochov:
Let me clear this up, and put it bed, we are not getting a separate repo for this for the time being.
* We can have a buildable project regardless of the selenium module if we separate its tests from the ones run with a regular build using one of the following:
  * different maven profiles 
  * running them as integration tests with maven's failsafe plugin (which needs to be explicitly called), rather than running them as unit tests with maven's surefire plugin
* Either of these solutions gives you the freedom to separate out the steps of the CI pipeline
* If we need to do regression testing with newer tests, we can just test against previous versions of the jar, we don't need to rebuild the whole project.

pom.xml
```diff
@@ -32,18 +32,31 @@
...
-		<module>standardization</module>
-		<module>dataModel</module>
-		<module>dao</module>
-		<module>menas</module>
-		<module>migrations</module>
-		<module>external/kafka</module>
-		<module>external/fixedWidth</module>
-		<module>selenium</module>
-	</modules>
+	<profiles>
```
GeorgiChochov: Let's leave the modules as they were without `selenium` and have a profile which specifies the selenium module.
The profiles usually go at the end of the pom.xml.

PR #16, opened 2018-11-26
Add basic build jenkinsfile

Zejnilovic:
Creates a basic Jenkins file with build job on it.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,14 @@
+pipeline {
+     agent {
+         docker {
+             image 'maven:3-alpine'
+         }
+     }
+     stages {
+         stage('Build') {
+             steps {
+                 sh 'mvn -B -DskipTests clean install'
+             }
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -0,0 +1,14 @@
+pipeline {
+     agent {
+         docker {
+             image 'maven:3-alpine'
+         }
+     }
+     stages {
+         stage('Build') {
+             steps {
+                 sh 'mvn -B -DskipTests clean install'
```
GeorgiChochov: Can you just drop a link to why we need `-B` here, for documentation?

Zejnilovic: The -B is a batch-mode. The batch-mode will automatically use default values instead of asking you via prompt for those values, which we don't want to happen in a Jenkins build container. The batch-mode can also be activated via --batch-mode or -B on the command line.

Issue #68, opened 2018-12-07
Set default Timezone for sample data suite


VinothBigEng:
Currently the way we deal with timestamps in test case samples is by multiplying value with long which  is initialized with UNIX time which is zoneless since it denotes a period of time passed after epoch.

PR #73, opened 2018-12-11
#68 Timezone bug fix for TestSamples

VinothBigEng:
when a time stamp is initialized with Long it is initialized with UNIX time which is zoneless since it denotes a period of time passed after epoch so removed multiplication with long and added long value to convert to timestamp, so the test case can be executed locally

### DIFF
--- a/standardization/src/test/scala/za/co/absa/enceladus/standardization/samples/TestSamples.scala
+++ b/standardization/src/test/scala/za/co/absa/enceladus/standardization/samples/TestSamples.scala
@@ -73,11 +73,11 @@ object TestSamples {
   val startDate = 1438387200000l //01/08/2015
 
   val resData = List(
-    new StdEmployee(name = "John0", surname = "Unknown Surname", hoursWorked = Some(List(8, 7, 8, 9, 12, 0)),
+     StdEmployee(name = "John0", surname = "Unknown Surname", hoursWorked = Some(List(8, 7, 8, 9, 12, 0)),
         employeeNumbers = List(EmployeeNumberStd("SAP", List(456, 123)), EmployeeNumberStd("WD", List(5))), startDate =  new java.sql.Date(startDate), errCol = List(ErrorMessage.stdNullErr("surname"), ErrorMessage.stdNullErr("hoursWorked[*]"))),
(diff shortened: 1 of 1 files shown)


standardization/src/test/scala/za/co/absa/enceladus/standardization/samples/TestSamples.scala
```diff
@@ -74,13 +76,13 @@ object TestSamples {
...
     new StdEmployee(name = "John1", surname = "Doe1", hoursWorked = Some(List(99, 99, 76, 12, 12, 24)), startDate = new java.sql.Date(0), errCol = List(ErrorMessage.stdCastErr("startDate", "Two Thousand Something"))),
-    new StdEmployee(name = "John2", surname = "Unknown Surname", hoursWorked = None, startDate = new java.sql.Date(startDate), updated = Some(new java.sql.Timestamp(1437053544l *1000l)), errCol = List(ErrorMessage.stdNullErr("surname"), ErrorMessage.stdNullErr("hoursWorked"))),
-    new StdEmployee(name = "John3", surname = "Unknown Surname", hoursWorked = Some(List()), startDate = new java.sql.Date(startDate), updated = Some(new java.sql.Timestamp(1437042744l * 1000l)), errCol = List(ErrorMessage.stdNullErr("surname"))))
+    new StdEmployee(name = "John2", surname = "Unknown Surname", hoursWorked = None, startDate = new java.sql.Date(startDate), updated = Some(new java.sql.Timestamp(1437053544000l)), errCol = List(ErrorMessage.stdNullErr("surname"), ErrorMessage.stdNullErr("hoursWorked"))),
```
benedeki: I see you replaced one magical constant for two. Why not to use _Timestamp.valueOf()_ instead (https://docs.oracle.com/javase/6/docs/api/java/sql/Timestamp.html#valueOf(java.lang.String))?

benedeki: Remove _new_ .
**StdEmployee** is a case class

VinothBigEng: yes, it make sense.

VinothBigEng: yes, removed

Issue #40, opened 2018-12-02
Entity names shouldn't have whitespace


GeorgiChochov:
Dataset/Mapping table/Schema names should not be allowed to have whitespace as that can cause issues

PR #87, opened 2019-01-02
Feature/#40 entity names whitespace check

VinothBigEng:
Commit  for #40 entity names whitespace check after review

### DIFF
--- a/menas/src/main/scala/za/co/absa/enceladus/rest/services/DatasetService.scala
+++ b/menas/src/main/scala/za/co/absa/enceladus/rest/services/DatasetService.scala
@@ -42,8 +42,11 @@ class DatasetService @Autowired()(datasetMongoRepository: DatasetMongoRepository
   }
 
   override def create(newDataset: Dataset, username: String): Future[Dataset] = {
+    println("i m here")
     val dataset = Dataset(
-      name = newDataset.name,
+      name = { if (newDataset.name.indexOf(" ") == -1 ) newDataset.name
+      else throw new Exception(s"Dataset name must not contain whitespace : ${newDataset.name}")
+      },
       version = 0,
       description = newDataset.description,
       hdfsPath = newDataset.hdfsPath,
--- a/menas/src/main/scala/za/co/absa/enceladus/rest/services/DatasetService.scala
+++ b/menas/src/main/scala/za/co/absa/enceladus/rest/services/DatasetService.scala
@@ -20,7 +20,9 @@ import org.springframework.stereotype.Service
 import za.co.absa.enceladus.model.{Dataset, UsedIn}
 import za.co.absa.enceladus.rest.repositories.DatasetMongoRepository
 
-import scala.concurrent.Future
+
+import scala.concurrent.duration.Duration
+import scala.concurrent.{Await, Future}
 
 @Service
 class DatasetService @Autowired()(datasetMongoRepository: DatasetMongoRepository)
@@ -42,11 +44,8 @@ class DatasetService @Autowired()(datasetMongoRepository: DatasetMongoRepository
   }
 
   override def create(newDataset: Dataset, username: String): Future[Dataset] = {
-    println("i m here")
-    val dataset = Dataset(
-      name = { if (newDataset.name.indexOf(" ") == -1 ) newDataset.name
-      else throw new Exception(s"Dataset name must not contain whitespace : ${newDataset.name}")
-      },
+      val dataset = Dataset(
+      name = validateEntityName(newDataset.name,"Dataset"),
       version = 0,
       description = newDataset.description,
       hdfsPath = newDataset.hdfsPath,
--- a/menas/src/main/scala/za/co/absa/enceladus/rest/services/DatasetService.scala
+++ b/menas/src/main/scala/za/co/absa/enceladus/rest/services/DatasetService.scala
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018 ABSA Group Limited
+ * Copyright 2018-2019 ABSA Group Limited
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -19,10 +19,7 @@ import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.stereotype.Service
 import za.co.absa.enceladus.model.{Dataset, UsedIn}
 import za.co.absa.enceladus.rest.repositories.DatasetMongoRepository
-
-
-import scala.concurrent.duration.Duration
-import scala.concurrent.{Await, Future}
+import scala.concurrent.Future
 
 @Service
 class DatasetService @Autowired()(datasetMongoRepository: DatasetMongoRepository)
@@ -44,8 +41,10 @@ class DatasetService @Autowired()(datasetMongoRepository: DatasetMongoRepository
   }
 
   override def create(newDataset: Dataset, username: String): Future[Dataset] = {
-      val dataset = Dataset(
-      name = validateEntityName(newDataset.name,"Dataset"),
+    validateEntityName(newDataset.name,"Dataset")
+
+    val dataset = Dataset(
+      name = newDataset.name,
       version = 0,
       description = newDataset.description,
       hdfsPath = newDataset.hdfsPath,
--- a/menas/src/main/scala/za/co/absa/enceladus/rest/services/MappingTableService.scala
+++ b/menas/src/main/scala/za/co/absa/enceladus/rest/services/MappingTableService.scala
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018 ABSA Group Limited
+ * Copyright 2018-2019 ABSA Group Limited
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -36,8 +36,10 @@ class MappingTableService @Autowired()(mappingTableMongoRepository: MappingTable
   }
 
   override def create(mt: MappingTable, username: String): Future[MappingTable] = {
+    validateEntityName(mt.name, "MappingTable")
+
     val mappingTable = MappingTable(
-      name = validateEntityName(mt.name ,"MappingTable"),
+      name = mt.name,
       description = mt.description,
       schemaName = mt.schemaName,
       schemaVersion = mt.schemaVersion,
--- a/menas/src/main/scala/za/co/absa/enceladus/rest/services/MappingTableService.scala
+++ b/menas/src/main/scala/za/co/absa/enceladus/rest/services/MappingTableService.scala
@@ -36,9 +36,13 @@ class MappingTableService @Autowired()(mappingTableMongoRepository: MappingTable
   }
 
   override def create(mt: MappingTable, username: String): Future[MappingTable] = {
-    val mappingTable = MappingTable(name = mt.name, description = mt.description, schemaName = mt.schemaName,
-      schemaVersion = mt.schemaVersion, hdfsPath = mt.hdfsPath)
-
+    val mappingTable = MappingTable(
+      name = validateEntityName(mt.name ,"MappingTable"),
+      description = mt.description,
+      schemaName = mt.schemaName,
+      schemaVersion = mt.schemaVersion,
+      hdfsPath = mt.hdfsPath
+    )
     super.create(mappingTable, username)
   }
 
--- a/menas/src/main/scala/za/co/absa/enceladus/rest/services/VersionedModelService.scala
+++ b/menas/src/main/scala/za/co/absa/enceladus/rest/services/VersionedModelService.scala
@@ -18,9 +18,9 @@ package za.co.absa.enceladus.rest.services
 import org.slf4j.LoggerFactory
 import org.springframework.security.core.context.SecurityContextHolder
 import org.springframework.security.core.userdetails.UserDetails
-import za.co.absa.enceladus.model.UsedIn
 import za.co.absa.enceladus.model.versionedModel.{VersionedModel, VersionedSummary}
 import za.co.absa.enceladus.rest.repositories.VersionedMongoRepository
+import za.co.absa.enceladus.model.UsedIn
 
 import scala.concurrent.Future
 
@@ -84,6 +84,13 @@ abstract class VersionedModelService[C <: VersionedModel](versionedMongoReposito
     } else {
       versionedMongoRepository.disableVersion(name, version, principal.getUsername)
     }
+  }
+
+  def validateEntityName(name: String , entityType:String ):String  = {
+    if (name.matches("""\w+""")) {
+      name
+    }
+     else throw new Exception(s"${entityType} name must not contain whitespace : ${name}")
   }
 
 }
--- a/menas/src/main/scala/za/co/absa/enceladus/rest/services/VersionedModelService.scala
+++ b/menas/src/main/scala/za/co/absa/enceladus/rest/services/VersionedModelService.scala
@@ -88,7 +88,7 @@ abstract class VersionedModelService[C <: VersionedModel](versionedMongoReposito
 
   def validateEntityName(name: String , entityType:String ): Unit  = {
     if (name.matches("""\w+""")) name
-      else throw new Exception(s"${entityType} name must not contain whitespace : ${name}")
+      else throw new Exception(s"$entityType name must not contain whitespace : $name")
   }
 
 }
--- a/menas/src/main/scala/za/co/absa/enceladus/rest/services/VersionedModelService.scala
+++ b/menas/src/main/scala/za/co/absa/enceladus/rest/services/VersionedModelService.scala
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018 ABSA Group Limited
+ * Copyright 2018-2019 ABSA Group Limited
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -86,11 +86,9 @@ abstract class VersionedModelService[C <: VersionedModel](versionedMongoReposito
     }
   }
 
-  def validateEntityName(name: String , entityType:String ):String  = {
-    if (name.matches("""\w+""")) {
-      name
-    }
-     else throw new Exception(s"${entityType} name must not contain whitespace : ${name}")
+  def validateEntityName(name: String , entityType:String ): Unit  = {
+    if (name.matches("""\w+""")) name
+      else throw new Exception(s"${entityType} name must not contain whitespace : ${name}")
   }
 
 }
--- a/menas/src/main/scala/za/co/absa/enceladus/rest/services/VersionedModelService.scala
+++ b/menas/src/main/scala/za/co/absa/enceladus/rest/services/VersionedModelService.scala
@@ -86,7 +86,7 @@ abstract class VersionedModelService[C <: VersionedModel](versionedMongoReposito
     }
   }
 
-  def validateEntityName(name: String , entityType:String ): Unit  = {
+  private[services] def validateEntityName(name: String , entityType:String ): Unit  = {
     if (name.matches("""\w+""")) name
       else throw new Exception(s"$entityType name must not contain whitespace : $name")
   }
--- a/menas/src/main/scala/za/co/absa/enceladus/rest/services/VersionedModelService.scala
+++ b/menas/src/main/scala/za/co/absa/enceladus/rest/services/VersionedModelService.scala
@@ -86,7 +86,7 @@ abstract class VersionedModelService[C <: VersionedModel](versionedMongoReposito
     }
   }
 
-  private[services] def validateEntityName(name: String , entityType:String ): Unit  = {
+  protected[services] def validateEntityName(name: String , entityType:String ): Unit  = {
     if (name.matches("""\w+""")) name
       else throw new Exception(s"$entityType name must not contain whitespace : $name")
   }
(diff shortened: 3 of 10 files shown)


menas/src/main/scala/za/co/absa/enceladus/rest/services/DatasetService.scala
```diff
@@ -42,8 +44,8 @@ class DatasetService @Autowired()(datasetMongoRepository: DatasetMongoRepository
   }
 
   override def create(newDataset: Dataset, username: String): Future[Dataset] = {
-    val dataset = Dataset(
-      name = newDataset.name,
+      val dataset = Dataset(
+      name = validateEntityName(newDataset.name,"Dataset"),
```
GeorgiChochov: How about we keep the the validation call outside of the entity's definition? The validation function returns its input value or throws, we don't need to clutter the entity declaration with this.

menas/src/main/scala/za/co/absa/enceladus/rest/services/VersionedModelService.scala
```diff
@@ -86,4 +86,11 @@ abstract class VersionedModelService[C <: VersionedModel](versionedMongoReposito
     }
   }
 
+  def validateEntityName(name: String , entityType:String ):String  = {
```
GeorgiChochov:
```suggestion
  def validateEntityName(name: String, entityType: String): String = {
```
Minor formatting issue.

menas/src/main/scala/za/co/absa/enceladus/rest/services/DatasetService.scala
```diff
@@ -20,7 +20,9 @@ import org.springframework.stereotype.Service
 import za.co.absa.enceladus.model.{Dataset, UsedIn}
 import za.co.absa.enceladus.rest.repositories.DatasetMongoRepository
 
-import scala.concurrent.Future
+
+import scala.concurrent.duration.Duration
+import scala.concurrent.{Await, Future}
```
lokm01: Why Await and Duration? We want stuff to be async

menas/src/main/scala/za/co/absa/enceladus/rest/services/VersionedModelService.scala
```diff
@@ -18,9 +18,9 @@ package za.co.absa.enceladus.rest.services
 import org.slf4j.LoggerFactory
 import org.springframework.security.core.context.SecurityContextHolder
 import org.springframework.security.core.userdetails.UserDetails
-import za.co.absa.enceladus.model.UsedIn
 import za.co.absa.enceladus.model.versionedModel.{VersionedModel, VersionedSummary}
 import za.co.absa.enceladus.rest.repositories.VersionedMongoRepository
+import za.co.absa.enceladus.model.UsedIn
```
lokm01: Is this import used?

menas/ui/components/dataset/datasetMain.controller.js
```diff
@@ -145,6 +145,12 @@ sap.ui.controller("components.dataset.datasetMain", {
                 "Dataset name '" + oDataset.name + "' already exists. Choose a different name.");
             isOk = false;
         }
+        if (/\W/.test(oDataset.name)) {
```
lokm01: Maybe we should create a "generic" name validation function... Same will apply for other controllers too

menas/ui/components/schema/schemaMain.controller.js
```diff
@@ -121,7 +121,14 @@ sap.ui.controller("components.schema.schemaMain", {
 			sap.ui.getCore().byId("newSchemaName").setValueStateText(
 					"Schema name '" + schema.name + "' already exists. Choose a different name.")
 			return false;
-		} else
+		}
+		if (/\W/.test(schema.name)) {
```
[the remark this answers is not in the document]
GeorgiChochov: Yes, it should all be just a series of ifs.

GeorgiChochov:
Also, please add a unit test for the validation check.

menas/src/main/scala/za/co/absa/enceladus/rest/services/DatasetService.scala
```diff
@@ -20,7 +20,9 @@ import org.springframework.stereotype.Service
 import za.co.absa.enceladus.model.{Dataset, UsedIn}
 import za.co.absa.enceladus.rest.repositories.DatasetMongoRepository
 
-import scala.concurrent.Future
+
```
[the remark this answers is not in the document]
VinothBigEng: Resolved

menas/src/main/scala/za/co/absa/enceladus/rest/services/DatasetService.scala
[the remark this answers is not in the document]
VinothBigEng: Resolved

menas/src/main/scala/za/co/absa/enceladus/rest/services/DatasetService.scala
VinothBigEng: Resolved

menas/src/main/scala/za/co/absa/enceladus/rest/services/VersionedModelService.scala
VinothBigEng: Yes, it is used

menas/src/main/scala/za/co/absa/enceladus/rest/services/DatasetService.scala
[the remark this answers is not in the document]
VinothBigEng: Resolved

menas/src/main/scala/za/co/absa/enceladus/rest/services/DatasetService.scala
```diff
@@ -42,8 +44,8 @@ class DatasetService @Autowired()(datasetMongoRepository: DatasetMongoRepository
   }
 
   override def create(newDataset: Dataset, username: String): Future[Dataset] = {
-    val dataset = Dataset(
-      name = newDataset.name,
+      val dataset = Dataset(
```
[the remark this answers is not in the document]
VinothBigEng: Resolved

menas/src/main/scala/za/co/absa/enceladus/rest/services/MappingTableService.scala
```diff
@@ -36,9 +36,13 @@ class MappingTableService @Autowired()(mappingTableMongoRepository: MappingTable
   }
 
   override def create(mt: MappingTable, username: String): Future[MappingTable] = {
-    val mappingTable = MappingTable(name = mt.name, description = mt.description, schemaName = mt.schemaName,
-      schemaVersion = mt.schemaVersion, hdfsPath = mt.hdfsPath)
-
+    val mappingTable = MappingTable(
+      name = validateEntityName(mt.name ,"MappingTable"),
```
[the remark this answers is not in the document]
VinothBigEng: Resolved

menas/src/main/scala/za/co/absa/enceladus/rest/services/VersionedModelService.scala
VinothBigEng: Made the changes.

menas/src/main/scala/za/co/absa/enceladus/rest/services/DatasetService.scala
VinothBigEng: thanks for the explanation, made the changes

menas/ui/components/schema/schemaMain.controller.js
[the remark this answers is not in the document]
benedeki: Definitely should be uniform aproach within the function (all `if`s or '`else if`s. Ideally I would prefer to have it similar to `components/mappingTable/mappingTableMain.controller.js->validateNewMappingTable' just one `return` at the end. (I know it's requesting change on old code, but applying _good boy scout_ principle 😉 )

menas/src/main/scala/za/co/absa/enceladus/rest/services/VersionedModelService.scala
benedeki: Make `protected`

menas/src/main/scala/za/co/absa/enceladus/rest/services/VersionedModelService.scala
```diff
@@ -86,4 +86,11 @@ abstract class VersionedModelService[C <: VersionedModel](versionedMongoReposito
     }
   }
 
+  def validateEntityName(name: String , entityType:String ):String  = {
+    if (name.matches("""\w+""")) {
```
benedeki: My suggestion: include check for uniqueness from `create` method within this function, call `validateEntityName` within `create` and remove the call from higher instances.

menas/ui/components/dataset/datasetMain.controller.js
benedeki: Let's create the _generic_ function only when needed and **if** the validation is the same 🤔

menas/src/main/scala/za/co/absa/enceladus/rest/services/VersionedModelService.scala
GeorgiChochov: It's just an import moved 2 lines down.

menas/ui/components/dataset/datasetMain.controller.js
GeorgiChochov: The point here is that, they are the same validations (checking the name of the entity, dataset/schema/mapping table), rather than repeat the same function (with the odd JS regex syntax) in three places, we can reduce code duplication.

menas/src/main/scala/za/co/absa/enceladus/rest/services/VersionedModelService.scala
```diff
@@ -86,4 +86,11 @@ abstract class VersionedModelService[C <: VersionedModel](versionedMongoReposito
...
+      name
+    }
+     else throw new Exception(s"${entityType} name must not contain whitespace : ${name}")
```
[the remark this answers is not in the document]
VinothBigEng: resolved

menas/ui/components/mappingTable/mappingTableMain.controller.js
```diff
@@ -321,6 +321,11 @@ sap.ui.controller("components.mappingTable.mappingTableMain", {
       sap.ui.getCore().byId("newMappingTableName").setValueStateText("Mapping Table name '" + oMT.name + "' already exists. Choose a different name.")
       isOk = false;
     }
+    if (/\W/.test(oMT.name)){
```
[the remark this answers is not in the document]
GeorgiChochov: Let's use all `if`s to find and highlight all validation errors in one go. If we return on the first validation error (or have `else`s), only it will be displayed.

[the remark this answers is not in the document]
benedeki: That's preferable for sure, but would require little more changes otherwise the later `if` overwrites the message of the earlier ones (`setValueStateText` on the same component)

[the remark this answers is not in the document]
GeorgiChochov: Right, should clarify (didn't peer into this too close): `if` per field with a series of `else`s for each validation on that field.

menas/ui/components/schema/schemaMain.controller.js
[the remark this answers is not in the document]
VinothBigEng: Resolved.

menas/ui/components/dataset/datasetMain.controller.js
VinothBigEng: resolved.

GeorgiChochov:
Please update the PR according to the new copyright template #96

menas/src/test/scala/za/co/absa/enceladus/service/DatasetServiceTest.scala
```diff
@@ -0,0 +1,39 @@
+/*
+ * Copyright 2018-${today.year} ABSA Group Limited
```
GeorgiChochov: I guess the intelligence copyright template didn't quite work out here

GeorgiChochov:
Please resolve the merge conflicts

menas/ui/components/schema/schemaMain.controller.js
```diff
@@ -111,18 +111,26 @@ sap.ui.controller("components.schema.schemaMain", {
...
 		var schema = this._model.getProperty("/newSchema")
+    var isOk = true
 
 		if (!schema.name || schema.name === "") {
 			sap.ui.getCore().byId("newSchemaName").setValueState(sap.ui.core.ValueState.Error)
 			sap.ui.getCore().byId("newSchemaName").setValueStateText("Schema name cannot be empty")
-			return false;
-		} else if (!schema.nameUnique) {
+      isOk = false;
+		} else { if (!schema.nameUnique) {
```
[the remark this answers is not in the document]
VinothBigEng: Corrected and resolved

menas/src/test/scala/za/co/absa/enceladus/service/DatasetServiceTest.scala
VinothBigEng: Resolved

menas/ui/service/GenericService.js
```diff
@@ -104,4 +104,10 @@ var GenericService = new function() {
 
         fnHelper("", sModelAcc, paths)
     };
-}();
\ No newline at end of file
+
+    this.isValidEntityName = function(sName) {
```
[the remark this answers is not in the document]
VinothBigEng: I agree,resolved

menas/ui/components/schema/schemaMain.controller.js
```diff
@@ -111,18 +111,26 @@ sap.ui.controller("components.schema.schemaMain", {
...
 					"Schema name '" + schema.name + "' already exists. Choose a different name.")
-			return false;
-		} else
-			return true;
+      isOk = false;
+		}
+    if (GenericService.isValidEntityName(schema.name)) {
+      sap.ui.getCore().byId("newSchemaName").setValueState(sap.ui.core.ValueState.Error)
+      sap.ui.getCore().byId("newSchemaName").setValueStateText(
+          "Schema name '" + schema.name  + "' should not have spaces. Please remove spaces and retry")
```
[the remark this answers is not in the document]
VinothBigEng: resolved

menas/ui/service/GenericService.js
```diff
@@ -104,4 +104,11 @@ var GenericService = new function() {
 
         fnHelper("", sModelAcc, paths)
     };
+
+    this.validEntityName = function(sName) {
```
Zejnilovic: `validateEntityName` not `validEntityName`

VinothBigEng: My bad :-(

menas/src/main/scala/za/co/absa/enceladus/rest/services/VersionedModelService.scala
benedeki: Make `protected`

menas/src/main/scala/za/co/absa/enceladus/rest/services/VersionedModelService.scala
benedeki: Please, if you don't agree with a comment don't just mark it as _Resolved_. I would welcome to learn why you think the suggested idea wouldn't work.

menas/src/main/scala/za/co/absa/enceladus/rest/services/VersionedModelService.scala
GeorgiChochov: `protected` isn't a thing in scala, I think oyu mean package-private, like `private[services]`

GeorgiChochov: My bad, I had completely forgotten about `protected` being legit in scala, haven't used it in ages. You're right, it fits well here.

menas/src/main/scala/za/co/absa/enceladus/rest/services/VersionedModelService.scala
VinothBigEng: Sorry, I meant to resolve some other conversation, I accidentally resolved this.

VinothBigEng: > My suggestion: include check for uniqueness from `create` method within this function, call `validateEntityName` within `create` and remove the call from higher instances.

Please could you elaborate, couldnt understand.

benedeki: > 
> 
> Sorry, I meant to resolve some other conversation, I accidentally resolved this.

NP 😃

menas/ui/service/GenericService.js
VinothBigEng: resolved

menas/src/main/scala/za/co/absa/enceladus/rest/services/VersionedModelService.scala
VinothBigEng: then it becomes Inaccessible for DatasetsServiceTest, Unless and until I extend the versionmodelservice. Please suggest.

menas/src/main/scala/za/co/absa/enceladus/rest/services/VersionedModelService.scala
benedeki: > 
> 
> > My suggestion: include check for uniqueness from `create` method within this function, call `validateEntityName` within `create` and remove the call from higher instances.
> 
> Please could you elaborate, couldnt understand.

```
  def create(item: C, username: String): Future[C] = {
   validateEntityName(item.name)
   versionedMongoRepository.create(item, username)
   getLatestVersion(item.name)
  }

  def validateEntityName(name: String ):Boolean  = {
    if (!isUniqueName(name)) {
      throw new Exception(s"Name already exists: $name}")
    }
    if (!name.matches("""\w+""")) {
      throw new Exception(s"Entity name must not contain whitespace : $name")
    }
    true
  }
```

Zejnilovic: We don't do it right now. That would be another/new issue.
The only place that these are checked is in JS.
But I agree that this should be implemented.

menas/src/main/scala/za/co/absa/enceladus/rest/services/VersionedModelService.scala
benedeki: In test you can create a foo class inheriting the base class to make the function accessible.

GeorgiChochov: You don't need to extend anything. Just put the test class in the same package.

GeorgiChochov: Tests and their respective classes need to have matching packages in `src/` and `test/`

menas/src/main/scala/za/co/absa/enceladus/rest/services/VersionedModelService.scala
VinothBigEng: @Zejnilovic Ok, can i go ahead and create a new issue and resolve this conversation?

Zejnilovic: If you ask me, then yes. What has the creator @benedeki to say?
And reference the issue here.

menas/src/main/scala/za/co/absa/enceladus/rest/services/VersionedModelService.scala
GeorgiChochov: Make it `private[services]`, `protected` in scala works differently than it does in java. In Java, protected methods can be accessed by other classes in the same package or subclasses in any pacakage, but this isn’t true in Scala, it's only accessible from subclasses.

benedeki: Unfortunately `private` and `protected` restricts the member within the class not the package. That leaves three options:
1) Once I suggested (and prefer) as the least evil making only the test less nice. `protected` + foo inheritance in test
2) Leave the method `public`. Ok for test, but opens the internals of the class "to the world"
3) Make it `protected` but assign accessibility to certain class for the sake of the test. IMO worst option, because it pollutes the production code with something not needed there.

VinothBigEng: > Make it `private[services]`, `protected` in scala works differently than it does in java. In Java, protected methods can be accessed by other classes in the same package or subclasses in any pacakage, but this isn’t true in Scala, it's only accessible from subclasses.

I perfer @GeorgiChochov suggestion, as it is clean

menas/src/main/scala/za/co/absa/enceladus/rest/services/VersionedModelService.scala
benedeki: @Zejnilovic was not aware that the uniqueness check was already there and that I only suggested unification of the two name checks (we clarified it in conversation as I didn't understand his comment)
I don't want to stall this PR on this, this is why I made it a suggestion only, not request. I believe it make the logic less scattered and that it's not a new issue.

VinothBigEng: Thanks Everyone, resolving this conversation

menas/src/main/scala/za/co/absa/enceladus/rest/services/VersionedModelService.scala
benedeki: FYI asked my Scala mentor (without telling him my preference) and from the three options he chose the same one as me 😉 But of course it can be a bias caused by his mentoring 😄

VinothBigEng: Thanks David. Please approve the PR.

VinothBigEng: resolving the conversation

PR #91, opened 2019-01-03
Add appropriate HTTP response codes

GeorgiChochov:
First step to making things a bit more properly RESTful. Not perfect, but let's get going.

### DIFF
--- a/menas/src/main/scala/za/co/absa/enceladus/rest/controllers/HDFSController.scala
+++ b/menas/src/main/scala/za/co/absa/enceladus/rest/controllers/HDFSController.scala
@@ -36,32 +36,36 @@ class HDFSController @Autowired() (fs: FileSystem) {
   private val logger = LoggerFactory.getLogger(this.getClass)
   
   @PostMapping(path = Array("/list"))
-  def getHDFSFolder(@RequestBody path: String): CompletableFuture[ResponseEntity[HDFSFolder]] = {
-    
-    val p = new Path(path)
-    if (fs.exists(p)) {
-      val res = if (!fs.isDirectory(p)) {
-        Future.successful(HDFSFolder(p.toUri().getPath, p.getName, None))
-      } else {
-        Future {
-          val status = fs.listStatus(p)
-          val children = if (status.isEmpty) None else {
-            Some(
-              status.map({ x =>
-                val child = x.getPath
+  def getHDFSFolder(@RequestBody pathStr: String): CompletableFuture[ResponseEntity[HDFSFolder]] = {
+    val path = new Path(pathStr)
 
-                HDFSFolder(child.toUri().getPath, child.getName,
-                  if (fs.listStatus(child).isEmpty) None else Some(Seq(HDFSFolder("", "", None))))
-
-              }).toSeq)
-          }
-          HDFSFolder(path, p.getName, children)
-        }
-      }
-
-      res.map(result => ResponseEntity.ok(result))
+    if (fs.exists(path)) {
+      getFolder(path).map(result => ResponseEntity.ok(result))
     } else {
       Future.successful(ResponseEntity.notFound().build[HDFSFolder]())
+    }
+  }
+
+  private def getFolder(path: Path) = {
+    if (!fs.isDirectory(path)) {
+      Future {
+        HDFSFolder(path.toUri().getPath, path.getName, None)
+      }
+    } else {
+      Future {
+        val status = fs.listStatus(path)
+        val children = if (status.isEmpty) None else {
+          Some(
+            status.map({ x =>
+              val child = x.getPath
+
+              HDFSFolder(child.toUri().getPath, child.getName,
+                if (fs.listStatus(child).isEmpty) None else Some(Seq(HDFSFolder("", "", None))))
+
+            }).toSeq)
+        }
+        HDFSFolder(path.toString, path.getName, children)
+      }
     }
   }
 
--- a/menas/src/main/scala/za/co/absa/enceladus/rest/controllers/HDFSController.scala
+++ b/menas/src/main/scala/za/co/absa/enceladus/rest/controllers/HDFSController.scala
@@ -17,12 +17,14 @@ package za.co.absa.enceladus.rest.controllers
 
 import za.co.absa.enceladus.model.api.HDFSFolder
 import java.util.concurrent.CompletableFuture
-import org.apache.hadoop.fs.{Path,FileSystem}
+
+import org.apache.hadoop.fs.{FileSystem, Path}
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.web.bind.annotation._
 
 import scala.concurrent._
 import org.slf4j.LoggerFactory
+import org.springframework.http.ResponseEntity
 
 @RestController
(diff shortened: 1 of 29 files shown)


menas/src/main/scala/za/co/absa/enceladus/rest/controllers/HDFSController.scala
```diff
@@ -34,28 +36,33 @@ class HDFSController @Autowired() (fs: FileSystem) {
...
+          HDFSFolder(path, p.getName, children)
         }
-        HDFSFolder(path, p.getName, children)
       }
-    }
 
-    res
+      res.map(result => ResponseEntity.ok(result))
+    } else {
+      Future.successful(ResponseEntity.notFound().build[HDFSFolder]())
```
Zejnilovic: Isn't this called a guard clause or something like that. Basically, you have a huge if and then a small else. What you should do is have an `if not` on top with an early `return`, a one-liner, and rest of the method doesn't need any `else` or things like that.

menas/ui/service/SchemaService.js
```diff
@@ -92,25 +92,26 @@ var SchemaService = new function() {
...
-				} else {
-					SchemaService.getSchemaList(true, false)
-				}
-			}
-		}, function() {
-			sap.m.MessageBox.error("Failed to disable schema. Ensure no mapping tables or datasets use this schema(and/or version)")
-		})	
-	};
+    Functions.ajax(uri , "GET", {}, function(oData) {
+      sap.m.MessageToast.show("Schema disabled.");
```
Zejnilovic: Is it just badly showing or is the indentation moved too far to the left?

menas/ui/service/SchemaService.js
```diff
@@ -92,25 +92,26 @@ var SchemaService = new function() {
...
+      if(window.location.hash != "#/schema") {
+        window.location.hash = "#/schema"
+      } else {
+        SchemaService.getSchemaList(true, false)
+      }
+    }, function(xhr) {
+      if (xhr.status === 400) {
+        var err = "Disabling schema failed. Clear the following dependencies first:\n";
+        var oData = JSON.parse(xhr.responseText);
```
Zejnilovic: This `var oData` should be `val oData`

menas/ui/service/SchemaService.js
benedeki: I think a `return` is bigger antipattern in _Scala_ then a long `if` with short `else`.

menas/ui/service/SchemaService.js
Zejnilovic: I was thinging of let in react

menas/ui/service/SchemaService.js
Zejnilovic: Return is not and antipatern and scala does emphasize on readability, which this provides.

GeorgiChochov: I think this conversation is taking place on the wrong thread. 
As to the original comment, the indentation is 2spaces (showing 2 spaces), the stuff around it is tabbed (showing 4-spaces). This is just the way gtihub is showing it. I didn't want to edit it beyond what was necessary because there's big changes in #79 and I didn't want to deal with the conflicts.

menas/ui/service/SchemaService.js
GeorgiChochov: Sure, could be `let`, we're using ES6 after all.

menas/src/main/scala/za/co/absa/enceladus/rest/controllers/HDFSController.scala
GeorgiChochov: I would agree with you if this was java, but returns are antithetical to scala. Speaking as someone who's had to disentangle exactly this sort of early return mess, I'd rather avoid them entirely than deal with unexpected (`non-local returns`)[https://tpolecat.github.io/2014/05/09/return.html].
However, it's probably a sign that a bit of refactoring is needed here, so thanks for pointing it out.

menas/src/main/scala/za/co/absa/enceladus/rest/controllers/HDFSController.scala
```diff
@@ -34,13 +36,26 @@ class HDFSController @Autowired() (fs: FileSystem) {
...
+    Future { fs.exists(path) }.flatMap { exists =>
+      if (exists) {
+        getFolder(path).map(result => ResponseEntity.ok(result))
+      } else {
+        Future.successful(ResponseEntity.notFound().build[HDFSFolder]())
+      }
+    }
+  }
+
+  private def getFolder(path: Path) = {
```
jozefbakus: What do you think about defining return type of the method ? (Consistency)
Do you have reason why return type of this method is Future ? 
I think you can get rid of them and return only HDFSFolder.
Then method `getHDFSFolder` will look like this: 
`def getHDFSFolder(@RequestBody pathStr: String): CompletableFuture[ResponseEntity[HDFSFolder]] = {
    val path = new Path(pathStr)

    Future {
      if (fs.exists(path)) {
        ResponseEntity.ok(getFolder(path))
      } else {
        ResponseEntity.notFound().build[HDFSFolder]()
      }
    }.flatMap(Future.successful)
  }`

menas/src/main/scala/za/co/absa/enceladus/rest/controllers/MappingTableController.scala
```diff
@@ -34,17 +35,25 @@ class MappingTableController @Autowired() (mappingTableService: MappingTableServ
   import scala.concurrent.ExecutionContext.Implicits.global
 
   @PostMapping(path = Array("/updateDefaults"))
-  def updateDefaults(@AuthenticationPrincipal user: UserDetails, @RequestBody upd: MenasObject[Array[DefaultValue]]): CompletableFuture[MappingTable] = {
+  def updateDefaults(@AuthenticationPrincipal user: UserDetails,
+                     @RequestBody upd: MenasObject[Array[DefaultValue]]): CompletableFuture[ResponseEntity[MappingTable]] = {
     mappingTableService.update(user.getUsername, upd.id.name) { mt =>
       mt.setDefaultMappingValue(upd.value.toList)
+    }.map {
+      case Some(entity) => ResponseEntity.ok().body(entity)
```
jozefbakus:
```
    ResponseEntity.ok().body(entity)
    ResponseEntity.ok(entity)
    new ResponseEntity(entity, HttpStatus.OK)
```
If you have no reason to use all of them then it would be nice to have only one. It causes lot of confusion.
And because of responses like: ` ResponseEntity.notFound().build[MappingTable]()` 
It would be better define your own "response codes" that will wrap builders. 
For example instead of this: ` ResponseEntity.notFound().build[MappingTable]()` can be NotFound[MappingTable]

GeorgiChochov: Really appreciate the feedback! I used to do it this way in Play, just got lazy here or curious to see how it feels doing it like this. The way we build `201 CREADTED` sticks out the most, since it accepts a location header, which we can't really supply without bending over backwards. Ideally we would provide the location header, but currently we return the created entity and that is enough for now.

Thanks for keeping me accountable!

PR #94, opened 2019-01-04
Bump Jackson to 2.9.8 because of security vulnerability

GeorgiChochov:


lokm01:
All tests green? I think this version was specifically chosen to avoid binary incompatibilities with sparks Jackson.

GeorgiChochov:
> All tests green? I think this version was specifically chosen to avoid binary incompatibilities with sparks Jackson.

Tests came up green, but I do remember something with the Spring/Spark Jackson dependencies. Spark 2.2.1 is using 2.6.5, Spring is using 2.9.x. I think it was more about getting Spark to comply with Jackson 2.9.x, rather than specifically Jackson 2.9.5. 
I remember I thought we might have to go back to Spring 4 to match Spark's Jackson 2.6.x, but you got it to work with Spring 5 (and hence Jackson 2.9.5). Jackson's following [Apache/Semantic versioning](https://github.com/FasterXML/jackson/wiki/Jackson-Releases), so if we got it to work with 2.9.5, 2.9.8 should be compatible.
Will test more on Monday.

GeorgiChochov:
Just ran a quick test with that tiny Spring/Spark thing I have (whose Jackson was already 2.9.6 rather than Enceladus's 2.9.5) and it delivers JSON from HDSF over HTTP just fine with Jackson 2.9.8. I reckon we're good.

PR #115, opened 2019-01-15
104 Add login page

GeorgiChochov:
- Login/Logout are now restful.
- Simple login page inside the sapUI5 front-end, with routing that leads to the login page if you're unauthenticated and doesn't allow you to go back to it if you are.
- Supports multiple tabs being open, session lives on until it expires or user logs out.
- If user has two tabs open and logs out from one, when they try to do anything on the other they are sent to log in.

### DIFF
--- a/menas/src/main/scala/za/co/absa/enceladus/rest/WebSecurityConfig.scala
+++ b/menas/src/main/scala/za/co/absa/enceladus/rest/WebSecurityConfig.scala
@@ -127,7 +127,7 @@ class WebSecurityConfig {
 //  }
 
   @Bean
-  def aZuthenticationFailureHandler(): SimpleUrlAuthenticationFailureHandler = {
+  def authenticationFailureHandler(): SimpleUrlAuthenticationFailureHandler = {
     new SimpleUrlAuthenticationFailureHandler()
   }
 
--- a/menas/src/main/scala/za/co/absa/enceladus/rest/WebSecurityConfig.scala
+++ b/menas/src/main/scala/za/co/absa/enceladus/rest/WebSecurityConfig.scala
@@ -137,19 +137,11 @@ class WebSecurityConfig {
 
   @Configuration
   @Order(1)
-  class ApiWebSecurityConfigurationAdapter() extends WebSecurityConfigurerAdapter {
-
-    @Autowired()
-    val restAuthenticationEntyPoint: RestAuthenticationEntryPoint = null
-
-    @Autowired()
-    val authenticationSuccessHandler: AuthSuccessHandler = null
-
-    @Autowired()
-    val authenticationFailureHandler: AuthenticationFailureHandler = null
-
-    @Autowired()
-    val logoutSuccessHandler: LogoutSuccessHandler = null
+  class ApiWebSecurityConfigurationAdapter @Autowired()(restAuthenticationEntyPoint: RestAuthenticationEntryPoint,
+                                                        authenticationSuccessHandler: AuthSuccessHandler,
+                                                        authenticationFailureHandler: AuthenticationFailureHandler,
+                                                        logoutSuccessHandler: LogoutSuccessHandler)
+    extends WebSecurityConfigurerAdapter {
 
     override def configure(http: HttpSecurity) {
       http
--- a/menas/src/main/scala/za/co/absa/enceladus/rest/WebSecurityConfig.scala
+++ b/menas/src/main/scala/za/co/absa/enceladus/rest/WebSecurityConfig.scala
@@ -37,8 +37,8 @@ import org.springframework.security.ldap.authentication.ad.ActiveDirectoryLdapAu
 import org.springframework.security.ldap.search.FilterBasedLdapUserSearch
 import org.springframework.security.ldap.userdetails.{LdapUserDetailsMapper, LdapUserDetailsService}
 import org.springframework.security.web.AuthenticationEntryPoint
-import org.springframework.security.web.authentication.logout.HttpStatusReturningLogoutSuccessHandler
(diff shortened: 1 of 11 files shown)


menas/src/main/scala/za/co/absa/enceladus/rest/WebSecurityConfig.scala
```diff
@@ -116,24 +126,53 @@ class WebSecurityConfig {
...
+
+    @Autowired()
+    val authenticationFailureHandler: AuthenticationFailureHandler = null
+
+    @Autowired()
+    val logoutSuccessHandler: LogoutSuccessHandler = null
+
     override def configure(http: HttpSecurity) {
       http
+        .csrf().ignoringAntMatchers("/api/login").and()
```
Zejnilovic: I would shift every `.XXX` onto new line, so it is the same as the rest

menas/ui/components/login/loginMain.controller.js
```diff
@@ -0,0 +1,137 @@
...
+      }
+    },
+
+    _resetFieldState: function (sField) {
+      this.byId(sField).setValueState(sap.ui.core.ValueState.None);
+      this.byId(sField).setValueStateText("");
+    },
+
+    _resetLoginFormState: function () {
+      this._resetFieldState("username");
```
Zejnilovic: Here comes the question/discussion. Do we want to be the guys that make you re-type even the username?

menas/ui/components/login/loginMain.controller.js
```diff
@@ -0,0 +1,137 @@
...
+        },
+        data: oData,
+        method: "POST",
+        success: fnSuccess,
+        error: fnError
+      })
+    },
+    /**
+     * Similar to onAfterRendering, but this hook is invoked before the controller's
+     * View is re-rendered (NOT before the first rendering! onInit() is used for
```
Zejnilovic: Is this left here on purpose?

menas/ui/generic/functions.js
```diff
@@ -13,34 +13,42 @@
...
-			method : sMethod,
-			success : fnSuccess,
-			error : fnError
-		})
-	}
+  this.ajax = function (sPath, sMethod, oData, fnSuccess, fnError, oControl) {
+    if (oControl) oControl.setBusy(true);
+
+    let oFormattedData = null;
+    if ((sMethod.toLowerCase() === "post" || sMethod.toLowerCase() === "put") && typeof (oData) === "object") {
```
Zejnilovic: maybe a new line after `&&` or something like that? And put the whole last type of in `()`

Zejnilovic: Is there a reason for `oData` to be in `()`

menas/ui/components/login/loginMain.view.xml
```diff
@@ -0,0 +1,29 @@
...
+
+<core:View id="loginMainView" xmlns:core="sap.ui.core" xmlns:mvc="sap.ui.core.mvc" xmlns="sap.m" xmlns:form="sap.ui.layout.form"
+		   controllerName="components.login.loginMain">
+	<Page title="Login" enableScrolling="true">
+		<content>
+			<VBox id="loginForm" fitContainer="true" justifyContent="Center" alignItems="Center">
+				<items>
+					<Input type="Text" id="username" placeholder="username" value="{login>/username}" submit="onLoginSubmit"/>
+					<Input type="Password" id="password" placeholder="password" value="{login>/password}" submit="onLoginSubmit"/>
+					<Button width="10rem" text="Login" type="Emphasized" id="loginSubmit"/>
```
Zejnilovic: I know there is no predetermined sequence of attributes, but I would like if we could keep it same for the whole view. So `type`, `id`, `..rest..`. Just for readability.

menas/ui/components/login/loginMain.controller.js
GeorgiChochov: Yeah, for consistency like in the other controllers. Maybe we can drop them, I don't know.

menas/ui/components/login/loginMain.view.xml
GeorgiChochov: Fair point.

menas/ui/generic/functions.js
GeorgiChochov: > Is there a reason for `oData` to be in `()`

Don't think so, just the way it was.

menas/ui/components/login/loginMain.controller.js
GeorgiChochov: You saying extract it? I can do that.

menas/ui/generic/functions.js
```diff
@@ -13,34 +13,45 @@
...
-			success : fnSuccess,
-			error : fnError
-		})
-	}
+  this.ajax = function (sPath, sMethod, oData, fnSuccess, fnError, oControl) {
+    if (oControl) oControl.setBusy(true);
+
+    let oFormattedData = null;
+    if ((sMethod.toLowerCase() === "post" || sMethod.toLowerCase() === "put")
+      && (typeof oData === "object")) {
```
Zejnilovic: I think `&&` should be the last thing on the previous line. So the programmer reading it know that it is continuing from a first glance.

GeorgiChochov: I tend to disagree, for the same reason method calls on new lines start with a dot (e.g. `.map(...)`) makes it clear that this line isn't a statement on its own. I look at the start of lines a lot more than their ends of I'm just scanning the code.

Zejnilovic: As this is subjective, dealer call. Just know, that people who look at the end exist 😄

PR #117, opened 2019-01-16
86 Allow Enceladus and Menas communication

GeorgiChochov:


dao/src/main/scala/za/co/absa/enceladus/dao/EnceladusRestDAO.scala
```diff
@@ -33,53 +31,89 @@ import com.fasterxml.jackson.databind.ObjectMapper
...
-      .configure(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS, false)
-  
+    .registerModule(DefaultScalaModule)
+    .registerModule(new JavaTimeModule())
+    .configure(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS, false)
+
+  def postLogin(username: String, password: String) = {
+    try {
+      val httpClient = HttpClients.createDefault
+      val url = s"$restBase/login?username=$username&password=$password&submit=Login"
```
Zejnilovic: Does this take into account special characters that users might have in their passwords?

dao/src/main/scala/za/co/absa/enceladus/dao/EnceladusRestDAO.scala
```diff
@@ -33,53 +31,89 @@ import com.fasterxml.jackson.databind.ObjectMapper
...
+      val httpPost = new HttpPost(url)
+
+      val response: CloseableHttpResponse = httpClient.execute(httpPost)
+      try {
```
Zejnilovic: I don't like nested try, but I wasn't yet able to figure the best solution. Leave this comment here, if I come up with something.

dao/src/main/scala/za/co/absa/enceladus/dao/EnceladusRestDAO.scala
```diff
@@ -33,53 +31,89 @@ import com.fasterxml.jackson.databind.ObjectMapper
...
 import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule
 import com.fasterxml.jackson.databind.SerializationFeature
+import za.co.absa.enceladus.dao.MenasRestDAO.log
 
 object EnceladusRestDAO extends EnceladusDAO {
   val conf = ConfigFactory.load()
   val restBase = conf.getString("menas.rest.uri")
   private val userName = LoggedInUserInfo.getUserName
 
+  var sessionCookie: String = ""
```
Zejnilovic: TBH I don't like `var`s either but I don't have a solution for it. But how about setting them `private` for the time being

standardization/src/main/scala/za/co/absa/enceladus/standardization/CmdConfig.scala
```diff
@@ -74,6 +76,10 @@ object CmdConfig {
         }
       )
 
+    opt[String]('m', "menas-credentials-path").required().action((path, config) =>
```
Zejnilovic: Same validation as first comment?

dao/src/main/scala/za/co/absa/enceladus/dao/EnceladusRestDAO.scala
GeorgiChochov: Yeah, I have an idea using closures, but they can't be private because MenasRestDao needs them too. There is no reason to keep the two separate, but I didn't want to do this heavy a refactoring with this PR.

Zejnilovic: If you are thinking about it. Then just resolve this conversation. I believe in you.

dao/src/main/scala/za/co/absa/enceladus/dao/EnceladusRestDAO.scala
GeorgiChochov: This is subject to the same heavy refactoring I'd like to do in a follow up. Do you think it makes sense to separate that from the current or am I being overly fragmantory?

PR #128, opened 2019-01-22
Feature/123b rest testing framework

Zejnilovic:
Added a rest testing feature. This will let us run rest tests as a spark job and log outputs pretty easily. The main point of this is the integration into the future pipeline.
Second, non-main, feature is that you can use its models and RestCaller object in unit tests, or others.
I will update the README.md soon. I just wanted to share this first.


testutils/src/main/scala/za/co/absa/enceladus/testutils/models/TestCaseResult.scala
```diff
@@ -0,0 +1,26 @@
...
+
+case class TestCaseResult(id: String,
+                          name: String,
+                          comparison: String,
+                          passed: Boolean,
+                          input: String,
+                          output: String,
+                          expected: String,
+                          statusCode: String,
+                          elapsedTime: Long) {}
```
GeorgiChochov: You don't need the empty parens `{}`

testutils/src/main/scala/za/co/absa/enceladus/testutils/models/TestRun.scala
```diff
@@ -0,0 +1,55 @@
...
+
+package za.co.absa.enceladus.testutils.models
+
+import org.joda.time.{DateTime, DateTimeZone}
+
+import scala.collection.mutable
+
+class TestRun(name: String) {
+  val startTime: DateTime = DateTime.now(DateTimeZone.UTC)
+  lazy val endTime: DateTime = DateTime.now(DateTimeZone.UTC)
```
GeorgiChochov: Why the `lazy val`? You only, use it in the `getFinalInfo` function. It could easily be a local variable and it won't need to be lazy.

testutils/src/test/scala/za/co/absa/enceladus/testutils/rest/RestCallerTest.scala
```diff
@@ -0,0 +1,67 @@
...
+import org.scalatest.{BeforeAndAfterEach, FunSuite}
+import za.co.absa.enceladus.testutils.models.{RestMethod, TestCase, TestCaseResult}
+
+class RestCallerTest extends FunSuite with BeforeAndAfterEach {
+  test("testRun") {
+    val testCase: TestCase = TestCase("001",
+      "Test001",
+      "Post",
+      "application/json",
+      "https://httpbin.org/post",
```
GeorgiChochov: Does this mean that in order to pass, this test needs an open connection to the internet?

testutils/src/main/scala/za/co/absa/enceladus/testutils/rest/RestRunnerJob.scala
```diff
@@ -0,0 +1,75 @@
...
+
+    val testRunName: String = cmd.testDataPath.split("/").last
+    val testRun: TestRun = new TestRun(testRunName)
+
+    val results: Array[TestCaseResult] = testData.collect.map({ RestCaller.run })
+    testRun.addTestCaseResults(results)
+
+    val finalInfo = testRun.getFinalInfo
+    val today = Calendar.getInstance.getTime
+    val dateFormat = new SimpleDateFormat("YYYY_MM_dd/hh_MM")
```
GeorgiChochov: Isn't this format `YEAR_MONTH_DAY/HOUR_MONTH`? I'd go with `yyyy_MM_dd/HH_mm`:
* `mm` for minutes
* `HH` for exact hour of day (2 digits, 00-23), [not `hh` (0-12 AM/PM)](https://docs.oracle.com/javase/6/docs/api/java/text/SimpleDateFormat.html)
* `yyyy` for calendar-based year over [week-based year (`YYY`)](https://stackoverflow.com/questions/15133549/difference-between-yyyy-and-yyyy-in-nsdateformatter)

testutils/src/main/scala/za/co/absa/enceladus/testutils/rest/RestCaller.scala
```diff
@@ -0,0 +1,88 @@
...
+    val headers: Map[String, String] = Map("content-type" -> contentType)
+    (method: @switch) match {
+      case RestMethod.POST => call_post(url, payload, headers)
+      case RestMethod.GET => call_get(url, payload, headers)
+      case RestMethod.DELETE => call_delete(url, payload, headers)
+      case RestMethod.PUT => call_put(url, payload, headers)
+    }
+  }
+
+  private def call_post(url: String, payload: String, headers: Map[String, String]): Response = {
```
GeorgiChochov: Don't use undersocre in names

testutils/src/main/scala/za/co/absa/enceladus/testutils/rest/RestCaller.scala
```diff
@@ -0,0 +1,88 @@
...
+
+import scala.annotation.switch
+
+
+object RestCaller {
+
+  def run(testCase: TestCase): TestCaseResult = {
+    lazy val callOutput = call(testCase.method, testCase.url, testCase.payload, testCase.contentType)
+    val time = calculateTime { callOutput }
+    val comparison = compareJsons(testCase.expectedOutput, callOutput.text)
```
GeorgiChochov: Why not call it `diff`, seems more meaningful than `comparison`.

testutils/pom.xml
```diff
@@ -40,9 +49,29 @@
...
+			<version>${diffson.version}</version>
+		</dependency>
+		<dependency>
+			<groupId>io.spray</groupId>
+			<artifactId>spray-json_${scala.compat.version}</artifactId>
+			<version>${spray.json.version}</version>
+		</dependency>
+		<dependency>
+			<groupId>org.apache.hadoop</groupId>
+			<artifactId>hadoop-common</artifactId>
```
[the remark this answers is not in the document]
Zejnilovic: We don't have hadoop-common in our dependencies yet. Hadoop-common is a compilation dependency od hadoop-client.

testutils/pom.xml
```diff
@@ -28,6 +28,15 @@
 		<version>0.99.0-SNAPSHOT</version>
 	</parent>
 
+	<properties>
+		<requests.version>0.1.6</requests.version>
+		<ujson.version>0.7.1</ujson.version>
+		<requests.version>0.1.6</requests.version>
+		<diffson.version>2.0.2</diffson.version>
+		<spray.json.version>1.3.5</spray.json.version>
+		<hadoop.common.version>3.1.1</hadoop.common.version>
```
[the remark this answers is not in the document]
Zejnilovic: I understood that I should move properties with versions to parent but leave dependencies as they are. Hope it is correct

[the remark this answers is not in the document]
GeorgiChochov: Could you just reuse the `<hadopp.version>` property, to avoid having a version clash?

testutils/src/main/scala/za/co/absa/enceladus/testutils/models/TestRun.scala
GeorgiChochov: It just seems like a flawed design since the API allows me to set `endTime` at any point and unintentionally calling the getter produces a side-effect - setting the value.

Zejnilovic:
I would like to ask for final comments or propositions. @Zejnilovic @DzMakatun @benedeki

testutils/src/main/scala/za/co/absa/enceladus/testutils/models/TestCase.scala
```diff
@@ -25,5 +23,12 @@ case class TestCase(id: String,
                     payload: String,
                     expectedOutput: String,
                     expectedStatusCode: String) {
-  val method: RestMethod = RestMethod.withName(methodString)
+//  val method: RestMethod = RestMethod(methodString)
```
GeorgiChochov: This is still commented out

testutils/src/main/scala/za/co/absa/enceladus/testutils/models/RestMethod.scala
```diff
@@ -15,12 +15,10 @@
 
 package za.co.absa.enceladus.testutils.models
 
-import za.co.absa.enceladus.testutils.models
+sealed abstract class RestMethod(val methods: String)
```
GeorgiChochov: `method` singular rather than `methods`

PR #134, opened 2019-01-24
Feature/125 add pk based comparison

Zejnilovic:
Based on a request from multiple sources.

Now with adding the keys option to the spark submit job, it generates one output `parquet` file with all differences grouped. All columns are renamed with prefixes of `actual_` or `expected_` depending if they are coming from `new-path` or `ref-path` respectively.

### DIFF
--- a/testutils/src/main/scala/za/co/absa/enceladus/testutils/DataframeReader.scala
+++ b/testutils/src/main/scala/za/co/absa/enceladus/testutils/DataframeReader.scala
@@ -55,16 +55,16 @@ class DataframeReader(datasetPath: String, inputSchema: Option[StructType] )
                                  sparkSession: SparkSession): DataFrameReader = {
     val dfReader = getStandardReader
 
-    if (inputSchema.isDefined) dfReader.schema(inputSchema.get)
-    else dfReader
+    if (inputSchema.isDefined) { dfReader.schema(inputSchema.get) }
+    else { dfReader }
   }
 
   private def getFixedWidthReader()(implicit dfReaderOptions: DataframeReaderOptions,
                                     sparkSession: SparkSession): DataFrameReader = {
     val dfReader = getStandardReader
 
-    if (dfReaderOptions.fixedWidthTrimValues.get) dfReader.option("trimValues", "true")
-    else dfReader
+    if (dfReaderOptions.fixedWidthTrimValues.get) { dfReader.option("trimValues", "true") }
+    else { dfReader }
   }
 
   private def getXmlReader()(implicit dfReaderOptions: DataframeReaderOptions,
--- a/testutils/src/main/scala/za/co/absa/enceladus/testutils/DataframeReader.scala
+++ b/testutils/src/main/scala/za/co/absa/enceladus/testutils/DataframeReader.scala
@@ -35,7 +35,8 @@ class DataframeReader(datasetPath: String, inputSchema: Option[StructType] )
     StructType(dataFrameSchema.map{ f => StructField(f.name, f.dataType, f.nullable) })
   }
 
-  private def getDataFrameReader(implicit dfReaderOptions: DataframeReaderOptions, sparkSession: SparkSession): DataFrameReader = {
+  private def getDataFrameReader(implicit dfReaderOptions: DataframeReaderOptions,
+                                 sparkSession: SparkSession): DataFrameReader = {
     (dfReaderOptions.rawFormat: @switch) match {
       case "csv" => getCsvReader
       case "xml" => getXmlReader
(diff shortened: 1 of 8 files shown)


testutils/src/main/scala/za/co/absa/enceladus/testutils/datasetComparison/ComparisonJob.scala
```diff
@@ -16,11 +16,33 @@
...
+      else {
+        dataSet(c).as(s"$prefix$c")
+      })
+  }
+
+  def getKeyBasedOutput(expectedMinusActual: Dataset[Row],
+                        actualMinusExpected: Dataset[Row],
+                        keys: Seq[String]): DataFrame = {
+    val renamedColumnsExpected = renameColumns(expectedMinusActual ,keys, "expected_")
+    val dfNewExpected = expectedMinusActual.select(renamedColumnsExpected: _*)
```
GeorgiChochov: These two lines can probably be extracted as a function, to make them a single logical unit for each dataframe.

testutils/src/main/scala/za/co/absa/enceladus/testutils/DataframeReader.scala
```diff
@@ -45,32 +46,37 @@ class DataframeReader(datasetPath: String, inputSchema: Option[StructType] )
...
   }
 
-  private def getParquetReader()(implicit dfReaderOptions: DataframeReaderOptions, sparkSession: SparkSession): DataFrameReader = {
+  private def getParquetReader()(implicit dfReaderOptions: DataframeReaderOptions,
+                                 sparkSession: SparkSession): DataFrameReader = {
     val dfReader = getStandardReader
 
-    if (inputSchema.isDefined) dfReader.schema(inputSchema.get)
-    else dfReader
+    if (inputSchema.isDefined) { dfReader.schema(inputSchema.get) }
```
GeorgiChochov: This is to please scalastyle, right?
I would do this with pattern matching:
```
inputSchema match {
  case Some(schema) => dfReader.schema(schema)
  case None         => dfReader
}
```
This way we get to keep the nice syntax of the line per condition without unnecessary parens and we get rid of `.isDefined` and `.get` calls.

Zejnilovic: Done in all cases

testutils/src/main/scala/za/co/absa/enceladus/testutils/datasetComparison/ComparisonJob.scala
```diff
@@ -53,11 +75,17 @@ object ComparisonJob {
...
+        joinedData.write.format("parquet").save(s"${cmd.outPath}")
+      } else {
+        expectedMinusActual.write.format("parquet").save(s"${cmd.outPath}/expected_minus_actual")
+        actualMinusExpected.write.format("parquet").save(s"${cmd.outPath}/actual_minus_expected")
+      }
+
       throw CmpJobDatasetsDifferException(cmd.refPath, cmd.newPath, cmd.outPath, expectedDf.count(), actualDf.count())
     } else {
-      System.out.println("Expected and actual datasets are the same.")
+      println("Expected and actual datasets are the same.")
```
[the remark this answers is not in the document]
Zejnilovic: Yes

testutils/src/main/scala/za/co/absa/enceladus/testutils/datasetComparison/ComparisonJob.scala
```diff
@@ -53,11 +76,17 @@ object ComparisonJob {
     val actualMinusExpected: Dataset[Row] = actualDf.except(expectedDf)
 
     if (expectedMinusActual.count() != 0 || actualMinusExpected.count() != 0) {
-      expectedMinusActual.write.format("parquet").save(s"${cmd.outPath}/expected_minus_actual")
-      actualMinusExpected.write.format("parquet").save(s"${cmd.outPath}/actual_minus_expected")
+      if (cmd.keys.isDefined) {
+        val joinedData = getKeyBasedOutput(expectedMinusActual, actualMinusExpected, cmd.keys.get)
```
DzMakatun: Already discussed, but just in case ) The datasets are not generally expected to have row duplicates. However, it may happen when there are aggregated datasets for multiple dates, or if there are bugs in the source system. So it would be useful to have a fail safe check at the beginning (before joining): e.g., group by keys each dataset and indicate (or maybe throw exception) if count > 1

testutils/src/main/scala/za/co/absa/enceladus/testutils/datasetComparison/ComparisonJob.scala
```diff
@@ -24,22 +24,22 @@ import za.co.absa.enceladus.testutils.{DataframeReader, DataframeReaderOptions}
 object ComparisonJob {
   val log: Logger = LogManager.getLogger("enceladus.testutils.ComparisonJob")
 
-  private def renameColumns(dataSet: Dataset[Row], keys: Seq[String], prefix: String): Array[Column] = {
-    dataSet.columns.map(c =>
+  private def renameColumns(dataSet: Dataset[Row], keys: Seq[String], prefix: String): DataFrame = {
+    val renamedColumnsExpected = dataSet.columns.map(c =>
```
GeorgiChochov: I guess the variable name should be more generic than `renamedColumnsExpected`, maybe just `renamedColumns`

GeorgiChochov: Also, I would use  braces rather than parens with multiline callbakcs: `.map { c =>` rather than `.map( c =>`

testutils/src/main/scala/za/co/absa/enceladus/testutils/datasetComparison/ComparisonJob.scala
Zejnilovic: It is done. But it works only if one or more keys (unique ids) are specified.

Zejnilovic: Will wait for your approval.

PR #138, opened 2019-01-29
Feature/110 optimize conformance further

yruslan:
Add tools for appending errors to an error column to nested mapping tools based on transform() and concat() high order functions.

Optimised:
- Casting conformance rule
- Negation conformance rule

### DIFF
--- a/conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/rules/CastingRuleInterpreter.scala
+++ b/conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/rules/CastingRuleInterpreter.scala
@@ -15,18 +15,14 @@
 
 package za.co.absa.enceladus.conformance.interpreter.rules
 
-import org.apache.spark.sql.api.java.UDF1
 import org.apache.spark.sql.functions._
-import org.apache.spark.sql.types.{ArrayType, StringType, StructType}
+import org.apache.spark.sql.types.StringType
 import org.apache.spark.sql.{Dataset, Row, SparkSession}
 import za.co.absa.enceladus.conformance.CmdConfig
 import za.co.absa.enceladus.conformance.interpreter.RuleValidators
 import za.co.absa.enceladus.dao.EnceladusDAO
 import za.co.absa.enceladus.model.conformanceRule.CastingConformanceRule
-import za.co.absa.enceladus.utils.error.ErrorMessage
-import za.co.absa.enceladus.utils.schema.SchemaUtils
-import za.co.absa.enceladus.utils.transformations.ArrayTransformations
-import za.co.absa.enceladus.utils.validation._
+import za.co.absa.enceladus.utils.transformations.DeepArrayTransformations
 
 case class CastingRuleInterpreter(rule: CastingConformanceRule) extends RuleInterpreter {
   final val ruleName = "Casting rule"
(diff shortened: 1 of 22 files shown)


conformance/src/test/scala/za/co/absa/enceladus/conformance/interpreter/rules/ConcatenationRuleSuite.scala
```diff
@@ -23,18 +23,22 @@ import za.co.absa.enceladus.samples.DeepArraySamples
...
+    hdfsPath = "src/test/testData/library", hdfsPublishPath =
     "testData/conformedLibrary",
     schemaName = "Library", schemaVersion = 1,
     conformance = concatRulesList)
 
+  // scalastyle:off line.size.limit
   private val conformedConcatOrdersJSON: String =
     """{"id":1,"name":"National Library","city":"Washington","address":"10 Linkoln ave.","rooms":[{"roomName":"History","label":"H1","capacity":25000,"books":[{"author":"Christopher Browning","name":"Ordinary Men","CombinedBooks":"Christopher BrowningOrdinary Men"},{"author":"Vasily Grossman","name":"Life and Fate","CombinedBooks":"Vasily GrossmanLife and Fate"},{"author":"Homer","name":"Illiad","CombinedBooks":"HomerIlliad"}],"CombinedLabel":"HistoryH1"},{"roomName":"Politics","label":"P2","capacity":15000,"books":[{"author":"Michael Lewis","name":"The Fifths Risk","CombinedBooks":"Michael LewisThe Fifths Risk"},{"author":"Aristotle","name":"Politics","CombinedBooks":"AristotlePolitics"}],"CombinedLabel":"PoliticsP2"},{"roomName":"Sociology","label":"S3","capacity":8000,"books":[],"CombinedLabel":"SociologyS3"}],"errCol":[],"CombinedName":"National LibraryWashington10 Linkoln ave."}
```
Zejnilovic: How would you feel about putting this into a file?

conformance/src/test/scala/za/co/absa/enceladus/samples/CastingRuleSamples.scala
```diff
@@ -84,25 +84,128 @@ object CastingRuleSamples {
...
-    "\"price\":99.9500000000,\"ConformedId\":28781,\"ConformedQty\":\"10001\",\"ConformedPrice\":99.95}],\"ConformedStr\":\"3\"," +
-    "\"ConformedDate\":\"2018-02-21\",\"ConformedTimestamp\":1519200617,\"errCol\":[]}\n{\"id\":4,\"date\":\"2018-03-30\"," +
-    "\"placed\":\"2018-03-30T00:22:10.121Z\",\"discount\":0.850123456789,\"items\":[{\"itemid\":\"-123k\",\"qty\":1,\"price\":33.6600000000," +
-    "\"ConformedQty\":\"1\",\"ConformedPrice\":33.66}],\"ConformedStr\":\"4\",\"ConformedDate\":\"2018-03-30\",\"ConformedTimestamp\":1522369330," +
-    "\"ConformedDecimal\":0.850123,\"errCol\":[{\"errType\":\"confCastError\",\"errCode\":\"E00003\",\"errMsg\":\"Conformance Error - Null returned" +
-    " by casting conformance rule\",\"errCol\":\"items.ConformedId\",\"rawValues\":[\"-123k\"],\"mappings\":[]}]}"
+  val conformedOrdersJSON: String = """[ {
+                                      |  "id" : 1,
+                                      |  "date" : "2018-02-10",
+                                      |  "num" : "888",
```
Zejnilovic: How would you feel about putting this into a file?

utils/src/main/scala/za/co/absa/enceladus/utils/general/JsonUtils.scala
```diff
@@ -0,0 +1,43 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.enceladus.utils.general
+
+import com.fasterxml.jackson.databind.ObjectMapper
+import org.apache.spark.sql.{DataFrame, SparkSession}
+
+object JsonUtils {
```
Zejnilovic: If you don't have test, I would at least append 2 sentences about what each is and why is it created.

utils/src/main/scala/za/co/absa/enceladus/utils/general/JsonUtils.scala
```diff
@@ -0,0 +1,43 @@
...
+
+    val jsonUnindented = mapper.readValue(jsonIn, classOf[Any])
+    val indented = mapper.writerWithDefaultPrettyPrinter.writeValueAsString(jsonUnindented)
+    indented.replace("\r\n", "\n")
+  }
+
+  def prettySparkJSON(jsonIn: String): String = {
+    val properJson = "[" + "}\n".r.replaceAllIn(jsonIn, "},\n") + "]"
+    prettyJSON(properJson)
+
```
Zejnilovic: Line more, then needed

utils/src/main/scala/za/co/absa/enceladus/utils/general/JsonUtils.scala
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  def getDataFrameFromJson(spark: SparkSession, json: String): DataFrame = {
```
Zejnilovic: isn't sparkSession usually passed as implicit? Could it be even here?

utils/src/main/scala/za/co/absa/enceladus/utils/transformations/DeepArrayTransformations.scala
```diff
@@ -161,16 +351,40 @@ object DeepArrayTransformations {
...
+              // Retain the original column
+              ensureOutputColumnNameValid(outputFieldName)
+              mappedFields += transform(curColumn, lambdaName, exp).as(outputFieldName)
+
+              // Handle error column for arrays of primitives
+              handleErrorColumnOfArraysOfPrimitives(errorExpression, doFlatten)
+              curColumn
           }
         } else {
           // This is the case then the caller requested to map a field that is a child of a primitive.
```
Zejnilovic: This is the case `when` the caller `is` requested

utils/src/test/scala/za/co/absa/enceladus/utils/transformations/DeepArraySamples.scala
```diff
@@ -0,0 +1,239 @@
...
+    TradeWithErrorColumn(3, Seq(
+      Leg(104, Seq(
+        Condition("if bid>12", "1OO", 100), Condition("if sell<7", "150x", 150), Condition("if sell<3", "-1000-", 1000))),
+      Leg(105, Seq(
+        Condition("if bid<52", "2OO", 200), Condition("if sell>32", "f175", 175), Condition("if sell>27", "225_", 225)))
+    ), Seq(ErrorMessage("myErrorType", "E-1", "Testing This stuff", "whatEvColumn", Seq("some value"))))
+  )
+
+  // scalastyle:on magic.number
+  // scalastyle:on line.size.limit
```
Zejnilovic: If needed, I would move this up

conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/rules/CastingRuleInterpreter.scala
```diff
@@ -15,18 +15,14 @@
...
 import za.co.absa.enceladus.dao.EnceladusDAO
 import za.co.absa.enceladus.model.conformanceRule.CastingConformanceRule
-import za.co.absa.enceladus.utils.error.ErrorMessage
-import za.co.absa.enceladus.utils.schema.SchemaUtils
-import za.co.absa.enceladus.utils.transformations.ArrayTransformations
-import za.co.absa.enceladus.utils.validation._
+import za.co.absa.enceladus.utils.transformations.DeepArrayTransformations
 
 case class CastingRuleInterpreter(rule: CastingConformanceRule) extends RuleInterpreter {
   final val ruleName = "Casting rule"
```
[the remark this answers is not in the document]
yruslan: Yeah, a good spot, Thanks. I'd rather not go into an automatic rule name generation for now as some of the rules have 2 word name and some only 1 word. But it makes sense to make `ruleName` an abstract method that can be overridden in all the rules.

utils/src/main/scala/za/co/absa/enceladus/utils/transformations/DeepArrayTransformations.scala
```diff
@@ -62,14 +190,105 @@ object DeepArrayTransformations {
       name
     }
 
-    // Returns true if a path consists only of 1 element meaning it is the leaf element of the input column path requested by the caller
-    def isLeafElement(path: Seq[String]): Boolean = path.lengthCompare(2) < 0 // More efficient version of path.length == 1
+    def ensureOutputColumnNameValid(columnName: String): Unit= {
```
benedeki: This method feels weird. It uses 3 variables, where two are input parameters of the encompassing function. The other could provide some flexibility but actually is always  `outputComunName`  a value fixated from another of the input parameters of the higher context as the first step there.

utils/src/main/scala/za/co/absa/enceladus/utils/transformations/DeepArrayTransformations.scala
```diff
@@ -62,14 +190,105 @@ object DeepArrayTransformations {
...
+      if (columnName.isEmpty) {
+        if (isAddColumn) {
+          throw new IllegalArgumentException(s"The column '$inputColumnName' already exists.")
+
+        } else {
+          throw new IllegalArgumentException(s"Output field cannot be empty when transforming an " +
```
benedeki: I guess the string concatenation is done because of the line length limit?
I think, just a suggestion, it would be better to either put the string to its own line as whole or breaking the limit.

utils/src/main/scala/za/co/absa/enceladus/utils/schema/SchemaUtils.scala
```diff
@@ -178,5 +179,39 @@ object SchemaUtils {
...
+    * Generate a unique column name
+    *
+    * @param prefix A prefix to use for the column name
+    * @param schema An optional schema to validate if the column already exists (a very low probability)
+    * @return A name that can be used as a unique column name
+    */
+  def getUniqueName(prefix: String, schema: Option[StructType]): String = {
+    schema match {
+      case None =>
+        s"${prefix}_${Random.nextLong().abs}"
```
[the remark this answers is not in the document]
yruslan: Static unique counters is almost as bad. I'll consider changing this, but not sure if I will end up doing it.

conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/rules/CastingRuleInterpreter.scala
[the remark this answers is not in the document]
yruslan: Sorry, I've looked at the code closely and changed my mind. Not all rules have to define names, only these that do additional validation that requires rule name in exception messages. And there are custom rules which I'd rather not touch for now. The idea of making this field abstract is still sound, but maybe it requires a separate task of tidying up conformance rules.

utils/src/main/scala/za/co/absa/enceladus/utils/transformations/DeepArrayTransformations.scala
```diff
@@ -44,14 +166,20 @@ object DeepArrayTransformations {
...
-  def nestedWithColumnMap(df: DataFrame,
-                          inputColumnName: String,
-                          outputColumnName: String,
-                          expression: Column => Column): DataFrame = {
+  private def nestedWithColumnMapHelper(df: DataFrame,
+                                        inputColumnName: String,
+                                        outputColumnName: String,
+                                        expression: Option[Column => Column] = None,
+                                        errorCondition: Option[Column => Column] = None,
+                                        isAddColumn: Boolean = false
```
benedeki: Suggestion/idea: I see this Boolean flag only affects the message of exception eventually thrown. And is used as `true` only on e place. Maybe omitting the flag, catching and re-raising the exception with the desired message on that one place would make the code less cumbersome.

conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/rules/CastingRuleInterpreter.scala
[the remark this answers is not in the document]
benedeki: Fair enough!

utils/src/main/scala/za/co/absa/enceladus/utils/general/JsonUtils.scala
yruslan: Yes, this test suite inherits SparkTestBase which creates a special instance of SPark session in local mode.

conformance/src/test/scala/za/co/absa/enceladus/conformance/interpreter/rules/NegationRuleSuite.scala
```diff
@@ -26,6 +26,10 @@ import za.co.absa.enceladus.samples.NegationRuleSamples
 import za.co.absa.enceladus.utils.testUtils.SparkTestBase
 
 class NegationRuleSuite extends FunSuite with SparkTestBase {
+  // scalastyle:off import.grouping
```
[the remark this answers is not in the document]
GeorgiChochov: #140

utils/src/main/scala/za/co/absa/enceladus/utils/schema/SchemaUtils.scala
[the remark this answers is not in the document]
yruslan: After a lot of thinking I'm inclined to leave it as is for now.

PR #140, opened 2019-02-01
139 Turn off scalastyle for tests

GeorgiChochov:
Scalastyle is already off for tests, style warnings are not raised there. You may need to change your IDE code-inspection settings though (these are two separate issues). 
In IntelliJ:
image

### DIFF
--- a/scalastyle-config.xml
+++ b/scalastyle-config.xml
@@ -152,5 +152,5 @@
       <parameter name="ignoreRegex">^""$</parameter>
     </parameters>
   </check>
-  <check level="warning" class="org.scalastyle.scalariform.ImportGroupingChecker" enabled="true"/>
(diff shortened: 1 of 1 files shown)


Zejnilovic:
Still, I ask everyone to not throw code-style out of the window when writing tests.

benedeki:
Thanks for the screenshot where to switch off the  Scala Style for IDE. 👍 
Which one is the setting switching it off for tests within the source-code?

GeorgiChochov:
> Thanks for the screenshot where to switch off the Scala Style for IDE. 👍
> Which one is the setting switching it off for tests within the source-code?

https://github.com/AbsaOSS/enceladus/blob/develop/pom.xml#L223

PR #143, opened 2019-02-05
Feature/142 upgrade spark 2.4

yruslan:
* Switched the project to Spark 2.4 (Thanks Marek & Georgi)
* Switched to Spline 0.3.6
* Used spark-hofs transform wrappers() to update deep array transformation facilities
* Reversed modules dependencies. Utils had a dependency on DAO. Now DAO depends on Utils.
* Moved Menas plugin from Utils to DAO because it depends on DAO and Utils.
* Added  a test ensuring Standardization can still use json files as a source.


dao/pom.xml
```diff
@@ -52,6 +52,17 @@
...
+        <dependency>
+            <groupId>org.apache.httpcomponents</groupId>
+            <artifactId>httpclient</artifactId>
+            <version>4.4.1</version>
+        </dependency>
+        <dependency>
+            <groupId>org.scalatest</groupId>
+            <artifactId>scalatest_${scala.compat.version}</artifactId>
+            <version>${scalatest.version}</version>
+            <scope>provided</scope>
```
GeorgiChochov: Shouldn't `scalatest` be scoped `test`?

GeorgiChochov: Also, don't you inherit this for free from [`parent-pom`](https://github.com/AbsaOSS/enceladus/blob/develop/pom.xml#L127)?

yruslan: Indeed. We don't need that dependency in DAO. Will remove it.

But curiously enough such exact 'provided' dependency is present in Utils. That's because of `SparkJobRunnerMethods` - a cute reflection/loader trick to run Spark jobs as tests so that provided dependencies are also loaded.

We don't use `SparkJobRunnerMethods` in DCE, but it is a beautiful piece of code that I cannot fully comprehend, but just use and enjoy.

Now here is a dilemma. We can
1. Leave `SparkJobRunnerMethods` in place so we know which project/module to search for when we need it. And let contributors marvel our god-like smartness. Utils is a collection of generic tools after all
2. Remove `SparkJobRunnerMethods` and scala-test dependency - the only non-inherited dependency of Utils.

I'm slightly more inclined towards 1.

GeorgiChochov: Agreed, let's keep `SparkJobRunnerMethods`!

Issue #84, opened 2018-12-28
Delete conformance rule


GeorgiChochov:
Users should be able to delete a conformance rule through the Menas UI.

PR #147, opened 2019-02-06
#84 Delete conformance rule

jozefbakus:
### [#84 Delete conformance rule](https://github.com/AbsaOSS/enceladus/issues/84)
- Concept used for implementation: Any change related to Dataset should be done via update Dataset endpoint.
- Implemented without validation. It will be implemented in separate future. 
- Warning fixes.
### Test plan: 
**1. Run tests:**
- go to enceladus dir
- mvn verify

**2. Remove rule:**
- Run application
- Log in 
- Go to Dataset tab 
- Go to Conformance Rules
- Create at least 2 rules 
- Delete one of them 
- **Expected result**: Rule should be removed. Dataset version should be increased.

### DIFF
--- a/menas/ui/components/dataset/datasetMain.controller.js
+++ b/menas/ui/components/dataset/datasetMain.controller.js
@@ -221,7 +221,7 @@ sap.ui.controller("components.dataset.datasetMain", {
   },
 
   onRuleMenuAction: function (oEv) {
-    let sAction = oEv.getParameter("item").data("action")
+    let sAction = oEv.getParameter("item").data("action");
     let sBindPath = oEv.getParameter("item").getBindingContext().getPath();
 
     if (sAction === "edit") {
@@ -238,10 +238,11 @@ sap.ui.controller("components.dataset.datasetMain", {
         onClose: function (oResponse) {
           if (oResponse === sap.m.MessageBox.Action.YES) {
             let toks = sBindPath.split("/");
-            let index = toks[toks.length - 1];
-            let currDataset = this._model.getProperty("/currentDataset");
-            let conformance = currDataset["conformance"].filter((el, ind) => ind !== parseInt(index));
-            // RuleService.editConformanceRules(currDataset.name, currDataset.version, conformance);
+            let ruleIndex = parseInt(toks[toks.length - 1]);
+            let currentDataset = this._model.getProperty("/currentDataset");
+            let newDataset = RuleService.removeRule(currentDataset, ruleIndex);
+
+            if(newDataset) DatasetService.editDataset(newDataset)
           }
         }.bind(this)
       });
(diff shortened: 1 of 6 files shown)


menas/ui/components/dataset/datasetMain.controller.js
```diff
@@ -238,10 +238,11 @@ sap.ui.controller("components.dataset.datasetMain", {
...
             let toks = sBindPath.split("/");
-            let index = toks[toks.length - 1];
-            let currDataset = this._model.getProperty("/currentDataset");
-            let conformance = currDataset["conformance"].filter((el, ind) => ind !== parseInt(index));
-            // RuleService.editConformanceRules(currDataset.name, currDataset.version, conformance);
+            let ruleIndex = parseInt(toks[toks.length - 1]);
+            let currentDataset = this._model.getProperty("/currentDataset");
+            let newDataset = RuleService.removeRule(currentDataset, ruleIndex);
+
+            if(newDataset) DatasetService.editDataset(newDataset)
```
benedeki: Why the `if` clause?
If it's to remain the please fix to adhere to coding standards:
```
if (newDataset) {
  DatasetService.editDataset(newDataset);
}
```

menas/ui/service/RuleService.js
```diff
@@ -51,30 +51,12 @@ var RuleService = new function () {
...
   };
 
-  this.disableRule = function (sId, iVersion) {
-    var uri = "api/rule/disable/" + encodeURI(sId);
-    if (typeof (iVersion) !== "undefined") {
-      uri += "/" + encodeURI(iVersion)
-    }
+  this.removeRule = function (oCurrentDataset, iRuleIndex) {
+    let conformance = oCurrentDataset["conformance"].filter((_, index) => index !== iRuleIndex);
+    let newDataset = oCurrentDataset;
```
benedeki: Question: Why the new variable declaration? I don't think the structure is copied, or is it?

menas/ui/components/dataset/datasetMain.controller.js
jozefbakus: In next iteration will removeRule return null because of validation.
Style fixed on if.

menas/ui/service/RuleService.js
jozefbakus: That is true, I did alias of that variable. It is bit confusing so I've changed it.

menas/ui/components/dataset/datasetMain.controller.js
benedeki: This future addressing development can be confusing. Not mentioning when the suggested happens it still needs to be checked if the calls handle the changed result.
Thanks for the style fix.

PR #151, opened 2019-02-06
85 Port Runs functionality over from old Menas

GeorgiChochov:
* Maintains the old Menas functionality complete with overwrites
* Introduces integration tests for the REST API and the mongo Repositories

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -7,7 +7,7 @@ pipeline {
      stages {
          stage('Build') {
              steps {
-                 sh 'mvn -B -DskipTests clean install'
+                 sh 'mvn -B clean install -Pintegration'
              }
          }
      }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -7,7 +7,7 @@ pipeline {
      stages {
          stage('Build') {
              steps {
-                 sh 'mvn -B clean install -Pintegration'
+                 sh 'mvn -B -DskipTests clean install'
              }
          }
      }
--- a/dataModel/src/main/scala/za/co/absa/enceladus/model/Run.scala
+++ b/dataModel/src/main/scala/za/co/absa/enceladus/model/Run.scala
@@ -26,5 +26,6 @@ case class Run
   splineRef: SplineReference,
   startDateTime: String,
   runStatus: RunStatus,
-  controlMeasure: ControlMeasure
+  controlMeasure: ControlMeasure,
+  username: String = ""
 )
--- a/dataModel/src/main/scala/za/co/absa/enceladus/model/Run.scala
+++ b/dataModel/src/main/scala/za/co/absa/enceladus/model/Run.scala
@@ -27,5 +27,5 @@ case class Run
   startDateTime: String,
   runStatus: RunStatus,
   controlMeasure: ControlMeasure,
-  username: String = ""
+  username: Option[String] = None
 )
--- a/menas/src/main/scala/za/co/absa/enceladus/rest/controllers/RunController.scala
+++ b/menas/src/main/scala/za/co/absa/enceladus/rest/controllers/RunController.scala
@@ -83,7 +83,11 @@ class RunController @Autowired()(runService: RunService) extends BaseController 
   }
 
   @PostMapping(Array("/updateControlMeasure/{uniqueId}"))
-  def updateControlMeasure(): CompletableFuture[ControlMeasure] = Future(null)
+  @ResponseStatus(HttpStatus.OK)
+  def updateControlMeasure(@PathVariable uniqueId: String,
+                           @RequestBody controlMeasure: ControlMeasure): CompletableFuture[Run] = {
+    runService.updateControlMeasure(uniqueId, controlMeasure)
+  }
 
   @PostMapping(Array("/updateSplineReference/{uniqueId}"))
   def updateSplineReference(): CompletableFuture[SplineReference] = Future(null)
--- a/menas/src/main/scala/za/co/absa/enceladus/rest/controllers/RunController.scala
+++ b/menas/src/main/scala/za/co/absa/enceladus/rest/controllers/RunController.scala
@@ -48,8 +48,11 @@ class RunController @Autowired()(runService: RunService) extends BaseController 
     runService.getByStartDate(startDate)
   }
 
-  @GetMapping(Array("/splineUrl/{datasetName}/{datsetVersion}/{runId}"))
-  def getSplineUrl(): CompletableFuture[String] = Future("")
+  @GetMapping(Array("/splineUrl/{datasetName}/{datasetVersion}/{runId}"))
+  def getSplineUrl(@PathVariable datasetName: String, @PathVariable datasetVersion: Int,
+                   @PathVariable runId: Int): CompletableFuture[String] = {
+    runService.getSplineUrl(datasetName, datasetVersion, runId)
+  }
 
   @GetMapping(Array("/{datasetName}/{datasetVersion}/latest"))
   def getLatestRun(): CompletableFuture[Run] = Future(null)
--- a/menas/src/main/scala/za/co/absa/enceladus/rest/controllers/RunController.scala
+++ b/menas/src/main/scala/za/co/absa/enceladus/rest/controllers/RunController.scala
@@ -67,7 +67,7 @@ class RunController @Autowired()(runService: RunService) extends BaseController 
     runService.getSplineUrl(datasetName, datasetVersion, runId)
   }
 
-  @PostMapping(Array("/create"))
+  @PostMapping()
   @ResponseStatus(HttpStatus.CREATED)
   def create(@RequestBody run: Run,
              @AuthenticationPrincipal principal: UserDetails): CompletableFuture[Run] = {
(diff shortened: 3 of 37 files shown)


menas/src/main/scala/za/co/absa/enceladus/rest/services/RunService.scala
```diff
@@ -15,14 +15,125 @@
...
+    runMongoRepository.getLatestRun(datasetName, datasetVersion).map {
+      case Some(run) => run
+      case None      => throw NotFoundException()
+    }
+  }
+
+  def getSplineUrl(datasetName: String, datasetVersion: Int, runId: Int): Future[String] = {
+    getRun(datasetName, datasetVersion, runId).map { run =>
+      val splineRef = run.splineRef
+      String.format(splineUrlTemplate, splineRef.outputPath, splineRef.sparkApplicationId)
```
Zejnilovic: What happens if the `getRun` returns nothing? Probably same as line 55-57

menas/src/main/scala/za/co/absa/enceladus/rest/services/RunService.scala
```diff
@@ -15,14 +15,125 @@
...
+      case Some(run) => run
+      case None      => throw NotFoundException()
+    }
+  }
+
+  def validate(run: Run): Future[Validation] = {
+    validateUniqueId(run)
+  }
+
+  private def validateUniqueId(run: Run) = {
```
Zejnilovic: Add return type annotations

Jenkinsfile
```diff
@@ -7,8 +7,8 @@ pipeline {
      stages {
          stage('Build') {
              steps {
-                 sh 'mvn -B -DskipTests clean install'
+                 sh 'mvn -B clean install -Pintegration'
```
Zejnilovic: I am not sure if we want this. This changes Jenkins file for everyone, so any time someone tries to build the app, it force runs integration tests. Tests should be run in parallel in stage different than build. I would create parallel stages for tests and each would run "one type". This will cut down on time needed to run tests and will give us more outputs.

menas/src/main/scala/za/co/absa/enceladus/rest/controllers/RunController.scala
```diff
@@ -18,51 +18,82 @@ package za.co.absa.enceladus.rest.controllers
...
+  def getByStartDate(@PathVariable startDate: String): CompletableFuture[Seq[Run]] = {
+    runService.getByStartDate(startDate)
+  }
 
-  @GetMapping(Array("/splineUrl/{datasetName}/{datsetVersion}/{runId}"))
-  def getSplineUrl(): CompletableFuture[String] = Future("")
+  @GetMapping(Array("/{datasetName}/{datasetVersion}/{runId}"))
+  @ResponseStatus(HttpStatus.OK)
+  def getRun(@PathVariable datasetName: String, @PathVariable datasetVersion: Int,
+             @PathVariable runId: Int): CompletableFuture[Run] = {
```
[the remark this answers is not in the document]
GeorgiChochov: Goes over the line limit and I really don't want to leave the return type hanging.

menas/src/main/scala/za/co/absa/enceladus/rest/services/RunService.scala
GeorgiChochov: `getRun` doesn't return `nothing`, it throws a `NotFoundException`. Exception gets handled by  `RestExceptionHandler` and returns 404. There's a bunch of integration tests covering this.

menas/src/main/scala/za/co/absa/enceladus/rest/controllers/RunController.scala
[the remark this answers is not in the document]
Zejnilovic: I meant like 
```
def getRun(@PathVariable datasetName: String, 
           @PathVariable datasetVersion: Int,
           @PathVariable runId: Int): CompletableFuture[Run] = {
```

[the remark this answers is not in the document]
GeorgiChochov: Ah, one per line, alright, sure.

menas/src/main/scala/za/co/absa/enceladus/rest/repositories/RunMongoRepository.scala
```diff
@@ -15,15 +15,140 @@
...
+
+  def getAllLatest(): Future[Seq[Run]] = {
+    getLatestOfEach()
+      .toFuture()
+      .map(_.map(bson => ControlUtils.fromJson[RunWrapper](bson.toJson).value))
+  }
+
+  def getByStartDate(startDate: String): Future[Seq[Run]] = {
+    getLatestOfEach()
+      .filter(regex("startDateTime", s"^$startDate\\s+"))
```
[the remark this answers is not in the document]
GeorgiChochov: Oversight, thanks for spotting it!

menas/src/main/scala/za/co/absa/enceladus/rest/services/RunService.scala
Zejnilovic: I mixed local `getRun` with `RunMongoRepository`'s `getRun`.

menas/src/main/scala/za/co/absa/enceladus/rest/repositories/RunMongoRepository.scala
```diff
@@ -15,15 +15,140 @@
...
+
+    collection
+      .mapReduce[BsonDocument](mapFn, reduceFn)
+      .finalizeFunction(finalizeFn)
+      .jsMode(true)
+  }
+
+  def getRun(datasetName: String, datasetVersion: Int, runId: Int): Future[Option[Run]] = {
+    val datasetFilter = getDatasetFilter(datasetName, datasetVersion)
+    val runIdEq = equal("runId", runId)
```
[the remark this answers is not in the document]
GeorgiChochov: It's just `Bson` for both which seems like noise to me. I think the variable names do a better job of explaining what's going on.
I'll add the explicit type if you think `Bson` would be informative though.

Jenkinsfile
GeorgiChochov: Okay, let's do this in a separate task for adjusting the Jenkinsfile.

menas/src/test/resources/application.properties
```diff
@@ -30,21 +30,23 @@ za.co.absa.enceladus.menas.auth.ldap.search.filter=(&(samAccountName={0}))
...
+za.co.absa.enceladus.menas.hadoop.auth.method=default
 za.co.absa.enceladus.menas.hadoop.auth.krb5.debug=true
 za.co.absa.enceladus.menas.hadoop.auth.krb5.realm=EXAMPLE.COM
 za.co.absa.enceladus.menas.hadoop.auth.krb5.kdc=localhost
 za.co.absa.enceladus.menas.hadoop.auth.krb5.username=hdfs@EXAMPLE.COM
 za.co.absa.enceladus.menas.hadoop.auth.krb5.keytab=hdfs.keytab
 
 za.co.absa.enceladus.menas.auth.inmemory.user=user
-za.co.absa.enceladus.menas.auth.inmemory.password=changeme
+za.co.absa.enceladus.menas.auth.inmemory.password=chang< )eme
```
jozefbakus: typo ?

GeorgiChochov: No, I wanted to have the tests run with a different password than what is default in the `template` and to include odd characters to be sure they are handled properly by the rest client.

dataModel/src/main/scala/za/co/absa/enceladus/model/Run.scala
```diff
@@ -26,5 +26,6 @@ case class Run
   splineRef: SplineReference,
   startDateTime: String,
   runStatus: RunStatus,
-  controlMeasure: ControlMeasure
+  controlMeasure: ControlMeasure,
+  username: String = ""
```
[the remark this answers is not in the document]
GeorgiChochov: That's not a bad idea, thanks!

menas/src/main/scala/za/co/absa/enceladus/rest/controllers/RunController.scala
```diff
@@ -18,51 +18,88 @@ package za.co.absa.enceladus.rest.controllers
...
+  def updateSplineReference(@PathVariable uniqueId: String,
+                            @RequestBody splineReference: SplineReference): CompletableFuture[Run] = {
+    runService.updateSplineReference(uniqueId, splineReference)
+  }
 
   @PostMapping(Array("/updateRunStatus/{uniqueId}"))
-  def updateRunStatus(): CompletableFuture[RunStatus] = Future(null)
+  @ResponseStatus(HttpStatus.OK)
+  def updateRunStatus(@PathVariable uniqueId: String,
+                      @RequestBody runStatus: RunStatus): CompletableFuture[Run] = {
```
jozefbakus: shouldn't be like this: 
`{
        "status": {
            "enumClass": "za.co.absa.atum.model.RunState",
            "value": "stageSucceeded"
        }
} `

Issue #149, opened 2019-02-06
Persist conformance rules order


jozefbakus:
Current state: All conformance rules contain -1 value in order field.
Correct state: Field order should have value dependent on order of conformance rules.

PR #161, opened 2019-02-11
#149 Persist conformance rules order

jozefbakus:
### [#149 Persist conformance rules order](https://github.com/AbsaOSS/enceladus/issues/149)
- Assign correct order number to each conformance rule

### Test plan: 
**1. Run tests:**
- go to enceladus dir
- mvn verify

**2. Create/Remove rules:**
- Run application
- Log in 
- Go to Dataset tab 
- Go to Conformance Rules
- Create at least 3 rules 
- **Expected result**: Each rule should have correct order number in DB (Create A, Create B, Create C => A.order:0, B.order:1, C.order:2). They should display in correct order.
- Delete middle rule
- **Expected result**: Each rule should have correct order number in DB (Create A, Create B, Create C => A.order:0, B.order:1, C.order:2 => Delete B => A.order:0, C.order:1). They should display in correct order.
- Delete first rule
- **Expected result**: Each rule should have correct order number in DB (Create A, Create B, Create C => A.order:0, B.order:1, C.order:2 => Delete B => A.order:0, C.order:1 => Delete A => C.order:0 ). They should display in correct order.

### DIFF
--- a/menas/ui/service/RuleService.js
+++ b/menas/ui/service/RuleService.js
@@ -55,7 +55,7 @@ var RuleService = new function () {
     let conformance = oCurrentDataset["conformance"]
       .filter((_, index) => index !== iRuleIndex)
       .sort((first, second) => first.order > second.order)
-      .map((currElement, index) =>  {return {...currElement, order: index}});
+      .map((currElement, index) => {return {...currElement, order: index}});
 
     return {...oCurrentDataset, conformance: conformance};
   };
@@ -80,7 +80,7 @@ var RuleService = new function () {
       delete oRule.joinConditions;
     }
 
-    Functions.ajax("api/dataset/" + oCurrentDataset.name + "/rule/create", "POST", oRule,
+    Functions.ajax("api/dataset/" + encodeURI(oCurrentDataset.name) + "/rule/create", "POST", oRule,
       function (oData) {
       DatasetService.getDatasetList();
       SchemaService.getSchemaVersion(oData.schemaName, oData.schemaVersion, "/currentDataset/schema");
--- a/menas/ui/service/RuleService.js
+++ b/menas/ui/service/RuleService.js
@@ -52,10 +52,12 @@ var RuleService = new function () {
   };
 
   this.removeRule = function (oCurrentDataset, iRuleIndex) {
-    let conformance = oCurrentDataset["conformance"].filter((_, index) => index !== iRuleIndex);
-    let newDataset = {...oCurrentDataset, conformance: conformance};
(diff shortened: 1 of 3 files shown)


menas/ui/service/RuleService.js
```diff
@@ -77,13 +79,12 @@ var RuleService = new function () {
       });
       delete oRule.joinConditions;
     }
-    console.log(JSON.stringify(oRule)); // TODO: remove debug
 
-    Functions.ajax("api/dataset/" + sDatasetName + "/rule/create", "POST", oRule,
+    Functions.ajax("api/dataset/" + oCurrentDataset.name + "/rule/create", "POST", oRule,
```
Zejnilovic: Not sure what dataset name can be, do we need to encode it?

GeorgiChochov: Encoding it, might not be a bad idea, any idea if ajax handles encoding automatically?

menas/ui/service/RuleService.js
```diff
@@ -52,10 +52,12 @@ var RuleService = new function () {
...
   this.removeRule = function (oCurrentDataset, iRuleIndex) {
-    let conformance = oCurrentDataset["conformance"].filter((_, index) => index !== iRuleIndex);
-    let newDataset = {...oCurrentDataset, conformance: conformance};
+    let conformance = oCurrentDataset["conformance"]
+      .filter((_, index) => index !== iRuleIndex)
+      .sort((first, second) => first.order > second.order)
+      .map((currElement, index) =>  {return {...currElement, order: index}});
 
-    return newDataset;
+    return {...oCurrentDataset, conformance: conformance};
```
GeorgiChochov: Don't know enough to make a judgement, just curious. Couldn't this be done like?
```
oCurrentDataset.conformance = conformance;
return oCurrentDataset;
```

menas/ui/service/RuleService.js
Zejnilovic: Seems like we should do it by hand. [StackOverflow](https://stackoverflow.com/questions/6544564/url-encode-a-string-in-jquery-for-an-ajax-request)

jozefbakus: Good idea.

menas/ui/service/RuleService.js
jozefbakus: This will create new object with old values except conformance.
It is more functional approach than mutate state of object. 
So yes it could be done like your suggestion. 
Should I change it ?

GeorgiChochov: Cool man, I like your approach!

PR #163, opened 2019-02-12
Feature 148 comparison job reporting nested differences

Zejnilovic:
Deep Comparison, will be done only if keys are specified.

For speed, I use `distinct` first so to see if there is even a reason for action. Then flattening and the deep comparison is done just on the remaining data.


testutils/src/main/scala/za/co/absa/enceladus/testutils/datasetComparison/Flatten.scala
```diff
@@ -0,0 +1,130 @@
...
+      * Aggregating arrays of primitives by projecting it's columns
+      *
+      * @param path path to an StructArray
+      * @param fieldNamePrefix Prefix for the field name
+      * @param structField StructField
+      * @param arrayType ArrayType
+      */
+    def flattenStructArray(path: String, fieldNamePrefix: String, structField: StructField, arrayType: ArrayType): Unit = {
+      val maxInd = df.agg(max(expr(s"size($path${structField.name})"))).collect()(0)(0).toString.toInt
+      var i = 0
```
jozefbakus: It looks like old Java way. 
You can use scalas "for loop with ranges".
It will look like: 
`for (i <- 0 to maxInd)
       doYourBusiness`
or `(0 to maxInd).map { i => doYourBusiness}`
or `(0 to maxInd).foreach { i => doYourBusiness}`
or `for{
  i <- (0 to maxInd)
} yield {
  doYourBusiness
}`

Zejnilovic: While is used because it's the fastest loop.

jozefbakus: https://github.com/scala/bug/issues/1338 scala issue for that
Why ? http://ochafik.com/p_806

testutils/src/main/scala/za/co/absa/enceladus/testutils/datasetComparison/ComparisonJob.scala
```diff
@@ -92,7 +91,7 @@ object ComparisonJob {
 
     if (errorsPresent) {
       if (cmd.keys.isDefined) {
-        handleKeyBasedDiff(cmd, expectedDf, expectedMinusActual, actualMinusExpected)
+        handleKeyBasedDiff(cmd.keys.get, cmd.outPath, expectedMinusActual, actualMinusExpected)
```
GeorgiChochov: Why not replace this `if/else` with pattern matching? `.isDefined` + `.get` is sort of an anti-pattern:
```
cmd.keys match {
  case Some(keys) => handleKeyBasedDiff(keys, cmd.outPath, expectedMinusActual, actualMinusExpected)
  case None       => handleSeparateDiff(...)
}
```

testutils/src/main/scala/za/co/absa/enceladus/testutils/datasetComparison/Flatten.scala
```diff
@@ -0,0 +1,130 @@
...
+package za.co.absa.enceladus.testutils.datasetComparison
+
+import org.apache.log4j.{LogManager, Logger}
+import org.apache.spark.sql.{Column, DataFrame}
+import org.apache.spark.sql.functions.{expr, max}
+import org.apache.spark.sql.types._
+
+import scala.collection.mutable
+
+object Flatten {
```
GeorgiChochov: I would probably call this `FlattenFuncitons` or something like that, maybe throw `helper` in the mix? Just a suggestion.

Zejnilovic: Done. I think.

Issue #168, opened 2019-02-18
IBM Date format doesn't expect FloatType


Zejnilovic:
**Describe the bug**
`TypeParser` in standardization for IBM date format only specifies what to do if the type is `DoubleType` or `DecimalType`. But there might be a chance of it being `Float`, which would then be sent with no conversion.

**Can be found in** `enceladus/standardization/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala:53`

PR #176, opened 2019-02-20
#168 Adding float type as an option for IBM timestamp input

benedeki:


benedeki:
There are unit tests covering the originally present double type. I didn't create any new ones for float type, considering it's just a type bit size difference.

GeorgiChochov:
> There are unit tests covering the originally present double type. I didn't create any new ones for float type, considering it's just a type bit size difference.

How about we have some just in case?

benedeki:
UT covering this type won't add much value. But @Zejnilovic is adding test that would capture the original bug. It's for the whole data standardization flow, not just this small component, that creates the transformation prescription.

PR #189, opened 2019-02-25
Feature/110 optimize mapping rule

yruslan:
The optimized mapping rule is an experimental feature and is disabled by default.

Also, we have to deal with Catalyst optimizer hanging issue, which I will file as a separate task (#190).

So, basically, the idea behind the optimization of mapping rule is that we can explode arrays needed for joining with mapping tables only once - before processing conformance rules. After we done with the rules we "implode" arrays back.

That's what this version is doing.

Use `--experimental-mapping-rule` command line parameter to a Conformance Spark job to turn this optimization on.


conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/rules/MappingRuleInterpreterOld.scala
```diff
@@ -0,0 +1,294 @@
...
+import za.co.absa.enceladus.utils.error._
+import za.co.absa.enceladus.model.{Dataset => ConfDataset}
+
+import scala.util.Try
+import scala.util.control.NonFatal
+import org.apache.spark.sql.Column
+import za.co.absa.enceladus.conformance.interpreter.RuleValidators
+import za.co.absa.enceladus.utils.validation._
+
+case class MappingRuleInterpreterOld(rule: MappingConformanceRule, conformance: ConfDataset) extends RuleInterpreter {
```
GeorgiChochov: Since this old one is the default, does it make sense, to leave this one as `MappingRuleInterperter` and rename the new one to `MappingRuleInterpreterExperimental`?

conformance/src/test/scala/za/co/absa/enceladus/conformance/interpreter/ArrayConformanceSuite.scala
```diff
@@ -50,7 +50,8 @@ class ArrayConformanceSuite extends FunSuite with SparkTestBase with BeforeAndAf
     val df = spark.createDataFrame(ArraySamples.testData)
     mockWhen(dao.getSchema("test", 0)) thenReturn df.schema
 
-    val conformedDf = DynamicInterpreter.interpret(ArraySamples.conformanceDef, df)(spark, dao, progArgs, false).cache()
+    val conformedDf = DynamicInterpreter.interpret(ArraySamples.conformanceDef,
+      df, experimentalMappingRule = false)(spark, dao, progArgs, enableCF = false).cache()
```
GeorgiChochov: Can we add a test case for `experimentalMappingRule = true` as well?

conformance/src/test/scala/za/co/absa/enceladus/conformance/interpreter/InterpreterSuite.scala
```diff
@@ -39,7 +40,7 @@ class InterpreterSuite extends FunSuite with SparkTestBase {
     spark.sessionState.conf.setConfString("co.za.absa.enceladus.confTest", "hello :)")
 
     implicit val dao: EnceladusDAO = mock(classOf[EnceladusDAO])
-    implicit val progArgs = CmdConfig(reportDate = "2017-11-01")
+    implicit val progArgs = CmdConfig(reportDate = "2017-11-01", experimentalMappingRule = true)
```
GeorgiChochov: Could we also have a test case for `experimentalMappingRule = false`?

conformance/src/test/scala/za/co/absa/enceladus/samples/TradeConformance.scala
```diff
@@ -0,0 +1,73 @@
...
+
+  val tradeDS = Dataset(name = "Trade Conformance", version = 1, hdfsPath = "src/test/testData/trade",
+    hdfsPublishPath = "testData/conformedTrade",
+      schemaName = "Employee", schemaVersion = 0,
+      conformance = List(countryRule, litRule, upperRule, lit2Rule, dropRule, concatRule, singleColRule) )
+
+  // The original sample used to create the parquet file
+  // import spark.implicits._
+  // val dfA = spark.read.json(sample.toDS)
+  // dfA.repartition(1).write.parquet("/Users/abri036/sources/enceladus/conformance/src/test/testData/trade/2017/11/01/")
```
GeorgiChochov: I don't think this comment belongs here, with an absolute path.

conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/DynamicInterpreter.scala
```diff
@@ -16,33 +16,39 @@
...
+    *
+    * @param conformance  The dataset object - this represents a data conformance workflow
+    * @param inputDf      The dataset to be conformed
+    * @param jobShortName A job name used for checkpoints
+    *
+    */
+  def interpret(conformance: ConfDataset,
+                inputDf: Dataset[Row],
+                jobShortName: String = "Conformance",
+                experimentalMappingRule: Boolean = true)
```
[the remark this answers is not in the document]
yruslan: Yeah, I guess so. Will remove the default value.

conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/DynamicInterpreter.scala
```diff
@@ -54,47 +60,98 @@ object DynamicInterpreter {
...
+  private def explodeAllMappingRuleArrays(inputDf: DataFrame,
+                                          steps: List[ConformanceRule]): (DataFrame, ExplosionContext) = {
+    var explodeContext: ExplosionContext = ExplosionContext()
+    var explodedDf = inputDf
+    steps.foreach {
+      case r: MappingConformanceRule =>
+        val (df, context) = ExplodeTools.explodeAllArraysInPath(r.outputColumn, explodedDf, explodeContext)
+        explodeContext = context
+        explodedDf = df
+      case _ => // Nothing to do
```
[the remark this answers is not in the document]
yruslan: `.map()` won't work here, more like `foldLeft()`. I'll think about it.

conformance/src/test/scala/za/co/absa/enceladus/samples/TradeConformance.scala
yruslan: oops

yruslan:
@Zejnilovic, nw the suggestions you and Georgi made already will keep me busy for a while :)
We don't rush to merge this, and I'd like it to be good and maintainable. Thanks for the time spent on this. So far the feedback absolutely makes sense to me.

conformance/src/test/scala/za/co/absa/enceladus/conformance/interpreter/InterpreterSuite.scala
yruslan: Agree

conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/rules/MappingRuleInterpreterOld.scala
yruslan: Yes, this will make much more sense

yruslan: Saša suggested that the experimental class will become the production one sooner or later. Then the class should be renamed again. So maybe the better approach is to rename the classes according to semantics of what they do, like `MappingRuleInterpreter` and `MappingRuleInterpreterNoExplode`. The "no explode" version is the new one because it does not explode, but expects the schema to be already exploded.

conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/rules/MappingRuleInterpreterNoExplode.scala
```diff
@@ -0,0 +1,275 @@
...
+    val mappingErrUdfCall = callUDF("confMappingErr", lit(rule.outputColumn),
+      array(rule.attributeMappings.values.toSeq.map(arrCol(_).cast(StringType)): _*),
+      typedLit(mappings))
+
+    val placedDf = ExplodeTools.nestedRenameReplace(joined, rule.outputColumn, rule.outputColumn)
+
+    val defaultMappingValue = if (defaultMappingValueMap.contains(rule.targetAttribute)) {
+      Some(defaultMappingValueMap(rule.targetAttribute))
+    } else {
+      None
```
Zejnilovic: Why not just use
```scala
    val defaultMappingValue: Option[String] = defaultMappingValueMap.get(rule.targetAttribute)
```

Get returns None if not found.

conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/rules/MappingRuleInterpreterNoExplode.scala
```diff
@@ -0,0 +1,275 @@
...
+  // Flattens a schema for join validation purposes.
+  private[rules] def flattenForJoin(schema: StructType, prefix: Option[String] = None): Seq[StructField] = {
+    schema.fields.flatMap(field => {
+      val fieldName = prefix.getOrElse("") + field.name
+      field.dataType match {
+        case s: StructType =>
+          val fld = field.copy(name = fieldName)
+          Seq(fld) ++ flattenForJoin(s, Some(fieldName + "."))
+        case _ =>
+          val fld = field.copy(name = fieldName)
```
Zejnilovic: this line is the same as 250, we could move it up.

utils/src/main/scala/za/co/absa/enceladus/utils/explode/ExplodeTools.scala
```diff
@@ -0,0 +1,390 @@
...
+    }
+  }
+
+  private def getRootLevelPrefix(fieldName: String, prefix: String, schema: StructType): String = {
+    getClosestUniqueName(s"${fieldName}_$prefix", schema)
+      .replaceAll("\\.", "_")
+  }
+
+  private def extructFieldFromStruct(df: DataFrame, structFieldName: String): (DataFrame, Column) = {
+    val tmpColName = getUniqueName("tmp_ext", Some(df.schema))
```
Zejnilovic: This method is never used
 \+ I think there is a typo. I think it should be `extractFieldFromStruct`

utils/src/main/scala/za/co/absa/enceladus/utils/schema/SchemaUtils.scala
```diff
@@ -144,6 +144,60 @@ object SchemaUtils {
...
+    */
+  def getDeepestCommonArrayPath(schema: StructType, fieldPaths: Seq[String]): Option[String] = {
+    val arrayPaths = fieldPaths.flatMap(path => getAllArraysInPath(path, schema)).distinct
+
+    if (arrayPaths.isEmpty) {
+      None
+    } else {
+      if (isCommonSubPath(arrayPaths: _*)) {
+        Some(arrayPaths.maxBy(_.length))
+      } else {
```
Zejnilovic: We could switch this double if for more readable one
```scala
if (arrayPaths.isEmpty && isCommonSubPath(arrayPaths: _*)) {
  Some(arrayPaths.maxBy(_.length))
} else {
  None
}
```

utils/src/main/scala/za/co/absa/enceladus/utils/schema/SchemaUtils.scala
```diff
@@ -144,6 +144,60 @@ object SchemaUtils {
...
+    * @param paths A list of paths to be analyzed
+    * @return true if for all pathe the above property holds
+    */
+  def isCommonSubPath(paths: String*): Boolean = {
+    def sliceRoot(paths: Seq[Seq[String]]): Seq[Seq[String]] = {
+      paths.map(path => path.drop(1)).filter(_.nonEmpty)
+    }
+
+    var isParentCommon = true // For Seq() the property holds by [my] convention
+    var restOfPaths = paths.map(_.split('.').toSeq).filter(_.nonEmpty)
```
Zejnilovic: Maybe it is just me but I find this to need annotations, so 
```scala
var restOfPaths: Seq[Seq[String]] = paths.map(_.split('.').toSeq).filter(_.nonEmpty)
```

utils/src/main/scala/za/co/absa/enceladus/utils/explode/ExplodeTools.scala
```diff
@@ -0,0 +1,390 @@
...
+    * @param inputDf    A dataframe to process
+    * @param columnName A nested column to process
+    * @return A transformed dataframe
+    **/
+  def deconstructNestedColumn(inputDf: DataFrame, columnName: String): DeconstructedNestedField = {
+    var transientColName: Option[String] = None
+    def processStruct(schema: StructType, path: Seq[String], parentCol: Option[Column]): Seq[Column] = {
+      val currentField = path.head
+      val isLeaf = path.lengthCompare(1) <= 0
+      val newFields = schema.fields.flatMap(field => {
```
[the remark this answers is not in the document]
yruslan: Possibly, both. Still looks more readable for me like that. What do you think?

utils/src/main/scala/za/co/absa/enceladus/utils/explode/ExplodeTools.scala
```diff
@@ -0,0 +1,390 @@
...
+          .otherwise(when(col(explosion.sizeFieldName) === 0, typedLit(Array())).otherwise(null))
+      )
+
+    val dfArraysRestored = nestedRenameReplace(nullsRestored, tmpColName2, explosion.arrayFieldName,
+      transientColumn)
+
+    val dfTransientRestored = explosion.superTransientFieldName match {
+      case None => dfArraysRestored
+      case Some(transientField) =>
+        DeepArrayTransformations.nestedDropColumn(dfArraysRestored, transientField)
```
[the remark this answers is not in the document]
yruslan: I guess it's like a microoptimization. For some reason I think that checking against concrete value (None) is faster than matching unapply (Some(x)). So I make faster check first.
But don't insist on that. If checking for Some(x) first is more idiomatic I can go for it.

utils/src/main/scala/za/co/absa/enceladus/utils/explode/ExplodeTools.scala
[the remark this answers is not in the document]
Zejnilovic: I am all up for readability 👍 
I would leave it

conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/rules/MappingRuleInterpreterOld.scala
GeorgiChochov: That makes sense.

PR #192, opened 2019-02-25
191 Add feature request template

GeorgiChochov:


.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,21 @@
...
+## Background
+A clear and concise description of where the limitation lies.
+
+## Feature
+A description of the requested feature.
+
+### *Example:*
+A simple example if applicable.
+
+## Proposed Solution
```
benedeki: I would mark this section as optional (maybe even the "Example" section).

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -1,29 +1,30 @@
...
-**To Reproduce**
+## To Reproduce
 Steps to reproduce the behavior OR commands run:
 1. Go to '...'
 2. Click on '....'
 3. Enter value '...'
 4. See error
 
-**Expected behavior**
+## Expected behavior
```
Zejnilovic: Can we change this to behaviour? My spellcheck is going nuts today. 😄

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,21 @@
...
+
+---
+
+## Background
+A clear and concise description of where the limitation lies.
+
+## Feature
+A description of the requested feature.
+
+### *Example:* [Optional]
```
Zejnilovic: I am ok with optional, but I think it is not needed. What is weird me though is that example is surrounded by ** and the proposed solution isn't and that it is on a different level. Why?

.github/ISSUE_TEMPLATE/bug_report.md
GeorgiChochov: Fair enough

.github/ISSUE_TEMPLATE/feature_request.md
GeorgiChochov: Good point, I wanted to place the `Example` as a subsection of the `Feature` section, but it just looks a bit odd, I guess.

Issue #80, opened 2018-12-28
Update confromance rules


GeorgiChochov:
Users should be able to update conformance rules through the Menas UI. This includes opening a dialog like the rule creation one but with the rule type uneditable.

PR #197, opened 2019-02-27
#80 Update conformance rules ui implementation

jozefbakus:
### [#80 Update conformance rules ui implementation](https://github.com/absaoss/enceladus/issues/80)
Be gentle! :D 
### Changes: 
- Update conformance rules ui implementation (New feature)
- A lot of small fixes and refactorings:
  - Update/Create conformance rules code moved to fragment's controller
  - ConcatenationConformanceRule and MappingConformanceRule fragment redesigne
    - List is used instead of in code generated fragment parts
    - New button with remove input fields functionality
  - Removed rule specific logic from RuleService
  - Others ...

### Test plan: 
**1. Run tests:**
- go to enceladus dir
- mvn verify

**2. Create rule for each rule type:**
- Run application
- Log in 
- Go to Dataset tab 
- Go to Conformance Rules
- Create rule 
- **Expected result**: 
  - Rule should be created/displayed with correct values. 

**2. Update rule (Test with each rule type):**
- Run application
- Log in 
- Go to Dataset tab 
- Go to Conformance Rules
- Click on update conformance rule button
- **Expected result**: 
  - Update fragment should be loaded with current values
  - Rule type select should be disabled
- Save changes
- **Expected result**: 
  - Rule should be updated/displayed with correct values.

### DIFF
--- a/menas/ui/components/dataset/conformanceRule/upsert.controller.js
+++ b/menas/ui/components/dataset/conformanceRule/upsert.controller.js
@@ -1,0 +1,265 @@
+/*
+ * Copyright 2018-2019 ABSA Group Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+sap.ui.define([
+  "sap/ui/core/mvc/Controller"
+], function (Controller) {
+  return Controller.extend("components.dataset.conformanceRule.upsert", {
+
+    rules: [
+      {_t: "CastingConformanceRule"},
+      {_t: "ConcatenationConformanceRule"},
+      {_t: "DropConformanceRule"},
+      {_t: "LiteralConformanceRule"},
+      {_t: "MappingConformanceRule"},
+      {_t: "NegationConformanceRule"},
+      {_t: "SingleColumnConformanceRule"},
+      {_t: "SparkSessionConfConformanceRule"},
+      {_t: "UppercaseConformanceRule"}
+    ],
+
+    dataTypes: [
+      {type: "boolean"},
+      {type: "byte"},
+      {type: "short"},
+      {type: "integer"},
+      {type: "long"},
+      {type: "float"},
+      {type: "double"},
+      {type: "decimal(38,18)"},
+      {type: "char"},
+      {type: "string"},
+      {type: "date"},
+      {type: "timestamp"}
+    ],
+
+    constructor: function () {
+      this._model = sap.ui.getCore().getModel();
+      this._model.setProperty("/rules", this.rules);
+      this._model.setProperty("/dataTypes", this.dataTypes);
+      this._formFragments = {};
+    },
+
+    onBeforeOpen: function () {
+      this._dialog = sap.ui.getCore().byId("upsertConformanceRuleDialog");
+      this._ruleForm = sap.ui.getCore().byId("ruleForm");
+      MappingTableService.getMappingTableList(true, true);
+      if(this._model.getProperty("/newRule/isEdit")) {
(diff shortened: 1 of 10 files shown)


menas/ui/components/dataset/conformanceRule/upsert.controller.js
```diff
@@ -0,0 +1,265 @@
...
+      } else {
+        this._model.setProperty("/newRule/_t", this.rules[0]._t);
+        this.showFormFragment(this.rules[0]._t);
+      }
+      this._dialog.setEscapeHandler(() => this.onClosePress());
+    },
+
+    onAfterOpen: function () {
+      let schemaFieldSelectorSupportedRules =
+        ["CastingConformanceRule", "NegationConformanceRule", "SingleColumnConformanceRule", "UppercaseConformanceRule"];
```
Zejnilovic: The above-mentioned rules could have an attribute `schemaFieldSelectorSupportedRules` which would be false/true. It would be globally accessible in case of future needs.

menas/ui/components/dataset/conformanceRule/upsert.controller.js
```diff
@@ -0,0 +1,265 @@
...
+        newRule.attributeMappings = {};
+        newRule.newJoinConditions.map(function(joinCondition) {
+          newRule.attributeMappings[joinCondition.mappingTableField] = joinCondition.datasetField
+        });
+        delete newRule.joinConditions;
+      }
+    },
+
+    addRule: function (currentDataset, newRule) {
+      currentDataset.conformance[newRule.order] = newRule;
```
GeorgiChochov: Pretty sure `updateRule` and `addRule` are the same function. It makes sense to have the logical separation fo the two functions, so we can use it to distinguish the two intended behaviours. This distinction will allow us to more easily refactor further, if we decide to slim down the front-end (I suggested we fatten it, but not so sure about that anymore).
The only change I would make here is, for now, to call `updateRule` from `addRule`, so when we change the front-end logic for `addRule` nothing else is affected.

menas/ui/components/dataset/conformanceRule/upsert.controller.js
jozefbakus: Fixed

menas/ui/components/dataset/conformanceRule/upsert.controller.js
```diff
@@ -0,0 +1,265 @@
...
+              datasetField: oAttributeMappings[key]
+            });
+          }
+          newRule.newJoinConditions = aNewJoinConditions;
+        }
+        MappingTableService.getAllMappingTableVersions(newRule.mappingTable, sap.ui.getCore().byId("schemaVersionSelect"));
+      }
+
+      if(!currentRule.isEdit)
+        newRule.order = this._model.getProperty("/currentDataset").conformance.length;
```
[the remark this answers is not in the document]
jozefbakus: Fixed

menas/ui/components/dataset/conformanceRule/upsert.controller.js
```diff
@@ -0,0 +1,265 @@
...
+      this.beforeShowFragmentChanges();
+    },
+
+    getFormFragment: function (sFragmentName) {
+      let oFormFragment = this._formFragments[sFragmentName];
+      if (oFormFragment) {
+        return oFormFragment;
+      }
+
+      oFormFragment = sap.ui.xmlfragment( sFragmentName,"components.dataset.conformanceRule." + sFragmentName + ".add", this);
```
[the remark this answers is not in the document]
jozefbakus: Fixed. Sorry for that but I haven't seen that because IndelliJ puts there parameter names.

menas/ui/components/dataset/conformanceRule/upsert.controller.js
```diff
@@ -0,0 +1,265 @@
...
+          }
+        })
+      });
+      this._ruleForm.removeAllContent();
+    },
+
+    preselectSchemaFieldSelector: function(ruleType) {
+      let sExpandTo= this._model.getProperty("/newRule/inputColumn");
+      let aTokens = sExpandTo.split(".");
+      let oCtl = sap.ui.getCore().byId(ruleType+"--schemaFieldSelector");
```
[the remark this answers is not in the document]
jozefbakus: Fixed.

menas/ui/components/dataset/conformanceRule/upsert.controller.js
jozefbakus: Fixed.

PR #132, opened 2019-01-24
Feature/29 audit trails

lokm01:
This PR implements the Audit Trails

PR #211, opened 2019-03-05
121 Show Used In dependencies on failed delete

GeorgiChochov:


menas/ui/service/MappingTableService.js
```diff
@@ -170,8 +170,11 @@ var MappingTableService = new function () {
       if (xhr.status === 400) {
         let err = "Disabling mapping table failed. Clear the following dependencies first:\n";
         let oData = JSON.parse(xhr.responseText);
-        for(let ind in oData) {
-          err += "\t - " + oData[ind].name + " (v. " + oData[ind].version + ")";
+
+        err += "Datasets:\n";
+        let datasets = oData["datasets"];
+        for(let ind in datasets) {
+          err += "- " + datasets[ind].name + " (v. " + datasets[ind].version + ")";
```
Zejnilovic: Please add new lines and maybe some tabs. This is not readable enough. Maybe organize them in a table?

This counts for others additions in this PR too.

screenshot 2019-03-05 at 17 58 37

GeorgiChochov: That's spot on, I completely missed I hadn't added the new line. Thanks!
Not sure what you mean by `This counts for others additions in this PR too.`, though.

Zejnilovic: There are 3 occurrences of the same error. Schemas, Datasets and MappingTables dependencies

GeorgiChochov: Ah, that was implicitly understood.

menas/ui/service/MappingTableService.js
```diff
@@ -170,8 +170,11 @@ var MappingTableService = new function () {
       if (xhr.status === 400) {
         let err = "Disabling mapping table failed. Clear the following dependencies first:\n";
         let oData = JSON.parse(xhr.responseText);
-        for(let ind in oData) {
-          err += "\t - " + oData[ind].name + " (v. " + oData[ind].version + ")";
+
+        err += "Datasets:\n";
+        let datasets = oData["datasets"];
+        for(let ind in datasets) {
+          err += "- " + datasets[ind].name + " (v. " + datasets[ind].version + ")\n";
```
Zejnilovic: If I understand this correctly, I should not be able to delete a mapping table if there is a dataset that is using it. 

I was able to delete a mapping table while the dataset was using it.

GeorgiChochov: Thanks for catching that! I'll prioritize the REST API tests to reduce the odds of things like this happening.

GeorgiChochov: Let's log this as a separate issue, it's a back-end problem that I think is addressed in #132 and we can revisit it after.

menas/ui/service/MappingTableService.js
yruslan: No sure I understand what the error message is trying to tell me. Why disabling failed and what do I need to do to "clear" the dependencies. Consider rephrasing/simplifying the message text.

Also, the list of entities (datasets and versions I presume) seems quite long. Do we need to list everything?

menas/ui/service/SchemaService.js
```diff
@@ -103,8 +103,21 @@ var SchemaService = new function() {
...
+
+        let datasets = oData["datasets"];
+        if (datasets.length !== 0) {
+          err += "Datasets:\n";
+          for (let ind in datasets) {
+            err += "- " + datasets[ind].name + " (v. " + datasets[ind].version + ")\n";
+          }
+        }
+
+        let mappingTables = oData["mappingTables"];
```
yruslan: This repeats lines 107-113 almost exactly. Consider combining the blocks in a function.

yruslan: Also, similar code is in MappingTableService.js:175. Looks like this could be a library function.

GeorgiChochov: Fair point, need to do a lot of refactoring in these services.

menas/ui/service/MappingTableService.js
GeorgiChochov: Good point, I think I'll reduce it to just the names of the dependencies and ask users to check the `Used In` tab for details (which will be formatted better in #213)

Zejnilovic:
There is a dash with empty line for no dependencies

image

Zejnilovic:
Sort the versions before printing them. Either by last update or name.

image

Zejnilovic:
I am missing A8 and A9 in the list but there is no one more.

image

GeorgiChochov:
> Looks great! The only thing is that there seems a merge conflict.

It's a rebase conflict, merge is fine (it will have to be one of the times we don't rebase and merge a PR, I guess)

PR #212, opened 2019-03-05
201 Order Used In entities ASC by name and version

GeorgiChochov:


GeorgiChochov:
> How about in future we let the user choose between ASC and DESC?

Yeah, that can be done entirely on the front-end, as well as grouping the dependencies by name, and allowing users to expand to see the versions. I'll add a task for it.

Issue #215, opened 2019-03-06
Enceladus build should fail if application.properties is not found


yruslan:
## Describe the bug
When running a Standardization job an exception is thrown
`
Exception in thread "main" com.typesafe.config.ConfigException$Missing: No configuration setting found for key 'enceladus'
	at com.typesafe.config.impl.SimpleConfig.findKeyOrNull(SimpleConfig.java:152)
	at com.typesafe.config.impl.SimpleConfig.findKey(SimpleConfig.java:145)
	at com.typesafe.config.impl.SimpleConfig.findOrNull(SimpleConfig.java:172)
	at com.typesafe.config.impl.SimpleConfig.find(SimpleConfig.java:184)
	at com.typesafe.config.impl.SimpleConfig.find(SimpleConfig.java:189)
	at com.typesafe.config.impl.SimpleConfig.getString(SimpleConfig.java:246)
	at za.co.absa.enceladus.standardization.StandardizationJob$.addPerformanceMetadata(StandardizationJob.scala:52)
	at 
`
This is caused by this line:
```scala
val enceladusVersion = conf.getString("enceladus.version")
```

## To Reproduce
- Clone fresh Enceladus
- Build the project via `mvn clean package`
- Run a Standardization job and wait for the exception.

## Expected behaviour
When building Enceladus it is required to setup `application.conf` for both Standardization and Conformance modules.
Maven build should fail if `application.conf` is not found at expected places.

## Desktop (please complete the following information):
 - OS: Linux lab cluster

PR #217, opened 2019-03-06
#215 Make Standardization and Conformance build fail if there is no application.conf

yruslan:
This makes build fail if there are no 'application.conf' inside Standardization and Conformance resources folder.

This is likely to require changes on Jenkins side.

### DIFF
--- a/conformance/pom.xml
+++ b/conformance/pom.xml
@@ -129,6 +129,35 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-enforcer-plugin</artifactId>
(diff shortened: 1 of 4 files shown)


yruslan:
For Standardization and Conformance the file name is 'application.conf' while for Menas it is 'application.properties'. Should we make 'application.properties' for every module?

GeorgiChochov:
> For Standardization and Conformance the file name is 'application.conf' while for Menas it is 'application.properties'. Should we make 'application.properties' for every module?

The difference is due to the framework reading the configuration files (`typesafe config` uses `.conf` while `spring` uses `.properties`). I'm sure we can align this better, but let's think about standardizing the configuration files in a separate PR.

yruslan:
This should be fixed. Did this for Menas as well.

Issue #228, opened 2019-03-08
Menas to StructField converter mishandles nested arrays


yruslan:
## Describe the bug
When Menas sends a schema to Standardization it
1. Gets the schema from the persistence layer (MongoDB)
2. Converts the json to the Menas schema model
3. Converts Menas schema to Spark's `Seq[StructField]`

At the step 3 all nested arrays of structs become arrays of `decimal(38,18)`.

## To Reproduce
Run a Standardization job using a dataset which has a schema containing a nested array of structs.

## Expected behaviour
All nested arrays of structs should be retained in StructFields

yruslan:
I confirm the fix. Have ran a Standardization job that failed before and the schema is proper now.

PR #229, opened 2019-03-08
#228 Fix nested arrays conversion from menas schema type to Spark Struct

yruslan:
This passes the fixed unit tests, which covers all of the possible cases.

### DIFF
--- a/menas/src/main/scala/za/co/absa/enceladus/rest/utils/converters/SparkMenasSchemaConvertor.scala
+++ b/menas/src/main/scala/za/co/absa/enceladus/rest/utils/converters/SparkMenasSchemaConvertor.scala
@@ -24,15 +24,32 @@ import org.springframework.stereotype.Component
 import java.io.ByteArrayOutputStream
 
 @Component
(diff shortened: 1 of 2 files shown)


menas/src/main/scala/za/co/absa/enceladus/rest/utils/converters/SparkMenasSchemaConvertor.scala
```diff
@@ -24,17 +24,34 @@ import org.springframework.stereotype.Component
...
 
+  /**
+    * Converts a menas array to a spark array
+    */
+  def convertMenasToSparkFields(arrayField: SchemaField): ArrayType = {
+    if (arrayField.`type` != "array") {
+      throw new IllegalStateException(s"An array is expected.")
+    }
+    arrayField.elementType match {
+      case Some("struct") => ArrayType.apply(StructType(convertMenasToSparkFields(arrayField.children)))
```
GeorgiChochov: No need to call `.apply` explicitly.

Issue #233, opened 2019-03-08
Spline initialization failed 


GeorgiChochov:
## Describe the bug
Running Standardization of Conformance fails to initialize Spline. As a result jobs run without lineage tracking.

## To Reproduce
Run any standardization or conformance job with the default spline.properties provided in the `spline.properties.template`

## Expected behaviour
Spline Lineage tracking should be enabled

## Screenshots
![image](https://user-images.githubusercontent.com/20905867/54037427-683e0f00-41be-11e9-81f8-26bcb420e4a2.png)

PR #234, opened 2019-03-08
#233 Fix Spline initialization

GeorgiChochov:
There was a superfluous `mongo-scala-driver` dependency in the `dataModel` module, which conflicted with the `mongo-java-driver` used by spline.
Many thanks to @wajda and @yruslan for the help in figuring this out!

### DIFF
--- a/dataModel/pom.xml
+++ b/dataModel/pom.xml
@@ -38,11 +38,6 @@
             <artifactId>jackson-module-scala_${scala.compat.version}</artifactId>
             <version>${jackson.version}</version>
         </dependency>
-        <dependency>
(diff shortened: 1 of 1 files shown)


Zejnilovic:
clean package, run menas, run standartization and I still get the same error

yruslan:
@Zejnilovic, let's double check that together when we have a chance. It worked perfectly fine after the fix. Unfortunately, I don't have Menas deployed locally (which I think I should do at some point in time).

Zejnilovic:
> These changes can be merged even if they do not fix all related issues.

Agreed. Deletion of useless dependency is still a step forward.

PR #236, opened 2019-03-12
Epic/175 better timestamp handling

benedeki:


benedeki:
There will be more commits related to other tasks in the Epic.

benedeki:
It has been brought to my attention this PR is unnecessary big, which is true.
I will split it by tasks/commits. Please don't bother reviewing this one. 😄

PR #240, opened 2019-03-13
#175 Fixes the build Enceladus on Windows

benedeki:


utils/src/main/scala/za/co/absa/enceladus/utils/error/UDFLibrary.scala
```diff
@@ -46,10 +44,11 @@ case class UDFLibrary()(implicit val spark: SparkSession) {
     ErrorMessage.confNegErr(errCol, rawValue)
   })
 
-  spark.udf.register("arrayDistinctErrors", (arr: mutable.WrappedArray[ErrorMessage]) => arr.distinct.filter((a: AnyRef) => a != null))
+  spark.udf.register("arrayDistinctErrors",
+    (arr: mutable.WrappedArray[ErrorMessage]) => arr.distinct.filter((a: AnyRef) => a != null))
 
-  val cleanErrCol = new UDF1[Seq[Row], Seq[Row]] {
-    override def call(t1: Seq[Row]) = {
+  private val cleanErrCol: UDF1[Seq[Row], Seq[Row]] = new UDF1[Seq[Row], Seq[Row]] {
```
GeorgiChochov: I'd leave this up to type inference, since on the right hand side we have the exact same type information with `new UDF1[Seq[Row], Seq[Row]]`

utils/src/main/scala/za/co/absa/enceladus/utils/error/UDFLibrary.scala
```diff
@@ -61,12 +60,12 @@ case class UDFLibrary()(implicit val spark: SparkSession) {
 
   spark.udf.register("cleanErrCol", cleanErrCol, ArrayType.apply(ErrorMessage.errorColSchema, false))
 
-  val errorColumnAppend = new UDF2[Seq[Row], Row, Seq[Row]] {
+  private val errorColumnAppend: UDF2[Seq[Row], Row, Seq[Row]] = new UDF2[Seq[Row], Row, Seq[Row]] {
```
GeorgiChochov: Same here

Issue #154, opened 2019-02-07
Configure Menas logging with log4j2


GeorgiChochov:
We should have a basic log4j2 configuration for handling logging in Menas.

PR #241, opened 2019-03-14
Feature/156 configuration of service time zone

benedeki:
The idea is to set every application/service/test to time zone as specified in `application.conf` defaulting to GMT (and preferably not other). Also Spark instances are to have their time zone changed accordingly.


utils/src/main/scala/za/co/absa/enceladus/utils/time/TimeZoneNormalizer.scala
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  def normalizeSessionTimeZone(spark: SparkSession): Unit = {
+    spark.conf.set("spark.sql.session.timeZone", timeZone)
+    log.debug(s"Spark session time zone of name ${spark.sparkContext.appName} set to $timeZone")
+  }
+
+  // scalastyle:off null
+  // Not nice, but we want to use null here in case no Spark session is available
+  def normalizeTimezone()(implicit spark: SparkSession = null): Unit = {
```
GeorgiChochov: Why give the `SparkSession` a default value? 
I think that implicit params should not have default values, since that further hides their implicit requirements.
Then you wouldn't need to handle `null` cases.

utils/src/main/scala/za/co/absa/enceladus/utils/time/TimeZoneNormalizer.scala
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+  * Sets the system time zone per application configuration, recommended value being UTC
+   */
+object TimeZoneNormalizer {
+  private val log: Logger = LogManager.getLogger(this.getClass)
+  private val timeZone: String = getConf("timezone", "UTC")
+  private var justBuilt: Boolean = false
+
+  setJVMTimeZone()
```
GeorgiChochov: why is this being called in the object's initialization block?

utils/src/main/scala/za/co/absa/enceladus/utils/error/UDFLibrary.scala
```diff
@@ -46,10 +44,11 @@ case class UDFLibrary()(implicit val spark: SparkSession) {
     ErrorMessage.confNegErr(errCol, rawValue)
   })
 
-  spark.udf.register("arrayDistinctErrors", (arr: mutable.WrappedArray[ErrorMessage]) => arr.distinct.filter((a: AnyRef) => a != null))
+  spark.udf.register("arrayDistinctErrors",
+    (arr: mutable.WrappedArray[ErrorMessage]) => arr.distinct.filter((a: AnyRef) => a != null))
 
-  val cleanErrCol = new UDF1[Seq[Row], Seq[Row]] {
-    override def call(t1: Seq[Row]) = {
+  private val cleanErrCol: UDF1[Seq[Row], Seq[Row]] = new UDF1[Seq[Row], Seq[Row]] {
```
GeorgiChochov: I remember this was changed in #240, can you rebase this branch onto develop?

utils/src/main/scala/za/co/absa/enceladus/utils/time/TimeZoneNormalizer.scala
```diff
@@ -0,0 +1,69 @@
...
+      config.getString(path)
+    } else {
+      log.warn(s"No time zone (TimeZone) setting found. Setting to default, which is $default.")
+      default
+    }
+  }
+
+  private def setJVMTimeZone(): Unit = {
+    if (justBuilt) {
+      justBuilt = false
```
GeorgiChochov: I don't understand what this method is doing. It is sets the timezone once on initialization, and prevents doing it again the second time it is called, but resets the `justBuilt` flag to false (which never gets sets to true again) and I can keep setting the timezone as many times as I want after that. 
Ultimately calling it more than once changes nothing, since every time I call this method it always sets the timezone to `UTC`, so I don't know why we are trying to prevent repeated calls. Is `TimeZone::setDefault` a heavy operation?

examples/data/.gitignore
```diff
@@ -0,0 +1 @@
+output
```
GeorgiChochov: This is interesting, I guess you ned to ignore some output from the examples (but I can't find where)? 
I would expect this at the root of the module then.

conformance/src/main/resources/log4j.properties
```diff
@@ -0,0 +1,21 @@
+#
```
GeorgiChochov: I'm not sure about overriding the default logging configuration of spark. 
@lokm01 @yruslan thoughts?

testutils/src/main/resources/application.conf.template
```diff
@@ -0,0 +1,16 @@
+#
```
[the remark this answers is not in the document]
benedeki: At the moment I don't think so (only tests). But if it would be added, it would be obvious this setting needs to be added.

examples/data/.gitignore
benedeki: The _CustomRuleSample4_ if run with the parameters from `README.md`creates the files.
But you bringing this to attention, I wonder if it's needed all that much (doubting my own decision 😉 )
On general note though - you would prefer the subdirectory to be ignored form the module root instead of locally?

conformance/src/main/resources/log4j.properties
benedeki: if this file ins not present, then `Logger` used in `TimeZoneNormalizer` fires warning on start-up.

conformance/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
```diff
@@ -38,67 +38,95 @@ import scala.util.control.NonFatal
...
-
     // die before performing any computation if the output path already exists
-    if (FileSystemVersionUtils.exists(publishPath)) throw new IllegalStateException(s"Path $publishPath already exists. Increment the run version, or delete $publishPath")
-
+    if (FileSystemVersionUtils.exists(publishPath)) {
+      throw new IllegalStateException(s"Path $publishPath already exists. Increment the run version, or delete $publishPath")
+    }
     // init performance measurer
-    val performance = new PerformanceMeasurer(spark.sparkContext.appName)
+    val performance: PerformanceMeasurer = new PerformanceMeasurer(spark.sparkContext.appName)
```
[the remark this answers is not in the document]
benedeki: Will remove

utils/src/main/scala/za/co/absa/enceladus/utils/time/TimeZoneNormalizer.scala
GeorgiChochov: My argument is still that `implicit parameters` should not have `default values` because it obscures their presence and makes it easy to have misbehaving code.

examples/data/.gitignore
GeorgiChochov: I think a `.gitignore` file belongs at the root of a project. If for whatever reason it's necessary to specifically ignore something int a submodule, that should ideally be in the general `.gitignore` pointing to the submodule or at the root of the specifc module. I've never seen `.gitignore` being placed in other places, I don't even know if that works. I know I'd have a hard time finding out why something is ignored if it's not in the root-level `.gitignore`

utils/src/main/scala/za/co/absa/enceladus/utils/time/TimeZoneNormalizer.scala
benedeki: In general I agree.
But here it's deliberate. The idea being: _"Normalize" as much time zone settings as possible wihtout the least possible hassle and need of revisit in case of code changes._

standardization/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
```diff
@@ -35,87 +35,105 @@ import za.co.absa.enceladus.utils.fs.FileSystemVersionUtils
...
       dfReaderConfigured
     }).load(s"$path/*")
+    ensureSplittable(dfWithSchema, path, schema)
+  }
 
-    val dfAll = ensureSplittable(dfWithSchema, path, schema)
-
-    val rawDirSize = FileSystemVersionUtils.getDirectorySize(path)
-
+  private def process(performance: PerformanceMeasurer, dfAll: DataFrame, schema: StructType, cmd: CmdConfig,
```
Zejnilovic: I would probably choose a different name. `process` doesn't tell much.

standardization/src/test/resources/application.conf
```diff
@@ -11,3 +11,5 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+#system-wide time zone
+timezone="UTC"
```
Zejnilovic: missing new line end of file

testutils/src/main/resources/application.conf.template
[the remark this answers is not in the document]
Zejnilovic: I am very happy that @benedeki is thinking about this.

Right now, there is no config file in use. But if we are normalizing time across the project, `TestUtils` shouldn't be left without a change, they are still spark jobs which should work in unison with the others.

conformance/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
```diff
@@ -38,67 +38,95 @@ import scala.util.control.NonFatal
...
+      stdPath, publishPath, LoggedInUserInfo.getUserName)
 
-    spark.enableControlMeasuresTracking()
-      .setControlMeasuresWorkflow("Conformance")
+    withPartCols.writeInfoFile(publishPath)
+    cmd.performanceMetricsFile.foreach(fileName => {
+      try {
+        performance.writeMetricsToFile(fileName)
+      } catch {
+        case NonFatal(_) => log.error(s"Unable to write performance metrics to file '$fileName'")
```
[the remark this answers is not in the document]
benedeki: Will add the exception message.

standardization/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
```diff
@@ -35,87 +35,105 @@ import za.co.absa.enceladus.utils.fs.FileSystemVersionUtils
...
     val performance = new PerformanceMeasurer(spark.sparkContext.appName)
+    val dfAll: DataFrame = prepareDataFrame(schema, cmd, path)
 
-    val dfReader = spark.read.format(cmd.rawFormat)
+    process(performance, dfAll, schema, cmd, path, stdPath)
+    cmd.performanceMetricsFile.foreach(fileName => {
+      try {
+        performance.writeMetricsToFile(fileName)
+      } catch {
+        case NonFatal(_) => log.error(s"Unable to write performance metrics to file '$fileName'")
```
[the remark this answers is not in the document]
benedeki: Will add the exception message.

standardization/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
```diff
@@ -35,87 +35,105 @@ import za.co.absa.enceladus.utils.fs.FileSystemVersionUtils
...
+      }
+    })
+  }
+
+  private def obtainSparkSession(cmd: CmdConfig): SparkSession = {
```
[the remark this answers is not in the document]
benedeki: Moving private functions bellow `main`

menas/src/main/resources/log4j.properties
```diff
@@ -0,0 +1,21 @@
+#
```
GeorgiChochov: Logging for Menas will be configured separately (#154). I don't think using the spark default configuration from `logj4.properties.template` is appropriate for our webapp.

examples/data/.gitignore
benedeki: Removing the file (forgot to do it in previous commits)

utils/src/main/scala/za/co/absa/enceladus/utils/time/TimeZoneNormalizer.scala
benedeki: OK, will convert it to explicit.

conformance/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
```diff
@@ -38,67 +38,50 @@ import scala.util.control.NonFatal
...
-
-    spark.enableControlMeasuresTracking()
-      .setControlMeasuresWorkflow("Conformance")
+  private val format = new SimpleDateFormat("yyyy-MM-dd")
+  private val infoDateColumn = "enceladus_info_date"
+  private val infoVersionColumn = "enceladus_info_version"
 
-    // Enable control framework performance optimization for pipeline-like jobs
-    Atum.setAllowUnpersistOldDatasets(true)
+  private val log: Logger = LogManager.getLogger("enceladus.conformance.DynamicConformanceJob")
```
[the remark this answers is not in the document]
benedeki: It was there like that before, but will change 👍

utils/src/main/scala/za/co/absa/enceladus/utils/time/TimeZoneNormalizer.scala
```diff
@@ -26,37 +26,31 @@ import org.apache.spark.sql.SparkSession
...
   }
 
-  // scalastyle:off null
-  // Not nice, but we want to use null here in case no Spark session is available
-  def normalizeTimezone()(implicit spark: SparkSession = null): Unit = {
-    setJVMTimeZone()
-    if (spark != null) {
-      normalizeSessionTimeZone(spark)
-    } // else even if no Spark session was provided, the JVM time zone was still normalized
+  def normalizeAll(spars: Seq[SparkSession]): Unit = {
```
GeorgiChochov: Why `Seq`?

standardization/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
```diff
@@ -35,87 +35,105 @@ import za.co.absa.enceladus.utils.fs.FileSystemVersionUtils
...
 import za.co.absa.atum.AtumImplicits
 import za.co.absa.enceladus.menasplugin.MenasPlugin
 import za.co.absa.enceladus.model.Dataset
+import za.co.absa.enceladus.utils.time.TimeZoneNormalizer
 
 object StandardizationJob {
 
-  val log: Logger = LogManager.getLogger("enceladus.standardization.StandardizationJob")
-  val conf: Config = ConfigFactory.load()
+  private val log: Logger = LogManager.getLogger("enceladus.standardization.StandardizationJob")
```
Zejnilovic: `private val log: Logger = LogManager.getLogger(this.getClass)`

testutils/src/main/scala/za/co/absa/enceladus/testutils/datasetComparison/ComparisonJob.scala
```diff
@@ -21,6 +21,7 @@ import org.apache.spark.sql._
 import org.apache.spark.sql.functions._
 import za.co.absa.enceladus.testutils.exceptions._
 import za.co.absa.enceladus.testutils.{DataframeReader, DataframeReaderOptions, HelperFunctions}
+import za.co.absa.enceladus.utils.time.TimeZoneNormalizer
 
 object ComparisonJob {
   private val log: Logger = LogManager.getLogger("enceladus.testutils.ComparisonJob")
```
Zejnilovic: Would you mind doing the update even to this? 
`private val log: Logger = LogManager.getLogger(this.getClass)`

utils/src/main/scala/za/co/absa/enceladus/utils/time/TimeZoneNormalizer.scala
benedeki: Allows to deal with multiple Spark sessions in case of need within one Interface. Also easily support no sessions as well, or `Option[Session]`.
`Seq` in this kind of interface I found to be a more general solution.

GeorgiChochov: I don't think multiple `SparkSessions` is a real concern for us.

benedeki: Perhaps.
Does that disapprove the universality of the function? Does the current solution make it unnecessary complicated or slow? 😉

Issue #162, opened 2019-02-12
Model new Jenkinsfile


Zejnilovic:
Guys from DevOps are moving at a steady pace, maybe this should be discussed now.

We need a Jenkinsfile that can be used in our environment and be easy to set up and scale later on. It shouldn't be anything overengineered. The bellow lines should be a simplified template for it.

```groovy
pipeline {
    agent any

    parameters {
        string(name: 'branch', defaultValue: 'develop', description: 'Which branch to build from')
    }

    stages {
        stage('Prepare ENV') {
            steps {
                git([url: 'git@github.com:AbsaOSS/enceladus.git', branch: params.branch])
            }
        }
        stage('Build') {
            steps {
                echo 'Building..'
            }
        }
        stage('Test') {
            steps {
                echo 'Testing..'
                echo 'Later on this can be paralelized'
            }
        }
        stage('Deploy') {
            steps {
                echo 'Deploying....'
            }
        }
    }
    post {
        always {
            junit '**/target/*.xml'
        }
        failure {
            mail to: team@example.com, subject: 'The Pipeline failed :('
        }
    }
}
```

jenkins-absa:
Agreed, we can build on this to develop job configurations per branch or even go so far as to have multiple with different goals in mind, e.g. general test run, quick package builds without tests and full test with package building with the goal of merging based on success. 
Might be superfluous but we can discuss the scope we need.

hamc17:
Agreed, we can build on this to develop job configurations per branch or even go so far as to have multiple with different goals in mind, e.g. general test run, quick package builds without tests and full test with package building with the goal of merging based on success. 
Might be superfluous but we can discuss the scope we need.

Zejnilovic:
I am assigning @hamc17 because in the end, he should be the one to say yay or nay on this stuff.

hamc17:
Below is an example DSL jenkinsfile which builds the exact same as the manually configured job.
Investigating a way to abstract out the implementation specific details so we could upload this generically.

`mavenJob('Enceladus_develop_build') {
    description 'Build and test the app.'
    label( 'redhat_7.6_maven_build_slave')
    logRotator(-1, 20, -1, -1)
    triggers {
        githubPush()
    }
    scm {
        git {
            remote {
                github('AbsaOSS/enceladus', protocol = 'https', host = 'github.com')
                url('https://github.com/AbsaOSS/enceladus.git')
                credentials('GitHubCreds')
                branch('origin/develop')                 
[...]

PR #242, opened 2019-03-18
#162 Update Jenkinsfile

hamc17:


Jenkinsfile
```diff
@@ -1,14 +1,41 @@
...
+            }
+        }
+        stage ('Build') {
+            steps {
+                configFileProvider([configFile(fileId: "${mavenSettingsId}", variable: 'MAVEN_SETTINGS_XML')]) {
+                    sh "mvn -s $MAVEN_SETTINGS_XML clean package ${mavenAdditionalSettings}"
+                }
+            }
+        }
+        stage ('Deploy') {
```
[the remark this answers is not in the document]
hamc17: Sounds good.
I've updated our Jenkinsfile to deploy to the snapshot repo only if the branch is develop and release only if branch is master.

The first step is clean package which is all that pull requests will run.
This will verify for each pull request that the modules build successfully and pass all tests.
Once they get merged to master, they will be deployed to the internal repos.

This can be all expanded on further without too much trouble!

Jenkinsfile
```diff
@@ -1,14 +1,69 @@
...
+                    deploymentStageRun = sh(returnStdout: true, script: "echo true")
+                }
+            }
+        }
+    }
+    post {
+        success {
+            script {
+                if (deploymentStageRun != 'true')
+                    sh "echo 'Deployment or artifacts to internal repository skipped, branch is not develop/master'"
```
GeorgiChochov:
```suggestion
                    sh "echo 'Deployment of artifacts to internal repository skipped, branch is not develop/master'"
```

Jenkinsfile
```diff
@@ -1,14 +1,69 @@
+def nexusSnapshotRepoDetails = getNexusSnapshotRepoDetails()
+def nexusReleaseRepoDetails = getNexusReleaseRepoDetails()
+def mavenSettingsId = getMavenSettingsId()
+def mavenAdditionalSettings = getMavenAdditionalSettings()
+def enceladusSlaveLabel = getEnceladusSlaveLabel()
+def deploymentStageRun = 'false'
```
Zejnilovic: Are you sure we want this hardcoded value in the Jenkinsfile which we commit? I would do it the same way as the `enceladusSlaveLabel` and others.

Jenkinsfile
```diff
@@ -1,14 +1,69 @@
...
-         }
-     }
- }
+    agent {
+        label "${enceladusSlaveLabel}"
+    }
+    tools {
+        jdk 'openjdk-1.8.0'
+        maven 'Maven-3.6.0'
+        git 'git-latest'
```
Zejnilovic: I would maybe parametrize the tools as well - but that might be too much

Jenkinsfile
```diff
@@ -1,14 +1,69 @@
...
+            }
+        }
+        stage ('Build') {
+            steps {
+                configFileProvider([configFile(fileId: "${mavenSettingsId}", variable: 'MAVEN_SETTINGS_XML')]) {
+                    sh "mvn -s $MAVEN_SETTINGS_XML clean package ${mavenAdditionalSettings}"
+                }
+            }
+        }
+        stage ('Deploy-to-Snapshot') {
```
Zejnilovic: Having these deploy as two separate stages looks bad in the pipeline view. I would put it in one stage.

Jenkinsfile
hamc17: The reason for this particular value is it's used as a flag for the post step on whether to print out that there was no deployment done to Nexus.
I don't mind abstracting it if you guys would like but I feel it's not worth it.

Jenkinsfile
hamc17: This I can indeed abstract further.

Jenkinsfile
hamc17: That was my original plan, but using more than one 'when' within a stage doesn't work; Jenkins will complain.
If you know of any other options I'm happy to take a look.

Jenkinsfile
hamc17: Value removed, resolving

Jenkinsfile
hamc17: Tool versions abstracted, resolving.

Jenkinsfile
hamc17: Updated to have one stage for build, resolving.

hamc17:
Merging!

Issue #158, opened 2019-02-08
Getting latests Runs should return a Run projection


GeorgiChochov:
Currently getting the latest `Runs` returns the full `Run` objects, it should return just the `datasetName` `datasetVersion` and `runId` for querying the `Run's` details. We should also replace the `mapReduce` mongo queries to be performed with `aggregations` for better performance.
When we have authentication set up we should also limit the the results per user permissions.

PR #257, opened 2019-03-20
#158 Add Run Summary endpoints

GeorgiChochov:


GeorgiChochov:
> Seems ok. I just have one question. Why all my runs have the same id? Zero that is.

I noticed the same thing with your mongodump. I assume you clean up the output every time you rerun your spark job? It shouldn't be 0 though. It's something I'd like to look at but not as a part of this task. Once I have an explanation we either log a bug for it (and I fix it) or we understand how it happened.

Issue #256, opened 2019-03-20
Materialize application.conf templates rather than failing the build


lokm01:
Let's replace the enforcer plugin, which fails the build if the application.conf is missing from the main resources.

Rather let's materialize the template on the first build, which will provide reasonable defaults without necessary providing custom configs while preserving existing files.

This means that people should be able to clone and build the project (incl tests) out of the box.

Zejnilovic:
This will also mean rewriting the templates, so they work out of the box.

PR #258, opened 2019-03-20
#256 Replace maven enforcer with antrun copy task

lokm01:
Let's replace the enforcer plugin, which fails the build if the
application.conf is missing from the main resources.

Rather let's materialize the template on the first build, which
will provide reasonable defaults without necessary providing
custom configs while preserving existing files.

This means that people should be able to clone and build the
project (incl tests) out of the box.

### DIFF
--- a/conformance/pom.xml
+++ b/conformance/pom.xml
@@ -140,6 +140,7 @@
                         <configuration>
                             <target>
                                 <copy file="${project.basedir}/src/main/resources/application.conf.template" tofile="${project.basedir}/src/main/resources/application.conf" overwrite="false" />
+                                <copy file="${project.basedir}/src/main/resources/spline.properties.template" tofile="${project.basedir}/src/main/resources/spline.properties" overwrite="false" />                                
                             </target>
                         </configuration>
                         <goals>
--- a/conformance/pom.xml
+++ b/conformance/pom.xml
@@ -139,9 +139,8 @@
                         <phase>validate</phase>
                         <configuration>
                             <target>
-                                <copy file="${project.basedir}/src/main/resources/application.conf.template" tofile="${project.basedir}/src/main/resources/application.conf" failonerror="false" overwrite="false" />
+                                <copy file="${project.basedir}/src/main/resources/application.conf.template" tofile="${project.basedir}/src/main/resources/application.conf" overwrite="false" />
                             </target>
-                            <failOnError>false</failOnError>
                         </configuration>
                         <goals>
                             <goal>run</goal>
--- a/conformance/pom.xml
+++ b/conformance/pom.xml
@@ -131,30 +131,21 @@
             </plugin>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-enforcer-plugin</artifactId>
-                <version>${maven.enforcer.version}</version>
+                <artifactId>maven-antrun-plugin</artifactId>
+                <version>${maven.ant.plugin.version}</version>
                 <executions>
                     <execution>
-                        <id>enforce-app-properties-exist</id>
-                        <goals>
-                            <goal>enforce</goal>
-                        </goals>
+                        <id>copy</id>
+                        <phase>validate</phase>
                         <configuration>
-                            <rules>
-                                <requireFilesExist>
-                                    <files>
-                                        <file>${project.basedir}/src/main/resources/application.conf</file>
-                                    </files>
-                                    <level>ERROR</level>
-                                    <message>/src/main/resources/application.conf not found.
-To build Conformance module it is required to have 'application.conf' in the resources directory.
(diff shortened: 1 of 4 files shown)


conformance/pom.xml
```diff
@@ -131,30 +131,21 @@
...
-Please, copy the template file 'application.conf.template' into 'application.conf' and make changes
-corresponding to your use case.
-                                    </message>
-                                </requireFilesExist>
-                            </rules>
-                            <fail>true</fail>
+                            <target>
+                                <copy file="${project.basedir}/src/main/resources/application.conf.template" tofile="${project.basedir}/src/main/resources/application.conf" failonerror="false" overwrite="false" />
+                            </target>
+                            <failOnError>false</failOnError>
```
benedeki: Don't we want to fail in case the copy fails? (of course unless the file was there before)

benedeki:
While I had raised this question originally as well and I am the proponent of "out of the box" executable, i wonder if some more sophisticated solution wouldn't be cool in future (also based on the discussions we had).
1) Template defaults might not be generally suitable
2) Later versions might add some new configurations, and a so far working app suddenly stops functioning correctly.

conformance/pom.xml
```diff
@@ -131,30 +131,21 @@
...
-                                    <message>/src/main/resources/application.conf not found.
-To build Conformance module it is required to have 'application.conf' in the resources directory.
-Please, copy the template file 'application.conf.template' into 'application.conf' and make changes
-corresponding to your use case.
-                                    </message>
-                                </requireFilesExist>
-                            </rules>
-                            <fail>true</fail>
+                            <target>
+                                <copy file="${project.basedir}/src/main/resources/application.conf.template" tofile="${project.basedir}/src/main/resources/application.conf" failonerror="false" overwrite="false" />
```
Zejnilovic: what about spline.properties?

menas/pom.xml
```diff
@@ -260,30 +260,21 @@
...
-                                    <message>/src/main/resources/application.properties not found.
-To build Menas it is required to have 'application.properties' in the resources directory.
-Please, copy the template file 'application.properties.template' into 'application.properties' and make changes
-corresponding to your use case.
-                                    </message>
-                                </requireFilesExist>
-                            </rules>
-                            <fail>true</fail>
+                            <target>
+                                <copy file="${project.basedir}/src/main/resources/application.conf.template" tofile="${project.basedir}/src/main/resources/application.conf" failonerror="false" overwrite="false" />
```
Zejnilovic: Menas template is not in a runnable state out of the box.

standardization/pom.xml
```diff
@@ -147,30 +147,21 @@
...
-                                    <message>/src/main/resources/application.conf not found.
-To build Standardization module it is required to have 'application.conf' in the resources directory.
-Please, copy the template file 'application.conf.template' into 'application.conf' and make changes
-corresponding to your use case.
-                                    </message>
-                                </requireFilesExist>
-                            </rules>
-                            <fail>true</fail>
+                            <target>
+                                <copy file="${project.basedir}/src/main/resources/application.conf.template" tofile="${project.basedir}/src/main/resources/application.conf" failonerror="false" overwrite="false" />
```
Zejnilovic: Again, what about the spline.properties?

conformance/pom.xml
lokm01: Agree. Changed.

lokm01:
> 2\. Later versions might add some new configurations, and a so far working app suddenly stops functioning correctly.

Yeah if we can come up with something more sophisticated (for updating existing configs automatically) then we definitely should.

I guess this is less evil than failing the build out of the box for now.

menas/pom.xml
GeorgiChochov: @lokm01 Let's update the `za.co.absa.enceladus.migrations.dump.filepath` as part of this task, it's the only property that seems to me, not set to a reasonable default. I'd suggest: `/tmp/mongo-backups/menas`.

Let's double check if there are others

menas/pom.xml
```diff
@@ -260,30 +260,20 @@
...
-                                    <message>/src/main/resources/application.properties not found.
-To build Menas it is required to have 'application.properties' in the resources directory.
-Please, copy the template file 'application.properties.template' into 'application.properties' and make changes
-corresponding to your use case.
-                                    </message>
-                                </requireFilesExist>
-                            </rules>
-                            <fail>true</fail>
+                            <target>
+                                <copy file="${project.basedir}/src/main/resources/application.conf.template" tofile="${project.basedir}/src/main/resources/application.conf" overwrite="false" />
```
GeorgiChochov:
```suggestion
                                <copy file="${project.basedir}/src/main/resources/application.properties.template" tofile="${project.basedir}/src/main/resources/application.properties" overwrite="false" />
```
Menas uses `.properties`

lokm01: Thanks!

PR #262, opened 2019-03-20
Feature/259 include examples in the build

lokm01:
Adding examples in the build.
Fixed some compile errors.
Added ant run into pom to materialize the config template

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -270,11 +270,5 @@
                 <module>selenium</module>
             </modules>
         </profile>
-        <profile>
-            <id>examples</id>
-            <modules>
-                <module>examples</module>
-            </modules>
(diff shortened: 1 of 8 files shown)


pom.xml
```diff
@@ -37,6 +37,7 @@
         <module>conformance</module>
         <module>standardization</module>
         <module>dataModel</module>
+        <module>examples</module>
```
GeorgiChochov: If we are adding it to the main build, we should remove the `examples` profile at the end of the pom file.

benedeki: With this added, please remove the `examples`  profile at the bottom (line 274)

lokm01: Done

Issue #11, opened 2018-11-19
View updates even from canceled edit window


Zejnilovic:
If a user clicks edit over Basic Info of any of the models (Schemas, Datasets, Mapping tables), changes description, and then clicks cancel, the change still gets shown in the current window. It almost seems as the update went through.

This goes away after window refresh and does not send any call to database or such.

Zejnilovic:
Component Menas

GeorgiChochov:
> Component Menas

Let's use labels for this

PR #266, opened 2019-03-20
#11 View updates even from canceled edit window

lokm01:
Perform deep copy of the current entity objects to ensure
that two-way binding doesn't affect the un-saved changes.

### DIFF
--- a/menas/ui/components/dataset/conformanceRule/upsert.controller.js
+++ b/menas/ui/components/dataset/conformanceRule/upsert.controller.js
@@ -107,7 +107,7 @@ sap.ui.define([
 
     onRuleSubmit: function () {
       let currentDataset = this._model.getProperty("/currentDataset");
-      let newRule = JSON.parse(JSON.stringify(this._model.getProperty("/newRule")));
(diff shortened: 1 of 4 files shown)


lokm01:
> We have other places where we use `JSON.parse(JSON.stringify(obj))`, not sure if we should normalize one way or the other.
> There's an interesting [discussion on best way to deep copy in JS](https://stackoverflow.com/questions/122102/what-is-the-most-efficient-way-to-deep-clone-an-object-in-javascript)

Yeah it seems that Object.assign will be the cleanest for shallow copies. I'd say the jQuery may be a bit cleaner than JSON ser/de. Suggestions?

GeorgiChochov:
> > We have other places where we use `JSON.parse(JSON.stringify(obj))`, not sure if we should normalize one way or the other.
> > There's an interesting [discussion on best way to deep copy in JS](https://stackoverflow.com/questions/122102/what-is-the-most-efficient-way-to-deep-clone-an-object-in-javascript)
> 
> Yeah it seems that Object.assign will be the cleanest for shallow copies. I'd say the jQuery may be a bit cleaner than JSON ser/de. Suggestions?

Nope, sounds good to me.

Issue #53, opened 2018-12-02
Runs view


GeorgiChochov:
The dataset should have a runs page showing run metadata, as per old Menas.
Reminder: add thousands separator.

PR #267, opened 2019-03-21
#53 Add Runs view

GeorgiChochov:


menas/ui/components/run/runMain.view.xml
```diff
@@ -0,0 +1,189 @@
...
+           xmlns:table="sap.ui.table"
+           xmlns:cust="http://schemas.sap.com/sapui5/extension/sap.ui.core.CustomData/1">
+    <Page title="Runs" enableScrolling="false">
+        <customHeader>
+            <Bar>
+                <contentMiddle>
+                    <Title text="Runs"/>
+                </contentMiddle>
+                <contentRight>
+                    <Text text="Logged in as:"/>
```
lokm01: Was there a fragment for these added recently?

menas/ui/components/run/runMain.view.xml
```diff
@@ -0,0 +1,189 @@
...
+                                                        source: {
+                                                            pattern: 'dd-MM-yyyy HH:mm:ss ZZZ'
+                                                        }
+                                                    }
+                                                }"/>
+                                                <Label text="Conformance duration"/>
+                                                <Text text="{run>/cfmTime/duration}"/>
+                                                <Label text="Error" visible="{= ${run>/status} === 'Failed'}"/>
+                                                <Text id="error" text="{run>/runStatus/error/description}"
+                                                      visible="{= ${run>/status} === 'Failed'}"/>
```
lokm01: Nice, didn't even know about this complex binding syntax :)

menas/ui/service/RunService.js
```diff
@@ -0,0 +1,215 @@
...
+    });
+
+    let controlNames = checkpoints.flatMap(checkpoint =>
+      checkpoint.controls.map(control => {
+        checkpoint[control.controlName] = control.controlValue;
+        return control.controlName;
+      })
+    );
+
+    oTable.removeAllColumns();
```
lokm01: Could this be done in the view?

menas/ui/components/run/runMain.view.xml
GeorgiChochov: Thanks for spotting that! Oversight on my part.

menas/ui/service/RunService.js
GeorgiChochov: I tried... I couldn't come up with a way to do it since the columns need to be dynamically generated, which I was only able to figure to do programatically.

Zejnilovic:
### 1. Sort the additional info
Additional info is currently unsorted and searching for something is hard. Even more so when std and conform are shuffled.
image

### 2. Sort the runs
In Runs view, the left panel should be sorted from the most recenent.
In Dataset view -> Runs tab it should be sorted in the same way, from the most recent.

### 3. Run table style in dataset view
Table style should be same as for the audit trails. 
TBH I would update both. Audit trails aren't perfect as well. New issue/feature request maybe?

GeorgiChochov:
> ### 3. Run table style in dataset view
> Table style should be same as for the audit trails.
> TBH I would update both. Audit trails aren't perfect as well. New issue/feature request maybe?

I'll align the table with the Audit Trails one. If we want to rework the UX aspect of the design, makes sense to do it in a future/separate task.

menas/ui/service/RunService.js
```diff
@@ -91,6 +91,7 @@ var RunService = new function () {
     oRunSummaries.forEach(run => {
       run.status = this._normalizeStatus(run.status)
     });
+    oRunSummaries.sort((a, b) => b.runId - a.runId);
```
Zejnilovic: please sort by the start time. Now it looks like this and it is still a bit hard to understand. Even more, since each version has it's own id's. So something run for the first time would end on the end.

image

menas/ui/service/RunService.js
```diff
@@ -0,0 +1,226 @@
...
+
+    oControl.setModel(new sap.ui.model.json.JSONModel(oRun), "run");
+    oControl.setModel(new sap.ui.model.json.JSONModel(oRun.controlMeasure.metadata), "metadata");
+  };
+
+  this._bindRunSummaries = function(oRunSummaries, oControl) {
+    oRunSummaries.forEach(run => {
+      run.status = this._normalizeStatus(run.status)
+    });
+    oRunSummaries.sort((a, b) => b.runId - a.runId);
```
Zejnilovic: please sort by the start time. Now it looks like this and it is still a bit hard to understand. Even more, since each version has it's own id's. So something run for the first time would end on the end.

image

Zejnilovic: Will be solved in later PR, which incorporates a more complex menu of runs.

GeorgiChochov: #273

Issue #260, opened 2019-03-20
RunId is always 0


GeorgiChochov:
## Describe the bug
Menas does not update the runId of newly created Runs for Datasets with previous Runs.

## To Reproduce
1. Run Standardization of a Dataset
2. Check Menas's Runs for that Dataset

## Expected behaviour
Menas should handle incrementing the runId for new Runs.

Issue #39, opened 2018-12-02
Implement optimistic locking or similar concurrency control


GeorgiChochov:

GeorgiChochov:
This is fixed by #690 and #694

PR #269, opened 2019-03-21
#260 Fix runId generation

GeorgiChochov:


GeorgiChochov:
> We have no locking strategy currently for IDs. I think there should be something.

I agree, but we first need to come up with a good way of handling concurrent DB operations. I'd like to do that for all write operations, for all entities in a separate task (#39).

Zejnilovic:
> > We have no locking strategy currently for IDs. I think there should be something.
> 
> I agree, but we first need to come up with a good way of handling concurrent DB operations. I'd like to do that for all write operations, for all entities in a separate task (#39).

Fair enough. This doesn't concern current PR

Issue #186, opened 2019-02-25
Epoch standardisation doesn't pass with pattern "epoch"


Zejnilovic:
**Describe the bug**
I get an error when I have a schema which standartizises from string to epoch timestamp

**To Reproduce**
Schema field of one of the failing fields
```json
{
  "name": "random_negative",
  "type": "timestamp",
  "nullable": false,
  "metadata": {
    "pattern": "epoch"
  }
}
```

**Log**
```bash
2019-02-25 10:32:15 INFO  ControlFrameworkREST:134 - HttpResponseProxy{HTTP/1.1 200  [X-Content-Type-Options: nosniff, X-XSS-Protection: 1; mode=block, Cache-Control: no-cache, no-store, max-age=0, must-revalidate, Pragma: no-cache, Expires: 0, X-Frame-Options: DENY, X-Content-Type-Options: nosniff, Content-Type: application/json;charset=UTF-8, Transfer-Encoding: chunked, Date: Mon, 25 Feb 2019 09:32:14 GMT] ResponseEntityProxy{[Content-Type: application/json;charset=UTF-8,Chunked: true]}}
2019-02-25 10:32:15 INFO  SparkLineageInitializer$:54 - Spline v0.3.6 is initializing...
2019-02-25 10:32:15 ERROR SparkLineageInitializer$:60 - Initialization failed! Spark Lineage tracking is DISABLED.
java.lang.IllegalArgumentException: requirement failed: Missing configuration property spline.persistence.factory
	at scala.Predef$.require(Predef.scala:224)
	at za.co.absa.spline.common.ConfigurationImplicits$ConfigurationRequiredWrapper.za$co$absa$spline$common$ConfigurationImplicits$ConfigurationRequiredWrapper$$getRequired(ConfigurationImplicits.scala:107)
	at za.co.absa.spline.common.ConfigurationImplicits$ConfigurationRequiredWrapper$$anonfun$getRequiredString$3.apply(ConfigurationImplicits.scala:40)
	at za.co.absa.spline.common.ConfigurationImplicits$ConfigurationRequiredWrapper$$anonfun$getRequiredString$3.apply(ConfigurationImplicits.scala:40)
	at za.co.absa.spline.core.conf.DefaultSplineConfigurer.persistenceFactory$lzycompute(DefaultSplineConfigurer.scala:92)
[...]

PR #272, opened 2019-03-21
#186: Epoch standardisation doesn't pass with pattern "epoch"

benedeki:
* DateTimeValidators changes to use the same conversion class as the standardization itself
* EnceladusDateParser class now has format method to convert TO string
* Unit tests for DateTimeValidators
* Enhanced unit tests for Format and EnceladusDateParser

### DIFF
--- a/utils/src/main/scala/za/co/absa/enceladus/utils/types/EnceladusDateParser.scala
+++ b/utils/src/main/scala/za/co/absa/enceladus/utils/types/EnceladusDateParser.scala
@@ -26,7 +26,6 @@ import java.time.format.DateTimeFormatter
   * @param pattern  the formatting string, in case it's an epoch format the values wil need to be convertible to Long
   */
 case class EnceladusDateParser(pattern: Format) {
-  //TODO use Option.when with Scala 2.13
   private val formatter: Option[SimpleDateFormat] = if (pattern.isEpoch) {
     None
   } else {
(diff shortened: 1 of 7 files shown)


utils/src/main/scala/za/co/absa/enceladus/utils/types/EnceladusDateParser.scala
```diff
@@ -18,21 +18,25 @@ package za.co.absa.enceladus.utils.types
...
+import java.time.format.DateTimeFormatter
 
 /**
   * Enables to parse string to date and timestamp based on the provided format
   * Unlike SimpleDateFormat it also supports an epoch ("epoch", "milliepoch"...) formatting strings
   * @param pattern  the formatting string, in case it's an epoch format the values wil need to be convertible to Long
   */
 case class EnceladusDateParser(pattern: Format) {
-  private val simpleDateFormat: Option[SimpleDateFormat] = if (pattern.isEpoch) {
+  //TODO use Option.when with Scala 2.13
```
Zejnilovic: Please don't commit TODO's, FIXME, etc. to code. Use issues or other platforms available.

utils/src/main/scala/za/co/absa/enceladus/utils/types/Format.scala
```diff
@@ -61,7 +67,15 @@ object Format {
   def epochFactor(format: String): Long = {
     format.toLowerCase match {
       case "epoch"      => 1
-      case "milliepoch" => 1000
+      case "milliepoch" => thousand
```
Zejnilovic: If you are trying to "bypass" magic numbers here, why not change the upper as well?

Issue #157, opened 2019-02-07
Add support for offset time zone notation in timestamp format


benedeki:
This form of timestamp with time zone is not supported right now:
2019-01-01 00:00:00+07:00

benedeki:
This pattern is actually already supported, just not well documented in Spark/Java.SimpleDateFormat.
The pattern expression for the offset time zone is _xxx_.
I will add unit tests for showing and proving this pattern

PR #274, opened 2019-03-25
#157: Add support for offset time zone notation in timestamp format

benedeki:
* Added unit tests showing and verifying the existing classes already support offset timezone

### DIFF
--- a/utils/src/test/scala/za/co/absa/enceladus/utils/types/EnceladusDateParserSuite.scala
+++ b/utils/src/test/scala/za/co/absa/enceladus/utils/types/EnceladusDateParserSuite.scala
@@ -51,7 +51,7 @@ class EnceladusDateParserSuite extends FunSuite{
     assert(resultTimestamp.toString == expectedTimestamp.toString)
   }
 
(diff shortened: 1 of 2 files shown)


utils/src/test/scala/za/co/absa/enceladus/utils/types/EnceladusDateParserSuite.scala
```diff
@@ -64,6 +64,32 @@ class EnceladusDateParserSuite extends FunSuite{
     assert(resultTimestamp.toString == expectedTimestamp.toString)
   }
 
+  test("EnceladusDateParser class actual pattern with standard time zone") {
+    val parser: EnceladusDateParser = EnceladusDateParser("yyyy-MM-dd-HH-mm-ss-zz")
+
+    val value: String = "2011-01-31-22-52-33-EST"
+    val resultDate: Date = parser.parseDate(value)
+    val expectedDate: Date = new Date(111, 1, 1) //2019-01-19
```
Zejnilovic: scala> Date(111, 1, 1)
res4: java.sql.Date = 2011-02-01

Issue #206, opened 2019-03-04
Refactor timestamp handling classes


benedeki:
Some classes handling the timestamp don't have the best name, best location.
Also agreed to change the keyword for milliseconds since epoch from `milliepoch` to `epochmilli`.

PR #275, opened 2019-03-25
#206: Refactor timestamp handling classes

benedeki:
* keyword milliepoch -> epochmilli
* class Format -> DateTimePattern
* class EnceladusDateParser -> EnceladusDateTimeParser
* moving DateTimePattern and EnceladusDateTimeParser to time folder
* adhering to scalastyle and addressing some compilation warnings


standardization/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
```diff
@@ -18,18 +18,22 @@ package za.co.absa.enceladus.standardization.interpreter.stages
...
 import org.apache.spark.sql.functions._
-import za.co.absa.enceladus.utils.types.{Defaults, Format}
+import za.co.absa.enceladus.utils.types.Defaults
+
 import scala.util.Random
 import za.co.absa.enceladus.utils.error.ErrorMessage
 import za.co.absa.enceladus.utils.schema.SchemaUtils.appendPath
 import za.co.absa.enceladus.utils.error.UDFLibrary
 import za.co.absa.spark.hofs._
+
```
Zejnilovic: What is the reason for adding newlines here in such a way?

benedeki: None. I blame the IDE. I rarely meddle with `import` directly except when removing unused items.

conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/DynamicInterpreter.scala
```diff
@@ -45,32 +44,16 @@ object DynamicInterpreter {
     * @return The conformed dataframe
     *
     */
-  def interpret(conformance: ConfDataset,
-                inputDf: Dataset[Row],
-                jobShortName: String = "Conformance",
+  def interpret(conformance: ConfDataset, inputDf: Dataset[Row], jobShortName: String = "Conformance",
                 experimentalMappingRule: Boolean)
```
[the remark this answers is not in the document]
benedeki: In this case it actually can be re-expanded and it just fits into the function length limit.

But I suggest we really revisit our method length limit in  `scalastyle_config.xml`. In a 8 parameter function just the declaration takes 16% of the quote then.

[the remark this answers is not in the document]
Zejnilovic: - there is loose hanging `experimentalMappingRule` because it wouldn't fit. Please return to the param per line.
- parameters should not and I think are not counted into the length of the method. And having 8 parameters is a bit of an overkill, but that is for another topic and time.

conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/DynamicInterpreter.scala
```diff
@@ -154,4 +139,33 @@ object DynamicInterpreter {
     }
   }
 
+  private def getStepsAndExplode(conformance: ConfDataset, inputDf: Dataset[Row], experimentalMappingRule: Boolean):
```
GeorgiChochov: Is this method extraction necessary?
1. Method names with `And` in them make me think a method is doing too much.
2. The expanded nesting of the `experimentalMappingRule` explosions makes me a little worried that we might be introducing more complexity than we are reducing by refactoring right now.

yruslan: This method seems having several unrelated responsibilities. Consider making each statement of this function a separate method, e.g.
```
getOrderedSteps(...)
addErrorColumnIfMissing(...)
explodeAllArrays(...)
```

Then, you can call each method one by one from `interpret()` or you can leave `getStpsAndExplode()` as a grouping method. Up to you.

standardization/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
benedeki: My wrong, function declaration is not counted into the limit. Please consider my previous comment moot. 😊

conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/DynamicInterpreter.scala
```diff
@@ -45,32 +44,16 @@ object DynamicInterpreter {
     * @return The conformed dataframe
     *
     */
-  def interpret(conformance: ConfDataset,
-                inputDf: Dataset[Row],
-                jobShortName: String = "Conformance",
+  def interpret(conformance: ConfDataset, inputDf: Dataset[Row], jobShortName: String = "Conformance",
                 experimentalMappingRule: Boolean)
                (implicit spark: SparkSession, dao: EnceladusDAO, progArgs: CmdConfig, enableCF: Boolean): DataFrame = {
```
yruslan: `enableCF` is an implicit boolean parameter. This might be easily cause errors if we have another method that having an implicit boolean which has a different meaning. Consider making this parameter explicit.

conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/DynamicInterpreter.scala
```diff
@@ -32,7 +32,6 @@ import za.co.absa.enceladus.utils.explode.{ExplodeTools, ExplosionContext}
 
 object DynamicInterpreter {
   private val log = LogManager.getLogger("enceladus.conformance.DynamicInterpreter")
-  private var enableControlFramework = true
```
yruslan: A nice spot!

external/fixedWidth/src/test/scala/za/co/absa/enceladus/fixedWidth/FixedWidthRelationSuite.scala
```diff
@@ -17,6 +17,6 @@ package za.co.absa.enceladus.fixedWidth
 
 import org.scalatest.FunSuite
 
-class FormatSuite extends FunSuite {
+class FixedWidthRelationSuite extends FunSuite {
```
yruslan: Empty test suite? Should we remove it? Or add a `ToDo` so we won't forget to add tests for handling fixed width text files.

utils/src/main/scala/za/co/absa/enceladus/utils/types/Defaults.scala
```diff
@@ -59,21 +65,22 @@ object Defaults {
       case _: DoubleType    => value.toDouble
       case _: LongType      => value.toLong
       case _: StringType    => value
-      case _: DateType      => EnceladusDateParser(dtPattern.getOrElse(throw new IllegalArgumentException("No date format specified."))).parseDate(value)
-      case _: TimestampType => EnceladusDateParser(dtPattern.getOrElse(throw new IllegalArgumentException("No timestamp format specified."))).parseTimestamp(value)
+      case _: DateType      => getParser(dtPattern,"No date format specified.").parseDate(value)
```
yruslan: `getParser()` seems like a vary generic name. Consider using more specific name, like `getDateTimeParser()`

conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/DynamicInterpreter.scala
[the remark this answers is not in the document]
benedeki: My wrong, function declaration is not counted into the limit. Please consider my previous comment moot. 😊

conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/DynamicInterpreter.scala
benedeki: I understand that this refactor has left much to the desired. I will rework it to include you remarks.

conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/DynamicInterpreter.scala
benedeki: Thank you.

external/fixedWidth/src/test/scala/za/co/absa/enceladus/fixedWidth/FixedWidthRelationSuite.scala
benedeki: IIRC I had created the empty test, because something somewhere complained about the folder not existing. And before this commit I noticed that there was a wrong class name within the "test"
I'll create a ticket for the test to be implemented, but would consider in a low prio,

utils/src/main/scala/za/co/absa/enceladus/utils/time/DateTimePattern.scala
```diff
@@ -13,74 +13,77 @@
...
     format.toLowerCase match {
       case "epoch"      => epochThousandFactor
-      case "milliepoch" => epochUnitFactor
+      case "epochmilli" => epochUnitFactor
       case _            => throw new InvalidParameterException(s"'$format' is not an epoch format")
     }
   }
 
-  implicit def format2String(format: Format): String = format.get
+  implicit def format2String(format: DateTimePattern): String = format.get
```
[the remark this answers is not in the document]
benedeki: Agreed

utils/src/main/scala/za/co/absa/enceladus/utils/time/EnceladusDateTimeParser.scala
```diff
@@ -13,19 +13,17 @@
...
+import java.sql.{Date, Timestamp}
 import java.text.SimpleDateFormat
-import java.sql.Date
-import java.sql.Timestamp
-import java.time.format.DateTimeFormatter
 
 /**
   * Enables to parse string to date and timestamp based on the provided format
-  * Unlike SimpleDateFormat it also supports an epoch ("epoch", "milliepoch"...) formatting strings
+  * Unlike SimpleDateFormat it also supports an epoch ("epoch", "epochmilli"...) formatting strings
```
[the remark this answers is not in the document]
benedeki: It's a comment. In case future keywords would be added, this probably wouldn't be updated.

utils/src/main/scala/za/co/absa/enceladus/utils/types/Defaults.scala
benedeki: Changing to shorter but less generic `getDTParser`

utils/src/main/scala/za/co/absa/enceladus/utils/types/Defaults.scala
```diff
@@ -59,21 +65,22 @@ object Defaults {
...
+      case _: TimestampType => getParser(dtPattern,"No timestamp format specified.").parseTimestamp(value)
       case _: BooleanType   => value.toBoolean
       case _: DecimalType   => new java.math.BigDecimal(value)
       case _                => throw new IllegalStateException(s"No default value defined for data type ${dt.typeName}")
     }
   }
 
-  /** Wrapper which returns the parsed default value for datetype and timestamptype returns ,default value for other primtive type
-    * from metadata first (if any) and the global one otherwise */
+  /** Wrapper which returns the parsed default value for datetype and timestamptype returns,
```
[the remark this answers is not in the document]
benedeki: Will do, and actually turn it into scaladoc valid entry

external/fixedWidth/src/main/scala/za/co/absa/enceladus/fixedWidth/types/SparkTypeResolver.scala
```diff
@@ -56,7 +57,7 @@ object SparkTypeResolver {
 
   private def castToDateTimeType(field: StructField, dateString: String, pattern: String): Any = {
     try {
-      val parser: EnceladusDateParser = EnceladusDateParser(pattern)
+      val parser: EnceladusDateTimeParser = EnceladusDateTimeParser(pattern)
```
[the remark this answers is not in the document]
benedeki: As you surely noticed I like and tend to have my things explicitly typed. Sure a habit from other languages.
Is it a problem in our code? I would have to conscientiously concentrate to avoid the usage.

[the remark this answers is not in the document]
GeorgiChochov: I get that but I would prefer a consistent style and very much like type inference, especially in cases like this.

utils/src/main/scala/za/co/absa/enceladus/utils/time/EnceladusDateTimeParser.scala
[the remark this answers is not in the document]
GeorgiChochov: I kind of think that comments should reflect the current state of the application, so this type of future-proofing is confusing in terms of what is currently being supported (makes it look like the documentation is  incomplete).

external/fixedWidth/src/main/scala/za/co/absa/enceladus/fixedWidth/types/SparkTypeResolver.scala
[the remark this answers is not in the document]
benedeki: OK, I will try to avoid type declarations on direct (case) class or primitive types assignment. 😄

utils/src/main/scala/za/co/absa/enceladus/utils/time/EnceladusDateTimeParser.scala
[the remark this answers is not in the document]
benedeki: Will reword to avoid the issue.

external/fixedWidth/src/main/scala/za/co/absa/enceladus/fixedWidth/types/SparkTypeResolver.scala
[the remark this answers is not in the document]
GeorgiChochov: Thanks!

Issue #173, opened 2019-02-20
Optional meta timezone for date standardization instruction


benedeki:
Even dates might need time zone adjustment. But it's optional and once standardized not carried further.

GeorgiChochov:
Closing as duplicate of #155 (the two are tightly connected and handled in the same PR #279)

Issue #155, opened 2019-02-07
 Optional meta timezone for timestamp standardization instruction


benedeki:
If the source data doesn't contain information of the timestamp's time zone, to properly standardize the timestamp the time zone should be specified in the standardization instruction. (In certain cases it's also valid for dates)

GeorgiChochov:
Covers #173 as well.

PR #279, opened 2019-03-28
#155, #173: Optional timezone meta for timestamp and date standardization

benedeki:
* Extensive refactor of TypePattern to allow structured standardization based on type
* Added time zone recognition into DateTimePattern
* Created ancestor class (TypePattern) for DateTimePattern extracting the
  pattern logic not specific to date&time
* Implementing the timezone meta usage for timestamp and date normalization
* Massive extension of TypeParser unit tests


standardization/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
```diff
@@ -15,185 +15,331 @@
...
+      * Float         | ->Decimal->String->to_date      | ->Decimal->String->to_timestamp->to_utc_timestamp->to_date
+      * Decimal       | ->String->to_date               | ->String->->to_timestamp->->to_utc_timestamp->to_date
+      * String        | ->to_date                       | ->to_timestamp->->to_utc_timestamp->to_date
+      * Timestamp     | ->to_date                       | ->to_utc_timestamp->to_date
+      * Date          | O                               | ->to_utc_timestamp->to_date
+      * Other         | ->String                        | ->String->to_timestamp->to_utc_timestamp->to_date
+      */
+
+    trait DateTimePW extends PrimitivePW {
+      protected def basicCastFunction: (Column, String) => Column  //for epoch casting
```
GeorgiChochov: I'm really confused by this `def`, which seems like a perfectly good `val` to me.

standardization/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
```diff
@@ -15,185 +15,331 @@
...
+      protected def epochPattern: String
+      protected val pattern: DateTimePattern = DateTimePattern(field)
+      protected lazy val defaultTimeZone: Option[String] = pattern.defaultTimeZone
```
GeorgiChochov: Why are there so many `lazy val`s everywhere? I feel like this would break in very unexpected ways, very easily.

standardization/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
```diff
@@ -15,185 +15,331 @@
...
   }
 
-  /** Defines the cast error logic for numeric and other primitive types **/
-  private def primitiveCastErrorLogic(field: StructField, origSchema: StructType, path: String, attr: Column): Column = {
-    val castedCol = primitiveCastLogic(field, origSchema, path, attr)
-    field.dataType match {
-      // here we also want to check the numeric overflow by comparing with the original value this could break with trailing or leading zeros
-      case _: IntegerType | _: ShortType | _: FloatType | _: DoubleType | _: DecimalType | _: ByteType | _: LongType => castedCol isNull
-      case _ => castedCol isNull
+  private object ParserWorker {
```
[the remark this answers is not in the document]
benedeki: If naming is the issue I wasn't very happy with the current name either. I welcome any suggestions. 😉

benedeki:
We agreed to go through this change together with @GeorgiChochov and @yruslan Monday after lunch. Others feel free to join, or we can schedule a different time slot.

utils/src/main/scala/za/co/absa/enceladus/utils/time/DateTimePattern.scala
```diff
@@ -45,44 +41,61 @@ class DateTimePattern(val pattern: Option[String], val forType: Option[DataType]
...
     }
   }
-}
+
+
+  def timeZoneInPattern(pattern: String): Boolean = {
+    if (isEpoch(pattern)) {
+      true
+    } else {
+      val sanitized = "(^|[^\\\\])'.*[^\\\\]'".r.replaceAllIn(pattern, "$1") //clear literals
```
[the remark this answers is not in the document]
GeorgiChochov: Nice! I think the below one might benefit from a name as well. Generally speaking regex should have a self-explanatory variable name.

utils/src/test/scala/za/co/absa/enceladus/utils/time/DateTimePatternSuite.scala
```diff
@@ -141,4 +140,99 @@ class DateTimePatternSuite extends FunSuite {
...
+    val dateTimePattern2 = DateTimePattern("'XXX: 'HH:mm:ss XX yyyy-MM-dd")
+    assert(dateTimePattern2.timeZoneInPattern)
+    val dateTimePattern3 = DateTimePattern("'Date:'yyyy-MM-dd HH:mm:ss\\'ZZ\\'")
+    assert(dateTimePattern3.timeZoneInPattern)
+  }
+
+  test("Default time zone - not present") {
+    val dateTimePattern1 = DateTimePattern("yyyy-MM-dd HH:mm:ss", Some(TimestampType))
+    assert(dateTimePattern1.defaultTimeZone.isEmpty)
+    val dateTimePattern2 = DateTimePattern("yyyy-MM-dd", Some(DateType), None)
```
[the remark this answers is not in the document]
GeorgiChochov: Good point! I would consider that a separate test case then, one specifically for checking the default value of the param.

utils/src/main/scala/za/co/absa/enceladus/utils/time/DateTimePattern.scala
```diff
@@ -45,44 +41,61 @@ class DateTimePattern(val pattern: Option[String], val forType: Option[DataType]
   def epochMilliFactor: Long = {
     DateTimePattern.epochMilliFactor(get)
   }
+
+  lazy val defaultTimeZone: Option[String] = assignedDefaultTimeZone.filterNot(_ => timeZoneInPattern)
```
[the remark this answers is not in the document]
GeorgiChochov: Please be very careful with `lazy val`s I think they cause much more unexpected issues than they help solve. I would generally avoid them.

standardization/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
```diff
@@ -15,185 +15,331 @@
...
+  private type ColumnTransformer = Column=>Column
+
+  private object Parent {
+    def array(column: Column): Parent = Some(Left(column))
+    def struct(column: Column): Parent = Some(Right(column))
   }
 
-  private def toDate(field: StructField, origSchema: StructType, path: String, attr: Column): Column = {
-    dateTimePreProcess(field: StructField, origSchema: StructType, path: String, attr: Column, epochDateFormat, to_date)
+  private sealed trait ParserWorker {
```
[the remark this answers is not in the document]
GeorgiChochov: True, but we could just as easily put all of our classes in the same file because they all stem from the `main` method needing them, directly or transitively. 
I'd suggest using `package` access modifier and separating things a bit.

standardization/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
```diff
@@ -15,185 +15,331 @@
...
+        )
+        // rebuild the struct
+        val str = struct(cols: _*).as(field.name)
+        ParseOutput(str, errs1)
+      }
+    }
+
+    trait PrimitivePW extends ParserWorker {
+      /** Defines the cast error logic for numeric and other primitive types **/
+      //the casting logic differs by class, but within is the same, no need to evalute more timez
```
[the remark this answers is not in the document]
GeorgiChochov: I think you still have a trailing `*`.

standardization/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
GeorgiChochov: By making the code difficult to debug, reason about and change in any way. Spread out reliance on `lazy val`s is a bit of a code smell to me.

standardization/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
```diff
@@ -15,185 +15,331 @@
...
+          when(size(err) === lit(0), null
+          // scalastyle:on null
+          ).otherwise( Defaults.getDefaultValue(field) )
+        ) as field.name
+
+        ParseOutput(std, err)
+      }
+    }
+
+    trait SimplePW extends  PrimitivePW {
```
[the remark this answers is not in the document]
GeorgiChochov: I think you misunderstood my point, an `abstract` type extending a `sealed` type breaks the seal unless the `abstract` subtype is also sealed. Specifying `sealed` on concrete types a bit unconventional but I guess it could make sense in some situations. 

However, `sealed` specifies something that is being extended by a predefined set of subclasses. But extending a `case class` is generally speaking a[ bad idea](https://stackoverflow.com/questions/34561614/should-i-use-the-final-modifier-when-declaring-case-classes) (I would even like to enforce a `final case class` convention)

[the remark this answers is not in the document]
benedeki: Yeah, I went little over-zealous with the `sealed`. 😊

utils/src/main/scala/za/co/absa/enceladus/utils/types/TypePattern.scala
```diff
@@ -0,0 +1,62 @@
...
+object TypePattern {
+  implicit def patternToString(pattern: TypePattern): String = pattern.pattern
+
+  private def getMetadata(structField: StructField, key: String): Option[String] = {
+    Try(structField.metadata.getString(key)).toOption
+  }
+
+  def apply(structField: StructField ):TypePattern = {
+    val dataType = structField.dataType
+    val patternString: Option[String] = getMetadata(structField, "pattern")
```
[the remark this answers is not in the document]
benedeki: Yeah, thanks for noticing.

benedeki:
Please, don't bother review in current state. Massive changes in pipe...

benedeki:
Improved the code per our Monday discussion except the `TypeParser` object (which is still in pipe).
I don't plan to change the other classes more, unless you find some issues. So feel free to review them or wait until `TypeParser` is ready too.

Zejnilovic:
> Improved the code per our Monday discussion except the `TypeParser` object (which is still in pipe).
> I don't plan to change the other classes more, unless you find some issues. So feel free to review them or wait until `TypeParser` is ready too.

There seems to be a typo of some kind. Please check your code once again.

TypeParses.scala:236
image

benedeki:
> 
> 
> > Improved the code per our Monday discussion except the `TypeParser` object (which is still in pipe).
> > I don't plan to change the other classes more, unless you find some issues. So feel free to review them or wait until `TypeParser` is ready too.
> 
> There seems to be a typo of some kind. Please check your code once again.
> 
> TypeParses.scala:236
> image

Thanks, should be fixed now. At least it compiles, Tests might not be passing, the class/object will be reworked.

utils/src/main/scala/za/co/absa/enceladus/utils/implicits/StringImplicits.scala
```diff
@@ -0,0 +1,66 @@
...
+object StringImplicits {
+  implicit class StringImprovements(string: String) {
+    def findFirstUnquoted(charsToFind: Set[Char], quoteChars: Set[Char], escape: Char = '\\' ): Option[Integer] = {
+      @tailrec
+      def scan(sub: String, index: Integer, outQuote: Option[Char], escaped: Boolean = false):Option[Integer] = {
+        //escaped flag defaults to false, as every non-escape character clears it
+        val head: Option[Char] = sub.find(_ => true) //get the first character of the string if it exists
+        (head, escaped) match {
+          case (None, _) => None // end of string
+          case (`outQuote`, false) if !outQuote.contains(escape)  => scan(sub.tail, index + 1, None) // end of quote,
```
Zejnilovic: Too many whitespaces between `(escape)  =>`

utils/src/main/scala/za/co/absa/enceladus/utils/implicits/StringImplicits.scala
```diff
@@ -0,0 +1,66 @@
...
+                                                                                                     // unless escaped
+                                                                                                     // or equals escape
+          case (`outQuote`, true) if outQuote.contains(escape) => scan(sub.tail, index + 1, None) // end of quote set
+                                                                                                  // by from escape
+          case (Some(`escape`), false) => scan(sub.tail, index + 1, outQuote, escaped = true) // this is the only case
+                                                                                              // when escape
+          case (_, _) if outQuote.nonEmpty => scan(sub.tail, index + 1, outQuote) // within quotes
```
Zejnilovic: Why do you use `(_, _)` here but just `_` at the end?

utils/src/main/scala/za/co/absa/enceladus/utils/implicits/StructFieldImplicits.scala
```diff
@@ -0,0 +1,35 @@
...
+      Try(structField.metadata.getString(key)).toOption
+    }
+
+    def readPatternInfo(): PatternInfo = {
+      val patternOpt: Option[String] = getMetadata("pattern")
+      PatternInfo(patternOpt.getOrElse(Defaults.getGlobalFormat(structField.dataType)), patternOpt.isEmpty)
+    }
+  }
+
+  case class PatternInfo (pattern: String, isDefault: Boolean)
```
Zejnilovic: I would expect case class on top

utils/src/main/scala/za/co/absa/enceladus/utils/time/DateTimePattern.scala
```diff
@@ -17,72 +17,140 @@ package za.co.absa.enceladus.utils.time
...
+
+  def timeZoneInPattern(pattern: String): Boolean = {
+    if (isEpoch(pattern)) {
+      true
+    } else {
+      pattern.hasUnquoted(patternTimeZoneChars, Set('''))
+    }
+  }
+
+  /* TODO ---
```
Zejnilovic: Does this have a purpose?

utils/src/main/scala/za/co/absa/enceladus/utils/time/DateTimePattern.scala
```diff
@@ -17,72 +17,140 @@ package za.co.absa.enceladus.utils.time
...
+  def apply(pattern: String,
+            assignedDefaultTimeZone: Option[String] = None): DateTimePattern = {
+    createDateTimePattern(pattern, assignedDefaultTimeZone, isDefault = false)
+  }
+
+  def fromStructField(structField: StructField ): DateTimePattern = {
+    if (!Seq(DateType, TimestampType).contains(structField.dataType)) {
+      val typeName = structField.dataType.typeName
+      throw new InvalidParameterException(
+        s"StrucField data type for DateTimePattern has to be DateType or TimestampType, instead $typeName was given."
```
Zejnilovic: Typo `StrucField` => `StructField`

utils/src/main/scala/za/co/absa/enceladus/utils/time/DateTimePattern.scala
```diff
@@ -17,72 +17,140 @@ package za.co.absa.enceladus.utils.time
...
+      case EpochMilliKeyword => epochUnitFactor
+      case _                 => 0
     }
   }
+
+  def timeZoneInPattern(pattern: String): Boolean = {
+    if (isEpoch(pattern)) {
+      true
+    } else {
+      pattern.hasUnquoted(patternTimeZoneChars, Set('''))
```
Zejnilovic: This could be just a oneliner `isEpoch(pattern) || pattern.hasUnquoted(patternTimeZoneChars, Set('''))`. I for some reason don't like the if X then true style.

benedeki:
The PR is now ready for full review.
I added the classed based `TypeParser` in _newVer_ sub-directory. Please feel free to compare with the original solution (with the new logic) and comment.
Both versions pass the - IMHO extensive - set of UTs.

utils/src/main/scala/za/co/absa/enceladus/utils/implicits/StringImplicits.scala
```diff
@@ -0,0 +1,66 @@
...
+    def findFirstUnquoted(charsToFind: Set[Char], quoteChars: Set[Char], escape: Char = '\\' ): Option[Integer] = {
+      @tailrec
+      def scan(sub: String, index: Integer, outQuote: Option[Char], escaped: Boolean = false):Option[Integer] = {
+        //escaped flag defaults to false, as every non-escape character clears it
+        val head: Option[Char] = sub.find(_ => true) //get the first character of the string if it exists
+        (head, escaped) match {
+          case (None, _) => None // end of string
+          case (`outQuote`, false) if !outQuote.contains(escape)  => scan(sub.tail, index + 1, None) // end of quote,
+                                                                                                     // unless escaped
+                                                                                                     // or equals escape
```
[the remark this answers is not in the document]
benedeki: Do you think it would make it easier to understand the logic? I found it this way it made it clearer which line the explanation belongs to. Also easier to compare the cases...
Although the one comment with same indentation separates the later cases from the earlier ones, again for easier understanding of the flow.

utils/src/main/scala/za/co/absa/enceladus/utils/implicits/StringImplicits.scala
benedeki: Came natural in this phase of `match`, perhaps I had one of the sides of the case bound too in one moment of the development. I will simplify it, thanks.

utils/src/main/scala/za/co/absa/enceladus/utils/implicits/StructFieldImplicits.scala
benedeki: I prefer putting `implicit` stuff above all else. 😄

utils/src/main/scala/za/co/absa/enceladus/utils/time/DateTimePattern.scala
benedeki: Yeah, to show my forgetfulness. 😄 Will remove

utils/src/main/scala/za/co/absa/enceladus/utils/time/DateTimePattern.scala
benedeki: It's much nicer indeed. But it either relies on `||` optimization when in standard form or it always runs into the `pattern.hasUnquoted` evaluation. So wondering, what is the less bad...

Zejnilovic: Both Java and Scala have `||` optimized and I would not worry about it being changed. It is a feature of the languages that gets a lot of praise.

utils/src/main/scala/za/co/absa/enceladus/utils/implicits/StructFieldImplicits.scala
Zejnilovic: I get that, but the case class is on the level with the object.

utils/src/main/scala/za/co/absa/enceladus/utils/time/DateTimePattern.scala
GeorgiChochov: This `|` is an eager OR (or a bitwise operator), which evaluates both sides of the operator, `||` is a short-circuit OR. No need to worry in C-derived languages.

benedeki: To by technically pedantic the `||` function operator can be overloaded, and then it stops to be smart enough not to evaluate the right side unless needed.
But on the other hand I think it's safe to assume that for `Boolean` it's not probable, and if done, the author deserves the consequences. Will change it, you were right.

standardization/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/newVer/TypeParser.scala
```diff
@@ -0,0 +1,370 @@
...
+      case t => throw new IllegalStateException(s"${t.typeName} is not a supported type in this version of Enceladus")
+    }
+    parserClass(field, path, origSchema, parent)
+  }
+
+  private final case class ArrayParser(field: StructField,
+                                       path: String,
+                                       origSchema: StructType,
+                                       parent: Parent) extends TypeParser {
+    private val fieldType: ArrayType = field.dataType.asInstanceOf[ArrayType]
```
GeorgiChochov: This could probably be left to type inference, since we explicitly cast it to `ArrayType`

standardization/src/test/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParserSuite.scala
```diff
@@ -16,28 +16,391 @@
...
+      case d: Date => s"DATE '${d.toString}'"
+      case t: Timestamp => s"TIMESTAMP('${t.toString}')"
+      case s: String => s
+      case x => x.toString
+    }
+
+    val castString = createCastTemplate(baseType, target.dataType, pattern, decimalSize, timezone).format(srcF)
+
+    val std = s"CASE WHEN ((size(CASE WHEN (($srcF IS NULL) AND $boolS) THEN array(stdNullErr($srcF)) ELSE CASE WHEN (($srcF IS NOT NULL) AND ($castString IS NULL)) THEN array(stdCastErr($srcF, CAST($srcF AS STRING))) ELSE [] END END) = 0) AND ($srcF IS NOT NULL)) THEN $castString ELSE CASE WHEN (size(CASE WHEN (($srcF IS NULL) AND $boolS) THEN array(stdNullErr($srcF)) ELSE CASE WHEN (($srcF IS NOT NULL) AND ($castString IS NULL)) THEN array(stdCastErr($srcF, CAST($srcF AS STRING))) ELSE [] END END) = 0) THEN NULL ELSE $default END END AS `${target.name}`"
+    val actual = output.stdCol.toString().replaceFirst("#\\d+$", "")
```
GeorgiChochov:
```suggestion
    val actual = output.stdCol.toString()
```
This seems to be unnecessary, as just below the same operation does not require it (`val errActual = output.errors.toString()`).

standardization/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/newVer/TypeParser.scala
```diff
@@ -0,0 +1,370 @@
...
+  val fieldName: String = SchemaUtils.getFieldNameOverriddenByMetadata(field)
+  val currentAttrPath: String = appendPath(path, fieldName) // absolute path to field to parse
+  val column: Column = parent.map {
+    case Left(x) => x //parent is array
+    case Right(x) => x(fieldName) //parent is struct
+  }.getOrElse(col(currentAttrPath)) // no parent
+  val isArrayElement: Boolean = parent.exists(_.isLeft)
+}
+
+object TypeParser {
```
GeorgiChochov: I like this solution! Much better now that there is no `lazy val` complexity.

standardization/src/test/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParserSuite.scala
```diff
@@ -16,28 +16,391 @@
...
+      timestampPatternField,
+      datePatternDefaultField,
+      timestampPatternDefaultField,
+      datePatternTmzField,
+      timestampPatternTmzField,
+      dateEpochField,
+      timestampEpochField
+    ))
+
+    var output: ParseOutput = TypeParser.standardize(sourceField, input.path, schema)
```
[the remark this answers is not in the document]
GeorgiChochov: A nested suites approach with a Suites hierarchy like we discussed would be perfect.

standardization/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
```diff
@@ -36,65 +35,122 @@ object TypeParser {
...
+    castFunction(sourceCol, pattern)
   }
 
+  private def castEpoch(attr: Column, epochFactor: Long, epochPattern: String, castFunction: colStrToColFnc): Column =
+    castFunction(from_unixtime(attr.cast("Long")  / epochFactor, epochPattern), epochPattern)
+
+  private def toTimestampWithDefaultTimeZone(col: Column, pattern: String, defaultTimeZone: String): Column =
+    to_utc_timestamp(to_timestamp(col, pattern), defaultTimeZone)
+
+  private def toDateWithDefaultTimeZone(col: Column, pattern: String, defaultTimeZone: String): Column =
```
Zejnilovic: The name `toDateWithDefaultTimeZone` and parameter `defaultTimeZone` would make people think they can alter the output of date with timezone.

standardization/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
```diff
@@ -178,20 +244,20 @@ object TypeParser {
...
                   origSchema: StructType,
                   parent: Option[Column] = None,
-                  isArrayElement: Boolean = false)
-                 (implicit spark: SparkSession, udfLib: UDFLibrary): ParseOutput = {
+                  isArrayElement: Boolean = false): ParseOutput = {
     // If the meta data value sourcecolumn is set to override the field name
     val fieldName = SchemaUtils.getFieldNameOverriddenByMetadata(field)
 
-    import spark.implicits._
+    //import spark.implicits._
```
Zejnilovic: Is this still needed here if it's commented out?

standardization/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
```diff
@@ -178,20 +244,20 @@ object TypeParser {
...
     val currentAttrPath = appendPath(path, fieldName) // calculate the absolute path to this attribute
-    val thisCol = if (isArrayElement) parent.get else if (parent.isDefined) parent.get(fieldName) else col(currentAttrPath)
-    val errField = unpath(currentAttrPath) + Random.nextLong().abs
+    val thisCol =
+      if (isArrayElement) parent.get else if (parent.isDefined) parent.get(fieldName) else col(currentAttrPath)
```
Zejnilovic: You know I love oneliners but I think this is a bit of overkill

standardization/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
```diff
@@ -36,65 +35,122 @@ object TypeParser {
...
-      case _: IntegerType | _: ShortType | _: FloatType | _: DoubleType | _: DecimalType | _: ByteType | _: LongType => castedCol isNull
+      // here we also want to check the numeric overflow by comparing with the original value this could break with
+      // trailing or leading zeros
+      case _: IntegerType |
+           _: ShortType |
+           _: FloatType |
+           _: DoubleType |
+           _: DecimalType |
+           _: ByteType |
+           _: LongType => castedCol isNull
```
Zejnilovic: Can't this be a `case _: NumericType`

standardization/src/test/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParserSuiteTemplate.scala
```diff
@@ -0,0 +1,237 @@
...
+import za.co.absa.enceladus.standardization.interpreter.dataTypes.ParseOutput
+import za.co.absa.enceladus.utils.types.Defaults
+import za.co.absa.enceladus.standardization.interpreter.stages.TypeParserSuiteTemplate._
+
+trait TypeParserSuiteTemplate extends FunSuite {
+  protected def createCastTemplate(toType: DataType, pattern: String, timezone: Option[String]): String
+
+  private val sourceFieldName = "sourceField"
+  private val dateEpochPattern = "epoch"
+  private val timestampEpochPattern = "epochmilli"
```
yruslan: Why redefine epoch constants here if they are already defined as `EpochKeyword` and `EpochMilliKeyword`?

standardization/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
```diff
@@ -15,185 +15,364 @@
...
-            throw new InvalidParameterException("Dates & times represented as numeric values need specified 'pattern' metadata")
-          }
-          attr.cast(StringType)
-        case _ => attr
-      }
-      castFunction(sourceCol, format)
-    }
+  sealed trait Parent {
+    val parentColumn: Column
+    val isArray: Boolean
```
yruslan: A very minor suggestion. This boolean seems just a convenience for distinguishing between `ArrayParent` and `StructParent` since it i always true for `ArrayParent` and always false for `StructParent`.

Consider removing the boolean and replacing its currant usage with pattern matching (or `isInstanceOf()`). Again, it is only a suggestion.

standardization/src/test/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParserSuiteTemplate.scala
```diff
@@ -0,0 +1,237 @@
...
+    s"CASE WHEN ((size(CASE WHEN (($srcField IS NULL) AND $nullableS) THEN array(stdNullErr($srcField)) ELSE CASE WHEN (($srcField IS NOT NULL) AND ($castS IS NULL)) THEN array(stdCastErr($srcField, CAST($srcField AS STRING))) ELSE [] END END) = 0) AND ($srcField IS NOT NULL)) THEN $castS ELSE CASE WHEN (size(CASE WHEN (($srcField IS NULL) AND $nullableS) THEN array(stdNullErr($srcField)) ELSE CASE WHEN (($srcField IS NOT NULL) AND ($castS IS NULL)) THEN array(stdCastErr($srcField, CAST($srcField AS STRING))) ELSE [] END END) = 0) THEN NULL ELSE $default END END AS `${target.name}`"
+  }
+
+  private def assemblyErrorExpression(srcField: String, nullableS: String, castS: String): String = {
+    s"CASE WHEN (($srcField IS NULL) AND $nullableS) THEN array(stdNullErr($srcField)) ELSE CASE WHEN (($srcField IS NOT NULL) AND ($castS IS NULL)) THEN array(stdCastErr($srcField, CAST($srcField AS STRING))) ELSE [] END END"
+  }
+
+  private def doAssert(expectedExpression: String, actualExpression: String): Unit = {
+    if (actualExpression != expectedExpression) {
+      //the expressions tend to be rather long, the assert most often cuts the beginning and/or end of the string
```
GeorgiChochov:
```suggestion
      // the expressions tend to be rather long, the assert most often cuts the beginning and/or end of the string
```

standardization/src/test/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParserSuiteTemplate.scala
```diff
@@ -0,0 +1,237 @@
...
+      // showing just the vicinity of the difference, therefor the output of the whole strings
```
GeorgiChochov:
```suggestion
      // showing just the vicinity of the difference, so we log the output of the entire expressions here
```

yruslan:
I've just finished running an end-to-end test for this PR and values for all datetime fields are saved to Parquet with a +2 hour shift.

I presume that the new implementation corrects the behaviour of handling datetime types without timezones. The previous implementation probably used local time zone, while the new implementation uses UTC. Please, ensure that the new behaviour is as expected.

Screenshot 2019-04-15 at 13 22 30

The test was done on of of our datasets. It uses XML input and the following schema definition:
```json
        {
            "name" : "endDateTime",
            "type" : "timestamp",
            "path" : "",
            "elementType" : null,
            "containsNull" : null,
            "nullable" : true,
            "metadata" : {
                "pattern" : "yyyy-MM-dd'T'HH:mm:ss"
            },
            "children" : []
        }, 
        {
            "name" : "expiryDate",
            "type" : "date",
            "path" : "",
            "elementType" : null,
            "containsNull" : null,
            "nullable" : true,
            "metadata" : {
                "pattern" : "yyyy-MM-dd"
            },
            "children" : []
        }, 
        {
            "name" : "expiryTime",
            "type" : "timestamp",
            "path" : "",
            "elementType" : null,
            "containsNull" : null,
            "nullable" : true,
            "metadata" : {
                "pattern" : "yyyy-MM-dd'T'HH:mm:ss"
            },
            "children" : []
        }, 
```

The raw data had the following values in the XML:
```xml
<record>
    <endDateTime>2000-05-30T00:00:00</endDateTime>
    <expiryDate>2000-05-30</expiryDate>
    <expiryTime>1970-01-01T02:00:00</>
</record>
```

GeorgiChochov:
@yruslan pointed out a very important piece of information to me. The timezone specified in the `metadata` needs to conform to what spark expects: 
```
   * @param tz A string detailing the time zone that the input belongs to, such as `Europe/London`,
   *           `PST` or `GMT+5`
```

We should document this is the README and the maybe on the DateTimePattern class's scaladoc.

yruslan:
Can confirm:
1. Without a default timezone specified in the schema the job uses configured timezone which is 'UTC' by default.
2. When a default timezone is specified for a column the job uses it correctly for the field. Need to be careful how we specify timezones. `Africa/Johannesburg` worked for me, while `SAST`, `UTC+2:00`, `+02:00` didn't. Maybe we need to validate this property and provide documentation what time zone specs are valid. Maybe this needs a dedicated issue.
3. When reading from a parquet file Spark uses current settings when converting to JSON. Since I've did the comparison on my local machine which has `CET` timezone, the winter times were in `UTC+1:00` and the summer times were in `UTC+2:00` (thanks @GeorgiChochov for the that insight).
4. After adding default `timezone` parameter to the timestamp fields the results of the end-to-end tests are the same.

utils/src/main/scala/za/co/absa/enceladus/utils/implicits/StringImplicits.scala
```diff
@@ -0,0 +1,94 @@
...
+ */
+
+package za.co.absa.enceladus.utils.implicits
+
+import java.security.InvalidParameterException
+
+import scala.annotation.tailrec
+
+object StringImplicits {
+  implicit class StringImprovements(string: String) {
```
GeorgiChochov: Just one small thing, the convention for implicits is to call things usually goes "RichX" or "WrappedX" or "xOps". ([Like it's described here](https://stackoverflow.com/questions/28061661/is-there-a-naming-convention-for-implicit-classes))

benedeki:
>We should document this is the README and the maybe on the DateTimePattern class's scaladoc (or where appropriate).
@GeorgiChochov  Created the #353 for that

Issue #178, opened 2019-02-25
Button with no function in Menas Dataset creation


Zejnilovic:
**Describe the bug**
There is a button with no function to it, in Menas New/Edit Dataset. The code for the button in Menas `schemaSelector.fragment.xml` line 31-32 is empty. 
```
<Button icon="sap-icon://action">
</Button>
```
Even if it is supposed to do something later on, it should be somehow documented and show tooltip on hover or have some kind of label.

**To Reproduce**
1. Go to Menas
2. Open and edit any dataset or create a new one
3. Scroll to the line with `Schema Version:`
4. On the right side

**Expected behaviour**

**Screenshots**
![image](https://user-images.githubusercontent.com/17467768/53318399-5f506200-38cf-11e9-9630-78101a927d04.png)

Issue #278, opened 2019-03-26
Improve Navigation


GeorgiChochov:
We currently have two overlapping menus in Menas. One Master (navigation) menu that hides the underlying Detail (run/dataset/schema/mapping table) list menu. We can combine these (the Navigation menu can transition into the Entity List menu) and simplify the UI, improving UX.

PR #289, opened 2019-04-04
#178 Remove unused button from add/edit dialogs

GeorgiChochov:
This was probably intended to provide a popup view of the selected schema, but was never implemented. It doesn't make sense to have a functionless button until we do implement this feature (if we do, we should have create a task for it).


Zejnilovic:
> There seems to be something that was depending on that button. Now button new data set doesn't work.
> 
> image
> 
> mvn clean package
> deployed through IntelliJ

My fault. Error actually introduced to `Develop` branch already.

GeorgiChochov:
> > There seems to be something that was depending on that button. Now button new data set doesn't work.
> > image
> > mvn clean package
> > deployed through IntelliJ
> 
> My fault. Error actually introduced to `Develop` branch already.

Yep, I messed something up in #278, will address in #292.

Issue #292, opened 2019-04-04
Unable to create new Dataset or Mapping table


Zejnilovic:
## Describe the bug
When I log in freshly and try to create a dataset or a mapping table I get an error to console. If I refresh the page it lets me create them.

## Screenshots
![image](https://user-images.githubusercontent.com/17467768/55545519-acdd9d00-56cd-11e9-9322-8b8bd12d1a2c.png)

## Additional context
This blocks PR #289 and #290 from being properly tested.

PR #298, opened 2019-04-04
#292 Fetch schemas for add/edit dialog via callback on press

GeorgiChochov:


lokm01:
Can we also turn on busy indicator while we're fetching schemas? Or maybe disable the drop down?

GeorgiChochov:
> Can we also turn on busy indicator while we're fetching schemas? Or maybe disable the drop down?

Currently it waits for the schemas to be fetched before it opens the dialog. The busy indicator made the dialog pop-up, hide and reappear on click for some reason (I think it interfered with the HDFS browser's busy indicator). I'll try to change it.

GeorgiChochov:
> This works as intended. But I have noticed that we sent quite a lot of `CSRF: X-CSRF-TOKEN -> XXXXXX` with each click. Four with each click, to be precise. Is that normal?

Yes, we are making way too many requests (I've been keeping eye on the CSRF log as wellas an indicator of request frequency)... I'll come up with a way to reduce this. For now it's normal.

lokm01:
LGTM

Issue #281, opened 2019-04-02
Navigation should be always visible


Zejnilovic:
## Background
New navigation implemented by #280 should be always visible if the user has more then 1280px window size. This is a size enough for all the current views.

If a user has smaller window size I would fold the menu.

Issue #216, opened 2019-03-06
Active hamburger menu on login page


Zejnilovic:
## Describe the bug
There is classic navigation hamburger menu on the login page and it is active during logged off period. If I click it and pick an item, I get a message session expired. 

## Expected behaviour
I would make the button invisible or disabled on the login page

## Desktop (please complete the following information):
- OS: macOS 10.14.3
- Browser Chrome Version 72.0.3626.109 (Official Build) (64-bit)

PR #302, opened 2019-04-08
#281 Show Navigation menu

GeorgiChochov:


GeorgiChochov:
> This PR does what it is supposed to do. But now I am a bit bothered by the empty menu on the left when logging in. I couldn't find a trivial solution though, so I will think about it at a later date.

It bothers me too, but I can't find a way to hide it... We actually have an issue for this already (#216), so I thought we'd figure it out in that one and gave up for now.

PR #306, opened 2019-04-09
Bugfix/228 some errors not populting

yruslan:


conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/DynamicInterpreter.scala
```diff
@@ -214,43 +213,20 @@ object DynamicInterpreter {
...
   }
 
+  /**
+    * Adds an explosion pseudo-rule before each group of more than 1 mapping rule.
+    * Adds a collection pseudo-rule after each group of more than 1 mapping rule
+    *
+    * @param ruleGroups A list of rules grouped together if they are mapping rules operating at the same array level
+    * @param schema     A schema of a dataset
+    * @return The new list of rules that might contain pseudo-rules
+    */
```
Zejnilovic: We talked that private API doesn't have to be commented, but I really appreciate you taking time and doing so.

yruslan: Thanks!

conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/DynamicInterpreter.scala
```diff
@@ -214,43 +213,20 @@ object DynamicInterpreter {
...
-          pushGroup()
-          groups += List(a)
-      }
-    }
-    pushGroup()
-    groups.toList
+    Algorithms.stableGroupByOption[ConformanceRule, String](rules, {
+      case m: MappingConformanceRule => SchemaUtils.getDeepestArrayPath(schema, m.outputColumn)
+      case _ => None
+    }).map(_.toList).toList
```
GeorgiChochov: I get that you are mapping `Seq` to `List` here, but why not have the more abstract return type in the function: `Seq[Seq[ConformanceRule]]` we are not making use of some `List`-specific properties with the return of this function (we just `foldLeft` from what I could trace). I'd substitute it for the more abstract `Seq` type.

utils/src/main/scala/za/co/absa/enceladus/utils/general/Algorithms.scala
```diff
@@ -0,0 +1,143 @@
...
+    * A stable group by takes a collection of items and groups items when consecutive items have the same
+    * grouping value. A grouping value is provided by a callback function.
+    *
+    * The function does not change the original order of the elements.
+    *
+    * @param collection     An input collection
+    * @param groupedElement A callback that for a given item produces a value by which items should be grouped
+    * @return A sequence of groups of the original items
+    */
+  def stableGroupBy[T, A](collection: Seq[T], groupedElement: T => A): Seq[Seq[T]] = {
```
GeorgiChochov:
```suggestion
  def stableGroupBy[T, A](collection: Seq[T], groupBy: T => A): Seq[Seq[T]] = {
```
The second parameter is a callback function, I'd give it a more "verb-y" name  to reflect that. (The suggestion above is not the best option, I'm sure you can come up with something more appropriate)

conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/DynamicInterpreter.scala
GeorgiChochov: Also, the implementation underneath is already a `List` when you look at the implementation of this algorithm.

utils/src/main/scala/za/co/absa/enceladus/utils/general/Algorithms.scala
yruslan: In Scala standard library this callback function is named `f`. We can follow them...

PR #309, opened 2019-04-09
Added Build Status Badge for Develop Branch

hamc17:
This is a build status badge which will show the current build status on Jenkins for the develop branch.

Users can view the last build on successful login to Jenkins. 
No details can be seen until the user logs in.


Zejnilovic:
Put it into a table and have the status for both develop and master?

hamc17:
> Put it into a table and have the status for both develop and master?

Can do

hamc17:
> > Put it into a table and have the status for both develop and master?
> 
> Can do

Actually, on further inspection I realised that for now there is no Jenkinsfile on the master branch, thus no build job in Jenkins for master.
It can be added later as another column in the table but for now it doesn't make sense as the link won't work.
I could put in some kind of placeholder but I'm not sure that's the best.
Thoughts?

Issue #204, opened 2019-03-04
HDFS Browser can't show files with longer names


Zejnilovic:
## Describe the bug
When I try to open or show something in HDFS Browser in Menas (for example in Dataset - Raw folder) it just reloads the view and shows empty lines.

## To Reproduce
1. Go to Dataset
2. Click edit
3. Find in HDFS Browser a folder containing a file with a long name
4. Unfold the structure / try to show the file

## Expected behaviour
I would expect some type of folding or splitting of the name

## Screenshots
![mar-04-2019 11-07-02](https://user-images.githubusercontent.com/17467768/53726400-2a638280-3e6e-11e9-9f93-7946b44d4217.gif)

## Desktop (please complete the following information):
- OS: macOS 10.14.3
- Browser Chrome Version 72.0.3626.109 (Official Build) (64-bit)22]

Issue #179, opened 2019-02-25
HDFSBrowser component stuck after folding root structure


Zejnilovic:
**Describe the bug**
In Menas Dataset creation/editing, in a HDFSBrowser component, if I click to fold the root structure I lose the ability to click any other element in the browser. Then I need to click cancel or save. But if I click cancel I lose the other stuff I wrote in it, and if I click save, I get a version witch point to the root of hdfs.

Gif in Screenshots speaks more than a thousand words.

**To Reproduce**
Steps to reproduce the behavior OR commands run:
1. Go to Menas
2. Open and edit any dataset or create a new one
3. Click to fold the root structure in HDFSBrowser
4. Try to click any other

**Expected behaviour**
I expect to still be able to fold/expand and choose other things.

**Screenshots**
![feb-25-2019 07-43-05](https://user-images.githubusercontent.com/17467768/53319157-fcac9580-38d1-11e9-9aa9-04e11a3b2409.gif)

**Desktop (please complete the following information):**
 - OS: macOS 10.14.3
 - Browser Chrome Version 72.0.3626.109 (Official Build) (64-bit)

lokm01:
This seems to be fixed in PR #316

Zejnilovic:
Yes, it does. I will close the issue.

GeorgiChochov:
In the future, let's not close issues until their PR is merged.

Zejnilovic:
> In the future, let's not close issues until their PR is merged.

I agree, I thought it already was and when I realized my mistake I was keeping an active eye on it.

PR #316, opened 2019-04-10
#204 HDFS Browser can't show files with longer names

lokm01:


Zejnilovic:
This PR is correcting the behaviour of #179

menas/pom.xml
```diff
@@ -240,10 +238,7 @@
                             <filtering>true</filtering>
                         </resource>
                     </webResources>
-                    <packagingExcludes>
-                        META-INF/*.SF,
-                        META-INF/*.DSA,
-                        META-INF/*.RSA
+                    <packagingExcludes> META-INF/*.SF, META-INF/*.DSA, META-INF/*.RSA
```
GeorgiChochov: If we are inlining this, let's bring up the closing tag to be on the same line as well.

menas/ui/components/hdfs/HDFSBrowser.js
```diff
@@ -200,25 +196,22 @@ sap.ui.define([], function() {
...
+            that._tree.expand(index);
           }
           // also select the correct list item
           if (hdfsPath === sPath) {
             items[i].setSelected(true);
           }
         }
-        if(rev.length === 0) {
-          // this binding seems weird on these nested models
+        if(rev.length > 0) fnHelper(sNewPath, sModelPath, rev.reverse());
```
GeorgiChochov:
```suggestion
        if(rev.length > 0) {
          fnHelper(sNewPath, sModelPath, rev.reverse());
        } else {
```
Let's follow the agreed on style for multi-line `if-else` and add curly braces

PR #326, opened 2019-04-11
325 Changed calls to Nexus to call Artifactory

hamc17:
On a merge with these changes the groovyUtils repo update will be merged


Zejnilovic:
This should be in my opinion more generic like `getArtifactRepoUrl` etc. So we don't have to change it every time.

Zejnilovic:
> This should be in my opinion more generic like `getArtifactRepoUrl` etc. So we don't have to change it every time.

Created the issue #328, so the change gets properly logged.

PR #327, opened 2019-04-11
169 Editing dataset with wrong paths in HDFS makes user change the path to root first

lokm01:
- Changed Text to Input component for manual entry
- Added better validation directly to the HDFS browser
- Changed valueState to Warning for invalid HDFS paths

### DIFF
--- a/menas/ui/components/dataset/addDataset.fragment.xml
+++ b/menas/ui/components/dataset/addDataset.fragment.xml
@@ -31,17 +31,17 @@
                         <Label text="HDFS raw data folder path"/>
                         <hdfs:HDFSBrowser id="newDatasetRawHDFSBrowser" height="300px" width="100%"
                                           busyControl="addDatasetDialog"
-                                          pathLabel="selectedRawHDFSPathLabel" restURI="api/hdfs/list"
+                                          pathInput ="selectedRawHDFSPathLabel" restURI="api/hdfs/list"
                                           HDFSPath="{entity>/hdfsPath}"/>
                         <Label text="RAW HDFS Path"/>
-                        <Text id="selectedRawHDFSPathLabel"/>
+                        <Input type="Text" id="selectedRawHDFSPathLabel"/>
                         <Label text="HDFS conformed data publish folder path"/>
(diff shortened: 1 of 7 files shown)


Zejnilovic:
## UI Comments
### 1. In the case of dataset only one message
If I have two hdfs fields both yet non-existing I get two confirmation messages. Can we make it just one?

### 2. A prompt asking if we want to continue
image

This prompt is not like the others. Can we use [DialogConfirm](https://sapui5.hana.ondemand.com/#/entity/sap.m.Dialog/sample/sap.m.sample.DialogConfirm) from sapUI5?

### 3. Location of the text field and hdfs browser
What about moving the text field above the hdfs browser? I thought we agreed on this earlier. Is there an issue with it?

GeorgiChochov:
I'm being nitpicky but our convention is for the branch name to be something like `bugfix/169-hdfs-browser-invalid-path`

lokm01:
> I'm being nitpicky but our convention is for the branch name to be something like `bugfix/169-hdfs-browser-invalid-path` (just for future reference)

Sure..

lokm01:
Sorry, closed this accidentally

lokm01:
> ## UI Comments
> ### 1. In the case of dataset only one message
> If I have two hdfs fields both yet non-existing I get two confirmation messages. Can we make it just one?
> 
> ### 2. A prompt asking if we want to continue
> image
> 
> This prompt is not like the others. Can we use [DialogConfirm](https://sapui5.hana.ondemand.com/#/entity/sap.m.Dialog/sample/sap.m.sample.DialogConfirm) from sapUI5?
> 
> ### 3. Location of the text field and hdfs browser
> What about moving the text field above the hdfs browser? I thought we agreed on this earlier. Is there an issue with it?

1. Ok, will only trigger one confirm message (if the first one is cancelled). I believe the second one is still reasonable if the second path is invalid as well and the user confirms the first one. Thoughts?

2. Unless we rewrite the validations, I need a blocking confirmation dialog. Let's move this to a separate issue

3. That's still the plan, wanna do it in a separate PR

lokm01:
Raised isssue #330 for 2)

lokm01:
Raised #331 for 3)

PR #332, opened 2019-04-12
Feature/331 collapsible HDFS browser

lokm01:


lokm01:
The re-render of the control after expanding seems too strange.

Let's just switch the order of the input field and the tree for now.

Issue #329, opened 2019-04-12
Arrays are reordered in the new implementation of the mapping rule


yruslan:
## Describe the bug
During an end-to-end testing it was noticed that arrays of structs were reordered after applying mapping rules to its fields.

## To Reproduce
Run a Conformance job with `--experimental-mapping-rule` flag on a dataset containing an array of structs. A dataset should have a mapping rule applied to one of the fields of the array.

## Expected behaviour
Arrays should not be reordered

PR #334, opened 2019-04-12
#329 Fix array reordering in the experimental mapping rule implementation

yruslan:
This is tested end-to-end

### DIFF
--- a/utils/src/main/scala/za/co/absa/enceladus/utils/explode/ExplodeTools.scala
+++ b/utils/src/main/scala/za/co/absa/enceladus/utils/explode/ExplodeTools.scala
@@ -147,20 +147,19 @@ object ExplodeTools {
 
     val isNested = explosion.arrayFieldName.contains('.')
 
-    val (decDf, decField, transientColumn) = if (isNested) {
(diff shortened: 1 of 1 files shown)


utils/src/main/scala/za/co/absa/enceladus/utils/explode/ExplodeTools.scala
```diff
@@ -213,9 +214,9 @@ object ExplodeTools {
       // Drop the array size column
       .drop(col(explosion.sizeFieldName))
       // restore original record order
-      .orderBy(groupedCol)
+      .orderBy(orderByRecordCol)
       // remove monotonic id created during explode
-      .drop(groupedCol)
+      .drop(orderByRecordCol)
   }
   // scalastyle:on method.length
```
lokm01: is this IDE configuration?

GeorgiChochov: Not just IDE, scalastyle produces a report with warnings and errors found, unless they are wrapped in comments like these. For legitimate cases (a method can only be as simple as its use case) where exceptions need to be made, this prevents unnecessary warnings from stacking up.

Issue #348, opened 2019-04-16
Add _INFO comparison


Zejnilovic:
## Background
Add _INFO file comparison to the test utils suite and enable it in the E2ERunner. This should not compare dates and usernames but focus on the checkpoints and data in general.

PR #354, opened 2019-04-26
#348 Implement _INFO file comparison

Zejnilovic:
For its heavy dependence on the Atum models, I am quite unsure if we should have it here or in Atum.

### DIFF
--- a/testutils/src/main/scala/za/co/absa/enceladus/testutils/exceptions/InfoFileComparisonExceptions.scala
+++ b/testutils/src/main/scala/za/co/absa/enceladus/testutils/exceptions/InfoFileComparisonExceptions.scala
@@ -15,9 +15,9 @@
 
 package za.co.absa.enceladus.testutils.exceptions
 
-final case class InfoFilesDifferException(private val refPath: String,
-                                               private val stdPath: String,
-                                               private val outPath: String)
+final case class InfoFilesDifferException(refPath: String,
+                                          stdPath: String,
+                                          outPath: String)
   extends Exception("Expected and actual info files differ.\n" +
     s"Reference path: $refPath\n" +
     s"Actual dataset path: $stdPath\n" +
--- a/testutils/src/main/scala/za/co/absa/enceladus/testutils/exceptions/InfoFileComparisonExceptions.scala
+++ b/testutils/src/main/scala/za/co/absa/enceladus/testutils/exceptions/InfoFileComparisonExceptions.scala
@@ -1,0 +1,24 @@
+/*
+ * Copyright 2018-2019 ABSA Group Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.enceladus.testutils.exceptions
+
+final case class InfoFilesDifferException(private val refPath: String,
+                                               private val stdPath: String,
+                                               private val outPath: String)
+  extends Exception("Expected and actual info files differ.\n" +
+    s"Reference path: $refPath\n" +
+    s"Actual dataset path: $stdPath\n" +
+    s"Difference written to: $outPath")
--- a/testutils/src/main/scala/za/co/absa/enceladus/testutils/infoFileComparison/AtumModelUtils.scala
+++ b/testutils/src/main/scala/za/co/absa/enceladus/testutils/infoFileComparison/AtumModelUtils.scala
@@ -59,8 +59,8 @@ object AtumModelUtils {
     def compareWith(otherControlMeasure: ControlMeasure): List[ModelDifference[_]] = {
       val metadataDifferences = controlMeasure.metadata.compareWith(otherControlMeasure.metadata, "metadata")
       val checkpointsDifferences = controlMeasure.checkpoints.zipWithIndex.foldLeft(List[ModelDifference[_]]()) {
-        case (agg, (value, index)) =>
-          value.compareWith(otherControlMeasure.checkpoints(index), s"checkpoints[$index]") ::: agg
+        case (agg, (checkpointValue, index)) =>
+          checkpointValue.compareWith(otherControlMeasure.checkpoints(index), s"checkpoints[$index]") ::: agg
       }
       metadataDifferences ::: checkpointsDifferences
     }
@@ -71,6 +71,9 @@ object AtumModelUtils {
     * @param metadata ControlMeasureMetadata instance
     */
   implicit class ControlMeasureMetadataOps (metadata: ControlMeasureMetadata){
+    private val stdVersion = "std_enceladus_version"
+    private val confVersion = "conform_enceladus_version"
+
     /**
       * Compare this ControlMeasureMetadata with the one passed in
       * @param otherMetadata Second control measure metadata
@@ -92,7 +95,7 @@ object AtumModelUtils {
         otherMetadata.additionalInfo,
         s"$curPath.additionalInfo")
 
-      (diffs flatten) ::: additionalInfoDiff
+      diffs.flatten ::: additionalInfoDiff
     }
 
     /**
@@ -106,10 +109,12 @@ object AtumModelUtils {
                                          is: Map[String,String],
                                          curPath: String): List[ModelDifference[_]] = {
       was.flatMap {
-        case (wasKey, wasValue) if wasKey == "conform_enceladus_version" =>
+        case (wasKey, wasValue) if wasKey == confVersion =>
           logVersionAndContinue("Conformance", wasValue, is.get(wasKey))
-        case (wasKey, wasValue) if wasKey == "std_enceladus_version" =>
+          None
+        case (wasKey, wasValue) if wasKey == stdVersion =>
           logVersionAndContinue("Standartization", wasValue, is.get(wasKey))
+          None
         case (wasKey, _) if wasKey == "std_application_id" || wasKey == "conform_application_id" => None
         case (wasKey, wasValue) =>
           is.get(wasKey) match {
@@ -117,14 +122,13 @@ object AtumModelUtils {
             case None                                 => Some(ModelDifference(s"$curPath.$wasKey", wasValue, "Null"))
             case _                                    => None
           }
-      }.toList
+      }.toList.sortBy({ f: ModelDifference[String] => f.path })
     }
 
-    private def logVersionAndContinue(name: String, refVersion: String, newVersion: Option[String]): Option[Nothing] = {
+    private def logVersionAndContinue(name: String, refVersion: String, newVersion: Option[String]): Unit = {
       logger.info(s"$name versions is:")
       logger.info(s"Reference - $refVersion")
-      logger.info(s"New - $newVersion")
-      None
+      logger.info(s"New - ${newVersion.getOrElse("NOT SPECIFIED")}")
     }
   }
 
@@ -146,9 +150,9 @@ object AtumModelUtils {
         simpleCompare(checkpoint.order, otherCheckpoint.order, s"$curPath.order") :: Nil
 
       val controls = checkpoint.controls.zipWithIndex.foldLeft(List[ModelDifference[_]]()) {
-        case (agg, (value, index)) =>
+        case (agg, (checkpointValue, index)) =>
           val nextPath = s"$curPath.controls[$index]"
-          value.compareWith(otherCheckpoint.controls(index), nextPath) ::: agg
+          checkpointValue.compareWith(otherCheckpoint.controls(index), nextPath) ::: agg
       }
 
       diffs.flatten ::: controls
--- a/testutils/src/main/scala/za/co/absa/enceladus/testutils/infoFileComparison/AtumModelUtils.scala
+++ b/testutils/src/main/scala/za/co/absa/enceladus/testutils/infoFileComparison/AtumModelUtils.scala
@@ -20,7 +20,7 @@ import za.co.absa.atum.model.{Checkpoint, ControlMeasure, ControlMeasureMetadata
 import za.co.absa.atum.utils.ControlUtils
 
 /**
-  *
+  * A case class for the differences and the JSONPath like path to them
   * @param path Path to the difference through the model
   * @param was Value it had in reference
   * @param is Value it has now
@@ -35,8 +35,8 @@ object ModelDifferenceParser {
   /**
     * The method returns JSON representation of a [[ModelDifference]] object
     */
-  def asJson(modelDifference: List[ModelDifference[_]]): String = {
-    ControlUtils.asJsonPretty[List[ModelDifference[_]]](modelDifference)
+  def asJson(modelDifferences: List[ModelDifference[_]]): String = {
+    ControlUtils.asJsonPretty[List[ModelDifference[_]]](modelDifferences)
   }
 }
 
@@ -53,14 +53,14 @@ object AtumModelUtils {
   implicit class ControlMeasureOps (controlMeasure: ControlMeasure){
     /**
       * Compare this ControLMeasure with the one passed in
-      * @param secondControlMeasure Second control measure
+      * @param otherControlMeasure Second control measure
       * @return Returns a list of model differences
       */
-    def compareWith(secondControlMeasure: ControlMeasure): List[ModelDifference[_]] = {
-      val metadataDifferences = controlMeasure.metadata.compareWith(secondControlMeasure.metadata, "metadata")
+    def compareWith(otherControlMeasure: ControlMeasure): List[ModelDifference[_]] = {
+      val metadataDifferences = controlMeasure.metadata.compareWith(otherControlMeasure.metadata, "metadata")
       val checkpointsDifferences = controlMeasure.checkpoints.zipWithIndex.foldLeft(List[ModelDifference[_]]()) {
         case (agg, (value, index)) =>
-          value.compareWith(secondControlMeasure.checkpoints(index), s"checkpoints[$index]") ::: agg
+          value.compareWith(otherControlMeasure.checkpoints(index), s"checkpoints[$index]") ::: agg
       }
       metadataDifferences ::: checkpointsDifferences
     }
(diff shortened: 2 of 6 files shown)


testutils/src/main/scala/za/co/absa/enceladus/testutils/infoFileComparison/AtumModelUtils.scala
```diff
@@ -0,0 +1,192 @@
...
+  implicit class ControlMeasureOps (controlMeasure: ControlMeasure){
+    /**
+      * Compare this ControLMeasure with the one passed in
+      * @param secondControlMeasure Second control measure
+      * @return Returns a list of model differences
+      */
+    def compareWith(secondControlMeasure: ControlMeasure): List[ModelDifference[_]] = {
+      val metadataDifferences = controlMeasure.metadata.compareWith(secondControlMeasure.metadata, "metadata")
+      val checkpointsDifferences = controlMeasure.checkpoints.zipWithIndex.foldLeft(List[ModelDifference[_]]()) {
+        case (agg, (value, index)) =>
```
GeorgiChochov:
```suggestion
        case (agg, (checkpoint, index)) =>
```
I think calling it `checkpoint` might improve readability a little bit.

testutils/src/main/scala/za/co/absa/enceladus/testutils/infoFileComparison/AtumModelUtils.scala
```diff
@@ -0,0 +1,192 @@
...
+      simpleCompare(metadata.sourceType, secondMetadata.sourceType, s"$curPath.sourceType") ::
+      simpleCompare(metadata.version, secondMetadata.version, s"$curPath.version") ::
+      simpleCompare(metadata.informationDate, secondMetadata.informationDate, s"$curPath.informationDate") ::
+        Nil
+
+      val additionalInfoDiff = additionalInfoComparison(metadata.additionalInfo,
+        secondMetadata.additionalInfo,
+        s"$curPath.additionalInfo")
+
+      (diffs flatten) ::: additionalInfoDiff
```
GeorgiChochov: I'm not a huge fan of postfix notation. And it is [actively discouraged](https://docs.scala-lang.org/style/method-invocation.html):

> This style is unsafe, and should not be used. Since semicolons are optional, the compiler will attempt to treat it as an infix method if it can, potentially taking a term from the next line.
This may result in unexpected compile errors at best, and happily compiled faulty code at worst. Although the syntax is used by some DSLs, it should be considered deprecated, and avoided.


Since scala 2.10 it produces a warning and since [2.13 it won't compile outright](https://github.com/scala/scala/pull/6559) (unless you specify the cpmpiler option `postfixOps` to enable its parsing)

testutils/src/main/scala/za/co/absa/enceladus/testutils/infoFileComparison/AtumModelUtils.scala
```diff
@@ -0,0 +1,192 @@
...
+      * @param was     Value it had in reference
+      * @param is Value it has now
+      * @param curPath Path to the AdditionalInfo through the model
+      * @return Returns a list of model differences
+      */
+    private def additionalInfoComparison(was: Map[String, String],
+                                         is: Map[String,String],
+                                         curPath: String): List[ModelDifference[_]] = {
+      was.flatMap {
+        case (wasKey, wasValue) if wasKey == "conform_enceladus_version" =>
```
GeorgiChochov: I have a feeling this might be abusing pattern matching in a very seductive way. The guard clauses can easily be turned into a regular `if-else` thing, but I think this is much more readable. I like it quite a bit and would keep it as is, but I'm curious to get @wajda 's opinion (bring it up in CQC?)

testutils/src/main/scala/za/co/absa/enceladus/testutils/infoFileComparison/AtumModelUtils.scala
```diff
@@ -0,0 +1,192 @@
...
+      logger.info(s"New - $newVersion")
+      None
+    }
+  }
+
+  /**
+    * Checkpoint's class extension adding compareWith
+    * @param checkpoint Checkpoint instance
+    */
+  implicit class CheckpointOps (checkpoint: Checkpoint){
```
GeorgiChochov:
```suggestion
  implicit class CheckpointOps (checkpoint: Checkpoint) {
```
Tiny formatting.

testutils/src/main/scala/za/co/absa/enceladus/testutils/infoFileComparison/AtumModelUtils.scala
```diff
@@ -0,0 +1,192 @@
...
+    /**
+      * Compare this Checkpoint with the one passed in
+      * @param secondCheckpoint Second checkpoint
```
GeorgiChochov: I'd call it `otherCheckpoint`

testutils/src/main/scala/za/co/absa/enceladus/testutils/infoFileComparison/AtumModelUtils.scala
```diff
@@ -0,0 +1,192 @@
...
+  /**
+    *
+    * @param first First value or the ref value
+    * @param second Second value or the new value
+    * @param curPath Current path to the values, so they are traceable
+    * @tparam T Any value that has == implemented
+    * @return Returns an Option of ModelDifference. If None is returned, there is no difference in the two values
+    */
+  private def simpleCompare[T](first: T, second: T, curPath: String): Option[ModelDifference[T]] = {
+    if (first != second) Some(ModelDifference(curPath, first, second))
```
GeorgiChochov: Our Scalastyle is set up to require curly braces for multiline `if-else`

testutils/src/main/scala/za/co/absa/enceladus/testutils/infoFileComparison/InfoFileComparisonJob.scala
```diff
@@ -0,0 +1,87 @@
...
+import za.co.absa.atum.utils.ARMImplicits
+import za.co.absa.enceladus.testutils.exceptions.InfoFilesDifferException
+import za.co.absa.enceladus.testutils.infoFileComparison.AtumModelUtils._
+import za.co.absa.enceladus.utils.time.TimeZoneNormalizer
+
+object InfoFileComparisonJob {
+  private val log: Logger = LogManager.getLogger(this.getClass)
+
+  def main(args: Array[String]): Unit = {
+    val cmd: CmdConfig = CmdConfig.getCmdLineArguments(args)
```
GeorgiChochov: I don't think the type annotation is necessary

testutils/src/main/scala/za/co/absa/enceladus/testutils/infoFileComparison/InfoFileComparisonJob.scala
```diff
@@ -0,0 +1,87 @@
...
+      .getOrCreate()
+    TimeZoneNormalizer.normalizeAll(Seq(sparkSession))
+
+    implicit val sc: SparkContext = sparkSession.sparkContext
+    val hadoopConfiguration = sparkSession.sparkContext.hadoopConfiguration
+
+    val newPathLoader = new ControlMeasuresLoaderJsonFile(hadoopConfiguration, new Path(cmd.newPath))
+    val refPathLoader = new ControlMeasuresLoaderJsonFile(hadoopConfiguration, new Path(cmd.refPath))
+
+    val newControlMeasure: ControlMeasure = newPathLoader.load()
```
GeorgiChochov: I think the variable name is descriptive enough, no need for the type annotation.

testutils/src/main/scala/za/co/absa/enceladus/testutils/infoFileComparison/InfoFileComparisonJob.scala
```diff
@@ -0,0 +1,87 @@
...
+    val enableWholeStage = false //disable whole stage code gen - the plan is too long
+
+    implicit val sparkSession: SparkSession = SparkSession.builder()
+      .appName(s"_INFO file comparison - '${cmd.newPath}' and '${cmd.refPath}'")
+      .config("spark.sql.codegen.wholeStage", enableWholeStage)
+      .getOrCreate()
+    TimeZoneNormalizer.normalizeAll(Seq(sparkSession))
+
+    implicit val sc: SparkContext = sparkSession.sparkContext
+    val hadoopConfiguration = sparkSession.sparkContext.hadoopConfiguration
```
GeorgiChochov: Is this the only reason for the `SparkSession`? I feel like this doesn't need to be a spark job, it does only with metadata, no actual data. You could create a `hadoopConfiguration` from `HADOOP_CONF_DIR`'s `core-site.xml` and `hdfs-site.xml` I think.

testutils/src/main/scala/za/co/absa/enceladus/testutils/exceptions/InfoFileComparisonExceptions.scala
```diff
@@ -0,0 +1,24 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.enceladus.testutils.exceptions
+
+final case class InfoFilesDifferException(private val refPath: String,
+                                               private val stdPath: String,
```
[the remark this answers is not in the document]
Zejnilovic: TBH, I did not give it a thought, I just did, what I did in `CompareJobExceptions.scala`. But I can change this + the alignment.

testutils/src/main/scala/za/co/absa/enceladus/testutils/infoFileComparison/AtumModelUtils.scala
Zejnilovic: so `otherCheckpoint`, `otherMeasurement`, etc.?

testutils/src/main/scala/za/co/absa/enceladus/testutils/infoFileComparison/AtumModelUtils.scala
Zejnilovic: make it a compromise? `checkpointValue`?
`checkpoint` on its own could be misleading when it is only some string that it is going to be compared to another.

Zejnilovic:
Current state is

```
─$ hadoop jar $TEST_UTILS_JAR za.co.absa.enceladus.testutils.infoFileComparison.InfoFileComparisonJob
Exception in thread "main" java.io.IOException: Mkdirs failed to create /var/folders/83/gtj4k16s04z6kw77rxqhk8f10dz_4k/T/hadoop-unjar7830096556776904179/META-INF/license
	at org.apache.hadoop.util.RunJar.ensureDirectory(RunJar.java:129)
	at org.apache.hadoop.util.RunJar.unJar(RunJar.java:104)
	at org.apache.hadoop.util.RunJar.unJar(RunJar.java:81)
	at org.apache.hadoop.util.RunJar.run(RunJar.java:209)
	at org.apache.hadoop.util.RunJar.main(RunJar.java:136)
```

GeorgiChochov:
Thanks for the effort! Let's check it out when we're back on the 8th. Tagging it as "help wanted" and "work in progress" and moving it to "icebox" for now. You can continue with other tasks.

yruslan:
Will take a look at the code on Monday and we can investigate the Hadoop configuration issues as well.

testutils/src/main/scala/za/co/absa/enceladus/testutils/infoFileComparison/AtumModelUtils.scala
yruslan: @GeorgiChochov Are you sure `:::` is a postfix notation? Isn't it just a binary operator?

testutils/src/main/scala/za/co/absa/enceladus/testutils/infoFileComparison/AtumModelUtils.scala
```diff
@@ -0,0 +1,196 @@
...
+      * Compare additional information from ControlMeasureMetadata
+      * @param was     Value it had in reference
+      * @param is Value it has now
+      * @param curPath Path to the AdditionalInfo through the model
+      * @return Returns a list of model differences
+      */
+    private def additionalInfoComparison(was: Map[String, String],
+                                         is: Map[String,String],
+                                         curPath: String): List[ModelDifference[_]] = {
+      was.flatMap {
```
yruslan: Because `was` is a Map it can be a hash map. So the order of elements can differ depending on various circumstances. So from time to time, the list of differences can appear in a different order.

Consider either
* Convert it to a List first, sort it, and then apply `flatMap` or
* Sort the final list of `ModelDifference` instances before returning.

testutils/src/main/scala/za/co/absa/enceladus/testutils/infoFileComparison/AtumModelUtils.scala
GeorgiChochov: `:::` is fine, `diffs flatten` is postfix notation (`diffs.flatten()` is preferable).

testutils/src/main/scala/za/co/absa/enceladus/testutils/infoFileComparison/AtumModelUtils.scala
```diff
@@ -0,0 +1,196 @@
...
+        case (wasKey, wasValue) if wasKey == "conform_enceladus_version" =>
+          logVersionAndContinue("Conformance", wasValue, is.get(wasKey))
+        case (wasKey, wasValue) if wasKey == "std_enceladus_version" =>
+          logVersionAndContinue("Standartization", wasValue, is.get(wasKey))
```
yruslan: Consider creating constants for  `"std_enceladus_version"` and `"conform_enceladus_version"`.

testutils/src/main/scala/za/co/absa/enceladus/testutils/infoFileComparison/AtumModelUtils.scala
yruslan: Oh... another reason not to use them. I didn't notice it and saw `diffs_flatten`

testutils/src/main/scala/za/co/absa/enceladus/testutils/infoFileComparison/AtumModelUtils.scala
```diff
@@ -0,0 +1,196 @@
...
+        case (wasKey, wasValue) =>
+          is.get(wasKey) match {
+            case Some(isValue) if wasValue != isValue => Some(ModelDifference(s"$curPath.$wasKey", wasValue, isValue))
+            case None                                 => Some(ModelDifference(s"$curPath.$wasKey", wasValue, "Null"))
+            case _                                    => None
+          }
+      }.toList
+    }
+
+    private def logVersionAndContinue(name: String, refVersion: String, newVersion: Option[String]): Option[Nothing] = {
```
yruslan: I'd suggest this function return a `Unit` and insert explicit `None` to pattern matching lines.
The reason being 2 surprises that gave me a hard time reading the code:
1. I was wondering why several cases invoke `logVersionAndContinue()` while others return an Optional.
2. `Optional[Nothing]` is a surprise by itself. Looks like a boolean, but for some reason, it is not :)

testutils/src/main/scala/za/co/absa/enceladus/testutils/infoFileComparison/AtumModelUtils.scala
yruslan: I also like it as is

yruslan:
Ran the INFO file comparison tool on Asgard. Worked perfectly when used `spark-submit`:
```
/opt/spark-2.4.3/bin/spark-submit --class za.co.absa.enceladus.testutils.infoFileComparison.InfoFileComparisonJob enceladus-testUtils-0.99.0-SNAPSHOT.jar  --ref-path info1.json --new-path info2.json --out-path info3.json
```

When running using `hadoop` it failed on Scala being not in the classpath:
```
hadoop jar enceladus-testUtils-0.99.0-SNAPSHOT.jar za.co.absa.enceladus.testutils.infoFileComparison.InfoFileComparisonJob  --ref-path info1.json --new-path info2.json --out-path info3.json
```

```
Exception in thread "main" java.lang.NoClassDefFoundError: scala/collection/Seq
        at za.co.absa.enceladus.testutils.infoFileComparison.InfoFileComparisonJob.main(InfoFileComparisonJob.scala)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
```

The command that worked for me adds Scala jars to the classpath:
```
java -cp /opt/spark-2.4.0/jars/*:enceladus-testUtils-0.99.0-SNAPSHOT.jar za.co.absa.enceladus.testutils.infoFileComparison.InfoFileComparisonJob  --ref-path info1.json --new-path info2.json --out-path info3.json
```

Currently, on our environments Scala libraries are installed only as a part of Spark.

So let's discuss what is the best approach to launch the INFO file comparison tool.
For me it seems that because the tool has Hadoop dependency and Scala dependency, it might not be a bad idea to run it using `spark-submit`...

yruslan:
As a separate thought I'm wondering it the tool can also support loading files from a local file system. In Cobrix we have a convention if a path starts with `file://` it points to a local file system.

Having a local filesystem INFO file comparison can help with automating tests that rely on Spark running in local more in an environment without HDFS.

yruslan:
Added a commit that bundles Scala and json4s+jackson into testUtils jar file.

Now the tool should be runnable from the command line via 'hadoop' and 'java -cp'. I think bundling Scala in this case is completely reasonable. And the tool does not rely on Spark being installed. Moreover, it worked on my local filesystem out of the box when running via `hadoop` command.

Zejnilovic:
> Added a commit that bundles Scala and json4s+jackson into testUtils jar file.
> 
> Now the tool should be runnable from the command line via 'hadoop' and 'java -cp'. I think bundling Scala in this case is completely reasonable. And the tool does not rely on Spark being installed. Moreover, it worked on my local filesystem out of the box when running via `hadoop` command.

To be honest I still get some errors. Let's check it on Thursday
```bash
Exception in thread "main" java.lang.NoClassDefFoundError: org/apache/hadoop/conf/Configuration
	at za.co.absa.enceladus.testutils.infoFileComparison.InfoFileComparisonJob$.getHadoopConfiguration(InfoFileComparisonJob.scala:80)
	at za.co.absa.enceladus.testutils.infoFileComparison.InfoFileComparisonJob$.main(InfoFileComparisonJob.scala:33)
	at za.co.absa.enceladus.testutils.infoFileComparison.InfoFileComparisonJob.main(InfoFileComparisonJob.scala)
Caused by: java.lang.ClassNotFoundException: org.apache.hadoop.conf.Configuration
	at java.net.URLClassLoader.findClass(URLClassLoader.java:382)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:349)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	... 3 more
```

Zejnilovic:
> As a separate thought I'm wondering it the tool can also support loading files from a local file system. In Cobrix we have a convention if a path starts with `file://` it points to a local file system.
> 
> Having a local filesystem INFO file comparison can help with automating tests that rely on Spark running in local more in an environment without HDFS.

@yruslan please check the last [commit](https://github.com/AbsaOSS/enceladus/pull/354/commits/d32009997829febf331dacdcf5778a7f2f544beb). Is this what you wanted?

yruslan:
@Zejnilovic, yes, looks great! Worked perfectly on my local files.

Issue #224, opened 2019-03-07
Refactor JS services


GeorgiChochov:
## Background
Currently our JS services in Menas have a lot of common functionality which is duplicated instead of being reused in a generic way. This makes it difficult to maintain. 

## Example
From:
```
    this._model.setProperty("/newMappingTable", {
      name: "",
      description: "",
      schemaName: oFirstSchema._id,
      schemaVersion: oFirstSchema.latestVersion,
      hdfsPath: "/",
      isEdit : false,
      title : "Add"
    });
```
To:
```
    this._addDialog.setModel(new sap.ui.model.json.JSONModel({
      name: "",
      description: "",
      schemaName: oFirstSchema._id,
      schemaVersion: oFirstSchema.latestVersion,
      hdfsPath: "/",
      isEdit : false,
      title : "Add"
    }), "entity");
```

## Proposed Solution 
1. Use ECMAScript 6 style class inheritance
2. Bind entity models to controls rather than the root model

Issue #363, opened 2019-05-04
Editing Conformance Rules does not display preselected input field


GeorgiChochov:
## Describe the bug
Editing conformance rules which have an input column do not show it as preselected. Instead, no column is selected in the dialog.

## To Reproduce
Steps to reproduce the behavior OR commands run:
1. Go to any dataset with a conformance rule that has an input column (e.g., CastingConfromanceRule)
2. Click on the `Conformance Rules` tab
3. Open the rule's `Edit` dialog
4. See no input column is preselected
5. Open browser's developer tools
6. See exception

## Expected behaviour
The input column should be preselected on edit.

## Screenshots
![image](https://user-images.githubusercontent.com/20905867/57173071-c6d3e200-6e29-11e9-86f2-683f6377fcf0.png)

## Additional context
Exception in browser console:
```
Uncaught Error: Property "title" does not exist in Element sap.m.CustomTreeItem#__item20-CastingConformanceRule--schemaFieldSelector-0
    at f.j.getProperty (ManagedObject-dbg.js:1338)
    at eval (upsert.controller.js?eval:270)
    at Array.find (<anonymous>)
    at _preselectRadioButton (upsert.controller.js?eval:270)
    at constructor.preselectSchemaFieldSelector (upsert.controller.js?eval:289)
    at constructor.onAfterOpen (upsert.controller.js?eval:100)
    at f.b.fireEvent (EventProvider-dbg.js:228)
    at f.b.fireEvent (Element-dbg.js:531)
    at f.fireAfterOpen (ManagedObjectMetadata-dbg.js:762)
    at f.j._handleOpened (Dialog-dbg.js:648)
```

GeorgiChochov:
Resolved in #393

PR #369, opened 2019-05-06
 #224 Refactor EntityServices

GeorgiChochov:
This is a far from perfect refactoring of the JS entity services. To do better would require breaking everything down into components that have their domain-specific model and behaviour encapsulated rather than these bloated layers which still have leaky responsibilities. This PR is just to set up exactly this sort of further refactoring.


GeorgiChochov:
2. Is not new #363 
I'll check out 1. that doesn't seem right. Thanks!

Zejnilovic:
> I can't reproduce 1.
> I think 2. is just a variation of the same error as #363, unable to preselect the input field.

1. seems like lingering cache problem. I take it back.

Issue #374, opened 2019-05-06
Extract Dataset Info component


GeorgiChochov:
## Background
See #371 

## Feature
Extract Dataset Info component.

PR #383, opened 2019-05-10
#374 Extract Dataset Info component

GeorgiChochov:


menas/ui/components/dataset/datasetDetail.controller.js
```diff
@@ -15,8 +15,10 @@
 sap.ui.define([
   "sap/ui/core/mvc/Controller",
   "sap/ui/core/Fragment",
-  "sap/m/MessageBox"
-], function (Controller, Fragment, MessageBox) {
+  "sap/m/MessageItem",
+  "sap/m/MessageBox",
+  "sap/m/MessagePopover"
```
Zejnilovic: Is there a future reason to have these three here?

menas/ui/components/dataset/datasetDetail.controller.js
```diff
@@ -164,7 +166,7 @@ sap.ui.define([
     },
 
     fetchSchema: function (oEv) {
-      let dataset = sap.ui.getCore().getModel().getProperty("/currentDataset");
+      let dataset = this._model.getProperty("/currentDataset");
```
Zejnilovic: Would it be ok, to apply this change to two more occurrences in the file?

menas/ui/components/dataset/datasetDetail.controller.js
```diff
@@ -206,15 +208,21 @@ sap.ui.define([
...
+        this._datasetService.getLatestByName(oParams.id).then(() => this.load())
       } else {
-        this._datasetService.getByNameAndVersion(oParams.id, oParams.version)
+        this._datasetService.getByNameAndVersion(oParams.id, oParams.version).then(() => this.load())
       }
       this.byId("datasetIconTabBar").setSelectedKey("info");
     },
 
+    load: function() {
+      let current = this._model.getProperty("/currentDataset");
```
Zejnilovic: This variable is somewhere called current and somewhere dataset. Having it named the same way might improve readability, but generally, I don't mind keeping it as is, because they have quite the small scope.

menas/ui/components/dataset/datasetDetail.controller.js
GeorgiChochov: No, that must have been a left over. Thanks for spotting it.

Issue #377, opened 2019-05-06
Extract Schema Table


GeorgiChochov:
## Background
See #371 

## Feature
Extract Schema Table component to be reused across the different entity views.

PR #388, opened 2019-05-10
 #377 Extract SchemaTable component

GeorgiChochov:


menas/ui/components/SchemaTable.js
```diff
@@ -0,0 +1,67 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+class SchemaTable {
+
+  constructor(oController) {
+    this._oController = oController;
+    this._model;
```
Zejnilovic: Is this intentional?

menas/ui/components/SchemaTable.js
```diff
@@ -0,0 +1,67 @@
...
+    this._schemaTable = oController.byId("schemaFieldsTreeTable")
```
Zejnilovic: missing `;`

menas/ui/components/SchemaTable.js
```diff
@@ -0,0 +1,67 @@
...
+  get oController() {
+    return this._oController;
+  }
+
+  get model() {
+    return this._model
+  }
+
+  set model(newModel) {
+    newModel = new sap.ui.model.json.JSONModel(newModel)
```
Zejnilovic: missing `;`

menas/ui/components/SchemaTable.js
```diff
@@ -0,0 +1,67 @@
...
+    this.schemaTable.setModel(newModel, "schema");
+    this._model = newModel
+  }
+
+  metadataPress(oEv) {
+    let binding = oEv.getSource().getBindingContext("schema").getPath() + "/metadata";
+    let bindingArr = binding + "Arr";
+    //hmm bindAggregation doesn't take formatter :-/
+    const model = this.oController.byId("schemaFieldsTreeTable").getModel("schema")
```
Zejnilovic: missing `;`

menas/ui/components/SchemaTable.js
GeorgiChochov: It was, but I guess I don't need to declare the field in the constructor if I leave it undefined.

menas/ui/components/SchemaTable.js
```diff
@@ -31,11 +30,11 @@ class SchemaTable {
   }
 
   get model() {
-    return this._model
+    return this._model;
```
Zejnilovic: I think we don't need a semicolon here. As to why I refer to [EC6 Language Specification](https://www.ecma-international.org/ecma-262/7.0/index.html#sec-automatic-semicolon-insertion)

GeorgiChochov: Good point. (I'm actually leaning towards forgoing the semicolons entirely)

Zejnilovic:
There is an issue introduced by this PR it seems (as in I cannot reproduce it on develop)
When I try to change the schema in new/edit pop-up, the versions don't update.

May-10-2019 18-23-04

menas/ui/components/SchemaTable.js
Zejnilovic: I like them to be honest. But I have no real reason to be against them. It is just a readability thing for me. It eases my mind when I am sure that the line ends there.

GeorgiChochov:
> There is an issue introduced by this PR it seems (as in I cannot reproduce it on develop)
> When I try to change the schema in new/edit pop-up, the versions don't update.
> 
> May-10-2019 18-23-04

Thanks for spotting it!

Zejnilovic:
Second Issue. When I update a dataset's schema while having the schema tab open, it does not update.

Zejnilovic:
First issue solved.

Zejnilovic:
There still seems to be a problem with schemas view as to it not updating. After refresh I had version 7.

May-13-2019 16-08-23

Zejnilovic:
Now I am not sure if this is a feature or another issue, but when you update any model, the left menu with others and their latest versions doesn't update

I am talking about this left menu.
image

Issue #376, opened 2019-05-06
Extract Entity Dialog components


GeorgiChochov:
## Background
See #371. Entity Dialogs (Add/Edit) are nearly isolated as components, but they have a lot of duplicated logic, which can be extracted to a common superclass.

## Feature
Extract Entity Dialog components in a class hierarchy.

PR #389, opened 2019-05-13
#376 Extract Entity Dialog components

GeorgiChochov:


menas/ui/service/EntityDialog.js
```diff
@@ -0,0 +1,352 @@
...
+    this.oController.byId("newSchemaName").setValueStateText("");
+  }
+
+  isValid(oSchema) {
+    this.resetValueState();
+
+    let hasValidName = EntityValidationService.hasValidName(oSchema, "Schema",
+      this.oController.byId("newSchemaName"));
+
+    return hasValidName;
```
Zejnilovic: The variable assignment is not needed here.

menas/ui/service/EntityDialog.js
```diff
@@ -0,0 +1,352 @@
...
+    }
+  }
+
+}
+
+class AddSchemaDialog extends SchemaDialog {
+
+  constructor(oDialog, schemaService, oController) {
+    super(oDialog, schemaService, oController);
+  }
```
Zejnilovic: This constructor is not necessary.

menas/ui/service/EntityDialog.js
```diff
@@ -0,0 +1,352 @@
...
+    this.oDialog.open();
+  }
+
+}
+
+class EditSchemaDialog extends SchemaDialog {
+
+  constructor(oDialog, schemaService, oController) {
+    super(oDialog, schemaService, oController);
+  }
```
Zejnilovic: This constructor is not necessary.

menas/ui/service/EntityDialog.js
```diff
@@ -0,0 +1,352 @@
...
+
+}
+
+class SchemaDialog extends EntityDialog {
+
+  constructor(oDialog, schemaService, oController) {
+    super(oDialog, schemaService, oController)
+    oController.byId("newSchemaAddButton").attachPress(this.submit, this)
+    oController.byId("newSchemaCancelButton").attachPress(this.cancel, this)
+    oController.byId("newSchemaName").attachChange(this.onNameChange, this)
```
Zejnilovic: Missing `;` 3x

menas/ui/service/EntityDialog.js
```diff
@@ -0,0 +1,352 @@
...
+
+class MappingTableDialog extends EntityDialog {
+
+  constructor(oDialog, mappingTableService, schemaService, oController) {
+    super(oDialog, mappingTableService, oController)
+    this._schemaService = schemaService;
+    oController.byId("newMappingTableAddButton").attachPress(this.submit, this)
+    oController.byId("newMappingTableCancelButton").attachPress(this.cancel, this)
+    oController.byId("newMappingTableName").attachChange(this.onNameChange, this)
+    oController.byId("schemaNameSelect").attachChange(this.onSchemaSelect, this)
```
Zejnilovic: Missing `;` 4x
Line `253` theoretically doesn't need it, because it is the last line of a function

menas/ui/service/EntityDialog.js
```diff
@@ -0,0 +1,352 @@
...
+    })
+  }
+
+}
+
+class EditMappingTableDialog extends MappingTableDialog {
+
+  constructor(oDialog, datasetService, schemaService, oController) {
+    super(oDialog, datasetService, schemaService, oController);
+  }
```
Zejnilovic: This constructor is not necessary.

menas/ui/service/EntityDialog.js
```diff
@@ -0,0 +1,352 @@
...
+      this.oDialog.getModel("entity"), "/schemaVersion")
+  }
+
+}
+
+class AddMappingTableDialog extends MappingTableDialog {
+
+  constructor(oDialog, datasetService, schemaService, oController) {
+    super(oDialog, datasetService, schemaService, oController);
+  }
```
Zejnilovic: This constructor is not necessary.

menas/ui/service/DialogFactory.js
```diff
@@ -0,0 +1,133 @@
...
+  get oDialog() {
+    return this._oDialog;
+  }
+
+}
+
+class SchemaDialogFactory extends EntityDialogFactory {
+
+  constructor(oController, fnLoad) {
+    super(oController, fnLoad, "components.schema.addSchema", "addSchemaDialog")
```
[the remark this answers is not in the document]
GeorgiChochov: Not final, but initially we went with always add `;` at the end, so as to not have to worry about the rules for when they are inferred and when not. If we started this fresh, I'd probably avoid them all together. The thing is, if you know the rules for when they are necessary and when they can be safely left out, why bother adding them at all? It only makes sense to add them out of fear that you'll miss a case where the semicolon cannot be inferred.

menas/ui/service/DialogFactory.js
```diff
@@ -0,0 +1,133 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+class EntityDialogFactory {
+
+  constructor(oController, fnLoad, fragmentName, dialogId) {
```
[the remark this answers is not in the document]
GeorgiChochov: It's a callback for `Fragment.load`, `fn` means `function` in Hungarian notation (which I'd like to get rid of at some point, but can't be bothered to do it everywhere)

menas/ui/service/EntityDialog.js
GeorgiChochov: I know, I left it there for readability and to make it easy to just add more validations if necessary.

GeorgiChochov:
This implicit constructor inheritance is very odd to me, but I'll go with it for now. In JS, the constructor is just a function, so it makes sense that it's inherited, but it makes the Java part of my brain hurt.

menas/ui/service/DialogFactory.js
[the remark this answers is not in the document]
Zejnilovic: > ... The thing is, if you know the rules for when they are necessary and when they can be safely left out, why bother adding them at all? ...

It might because it is then easier to do changes, not being concerned about lines around your code. And probably for new people, it is also more "secure"

PR #397, opened 2019-05-15
Add support for the old Menas JSON schema format for schema upload

yruslan:
So this is actually not a separate conversion tool but the support of old Menas schema JSON format support by the schema uploader.

### DIFF
--- a/menas/src/main/scala/za/co/absa/enceladus/menas/utils/converters/SparkMenasSchemaConvertor.scala
+++ b/menas/src/main/scala/za/co/absa/enceladus/menas/utils/converters/SparkMenasSchemaConvertor.scala
@@ -39,7 +39,8 @@ class SparkMenasSchemaConvertor @Autowired()(val objMapper: ObjectMapper) {
         convertMenasModel0JsonToStructType(inputJson) match {
           case Left(schema) => schema
           case Right(message2) =>
-            throw new IllegalStateException(s"StructType serializer: $message1\nMenas serializer: $message2")
+            throw new IllegalStateException(
+              s"Unable to parse schema JSON\nStructType serializer: $message1\nMenas serializer: $message2")
         }
     }
   }
--- a/menas/src/main/scala/za/co/absa/enceladus/menas/utils/converters/SparkMenasSchemaConvertor.scala
+++ b/menas/src/main/scala/za/co/absa/enceladus/menas/utils/converters/SparkMenasSchemaConvertor.scala
@@ -59,11 +59,11 @@ class SparkMenasSchemaConvertor @Autowired()(val objMapper: ObjectMapper) {
   }
 
   /**
-    * Converts a JSON in Menas 0.9.* (Model 0) format into an instance of Spark's StructType
+    * Converts a model 0 JSON (pre-release Menas) format into an instance of Spark's StructType
     */
   def convertMenasModel0JsonToStructType(inputJson: String): Either[StructType, String] = {
     try {
-      val schema = model0.Serializer.convertFromModel0ToStructType(inputJson)
+      val schema = model0.Serializer.convertToStructType(inputJson)
       Left(schema)
     }
     catch {
--- a/menas/src/main/scala/za/co/absa/enceladus/menas/utils/converters/SparkMenasSchemaConvertor.scala
+++ b/menas/src/main/scala/za/co/absa/enceladus/menas/utils/converters/SparkMenasSchemaConvertor.scala
@@ -23,8 +23,27 @@ import com.fasterxml.jackson.databind.ObjectMapper
 import org.springframework.stereotype.Component
 import java.io.ByteArrayOutputStream
 
+import scala.util.control.NonFatal
+
 @Component
 class SparkMenasSchemaConvertor @Autowired()(val objMapper: ObjectMapper) {
+
+  def convertAnyToStructType(inputJson: String): StructType = {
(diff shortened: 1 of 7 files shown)


menas/src/main/scala/za/co/absa/enceladus/menas/utils/converters/model0/VersionedModel.scala
```diff
@@ -0,0 +1,26 @@
...
+ * limitations under the License.
+ */
+
+package za.co.absa.enceladus.menas.utils.converters.model0
+
+trait VersionedModel {
+  val name: String
+  val version: Int
+
+  /*  val dateCreated: LocalDateTime
```
GeorgiChochov: I think we can remove these commented out bits.

menas/src/main/scala/za/co/absa/enceladus/menas/utils/converters/model0/Serializer.scala
```diff
@@ -0,0 +1,113 @@
...
+    .registerModule(DefaultScalaModule)
+    .registerModule(new JavaTimeModule())
+    .configure(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS, false)
+
+  implicit private val formatsJson: Formats = Serialization.formats(NoTypeHints).withBigDecimal
+
+  /**
+    * Converts a JSON in Model 0 format to a Spark StructType
+    */
+  def convertFromModel0ToStructType(json: String): StructType = convertToStructType(deserialize(json))
```
GeorgiChochov: I have a feeling `Model0` might be a bit too much concrete detail. It tells me that this serializer might also handle conversion from `Model1`, `Model2`, etc, but why is it in package `model0` then? If the package isn't explicit enough, I think this information should be made clear in the class name. I like how you use `model0.Serializer` where this is being called, so I don't think even that is necessary.

This method seems to introduce a noisy wrapper for `convertToStructType`. I would rename it to overload `convertToStructType(json: String)`, thus giving options if you pass in the Schema or the json directly.

menas/src/test/scala/za/co/absa/enceladus/menas/schema/SchemaConvertersSuite.scala
```diff
@@ -0,0 +1,412 @@
...
+          StructField("field14",StringType,true))),true),true))),true),
+      StructField("field15",StringType,true)
+    ))
+
+    val actualSchema = sparkConverter.convertAnyToStructType(inputJson)
+
+    assert(actualSchema.prettyJson == expectedSchema.prettyJson)
+  }
+
+  test("Test 0.9.* Menas to StructType schema conversion") {
```
GeorgiChochov: I would keep this version number out of the project, just call it `alpha Menas` or `pre-release Menas`.

menas/src/main/scala/za/co/absa/enceladus/menas/utils/converters/SparkMenasSchemaConvertor.scala
```diff
@@ -23,9 +23,54 @@ import com.fasterxml.jackson.databind.ObjectMapper
...
+      val schema = DataType.fromJson(inputJson).asInstanceOf[StructType]
+      Left(schema)
+    }
+    catch {
+      case NonFatal(e) => Right(e.getMessage)
+    }
+  }
+
+  /**
+    * Converts a JSON in Menas 0.9.* (Model 0) format into an instance of Spark's StructType
```
GeorgiChochov: Not keen on this `Menas version` bit, see other comment.

menas/src/main/scala/za/co/absa/enceladus/menas/utils/converters/model0/Serializer.scala
```diff
@@ -0,0 +1,113 @@
...
+    * Converts a Model 0 shema object to Spark StructType
+    */
+  def convertToStructType(schema: Schema): StructType = {
+    StructType(schema.fields.map({ menas =>
+      convertMenasToSparkField(menas)
+    }))
+  }
+
+  /**
+    * Converts a Mode 0 struct field to a Spark's StructField
```
Zejnilovic: `Mode` -> `Model`.
2 more occurrences down the line, at the same places.

menas/src/test/scala/za/co/absa/enceladus/menas/schema/SchemaConvertersSuite.scala
yruslan: yeah... it is actually 0.4.180. Not sure if 'alpha' is a proper name though, it is 1 year in prod.

menas/src/main/scala/za/co/absa/enceladus/menas/utils/converters/model0/Serializer.scala
```diff
@@ -0,0 +1,113 @@
...
+    */
+  def deserialize(json: String): Schema = {
+    Serialization.read[Schema](json)
+  }
+
+  /**
+    * Converts a Model 0 shema object to Spark StructType
+    */
+  def convertToStructType(schema: Schema): StructType = {
+    StructType(schema.fields.map({ menas =>
```
yruslan: As per Sasa's comment, check if `convertMenasToSparkFields()` can be used here instead of map.

yruslan:
@Zejnilovic, I actually liked your comment. Will take a look if it could be simplified. Good spot!

Issue #399, opened 2019-05-16
HDFS Browser fails to retrieve folder listing due to permissions subfolders


lokm01:
## Describe the bug
HDFS Browser explodes when the user running menas doesn't have execute permission on one of the subfolders. This is due to the fact that we need to indicate whether it's empty or not (display the expansion arrow or not).

PR #401, opened 2019-05-16
#399 Add simple Hadoop authentication method

yruslan:


menas/src/main/scala/za/co/absa/enceladus/menas/HDFSConfig.scala
```diff
@@ -66,20 +62,32 @@ class HDFSConfig @Autowired()(spark: SparkSession) {
...
 
       FileSystem.get(c)
-      
-    } else FileSystem.get(hadoopConf)
+
+    } else if (authMethod == "simple") {
+      logger.info(s"Using simple hadoop authentication as $authUser")
+      val conf = hadoopConf()
+      conf.set("hadoop.security.authentication", "simple")
+      SecurityUtil.setAuthenticationMethod(AuthenticationMethod.SIMPLE, conf)
```
lokm01: Aren't those two lines doing the same thing?

yruslan: It is very likely - will remove the first one

menas/src/main/scala/za/co/absa/enceladus/menas/HDFSConfig.scala
```diff
@@ -66,20 +62,31 @@ class HDFSConfig @Autowired()(spark: SparkSession) {
...
       UserGroupInformation.reset()
-      UserGroupInformation.setConfiguration(c);
+      UserGroupInformation.setConfiguration(c)
       UserGroupInformation.loginUserFromKeytab(krb5username, krb5keytab)
 
       FileSystem.get(c)
-      
-    } else FileSystem.get(hadoopConf)
+
+    } else if (authMethod == "simple") {
```
GeorgiChochov: This `if-else` situation is starting to look a lot like pattern-matching, you could also separate each of the cases into their own methods for readability.

menas/src/main/scala/za/co/absa/enceladus/menas/HDFSConfig.scala
```diff
@@ -15,29 +15,24 @@
...
 
 @Configuration
 class HDFSConfig @Autowired()(spark: SparkSession) {
   private val logger = LoggerFactory.getLogger(this.getClass)
 
-  @Value("${za.co.absa.enceladus.menas.hadoop.auth.method}") 
+  @Value("${za.co.absa.enceladus.menas.hadoop.auth.method}")
   val authMethod: String = ""
-  @Value("${za.co.absa.enceladus.menas.hadoop.auth.krb5.debug:}") 
+  @Value("${za.co.absa.enceladus.menas.hadoop.auth.user}")
```
GeorgiChochov:
```suggestion
  @Value("${za.co.absa.enceladus.menas.hadoop.auth.user:}")
```
Shouldn't this be an optional configuration, since this is only used in the "simple" auth case? 
In Spring the expression language (`Value("${...}") allows you to specify that the property may not be set by appending a colon `:` at the end so it won't blow up on start-up.

menas/src/main/scala/za/co/absa/enceladus/menas/HDFSConfig.scala
yruslan: Thanks for the suggestion. Definitely can make the code more readable.

Issue #251, opened 2019-03-19
Check numeric type oveflow in Standardization


benedeki:
## Describe the bug
Standardization code (`TypeParser`) hints it should check on numeric types overflow/underflow. But nothing special is done there. 

## To Reproduce
File _za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala_

## Expected behaviour
Either implement the check or if the correct behavior is there by default remove the useless code fork.

## Code in mention
```
        field.dataType match {
          // here we also want to check the numeric overflow by comparing with the original value this could break with
          // trailing or leading zeros
          case _: IntegerType |
               _: ShortType |
               _: FloatType |
               _: DoubleType |
               _: DecimalType |
               _: ByteType |
               _: LongType => castedCol isNull 
          case _ => castedCol isNull
        }
```

benedeki:
Despite not directly related, it was discovered during #175 implementation. Including it in that Epic therefore.

Issue #291, opened 2019-04-04
Remove print lines from UnitTests


Zejnilovic:
## Describe the bug
Remove standard print lines from unit tests. If the print line is needed make it into a `log.debug`.
This should improve compilation time as well.

Zejnilovic:
@benedeki I think you fixed this already. Did you not?

benedeki:
Yes. I believe I addressed all occurrences - both of `print` and `dataframe.show`.
Closing the ticket, feel free to reopen in case you find it premature.

benedeki:
For the record it was done as part of several of #717 tickets

PR #407, opened 2019-05-20
#251: Handle Numeric types overflow

benedeki:
* Detecting over-/underflow and loss of precision of integer and float types
* Uni tests for the new logic
* Old Unit test modified to cope with new logic
* Support for metadata allowing +/-Infinite in float types
* Implicit method to read Boolean metadata from StructField class
* SparkTestBaseshowDateFrame method added


benedeki:
If you would come up with some other  ways how to break/verify the under-/overflow logic, I would sincerly  welcome it. Call out namely to @Zejnilovic as our head tester 😉

standardization/src/test/scala/za/co/absa/enceladus/standardization/interpreter/StandardizationInterpreter_DecimalSuite.scala
```diff
@@ -0,0 +1,144 @@
...
+    val src = seq.toDF("description","small", "big")
+    showDateFrame(src)
+
+    val std = StandardizationInterpreter.standardize(src, desiredSchema, "").cache()
+    showDateFrame(std)
+
+    val exp = Seq(
+      DecimalRow("01-Pi", Option(bd(3.14)), Option(bd(3.14))),
+      DecimalRow("02-Null", Option(zero), None, Seq(
+        ErrorMessage("stdNullError", "E00002", "Standardization Error - Null detected in non-nullable attribute", "small", Seq("null")))),
```
GeorgiChochov: Maybe it's time to extract constants for these Error codes. Too much repetition in all strings I'd say.

benedeki: Thanks, actually the `ErrorMessage` companion object has a methods for these standard type ones. I am going to use those.

utils/src/main/scala/za/co/absa/enceladus/utils/implicits/ColumnImplicits.scala
```diff
@@ -0,0 +1,36 @@
...
+ * limitations under the License.
+ */
+
+package za.co.absa.enceladus.utils.implicits
+
+import org.apache.spark.sql.Column
+import org.apache.spark.sql.functions._
+
+object ColumnImplicits {
+  implicit class RichColumn(column: Column) {
```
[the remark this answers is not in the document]
benedeki: Will use `Enhancements`. All around. E.g. `ColumnEnhancements`

standardization/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
```diff
@@ -211,10 +223,60 @@ object TypeParser {
     override def assemblePrimitiveCastLogic: Column = column.cast(field.dataType)
   }
 
-  private final case class NumericParser(field: StructField,
-                                         path: String,
-                                         origSchema: StructType,
-                                         parent: Option[Parent]) extends ScalarParser
+  private trait NumericParser extends ScalarParser
```
Zejnilovic: Is there a reason to keep NumericParser around?

standardization/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
```diff
@@ -270,23 +332,26 @@ object TypeParser {
...
       origType match {
         case _: DateType                  => castDateColumn(column)
         case _: TimestampType             => castTimestampColumn(column)
         case _: StringType                => castStringColumn(column)
-        case _: DoubleType | _: FloatType =>
+        case ot: DoubleType               =>
+          // this case covers some IBM date format where it's represented as a double ddmmyyyy.hhmmss
+          patternNeeded(ot)
+          castFractionalColumn(column, ot)
+        case ot: FloatType =>
```
Zejnilovic:
```suggestion
        case ot: FloatType                =>
```

standardization/src/test/scala/za/co/absa/enceladus/standardization/interpreter/StandardizationInterpreter_DecimalSuite.scala
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  test("From String") {
+    val seq = Seq(
+      ("01-Pi", "3.14", "3.14"),
+      ("02-Null", null, null),
+      ("03-Long", Long.MaxValue.toString, Long.MinValue.toString),
+      ("04-infinity", "-Infinity", "Infinity"),
+      ("05-Really big", "123456789123456791245678912324789123456789123456789.12",
+        "12345678912345679124567891232478912345678912345678912345678912345678912345678912345678912345678912345678912345678912345678912345678912345678912345678912345678912345678912345678912345678912345678912345678912346789123456789123456789123456789123456791245678912324789123456789123456789123456789123456789123456791245678912324789123456789123456789123456789123456789123456789123456789123456789.1"),
```
Zejnilovic: Could you use concatenation and cut the string?

standardization/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
benedeki: I anticipate it could be useful for patterns in numbers. That's why I kept it still there.

standardization/src/test/scala/za/co/absa/enceladus/standardization/interpreter/StandardizationInterpreter_DecimalSuite.scala
benedeki: I could. What benefit do you see in that?
The point is, that it's something really big value, the actual value doesn't really matter to be seen, IMO.

standardization/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
```diff
@@ -85,17 +83,30 @@ object TypeParser {
...
+      case _: IntegerType               =>
+        IntegralParser(field, path, origSchema, parent, Set(LongType), Int.MinValue, Int.MaxValue)
+      case  _: LongType                 =>
+        IntegralParser(field, path, origSchema, parent, Set.empty, Long.MinValue, Long.MaxValue)
+      case _: FloatType | _: DoubleType => FractionalParser(field, path, origSchema, parent)
+      case _: DecimalType               => DecimalParser(field, path, origSchema, parent)
+      case _: StringType                => StringParser(field, path, origSchema, parent)
+      case _: BooleanType               => BooleanParser(field, path, origSchema, parent)
+      case _: DateType                  => DateParser(field, path, origSchema, parent)
+      case _: TimestampType             => TimestampParser(field, path, origSchema, parent)
```
Zejnilovic: If I can have a full on a suggestion about keeping the repeating part minimal. I liked the previous approach a lot. The styling from the example below isn't perfect but should showcase what I mean and I know this is not a full solution, but I hope it is enough to illustrate.

```scala
val parserClass = field.dataType match {
      case _: ArrayType     => ArrayParser
      case _: StructType    => StructParser
      case _: ByteType      => IntegralParser(Set(ShortType, IntegerType, LongType), Byte.MinValue, Byte.MaxValue)
      case _: ShortType     => IntegralParser(Set(IntegerType, LongType), Short.MinValue, Short.MaxValue)
      case _: IntegerType   => IntegralParser(Set(LongType), Int.MinValue, Int.MaxValue)
      case  _: LongType     => IntegralParser(Set.empty, Long.MinValue, Long.MaxValue)
      case _: FloatType |
           _: DoubleType    => FractionalParser
      case _: DecimalType   => DecimalParser
      case _: StringType    => StringParser
      case _: BooleanType   => BooleanParser
      case _: DateType      => DateParser
      case _: TimestampType => TimestampParser
      case t =>
        throw new IllegalStateException(s"${t.typeName} is not a supported type in this version of Enceladus")
    }
    parserClass(field, path, origSchema, parent)
```
and IntegralParser

```scala
private final case class IntegralParser(field: StructField,
                                          path: String,
                                          origSchema: StructType,
                                          parent: Option[Parent]
                                         )(overflowableTypes: Set[DataType ],
                                          minValue: Long,
                                          maxValue: Long) extends NumericParser
```

standardization/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
```diff
@@ -211,10 +223,60 @@ object TypeParser {
...
+          basicLogic or (column =!= castedCol.cast(dt))
+        case DoubleType | FloatType =>
+          // same as Decimal but directly comparing fractional values is not reliable,
+          // best check for whole number is considered modulo 1.0
+          basicLogic or (column % 1.0 =!= 0.0) or column > maxValue or column < minValue
+        case ot if overflowableTypes.contains(ot) =>
+          // from these types there is the possibility of under-/overflow, extra check is needed
+          basicLogic or (castedCol =!= column.cast(LongType))
+        case StringType => basicLogic or column.contains(".")// string of decimals are not allowed
+        case _ => basicLogic
```
Zejnilovic: I would stick to keeping every implementation on a new line.

standardization/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
```diff
@@ -270,23 +332,26 @@ object TypeParser {
...
-          patternNeeded(origType)
-          castFractionalColumn(column, origType)
-        case _                            =>
-          patternNeeded(origType)
-          castNonStringColumn(column, origType)
+          patternNeeded(ot)
+          castFractionalColumn(column, ot)
+        case ot                            =>
+          patternNeeded(ot)
+          castNonStringColumn(column, ot)
```
Zejnilovic: I  would stick to keeping every implementation on a new line.

standardization/src/test/scala/za/co/absa/enceladus/standardization/interpreter/StandardizationInterpreter_DecimalSuite.scala
Zejnilovic: I am moving through the file using arrows with the cursor being at the end of the line or I am moving the lines. I need to scroll to check if brackets and commas are correct and what is at the end.

standardization/src/test/scala/za/co/absa/enceladus/standardization/interpreter/StandardizationInterpreter_DecimalSuite.scala
```diff
@@ -0,0 +1,144 @@
...
+    val seq = Seq(
+      new InputRowDoublesForDecimal("01-Pi", Math.PI),
+      InputRowDoublesForDecimal("02-Null", None, None),
+      InputRowDoublesForDecimal("03-Long", Option(Long.MaxValue.toFloat), Option(Long.MinValue.toDouble)),
+      InputRowDoublesForDecimal("04-Infinity", Option(Float.NegativeInfinity), Option(Double.PositiveInfinity)),
+      new InputRowDoublesForDecimal("05-Really big", reallyBig),
+      InputRowDoublesForDecimal("06-NaN", Option(Float.NaN), Option(Double.NaN))
+    )
+    val src = spark.createDataFrame(seq)
+    showDataFrame(src)
```
Zejnilovic: Please remove these. The issue about this #291

standardization/src/test/scala/za/co/absa/enceladus/standardization/interpreter/StandardizationInterpreter_DecimalSuite.scala
```diff
@@ -0,0 +1,144 @@
...
+
+    assertResult(exp)(std.as[DecimalRow].collect().sortBy(_.description).toList)  }
+}
+
+case class DecimalRow(
+                          description: String,
+                          small: Option[BigDecimal],
+                          big: Option[BigDecimal],
+                          errCol: Seq[ErrorMessage] = Seq.empty
+                        )
```
Zejnilovic: Please fix the formatting here

standardization/src/test/scala/za/co/absa/enceladus/standardization/interpreter/StandardizationInterpreter_FractionalSuite.scala
```diff
@@ -0,0 +1,227 @@
...
+  }
+
+  test("With infinity from double") {
+    val reallyBig = Double.MaxValue
+    val seq = Seq(
+      new InputRowDoublesForFractional("01-Euler", Math.E),
+      InputRowDoublesForFractional("02-Null", None, None),
+      InputRowDoublesForFractional("03-Long", Option(Long.MaxValue.toFloat), Option(Long.MinValue.toDouble)),
+      InputRowDoublesForFractional("04-Infinity", Option(Float.NegativeInfinity), Option(Double.PositiveInfinity)),
+      new InputRowDoublesForFractional("05-Really big", reallyBig),
```
Zejnilovic: I don't find this appealing and/or easily understandable. 
+ I would move class definitions to the beginning.

standardization/src/test/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParserSuite.scala
```diff
@@ -25,9 +25,9 @@ class TypeParserSuite extends FunSuite with SparkTestBase {
   private implicit val udfLib: UDFLibrary = new za.co.absa.enceladus.utils.error.UDFLibrary
 
   test("Test standardize with sourcecolumn metadata") {
-    val structFieldNoMetadata = StructField("a", IntegerType)
-    val structFieldWithMetadataNotSourceColumn = StructField("b", IntegerType, nullable = false, new MetadataBuilder().putString("meta", "data").build)
-    val structFieldWithMetadataSourceColumn = StructField("c", IntegerType, nullable = false, new MetadataBuilder().putString("sourcecolumn", "override_c").build)
+    val structFieldNoMetadata = StructField("a", StringType)
+    val structFieldWithMetadataNotSourceColumn = StructField("b", StringType, nullable = false, new MetadataBuilder().putString("meta", "data").build)
+    val structFieldWithMetadataSourceColumn = StructField("c", StringType, nullable = false, new MetadataBuilder().putString("sourcecolumn", "override_c").build)
```
Zejnilovic: Why do this?
+ in general, this suite seems a bit weird in the way of what it checks.

standardization/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
benedeki: I find it a prolongation of the code that does not bring extra clarity. 🤔

standardization/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
Zejnilovic: I find this to be more the functional way and there is less white noise and fewer places for refactoring

standardization/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
Zejnilovic: Keeping one style is my point so that the reader doesn't have to switch

standardization/src/test/scala/za/co/absa/enceladus/standardization/interpreter/StandardizationInterpreter_DecimalSuite.scala
Zejnilovic: Now I am looking better at the method. How do you switch it on?

standardization/src/test/scala/za/co/absa/enceladus/standardization/interpreter/StandardizationInterpreter_FractionalSuite.scala
Zejnilovic: You never have unimportant stuff in your code. And these are definitions needed for the understanding of tests. This is like reading a book backwards.

utils/src/main/scala/za/co/absa/enceladus/utils/testUtils/SparkTestBase.scala
```diff
@@ -69,12 +71,19 @@ trait SparkTestBase { self =>
   Logger.getLogger("org").setLevel(Level.WARN)
   Logger.getLogger("akka").setLevel(Level.WARN)
 
+  protected val showIsActive = false
```
yruslan: Please, add a comment explaining the purpose of `showIsActive` and `showDataFrame`.
I'm guessing the usage is changing it to `true` every time we want to see the debug output?

Why not just use `logger.debug(...)`? You can redirect `df.show(...)` to a string and then use the debug logger. The schema can also be exported to a string with `df.schema.treeString`

utils/src/main/scala/za/co/absa/enceladus/utils/implicits/ColumnImplicits.scala
```diff
@@ -0,0 +1,35 @@
...
+  implicit class ColumnEnhancements(column: Column) {
+    def isInfinite: Column = {
+      column.isin(Double.PositiveInfinity, Double.NegativeInfinity)
+    }
+
+    def isNotNumeric: Column = {
+      column.isNull or column.isNaN or column.isin(Double.PositiveInfinity, Double.NegativeInfinity)
+    }
+
+    def isNotNumericNorInfinity: Column = {
```
yruslan: Not sure the implementation matches the method name.

standardization/src/test/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParserSuites/TypeParserFromLongTypeSuite.scala
```diff
@@ -45,8 +46,21 @@ class TypeParserFromLongTypeSuite extends TypeParserSuiteTemplate {
     }
   }
 
-  test("Within the column - type stays") {
-    doTestWithinColumn(input)
+  override protected def createErrorCondition(srcField: String, target: StructField, castS: String): String = {
```
yruslan: Why you are overriding this method in several test suites with identical implementation?

standardization/src/test/scala/za/co/absa/enceladus/standardization/interpreter/StandardizationInterpreter_FractionalSuite.scala
yruslan: If these case classes are used for creating datasets we prefer to keep them in a separate file so several test suites may use them for different purposes. We consider such case classes an example model.

standardization/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
```diff
@@ -211,10 +223,60 @@ object TypeParser {
...
+      val basicLogic: Column = super.assemblePrimitiveCastErrorLogic(castedCol)
+
+      origType match {
+        case  dt: DecimalType =>
+          // decimal can be too big, to catch overflow or imprecision  issues compare to original
+          basicLogic or (column =!= castedCol.cast(dt))
+        case DoubleType | FloatType =>
+          // same as Decimal but directly comparing fractional values is not reliable,
+          // best check for whole number is considered modulo 1.0
+          basicLogic or (column % 1.0 =!= 0.0) or column > maxValue or column < minValue
```
yruslan: Not sure how this works. Let's discuss

utils/src/main/scala/za/co/absa/enceladus/utils/implicits/ColumnImplicits.scala
yruslan: Agreed to put the expressions directly at the usage places

utils/src/main/scala/za/co/absa/enceladus/utils/testUtils/SparkTestBase.scala
yruslan:
```scala
  /** Same as df.show(), bu the output goes to a string */
  def showString(df: DataFrame, numRows: Int = 20): String = {
    val outCapture = new ByteArrayOutputStream
    Console.withOut(outCapture) {
      df.show(numRows, truncate = false)
    }
    new String(outCapture.toByteArray).replace("\r\n", "\n")
  }
```

benedeki:
> 
> 
> `HelperFunctionsSuite` does not compile, as you've commented it out.

Yeah, omitted the comment opening in merge commit (shouldn't have committed this file here at first place). Fixing.

Issue #408, opened 2019-05-20
prettyPrintElapsedTime in different Locale


benedeki:
## Describe the bug
If a system Locale uses a different decimal delimiter, the output of `HelperFunctions.prettyPrintElapsedTime` is other then expected.

## To Reproduce
Change the decimal delimiter in system to _comma_. Output is considerably different from the expected one.

## Expected behaviour
Reliable result regardless of system Locale.

PR #409, opened 2019-05-20
#408: prettyPrintElapsedTime in different Locale

benedeki:
* Within the function US Locale is enforced
* Less type conversions
* Using object constants instead of computing them on every function run
* Using existing function on Integral implicit instead of defining own function of same result
* 0 milliseconds doesn't result in empty string anymore
* Swap of parameters in unit tests assert (to match expected and actual)

### DIFF
--- a/testutils/src/main/scala/za/co/absa/enceladus/testutils/HelperFunctions.scala
+++ b/testutils/src/main/scala/za/co/absa/enceladus/testutils/HelperFunctions.scala
@@ -15,6 +15,8 @@
 
 package za.co.absa.enceladus.testutils
 
+import java.util.Locale
+
 import org.apache.log4j.{LogManager, Logger}
 import org.apache.spark.sql.functions.{expr, max}
 import org.apache.spark.sql.types.{ArrayType, StructField, StructType}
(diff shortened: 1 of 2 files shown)


testutils/src/main/scala/za/co/absa/enceladus/testutils/HelperFunctions.scala
```diff
@@ -149,37 +151,39 @@ object HelperFunctions {
...
-    val hourMultiplier = minuteMultiplier * 60
-    val dayMultiplier = hourMultiplier * 24
-
-    val (numberOfDays, remainingAfterDay) = extractTimeSegment(elapsedTime, dayMultiplier)
-    val (numberOfHours, remainingAfterHours) = extractTimeSegment(remainingAfterDay, hourMultiplier)
-    val (numberOfMinutes, remainingAfterMinutes) = extractTimeSegment(remainingAfterHours, minuteMultiplier)
-    val numberOfSeconds = remainingAfterMinutes / 1000.0
+    val (numberOfDays, remainingAfterDay) = elapsedTime /% millisecondsPerDay
+    val (numberOfHours, remainingAfterHours) = remainingAfterDay /% millisecondsPerHour
+    val (numberOfMinutes, numberOfMilliseconds) = remainingAfterHours  /%  millisecondsPerMinute
```
Zejnilovic: This is quite an amazing method.

Issue #414, opened 2019-05-24
Automatically log out user is unable to logout and every click produces an error


Zejnilovic:
## Describe the bug
In case I get automatically logged out for inactivity in Menas. I am unable to log out. When selecting something I get error messages that are inaccurate. 

The way to log out is to refresh the page.

## Expected behaviour
When automatically logged out, I expect that on click I will be redirected to log in page and after successful login, my previous query would be executed.

## Screenshots
![May-24-2019 13-45-16](https://user-images.githubusercontent.com/17467768/58325545-54ec2880-7e2a-11e9-8014-7f2b1ec8d368.gif)

PR #418, opened 2019-05-27
#414 Automatically log user out after session timeout

GeorgiChochov:


Zejnilovic:
If I press edit while I am logged out I still get an error

image

GeorgiChochov:
> If I press edit while I am logged out I still get an error
> 
> image

Not all REST calls are handled by the new RestClient. Let's fix this in a separate task (`Refactor old REST calls`)?

GeorgiChochov:
Created #424 to address the above.

Issue #341, opened 2019-04-15
Improve Mapping Conformance Rule


Zejnilovic:
## Description
Let users choose the columns for join conditions from a schema picker. This picker will need to be aware of the dataset's schema and mapping table's schema.

GeorgiChochov:
When a join condition dialog is reopened to edit the join, the fields are preselected but that preselection is not reflected in the model. Pressing `accept` without changes results in no fields selected for the join condition.

PR #428, opened 2019-05-28
#341 Improve MappingConformanceRule dialog to use selectors

GeorgiChochov:


menas/ui/components/dataset/conformanceRule/ConformanceRuleDialog.js
```diff
@@ -0,0 +1,345 @@
...
+  get dataTypes() {
+    return [
+      {type: "boolean"},
+      {type: "byte"},
+      {type: "short"},
+      {type: "integer"},
+      {type: "long"},
+      {type: "float"},
+      {type: "double"},
+      {type: "decimal(38,18)"},
```
lokm01: Should the precision and scale be hardcoded here?

GeorgiChochov: It's a remnant from the old Menas. Of course it would be better to make these editable, but that's for a different task.

menas/ui/service/SchemaFieldSelector.js
```diff
@@ -106,12 +111,23 @@ class SchemaFieldSelector {
     return rawTitle.match(/<strong>(?<title>.+)<\/strong>/).groups.title;
   }
 
+  reset(tree) {
+    // This is a workaround for a bug in the Tree component of 1.56.x and 1.58.x
+    // which throws "Cannot read property 'setSelectedIndex' of undefined" error
+    // TODO: verify whether this was fixed in the subsequent versions
```
Zejnilovic: Create an Issue as we talked about?

GeorgiChochov: The workaround is good enough as is. Since we can't really fix it we don't need a task for it. And I personally don't intend to keep track of when this is fixed if ever so I don't need the TODO either.

Zejnilovic:
When I save a mapping rule it shows `^` instead of `.` in structure path for mapping table path. This vanishes after a refresh.

image

Zejnilovic:
> When I save a mapping rule it shows `^` instead of `.` in structure path for mapping table path. This vanishes after a refresh.
> 
> image

Created #431

Issue #339, opened 2019-04-15
Improve Concatenation Conformance Rule


Zejnilovic:
## Description
The selection of columns for concatenation should come from a schema picker.

Issue #314, opened 2019-04-09
Bundle Menas resources


GeorgiChochov:
## Background
Currently, Menas has a lot of `js` files which are fetched separately on load. This abundance of HTTP requests slows down loading the application. This will only grow over time. 

## Feature
We should bundle our resources and serve them all at once.

## Proposed Solution [Optional]
We can use something like
1. https://webpack.js.org
2. https://parceljs.org (less configuration than webpack, an interesting choice)
3. https://github.com/eirslett/frontend-maven-plugin (maven plugin to run npm commands as part of the build)

lokm01:
I guess this (for most part) solved by the component preload.

GeorgiChochov:
> I guess this (for most part) solved by the component preload.

Not quite, our source files are still not bundled in. #478 certainly helps, but is a different thing. A bundler would also allow transpilation (and therefore a wider browser support).

Let's put this in the icebox, but not close it.

Issue #82, opened 2018-12-28
Conformance rule output column should be reflected in schema 


GeorgiChochov:
When adding a new conformance rule, the schema presented in any input column should reflect the dataset's schema as it will appear after all changes from preceding rules have been applied to it (i.e., the original schema + all output columns - all dropped columns).

Zejnilovic:
There is an issue found after extraction of components. Now the schema isn't reflected, sometimes only part of it. Needs more look into

GeorgiChochov:
Added #461 for further improvements.

PR #441, opened 2019-05-30
#339 Add schema selector for concatenation rule

GeorgiChochov:
* Fixed broken concatenation output nullability check
* Added Schema Field Selector

### DIFF
--- a/menas/ui/index.html
+++ b/menas/ui/index.html
@@ -38,6 +38,8 @@
     <script src="components/dataset/conformanceRule/schemaManager.js"></script>
     <script src="components/dataset/conformanceRule/MappingConformanceRule/JoinConditionDialog.js"></script>
     <script src="components/dataset/conformanceRule/MappingConformanceRule/JoinConditionDialogFactory.js"></script>
+    <script src="components/dataset/conformanceRule/ConcatenationConformanceRule/ConcatenationColumnDialog.js"></script>
+    <script src="components/dataset/conformanceRule/ConcatenationConformanceRule/ConcatenationColumnDialogFactory.js"></script>
     <script src="components/SchemaTable.js"></script>
     <script src="service/MessageProvider.js"></script>
     <script src="service/RestDAO.js"></script>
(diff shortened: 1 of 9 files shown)


GeorgiChochov:
Thanks, JS refactoring can sometimes go a little funny (unintentional method rename)

GeorgiChochov:
> Added 4th rule and got Harlem Shake
> 
> `Uncaught TypeError: Cannot read property 'collapseAll' of undefined`
> 
> In general, whenever I add another field it does this. Sometimes 3 times sometimes doesn't stop.

That's wild. I can't reproduce it...

Zejnilovic:
The Harlem Shake happens when the line of the concatenation field is too long it seems.

GeorgiChochov:
I'll finish up #82 before getting back to this.

menas/ui/index.html
```diff
@@ -38,6 +38,8 @@
     <script src="components/dataset/conformanceRule/schemaManager.js"></script>
     <script src="components/dataset/conformanceRule/MappingConformanceRule/JoinConditionDialog.js"></script>
     <script src="components/dataset/conformanceRule/MappingConformanceRule/JoinConditionDialogFactory.js"></script>
+    <script src="components/dataset/conformanceRule/ConcatenationConformanceRule/ConcatenationColumnDialog.js"></script>
+    <script src="components/dataset/conformanceRule/ConcatenationConformanceRule/ConcatenationColumnDialogFactory.js"></script>
```
lokm01: I wish ES6 imports worked without these... 🙄

GeorgiChochov: They do if you use a real bundler, but UI5 doesn't play ball with ES6 or any transpiling bundler (#314). That's something I'll figure out how to make work in the future. Already wasted 2 todays on it a while ago.

GeorgiChochov:
I really can't reproduce the Harlem Shake... But I fixed the preselect error.

Zejnilovic:
Can't reproduce on new commits either.

PR #442, opened 2019-05-30
Broken transitive schema

GeorgiChochov:
* Fixed Concatenation rule bug
* Fixed issue where multiple mapping rules' output columns override each other (all are present)
* Outstanding issue: mapping rule output columns are resolved after the schema is shown and get updated when the rule is changed (mapping rule output columns are delayed)


menas/ui/components/dataset/conformanceRule/schemaManager.js
```diff
@@ -319,12 +316,20 @@ class SchemaField {
...
-class ArrayUtils {
-
-  static applyOnCopy(input, fn) {
-    const copy = $.extend(true, [], input);
-    fn(copy);
-    return copy
-  }
-
-}
+// class ArrayUtils {
```
lokm01: Let's get rid of this if its not used anymore??

PR #452, opened 2019-06-04
Feature/311 mongodb migration framework

yruslan:
This PR includes:
- A MongoDB migration framework for Menas (but can be potentially reused elsewhere)
- Unit and integration tests for the framework
- The first actual migration from the old Menas to the new one. I call it model v0 and model v1 migration.

The first migration is complicated because from the old Menas to the new one we use different Web application framework (Play vs Spring), different JSON serializers (json4s vs jackson vs MongoDB Scala driver Bson).

Migrations are not destructive, they are safe to run on existing databases. Model v0 collections are expected to have no postfix, e.g. 'dataset', 'schema'. The model v1 collections are expected to have a version postfix, e.g. 'dataset_v1', 'schema_v1'. Several models can coexist in one database sharing all MongoDB connection settings. The latest usable model is specified in 'db_version' collection.

Object Ids are retained during migrations which opens a possibility to later have a continuous migration process.

Migrations are interruption tolerant. The last thing a migration does is writing a new version into 'db_version'. If a process was interrupted before completion the version in 'db_version' will still be old. When a new migration starts it will drop all partially migrated collections and will start a migration from the beginning.


yruslan:
Please, do not post comments before the CQC meetup.

yruslan:
Let's resume commenting to this PR. Even if we decide to do a live PR review with this PR next time, the comments might only add something to the discussion.

migrations/src/main/scala/za/co/absa/enceladus/migrations/framework/Migrator.scala
```diff
@@ -0,0 +1,207 @@
...
+    */
+  private def validateCollectionManipulationConsistency(targetDbVersion: Int): Unit = {
+    var currentVersionCollections = getCollectionNames(0)
+
+    if (currentVersionCollections.isEmpty) {
+      throw new IllegalStateException("No collection names are registered for db version 0.")
+    }
+
+    for (i <- 1 to targetDbVersion) {
+      val migrationsToExecute = migrations.filter(m => m.targetVersion == i)
```
Zejnilovic: if we already tested that the versions do go in order, is it still needed to have such a filter instead of getting it by index? You could then drop the `foreach` on lines `145` and `146`

migrations/src/main/scala/za/co/absa/enceladus/migrations/MongoMigratorApp.scala
```diff
@@ -0,0 +1,51 @@
...
+/**
+  * This is a command line tool for running migrations on a MongoDb database.
+  *
+  * Syntax:
+  *   java -cp enceladus-migrations.jar za.co.absa.enceladus.migrations.MongoMigratorApp <MongoDbURL> <DbName> <TargetVersion>
+  */
+object MongoMigratorApp {
+
+  def main(args: Array[String]) {
+    if (args.length < 3) {
```
Zejnilovic: Why not use our standard `scopt`?

migrations/src/main/scala/za/co/absa/enceladus/migrations/framework/MigrationUtils.scala
```diff
@@ -0,0 +1,37 @@
...
+
+package za.co.absa.enceladus.migrations.framework
+
+object MigrationUtils {
+
+  /**
+    * Gets versioned collection name given a collection name.
+    *
+    * A collection name can be 'schema' or 'dataset', for example. The corresponding versioned collections could
+    * be 'schema_v5' or 'dataset_v5'.
```
Zejnilovic: I am thinking about this and each migration creates a new set of collections with a coresponding number. Why not create a database with the number and keep the names the same, so it is just "plug and play" in our properties files.

migrations/src/main/scala/za/co/absa/enceladus/migrations/framework/Migrator.scala
```diff
@@ -0,0 +1,207 @@
...
+  private def validateTargetVersion(targetDbVersion: Int): Unit = {
+    if (!migrations.exists(_.targetVersion == targetDbVersion)) {
+      throw new IllegalStateException(s"The target database version ($targetDbVersion) is not reachable.")
+    }
+  }
+
+  /**
+    * Validates that values of a source and a target database versions make sense.
+    */
+  private def validateDbVersios(sourceDbVersion: Int, targetDbVersion: Int): Unit = {
```
Zejnilovic:
```suggestion
  private def validateDbVersions(sourceDbVersion: Int, targetDbVersion: Int): Unit = {
```

migrations/src/main/scala/za/co/absa/enceladus/migrations/framework/Migrator.scala
```diff
@@ -0,0 +1,207 @@
...
+    val sourceDbVersion = db.getVersion()
+    validateDbVersios(sourceDbVersion, targetDbVersion)
+
+    var currentVersionCollections = getCollectionNames(sourceDbVersion)
+
+    if (currentVersionCollections.isEmpty) {
+      throw new IllegalStateException(s"No collection names are registered for db version $sourceDbVersion.")
+    }
+
+    if (targetDbVersion > sourceDbVersion) {
```
Zejnilovic: I think it would not even get to this point if this would not be true. There are a lot of validations beforehand. Like line 46 of `MigrationService.scala`

migrations/src/main/scala/za/co/absa/enceladus/migrations/framework/Migrator.scala
```diff
@@ -0,0 +1,207 @@
...
+      throw new IllegalStateException(s"Source database version cannot be negative.")
+    }
+
+    if (targetDbVersion < sourceDbVersion) {
+      throw new IllegalStateException(s"A target database version cannot be less than a source version.")
+    }
+  }
+
+  /**
+    * Removes collections that were created during an interrupted migration process.
```
Zejnilovic: There should be a warning about this in the main class with the comments.

menas/src/main/scala/za/co/absa/enceladus/menas/services/MigrationService.scala
```diff
@@ -15,42 +15,39 @@
...
-    })
-
-    // Register all evolutions here as above
-    migrator.runEvolutions()
+  @PostConstruct
+  def init(): Unit = {
+    val mongoClient = MongoClient(connectionString)
+    val db = new MongoDb(mongoClient.getDatabase(database))
+
+    if (db.isCollectionExists("schema") || db.isCollectionExists("db_version")) {
```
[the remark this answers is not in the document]
yruslan: Yeah, the intention for this check is to distinguish between an empty database and an old database (that doesn't have `db_version` collection).

- If there is no `schema` and `db_version` => the database is clearly empty, no migrations needed
- If there is `schema` => the database is not empty
- If there is `db_version` => the database is not empty. The database might not have `schema`, but `schema_v21`. But in this case it should have `db_version` as well.

But I need to improve this part of the code so that all the checks make sense. I might create a function that checks and initializes an empty database.

migrations/src/main/scala/za/co/absa/enceladus/migrations/framework/MigrationUtils.scala
yruslan: The idea behind numbered collection names is that we might have no permissions for creating a db in a preprod/prod instance, but we always have permissions to create a collection in our db. But I might be too cautious.
Lets' discuss this with others.

migrations/src/main/scala/za/co/absa/enceladus/migrations/framework/Migrator.scala
yruslan: If `targetDbVersion == sourceDbVersion` all validations should pass and no migrations should run. I'll take a look if the `for(...)` below will work just fine without the `if`.

migrations/src/main/scala/za/co/absa/enceladus/migrations/framework/MigrationUtils.scala
Zejnilovic: Actually, that makes a lot of sense.

menas/src/main/scala/za/co/absa/enceladus/menas/services/MigrationService.scala
```diff
@@ -15,42 +15,39 @@
...
-    migrator.registerOne(Evolution(1, "Create index on evolutions collection", "chochovg").setEvolution { () =>
-      mongoDb.getCollection("evolutions").createIndex(Document("""{"order":1}"""))
-    })
-
-    // Register all evolutions here as above
-    migrator.runEvolutions()
+  @PostConstruct
+  def init(): Unit = {
+    val mongoClient = MongoClient(connectionString)
+    val db = new MongoDb(mongoClient.getDatabase(database))
```
lokm01: Why are we creating a new one here? Can we not use the autowired one?

migrations/src/main/scala/za/co/absa/enceladus/migrations/MongoMigratorApp.scala
```diff
@@ -0,0 +1,51 @@
...
+import za.co.absa.enceladus.migrations.framework.dao.MongoDb
+import za.co.absa.enceladus.migrations.migrations._
+
+/**
+  * This is a command line tool for running migrations on a MongoDb database.
+  *
+  * Syntax:
+  *   java -cp enceladus-migrations.jar za.co.absa.enceladus.migrations.MongoMigratorApp <MongoDbURL> <DbName> <TargetVersion>
+  */
+object MongoMigratorApp {
```
lokm01: Could we automatically check whether we need to migrate on start up?

yruslan: Yeah, it does it as well. This is just for testing/debugging.

migrations/src/main/scala/za/co/absa/enceladus/migrations/framework/migration/CollectionMigration.scala
```diff
@@ -0,0 +1,293 @@
...
+
+  /** Returns a list of indexes to be removed during the migration */
+  def getIndexesToRemove: List[Index] = indexesToDrop.toList
+
+  /**
+    * If a migration adds or removes collections it should provide a new list of collections based
+    * on the list of collections available for he previous version of a database.
+    */
+  override def applyCollectionChanges(collections: List[String]): List[String] = {
+    var newCollections = collections
```
Zejnilovic: If we used `Set` with `++` and `--` we could omit the var. But that would not keep the order.

migrations/src/main/scala/za/co/absa/enceladus/migrations/framework/migration/CollectionMigration.scala
```diff
@@ -0,0 +1,293 @@
...
+    })
+  }
+
+  /**
+    * Applies collection renames to the provided database.
+    */
+  private def applyRenameCollection(db: DocumentDb): Unit = {
+    collectionsToRename.foreach {
+      case (oldName, newName) =>
+        val oldVersionedCollection = MigrationUtils.getVersionedCollectionName(oldName, targetVersion)
```
Zejnilovic: I don't very much get the use of this method. If I am doing migration and I want to rename something, is the target version going to be the same or -1?

migrations/src/main/scala/za/co/absa/enceladus/migrations/framework/migration/JsonMigration.scala
```diff
@@ -0,0 +1,176 @@
...
+  *
+  *     transformJSON("collection2_name") (jsonIn => {
+  *       val jsonOut = collection2Transformations(jsonIn)
+  *       jsonOut
+  *     })
+  *   }
+  * }}}
+  *
+  * If a source document is invalid for some reason and should be dropped, just return `InvalidDocument`.
+  *
```
Zejnilovic: Unnecessary comment line. More occurrences in this document.

migrations/src/main/scala/za/co/absa/enceladus/migrations/framework/migration/JsonMigration.scala
```diff
@@ -0,0 +1,176 @@
...
+
+  /**
+    * This function is used by derived classes to add transformations for affected collections.
+    * This is used for complex migrations that requite complex model version maps.
+    *
+    * @param collectionName A collection name to be migrated
+    * @param f              A transformation to applied to each document of the collection
+    *
+    */
+  def transformJSON(collectionName: String)(f: String => String): Unit = {
```
Zejnilovic: This name sounds more like an alias to `applyTransformers`. Mabe `addTransformer`

migrations/src/main/scala/za/co/absa/enceladus/migrations/framework/migration/JsonMigration.scala
```diff
@@ -0,0 +1,176 @@
...
+                                     invalidDocumentsCount: Int): Unit = {
+    val sourceDocsCnt = db.getDocumentsCount(sourceCollection)
+    val targetDocsCnt = db.getDocumentsCount(targetCollection)
+    if (invalidDocumentsCount == 0) {
+      log.info(s"Migration '$sourceCollection' -> '$targetCollection' completed successfully. " +
+        s"Total documents migrated: $targetDocsCnt.")
+    } else {
+      if (invalidDocumentsCount == sourceDocsCnt) {
+        log.error(s"Migration '$sourceCollection' -> '$targetCollection' failed. " +
+          s"Unable to convert any of $sourceDocsCnt documents.")
```
Zejnilovic: Why have 2 levels of `if/else`

migrations/src/main/scala/za/co/absa/enceladus/migrations/migrations/model1/SchemaField.scala
```diff
@@ -0,0 +1,31 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.enceladus.migrations.migrations.model1
+
+case class SchemaField
+(
+  name: String,
+  `type`: String,
```
Zejnilovic: What is the use case of backticks for case class variable?

migrations/src/main/scala/za/co/absa/enceladus/migrations/migrations/model1/Serializer1.scala
```diff
@@ -0,0 +1,123 @@
...
+    val encodeContext = EncoderContext.builder.build
+
+    schemaCodec.encode(bsonWriter, schema, encodeContext)
+
+    bsonDocument.toJson
+
+    // alternative 1
+    //objectMapper.writeValueAsString(schema)
+    // alternative 2
+    //Serialization.write(schema)
```
Zejnilovic: Are we keeping these?

migrations/src/main/scala/za/co/absa/enceladus/migrations/migrations/model1/Serializer1.scala
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  /**
+    * Deserializes a Model 1 schema JSON
+    */
+  def deserializeSchema(json: String): Schema = {
+    objectMapper.readValue(json, classOf[Schema])
+    //Serialization.read[Schema](json)
```
Zejnilovic: Presumably a residual line

migrations/src/main/scala/za/co/absa/enceladus/migrations/MongoMigratorApp.scala
lokm01: 👍

migrations/src/main/scala/za/co/absa/enceladus/migrations/framework/dao/MongoDb.scala
```diff
@@ -0,0 +1,297 @@
...
+    if (!isCollectionExists(collectionName)) {
+      throw new IllegalStateException(s"Collection does not exist: '$collectionName'.")
+    }
+    db.getCollection(collectionName).countDocuments().execute()
+  }
+
+  /**
+    * Inserts a document into a collection.
+    */
+  override def insertDocument(collectionName: String, document: String): Unit = {
```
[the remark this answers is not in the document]
yruslan: Theoretically `insertDocumet` should create a collection automatically.

But I agree, will add checks to all related methods, including `insertDocumet`. The reason is that the migration framework's `DocumentDb` API is not a general purpose API for accessing MongoDB, but an API specifically targeted for migration purposes. So making it more strict won't hurt. From the migration perspective all methods you mentioned should be used only on existing collections.

migrations/src/main/scala/za/co/absa/enceladus/migrations/framework/migration/CollectionMigration.scala
```diff
@@ -0,0 +1,293 @@
...
+    *
+    * @param oldName A collection to be renamed
+    * @param newName A new name for the collection
+    */
+  def renameCollection(oldName: String, newName: String): Unit = {
+    if (collectionsToDrop.contains(oldName)) {
+      throw new IllegalArgumentException(s"Collection '$oldName' is in the removal list.")
+    }
+    if (collectionsToCreate.contains(oldName)) {
+      throw new IllegalArgumentException(s"Collection '$oldName' is in the list of new collections. Cannot rename it.")
```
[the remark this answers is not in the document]
yruslan: Now I'm thinking if we need to check this here at all since validation does more sophisticated things... Better to check more and fail soon. So will add checks for the $newName as well.

migrations/src/main/scala/za/co/absa/enceladus/migrations/framework/migration/CollectionMigration.scala
yruslan: It is better to keep order. Will switch to ArrayBuffer to remove the var since we remove items from the middle and insert at the end.

migrations/src/main/scala/za/co/absa/enceladus/migrations/framework/migration/CollectionMigration.scala
yruslan: Migrator always clones something first, and after that applies all renames, drops, etc.

migrations/src/main/scala/za/co/absa/enceladus/migrations/framework/migration/MigrationBase.scala
```diff
@@ -0,0 +1,54 @@
...
+  * Each migration should use different migration strategies as mixings. For example,
+  * in order to create an instance of JSON to JSON transformer migration you need
+  * to define a migration like this:
+  *
+  * {{{
+  *   object MigrationTo1 extends MigrationBase with JsonMigration {...}
+  * }}}
+  */
+abstract class MigrationBase extends Migration {
+  private val log: Logger = LogManager.getLogger("MigrationBase")
```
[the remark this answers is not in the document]
yruslan: The log looked ugly during running tests since `this.getClass` returned class names with full qualifiers. But I agree, let's be consistent across the project.

migrations/src/main/scala/za/co/absa/enceladus/migrations/migrations/model0/Dataset.scala
```diff
@@ -0,0 +1,37 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.enceladus.migrations.migrations.model0
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties
+import za.co.absa.enceladus.migrations.migrations.model0.conformanceRule.ConformanceRule
+
+@JsonIgnoreProperties(ignoreUnknown = true)
```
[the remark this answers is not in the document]
yruslan: JSONs extracted from MongoDB include `_id` which is not a part of the model.

migrations/src/main/scala/za/co/absa/enceladus/migrations/migrations/model1/Dataset.scala
```diff
@@ -0,0 +1,46 @@
...
+  description: Option[String] = None,
+  
+  hdfsPath:        String,
+  hdfsPublishPath: String,
+
+  schemaName:    String,
+  schemaVersion: Int,
+
+  dateCreated: ZonedDateTime = ZonedDateTime.now(),
+  userCreated: String        = null,
```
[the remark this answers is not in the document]
yruslan: Yeah, this is probably not necessary. Just copied the current model as it. Will remove this.

migrations/src/main/scala/za/co/absa/enceladus/migrations/migrations/model1/SchemaField.scala
yruslan: Since `type` is a keyword in Scala it cannot be used as a attribute name. That's why the backticks.

migrations/src/main/scala/za/co/absa/enceladus/migrations/migrations/model1/Serializer1.scala
yruslan: Will remove this. It only shows my struggle trying out different serializers to output `ZonedDataTime` in the extended format.

menas/src/main/scala/za/co/absa/enceladus/menas/services/MigrationService.scala
```diff
@@ -15,42 +15,38 @@
...
-    migrator.runEvolutions()
+  def init(): Unit = {
+    val db = new MongoDb(mongoDb)
+    val mig = new Migrator(db, Migrations)
+
+    if (mig.isDatabaseEmpty()) {
+      log.warn(s"The database '${mongoDb.name}' is empty. Initializing...")
+      mig.initializeDatabase(ModelVersion)
+    } else {
+      if (mig.isMigrationRequired(ModelVersion)) {
```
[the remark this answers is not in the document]
Zejnilovic: I know this is nitpicky but now else is moved on line 47 but should be on line 46

dataModel/src/main/scala/za/co/absa/enceladus/model/package.scala
```diff
@@ -0,0 +1,34 @@
...
+    */
+  final val ModelVersion = 1
+
+  /**
+    * The collection postfix to use for a non-zero model version.
+    *
+    * For instance, if the collection's base name is 'dataset', for model version 0 the collection name will
+    * be 'dataset', for version 1 it will be 'dataset_v1', for version 2 it will be 'dataset_v2', etc.
+    */
+  final val CollectionPostfix = s"_v$ModelVersion"
```
GeorgiChochov: Suffix?

migrations/src/main/scala/za/co/absa/enceladus/migrations/framework/dao/DocumentDb.scala
```diff
@@ -0,0 +1,93 @@
...
+  /**
+    * Sets the version of the database. When a version is set it implies a migration to that version is completed
+    * successfully and the database can be used with the specified version of the schema.
+    */
+  def setVersion(version: Int): Unit
+
+  /**
+    * Returns true if the specified collection exists in the database.
+    */
+  def isCollectionExists(collectionName: String): Boolean
```
GeorgiChochov:
```suggestion
  def doesCollectionExist(collectionName: String): Boolean
```

Issue #456, opened 2019-06-06
Add browser check for Chromium based browsers


Zejnilovic:
## Background
We don't have time or the manpower to support other browsers. We are going to only support chromium based ones.

## Feature
Blow up on other browsers

benedeki:
I object against such behavior.
We are already getting to the situation like it was years ago with IE6. Some sites work only in chromium and it's not because it would stick to standards the most. With a feature like this, we just support browser monopoly.
Do we use any special controls or flows that fail in other browsers? Isn't our UX library actually supposed to address these - as one of the reasons to use such a library anyway?

Zejnilovic:
> I object against such behavior.
> We are already getting to the situation like it was years ago with IE6. Some sites work only in chromium and it's not because it would stick to standards the most. With a feature like this, we just support browser monopoly.
> Do we use any special controls or flows that fail in other browsers? Isn't our UX library actually supposed to address these - as one of the reasons to use such a library anyway?

We do have troubles with Edge and Safari of the major browser distributions. Views not updating after you select a first model, only that model updates.

lokm01:
> I object against such behavior.
> We are already getting to the situation like it was years ago with IE6. Some sites work only in chromium and it's not because it would stick to standards the most. With a feature like this, we just support browser monopoly.
> Do we use any special controls or flows that fail in other browsers? Isn't our UX library actually supposed to address these - as one of the reasons to use such a library anyway?

The whole point of this is that the project is developed and tested on chrome(ium) - or at least 95% of it is.

If we don't have time to test using other browsers (and right now we don't) the best thing we can do is to give them a warning at the very least.

PR #457, opened 2019-06-06
#456 Check for supported browsers

lokm01:
This PR adds a warning if user is using anything other than:

Chrome version >= 68


benedeki:
> 
> 
> Could you associate the popup only to the login page? This will show every time a person does a refresh and that is annoying.

If we are going to have this (see my comment in #456) I definitely support to have it only on the login page.

PR #460, opened 2019-06-11
Feature/445 support star default

yruslan:
If a target attribute of default value mapping is set to '*', the expression
will be used if there is no default value a target attribute selected in
the mapping rule.


conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/rules/MappingRuleInterpreter.scala
```diff
@@ -96,11 +88,10 @@ case class MappingRuleInterpreter(rule: MappingConformanceRule, conformance: Con
         when((col(s"`${rule.outputColumn}`") isNull) and inclErrorNullArr(mappings, datasetSchema), appendErrUdfCall).otherwise(col(ErrorMessage.errorColumnName)))
 
       // see if we need to apply default value
-      val resDf = if (defaultMappingValueMap.contains(rule.targetAttribute)) {
+      val resDf = if (defaultValueOpt.nonEmpty) {
```
GeorgiChochov: This looks like a perfect use case for pattern matching.

conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/rules/MappingRuleInterpreter.scala
```diff
@@ -126,6 +117,30 @@ case class MappingRuleInterpreter(rule: MappingConformanceRule, conformance: Con
...
+
+    val attributeDefaultValueOpt = defaultMappingValueMap.get(rule.targetAttribute)
+    val genericDefaultValueOpt = defaultMappingValueMap.get("*")
+
+    val defaultValueOpt = attributeDefaultValueOpt match {
+      case Some(_) => attributeDefaultValueOpt
+      case None => genericDefaultValueOpt
+    }
+
+    if (defaultValueOpt.nonEmpty) {
```
GeorgiChochov: It's also odd that you are using `nonEmpty` (a `TraversableOnce` function added to `Option` for API interoperability, I guess, but a little confusing when actually applied to such) on an `Option`, which is an alias for `isDefined` (an `Option`-specific function).

conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/rules/MappingRuleInterpreter.scala
```diff
@@ -126,6 +117,30 @@ case class MappingRuleInterpreter(rule: MappingConformanceRule, conformance: Con
...
+      val mappingTableSchema = dao.getSchema(mappingTableDef.schemaName, mappingTableDef.schemaVersion)
+      if (mappingTableSchema != null) {
```
GeorgiChochov: I would wrap this in an `Option` and pattern match. I'm a little surprised that `DAO` does not return an `Option`, but that's outside the scope of this task.

conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/rules/MappingRuleInterpreterGroupExplode.scala
```diff
@@ -118,23 +116,36 @@ case class MappingRuleInterpreterGroupExplode(rule: MappingConformanceRule,
...
-  private def validateMappingRule(df: Dataset[Row],
-                                  dao: EnceladusDAO,
-                                  mappingTableDef: MappingTable,
-                                  mapTable: Dataset[Row],
-                                  joinContidionStr: String,
-                                  defaultMappingValueMap: Map[String, String])
-                                 (implicit spark: SparkSession): Unit = {
-    // validate the default value against the mapping table schema
-    if (defaultMappingValueMap.contains(rule.targetAttribute)) {
+  private def getDefaultValue(mappingTableDef: MappingTable)
```
GeorgiChochov: This seems to be a complete duplicate of the method in `MappingRuleInterpreter`, can we abstract it out to reduce duplication?

conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/rules/MappingRuleInterpreterGroupExplode.scala
```diff
@@ -118,23 +116,36 @@ case class MappingRuleInterpreterGroupExplode(rule: MappingConformanceRule,
...
       }
     }
+    defaultValueOpt
+  }
+
+  private def validateMappingRule(df: Dataset[Row],
+                                  dao: EnceladusDAO,
+                                  mappingTableDef: MappingTable,
+                                  mapTable: Dataset[Row],
+                                  joinContidionStr: String)
```
Zejnilovic:
```suggestion
                                  joinConditionSlr: String)
```

conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/rules/MappingRuleInterpreterGroupExplode.scala
yruslan: Ok, will do. the idea was that one of the mapping rule classes will be removed at some point, just didn't want for these 2 implementations to have common dependencies.

Issue #81, opened 2018-12-28
Reorder conformance rules


GeorgiChochov:
Users should be able to reorder the conformance rules through the Menas UI. This is best achieved through either drag-and-drop or a set of "move-up/down" arrows on the side of the displayed rule.

Issue #461, opened 2019-06-11
Better transitive schema handling


GeorgiChochov:
## Background
Currently, the transitive schema is created on the fly based on the rules which need to be applied to the dataset. This includes fetching the mapping table schemas for each mapping rule in a blocking fashion every time an add/edit rule dialog is opened. This can become very slow and cumbersome when there are a lot of mapping rules.

## Feature
If we fetch all of the mapping table schemas on dataset selection we can fetch them once and keep using them for every rule dialog without incurring the cost of additional redundant fetches. There might be an issue if a different dataset is selected before the transitive schema requests of the previous selection complete and the new ones start...

PR #467, opened 2019-06-11
#81 Add conformance rule reordering

GeorgiChochov:


lokm01:
This doesn't seem to check the consistency of other rules. Is that coming in another PR?

Zejnilovic:
> This doesn't seem to check the consistency of other rules. Is that coming in another PR?

There are no validations currently in general. When we get validations functions, they should be called before each. So maybe do validation in a separate PR instead of ad-hoc ifs?

GeorgiChochov:
I'll stick it into this one.

Zejnilovic:
When changing the order, the audit trail is really hard to read. Is this automatic or can we change it a bit?

image

lokm01:
> When changing the order, the audit trail is really hard to read. Is this automatic or can we change it a bit?
> 
> image

It's automatic. There's an issue for making them more readable.

GeorgiChochov:
I'll get back to adding validation to this once #461 is merged.

Zejnilovic:
We have the same error message regardless of the direction of movement. Meaning, if I have dependency A -> B (B needs A to exist) and try moving A below B, it will tell me `Unable to move conformance rule as it depends on column "A"`, which would mean A is self-dependent.

Can we change the text for arrow down?

menas/ui/service/RuleService.js
```diff
@@ -48,17 +48,21 @@ class RuleService {
...
       });
+
     const schemas = [schema];
     try {
       SchemaManager.updateTransitiveSchemas(schemas, newRules);
     } catch (e) {
-      return new InvalidResult(`Unable to move conformance rule as it depends on column "${e.fieldPath}"`)
+      const smallerIndex = Math.min(ruleIndex, otherRuleIndex);
+      const largerIndex = Math.max(ruleIndex, otherRuleIndex);
+      return new InvalidResult(`Unable to move conformance rule as it would break the dependency on column "${e.fieldPath}" between rule ${smallerIndex} and ${largerIndex}`)
```
Zejnilovic: index + 1 for the message and rules headings.

GeorgiChochov: Fair point, thanks for keeping my laziness in check.

Issue #445, opened 2019-05-31
Need to support 'default default' value for mapping tables


yruslan:
## Background
The old Menas defines a single default value per mapping table. The new Menas supports a mapping between column names and default values.

During data model migration we don't know target columns for mapping tables.

## Feature
An easy workaround would be adding a default value that would work for whatever field is specified as the target column. For instance, if in the list of default values there is a mapping `"*" -> "SomeValue"`, the `SomeValue` value should be used as the default value for any target column if that target column does not have a default value.

So it is kind of 'default default' value.

## Implementation
The only thing that needs to be done is Conformance Spark job support.

PR #469, opened 2019-06-12
#445 Make default value for using the experimental mapping rule

yruslan:
This is a tiny thing to make experimental mapping rule configurable so we can turn it on by default on test environments.

### DIFF
--- a/conformance/src/main/scala/za/co/absa/enceladus/conformance/CmdConfig.scala
+++ b/conformance/src/main/scala/za/co/absa/enceladus/conformance/CmdConfig.scala
@@ -36,7 +36,7 @@ case class CmdConfig(datasetName: String = "",
                      performanceMetricsFile: Option[String] = None,
                      publishPathOverride: Option[String] = None,
                      folderPrefix: Option[String] = None,
-                     experimentalMappingRule: Boolean = false)
+                     experimentalMappingRule: Option[Boolean] = None)
 
 object CmdConfig {
 
@@ -96,8 +96,8 @@ object CmdConfig {
     opt[String]("folder-prefix").optional().action((value, config) =>
       config.copy(folderPrefix = Some(value))).text("Adds a folder prefix before the infoDateColumn")
 
-    opt[Unit]("experimental-mapping-rule").optional().action((value, config) =>
-      config.copy(experimentalMappingRule = true)).text("Use experimental optimized mapping conformance rule")
+    opt[Boolean]("experimental-mapping-rule").optional().action((value, config) =>
+      config.copy(experimentalMappingRule = Some(value))).text("Use experimental optimized mapping conformance rule")
 
     help("help").text("prints this usage text")
   }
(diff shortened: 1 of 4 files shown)


conformance/src/main/scala/za/co/absa/enceladus/conformance/CmdConfig.scala
```diff
@@ -96,8 +96,8 @@ object CmdConfig {
     opt[String]("folder-prefix").optional().action((value, config) =>
       config.copy(folderPrefix = Some(value))).text("Adds a folder prefix before the infoDateColumn")
 
-    opt[Unit]("experimental-mapping-rule").optional().action((value, config) =>
-      config.copy(experimentalMappingRule = true)).text("Use experimental optimized mapping conformance rule")
+    opt[Boolean]("experimental-mapping-rule").optional().action((value, config) =>
+      config.copy(experimentalMappingRule = Some(value))).text("Use experimental optimized mapping conformance rule")
```
GeorgiChochov: I would use `Option(value)` since a variable is being passed in and that variable could be `null`

conformance/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
```diff
@@ -97,6 +97,15 @@ object DynamicConformanceJob {
     processResult(result, performance, publishPath, stdPath)
   }
 
+  private def getExperimentalRuleEnabled()(implicit cmd: CmdConfig): Boolean = {
```
Zejnilovic: this sound like this will make it enabled. But it actually returns if it is or is not enabled.

conformance/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
```diff
@@ -97,6 +97,15 @@ object DynamicConformanceJob {
     processResult(result, performance, publishPath, stdPath)
   }
 
+  private def isExperimentalRuleEnabled()(implicit cmd: CmdConfig): Boolean = {
+    val enabled = cmd.experimentalMappingRule match {
+      case Some(b) => b
+      case None => conf.getBoolean("conformance.mapping.rule.experimental.implementation")
```
lokm01: Does this explode when its not provided as a cli argument and missing in props?

yruslan: Yeah, it can. Will make it more safe.

Zejnilovic: I had the same thought, but if we are providing default in the properties, it should be ok to leave it.

yruslan: This is fixed. Merging...

Issue #472, opened 2019-06-13
Add conformance rules inbetween other rules


GeorgiChochov:
## Background
Currently conformance rules can only be added to the end of the list.

## Feature
We need to allow users to add conformance rules before or after other rules that already exist in the list.

PR #475, opened 2019-06-13
#472 Insert conformance rule arbitrarily

GeorgiChochov:


Zejnilovic:
Just so you know this reintroduces a problem with needing validation for dropping the column. I am able to drop a column that is going to be used, which then breaks the transitive schema and prints an error in console. We are then unable to use schema selector any more and it always shows `No Data`

switching places of the conformance rule doesn't work as well, because the dependency validation works both ways and will tell me, there is a dependency between the two.

Zejnilovic:
> Just so you know this reintroduces a problem with needing validation for dropping the column. I am able to drop a column that is going to be used, which then breaks the transitive schema and prints an error in console. We are then unable to use schema selector any more and it always shows `No Data`
> 
> switching places of the conformance rule doesn't work as well, because the dependency validation works both ways and will tell me, there is a dependency between the two.

Could not simulate again. I have the data with the error but can't figure out how I created it...
Will retest in a week or two.

Issue #480, opened 2019-06-16
Basic search for versioned menas entities


lokm01:
Provide basic search for Datasets, Schemas, Mapping Tables.

PR #481, opened 2019-06-16
#480 Basic search for versioned menas entities

lokm01:


Zejnilovic:
## 1 cannot read property type
Getting new console error when searching and no data is found

image

Zejnilovic:
## 1 Suggestions don't auto refresh
If I create a new schema/dataset/mapping table and I don't do a full refresh I don't see the newly created item in the suggestions.

## 2 Suggestion suggest disabled items
Suggestions suggest something that was already deleted/disabled

menas/src/main/scala/za/co/absa/enceladus/menas/repositories/VersionedMongoRepository.scala
```diff
@@ -38,13 +38,24 @@ abstract class VersionedMongoRepository[C <: VersionedModel](mongoDb: MongoDatab
...
   private def getParent(oldEntity: C): MenasReference = {
     MenasReference(collection = Some(collectionBaseName), name = oldEntity.name, version = oldEntity.version)
   }
   
-  def getLatestVersions(): Future[Seq[VersionedSummary]] = {
+  def getDistinctNames(): Future[Seq[String]] = {
+    collection.distinct[String]("name").toFuture()
+  }
+
+  def getLatestVersions(searchQuery: Option[String] = None): Future[Seq[VersionedSummary]] = {
```
GeorgiChochov: Can we add some tests for this (just a few cases in `DatasetRepositoryIntegrationSuite`)?

menas/src/main/scala/za/co/absa/enceladus/menas/repositories/VersionedMongoRepository.scala
```diff
@@ -38,13 +38,24 @@ abstract class VersionedMongoRepository[C <: VersionedModel](mongoDb: MongoDatab
...
+    val searchFilter = searchQuery match {
+      case Some(search) => Filters.text(search)
```
GeorgiChochov: This `Filters.text` seems to act unpredictably. How about we just match on name for now and maybe use a `regex` filter as a start to find partial matches.

lokm01:
> ## 1 cannot read property type
> Getting new console error when searching and no data is found
> 
> image

Any chance this is related to https://github.com/lastpass/lastpass-cli/issues/428 ??

menas/src/main/scala/za/co/absa/enceladus/menas/repositories/VersionedMongoRepository.scala
lokm01: Cool will do

menas/src/main/scala/za/co/absa/enceladus/menas/repositories/VersionedMongoRepository.scala
lokm01: Yeah this works better

Zejnilovic:
## 4 Errors on start-up in private mode
I open a private mode. Go to Menas's login page and this is what shows.
image

## 1 Cannot read property type
In private mode, I see no such error. You were probably correct.

## 5 Suggestions don't show at all
Currently. No suggestions show up for me. But partial names work fine.

menas/src/test/scala/za/co/absa/enceladus/menas/integration/repositories/DatasetRepositoryIntegrationSuite.scala
```diff
@@ -398,4 +398,69 @@ class DatasetRepositoryIntegrationSuite extends BaseRepositoryTest {
...
+  "DatasetMongoRepository::getLatestVersions" should {
+    "return an empty Seq" when {
+      "no datasets exist and search query is provided" in {
+        val actual = await(datasetMongoRepository.getLatestVersions(Some("abc")))
+
+        assert(actual.isEmpty)
+      }
+    }
+    "return seq of versioned summaries matching the search query" when {
+      "datasets exist and search query is provided" in {
```
GeorgiChochov: Can we break this down into separate cases? A test case should really only test one case (very limited asserts).
1. test case for searching a disabled entity, should be in the "return an empty Seq" cases
1. full match (given several datasets with similar names, i.e. search("dataset1"), returns "dataset1" of "dataset1,dataset2,dataset,somethingelse")
1. partial match (given several datasets with similar names, i.e search("dataset") returns "dataset1,dataset2,dataset" of "dataset1,dataset2,dataset,somethingelse")
1. empty search (search("") returns all enabled datasets)

menas/src/test/scala/za/co/absa/enceladus/menas/integration/repositories/DatasetRepositoryIntegrationSuite.scala
```diff
@@ -398,4 +398,69 @@ class DatasetRepositoryIntegrationSuite extends BaseRepositoryTest {
...
+    "return an empty Seq" when {
+      "no datasets exist" in {
+        val actual = await(datasetMongoRepository.getDistinctNamesEnabled)
+
+        assert(actual.isEmpty)
+      }
+    }
+    "return distinct list of enabled names" when {
+      "datasets exist" in {
+        val dataset1 = DatasetFactory.getDummyDataset(name = "dataset1", version = 1,
```
GeorgiChochov: Same as for `DatasetMongoRepository::getLatestVersions`

lokm01:
Ok this last commit should fix 2) 4) 5)

Issue #485, opened 2019-06-17
HDFS inaccessible from Menas after some time


Zejnilovic:
## Describe the bug
There is an issue we don't know how to reproduce but after some time Menas becomes unable to access HDFS through HDFS browser.

Currently, we think it is some kind of a leas issue.

## Additional context
Stack Trace
```bash
java.lang.NoClassDefFoundError: Could not initialize class org.apache.htrace.Tracer
        at org.apache.htrace.Trace.isTracing(Trace.java:154)
        at org.apache.htrace.Trace.startSpan(Trace.java:84)
        at org.apache.htrace.Trace.startSpan(Trace.java:70)
        at org.apache.hadoop.hdfs.DFSClient.getPathTraceScope(DFSClient.java:3560)
        at org.apache.hadoop.hdfs.DFSClient.getFileInfo(DFSClient.java:2115)
        at org.apache.hadoop.hdfs.DistributedFileSystem$22.doCall(DistributedFileSystem.java:1305)
        at org.apache.hadoop.hdfs.DistributedFileSystem$22.doCall(DistributedFileSystem.java:1301)
        at org.apache.hadoop.fs.FileSystemLinkResolver.resolve(FileSystemLinkResolver.java:81)
        at org.apache.hadoop.hdfs.DistributedFileSystem.getFileStatus(DistributedFileSystem.java:1301)
        at org.apache.hadoop.fs.FileSystem.exists(FileSystem.java:1425)
        at za.co.absa.enceladus.menas.services.HDFSService$$anonfun$exists$1.apply$mcZ$sp(HDFSService.scala:31)
        at za.co.absa.enceladus.menas.services.HDFSService$$anonfun$exists$1.apply(HDFSService.scala:31)
        at za.co.absa.enceladus.menas.services.HDFSService$$anonfun$exists$1.apply(HDFSService.scala:31)
        at scala.concurrent.impl.Future$PromiseCompletingRunnable.liftedTree1$1(Future.scala:24)
        at scala.concurrent.impl.Future$PromiseCompletingRunnable.run(Future.scala:24)
        at scala.concurrent.impl.ExecutionContextImpl$AdaptedForkJoinTask.exec(ExecutionContextImpl.scala:121)
        at scala.concurrent.forkjoin.ForkJoinTask.doExec(ForkJoinTask.java:260)
        at scala.concurrent.forkjoin.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1339)
[...]

PR #493, opened 2019-06-18
#485 HDFS becomes inaccessible from Menas after some time

lokm01:
Added the dependency which seems to be missing based on the logs


lokm01:
> The project is dead but there still seems to be a newer HTrace 4.X version. What about that one?
> Sadly I am unable to find proper release notes...

This is the version hadoop & hdfs use. I don't want to risk binary incompatibility

Issue #482, opened 2019-06-17
Oozie Repository's validateProperties method always returns true


Zejnilovic:
## Describe the bug
In class `OozieRepository` there is a method `validateProperties` which calls `validateProperty` for each Oozie property, but this method always returns true. So the validation doesn't work as intended.

## Expected behaviour
Surround the return true in an else block

PR #496, opened 2019-06-19
#482 Oozie Repository validate props always returns true

lokm01:
Also submitted #495


Zejnilovic:
Now every UI render prints the `oozie support disable` warning

lokm01:
> Now every UI render prints the `oozie support disable` warning

You're killing me here 😄 

Ok should only log warns at spring start up

Issue #406, opened 2019-05-17
Pressing back after login breaks the UI


Zejnilovic:
## Describe the bug
If I log in and then press back the UI breaks and I see nothing. I need to log out and log in again.

## To Reproduce
Steps to reproduce the behavior OR commands run:
1. Login
2. Press Back
3. See error

## Expected behaviour
I would expect to be logged out or to just be re-logged in.

## Screenshots
The mouse disappearing is me clicking back
![May-17-2019 12-51-25](https://user-images.githubusercontent.com/17467768/57923417-8d2bbe00-78a2-11e9-8724-15e93e8aaa55.gif)

PR #497, opened 2019-06-19
#406 Pressing back after login breaks the UI

lokm01:


Zejnilovic:
Still breaks on my machine

Issue #187, opened 2019-02-25
File path to previously uploaded schema lingers


Zejnilovic:
**Describe the bug**
If I upload a schema JSON to one file, it is left and still usable in the upload scheme field. This should clear after the upload. It keeps the file path active even if I click to another schema and can be used to update it.

**To Reproduce**
Steps to reproduce the behavior OR commands run:
1. Go to schema1
2. Click on Upload New
3. Choose file
4. Click Upload File
5. Choose different schema
6. Click on Upload New
7. Click Upload File
8. Schemas are the same

**Expected behavior**
I expect the field to clear after each successful upload, or at least when I switch between schemas.

**Screenshots**
![feb-25-2019 11-12-21](https://user-images.githubusercontent.com/17467768/53330310-7488b900-38ee-11e9-8403-ea28ca52df19.gif)

**Desktop (please complete the following information):**
- OS: macOS 10.14.3
- Browser Chrome Version 72.0.3626.109 (Official Build) (64-bit)

PR #498, opened 2019-06-19
#187 File path to previously uploaded schema lingers

lokm01:


lokm01:
> Good enough. It persisting while I change schemas before submitting should be called a feature.

Yeah, let's call it that :)

Zejnilovic:
> > Good enough. It persisting while I change schemas before submitting should be called a feature.
> 
> Yeah, let's call it that :)

I am not joking. I find it very useful since I have datasetName1 .. datasetNameX. I misclick sometimes, but this keeps the file selected and I don't have to go through the file browser again

Issue #198, opened 2019-02-27
Clear password from Menas login form after login


Zejnilovic:
## Describe the bug
If I log out or get logged out due to inactivity in Menas, I still can see my user name and password in the form and one click to log back into the application.

## To Reproduce
1. Log in into Menas
2. Log out of Menas

## Expected behaviour
Password field clears after I successfully log in

## Additional context
It would seem maximum a person can do from this is log back in, but if someone has 2 minutes alone with the PC, he can steal the password too easily. You just change the type from `password` to `text`.

PR #500, opened 2019-06-19
#198 Clear password from Menas login form after login

lokm01:

### DIFF
--- a/menas/ui/components/login/loginDetail.controller.js
+++ b/menas/ui/components/login/loginDetail.controller.js
@@ -94,6 +94,7 @@ sap.ui.define([
       if (oControl) oControl.setBusy(true);
 
       let fnSuccess = (result, status, xhr) => {
+        this.byId("password").setValue("");
         let csrfToken = xhr.getResponseHeader("X-CSRF-TOKEN");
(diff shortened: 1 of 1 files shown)


menas/ui/components/login/loginDetail.controller.js
```diff
@@ -94,6 +94,7 @@ sap.ui.define([
       if (oControl) oControl.setBusy(true);
 
       let fnSuccess = (result, status, xhr) => {
+        this.byId("password").setValue("");
```
Zejnilovic: Seems like it does not work. And I cannot find any ID password in the DOM. Is `byId` grabbing by partial ID?

lokm01: Make sure to test with a user/password which isn't in your autocomplete.

lokm01: P.S its a ui5 element, you won't find it in the dom by this id.

Zejnilovic: > Make sure to test with a user/password which isn't in your autocomplete.

I am testing in a private window with tor. Works fine now.

Issue #247, opened 2019-03-19
Upload New Schema with expired session doesn't send you to login


Zejnilovic:
## Describe the bug
My session is expired and I am in the schemas view, looking at one schema. I click upload new. I am still ok, it didn't kick me out. I browse files, find a file and then I click upload. Nothing is happening.

## To Reproduce
1. Open schema view
2. Get your session expired
3. Try to upload a schema on the schema you are currently on

## Expected behaviour
I expect to get an error or logged off if I hit browse or upload.

## Screenshots
![Mar-19-2019 08-29-19](https://user-images.githubusercontent.com/17467768/54587983-f79abc00-4a21-11e9-8fb5-6d9d0d11f099.gif)

## Desktop (please complete the following information):
- OS: macOS 10.14.3
- Browser Version 0.60.48 Chromium: 72.0.3626.121 (Official Build) (64-bit)

## Additional context
Switching between tabs does not kick you, but that will be solved with the resolution of #239

PR #501, opened 2019-06-19
#247 Upload New Schema with expired session doesn't send to login

lokm01:


menas/ui/components/schema/schemaDetail.controller.js
```diff
@@ -122,6 +122,8 @@ sap.ui.define([
         this._eventBus.publish("schemas", "list");
         // nav back to info
         this.byId("schemaIconTabBar").setSelectedKey("info");
+      } else if (status === 401) {
```
Zejnilovic: I should have mentioned in the bug report that I get 403, not 401 when this happens.

lokm01: My bad, I was too lazy to wait for a timeout :)

Issue #440, opened 2019-05-30
Support cobol as a dataformat in enceladus & menas


lokm01:
We need to support cobrix as part of enceladus.

Definitely in the back-end, ideally in the scheduling UI too.

PR #502, opened 2019-06-19
#440 Add a support for the simplest use case of COBOL data format

yruslan:
This implementation is mostly for getting Cobrix dependencies.
COBOL options, like location of a copybook, etc will go to properties of dataset.

### DIFF
--- a/standardization/src/main/scala/za/co/absa/enceladus/standardization/CmdConfig.scala
+++ b/standardization/src/main/scala/za/co/absa/enceladus/standardization/CmdConfig.scala
@@ -163,11 +163,8 @@ object CmdConfig {
     help("help").text("prints this usage text")
 
     private def processCobolCmdOptions(): Unit = {
-      var copybookLocation: String = ""
-      var isXcom = false
       opt[String]("copybook").optional().action((value, config) => {
-        copybookLocation = value
-        config.copy(cobolOptions = Some(CobolOptions(copybookLocation, isXcom)))
+        config.copy(cobolOptions = cobolSetCopybook(config.cobolOptions, value))
       }).text("Path to a copybook for COBOL data format")
         .validate(value =>
           if (rawFormat.isDefined && rawFormat.get.equalsIgnoreCase("cobol"))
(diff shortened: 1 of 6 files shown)


yruslan:
It's only a beginning. It will get much much harder soon :)

standardization/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
```diff
@@ -129,7 +129,8 @@ object StandardizationJob {
       val dfReader1 = if (cmd.rowTag.isDefined) dfReader.option("rowTag", cmd.rowTag.get) else dfReader
       val dfReader2 = if (cmd.csvDelimiter.isDefined) dfReader1.option("delimiter", cmd.csvDelimiter.get) else dfReader1
       val dfReader3 = if (cmd.csvHeader.isDefined) dfReader2.option("header", cmd.csvHeader.get) else dfReader2
-      dfReader3
+      val dfReader4 = if (cmd.rawFormat.equalsIgnoreCase("cobol")) applyCobolOptions(dfReader3, cmd) else dfReader3
```
Zejnilovic: If you look on line 128 you have `val dfReader4` as well as on line 133

standardization/src/main/scala/za/co/absa/enceladus/standardization/CmdConfig.scala
```diff
@@ -158,6 +161,31 @@ object CmdConfig {
...
+        .validate(value =>
+          if (rawFormat.isDefined && rawFormat.get.equalsIgnoreCase("cobol"))
+            success
+          else
+            failure("The --copybook option is supported only for COBOL data format")
+        )
+
+      opt[Boolean]("is-xcom").optional().action((value, config) => {
+        isXcom = value
+        config.copy(cobolOptions = Some(CobolOptions(copybookLocation, isXcom)))
```
Zejnilovic: wouldn't this line as well as the line 170 profit from being at the end of this method and just once?

DzMakatun:
What happens if the copybook does not match menas schema?

Issue #509, opened 2019-06-20
Menas Credentials Authentication is broken


lokm01:
Exception in thread "main" org.springframework.web.client.HttpClientErrorException: 401 null

PR #511, opened 2019-06-20
#509 Enceladus DAO creds auth broken

lokm01:
- this was an error during PR refactoring...

### DIFF
--- a/dao/src/main/scala/za/co/absa/enceladus/dao/EnceladusRestDAO.scala
+++ b/dao/src/main/scala/za/co/absa/enceladus/dao/EnceladusRestDAO.scala
@@ -76,7 +76,8 @@ object EnceladusRestDAO extends EnceladusDAO {
     }
   }
 
-  private def restTemplateLogin(restTemplate: RestTemplate, url: String, username: String, reqBody: Option[LinkedMultiValueMap[String, String]] = None): Boolean = {
+  private def restTemplateLogin(restTemplate: RestTemplate, url: String, username: String,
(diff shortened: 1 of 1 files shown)


dao/src/main/scala/za/co/absa/enceladus/dao/EnceladusRestDAO.scala
```diff
@@ -75,11 +76,15 @@ object EnceladusRestDAO extends EnceladusDAO {
     }
   }
 
-  private def restTemplateLogin(restTemplate: RestTemplate, url: String, username: String): Boolean = {
+  private def restTemplateLogin(restTemplate: RestTemplate, url: String, username: String, reqBody: Option[LinkedMultiValueMap[String, String]] = None): Boolean = {
```
Zejnilovic: How about 1 param per line?

dao/src/main/scala/za/co/absa/enceladus/dao/EnceladusRestDAO.scala
```diff
@@ -111,9 +116,13 @@ object EnceladusRestDAO extends EnceladusDAO {
   }
 
   def postLogin(username: String, password: String) = {
```
Zejnilovic: I know it is a pain, but could you annotate this method

Zejnilovic:
Runs well

dao/src/main/scala/za/co/absa/enceladus/dao/EnceladusRestDAO.scala
lokm01: np

PR #514, opened 2019-06-20
Feature/455 oozie unable to run different report versions

lokm01:


conformance/src/main/scala/za/co/absa/enceladus/conformance/CmdConfig.scala
```diff
@@ -77,8 +77,8 @@ object CmdConfig {
           case _    => success
         })
 
-    opt[Int]('r', "report-version").action((value, config) =>
-      config.copy(reportVersion = value)).text("Report version")
+    opt[Int]('r', "report-version").optional().action((value, config) =>
```
Zejnilovic: Add to `.text` that if not provided it is inferred from previously run conformance.

conformance/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
```diff
@@ -69,9 +69,19 @@ object DynamicConformanceJob {
     // get the dataset definition
     val conformance = dao.getDataset(cmd.datasetName, cmd.datasetVersion)
     val dateTokens = cmd.reportDate.split("-")
+
+    val reportVersion = cmd.reportVersion match {
```
Zejnilovic: Why not do this in the `CmdConfig`? Would lower the hassle of changing a lot of code

yruslan:
Looks good from the code perspective.

I'm wondering about the feature itself. Inferring report version from the publish path adds new requirements that we didn't have before
- All report version numbers must be without gaps. For automatic jobs that run every couple of hours, it is theoretically possible that a batch has no data. I'm wondering if a publish folder will be created in this case?
- If a new raw data is landed before the previous one is published a wrong job might get triggered.

conformance/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
GeorgiChochov: I'm a little conflicted about this. On the one hand: 
1. `CmdConfig` is supposed to give you the provided command line args, if it starts fetching args from hadoop it might be breaking SRP a little bit.
1. `fsUtils` requires a `SparkSession` to read in the Hadoop configuration to check the latest version, so making `CmdConfig` dependent on a `SparkSession` seems like unwanted coupling to me.

On the other hand:
1. It would be nice to handle all the Spark Job arguments in a central place beforehand. As it stands they are being set in `CmdConfig` and `main()`.

Maybe a `CmdConfig` decorator which handles these additional responsibilities would solve this?

Zejnilovic: I would just like to note that `FileSystemVersionUtils` and `SparkSession` are already in CmdConfig from previous PRs. Se this wouldn't add a dependency to it, just move a method.

GeorgiChochov: I see. In that case sounds like a good suggestion, since they're already inextricable coupled.

lokm01: So firstly there's an issue created for refactoring all these things into common functionality across std & conf. I believe this would be an ideal place for such changes and I would very much like to see all this refactored too.

Secondly, the version inference required the dataset publish path, which requires the DAO to be logged in etc. It's too much of a hassle right now.

lokm01:
> Looks good from the code perspective.
> 
> I'm wondering about the feature itself. Inferring report version from the publish path adds new requirements that we didn't have before
> 
> * All report version numbers must be without gaps. For automatic jobs that run every couple of hours, it is theoretically possible that a batch has no data. I'm wondering if a publish folder will be created in this case?
> * If a new raw data is landed before the previous one is published a wrong job might get triggered.

These are very valid issues and I expected them to come up 😄 

Let me update the text of the parameter to specify that this is a an experimental feature etc.

Additionally the jobs log a WARN message, this was intended to suggest that something "funny" is happening. Let me update to include that it can lead to issues with concurrently running jobs on single dataset or if versions are being landed in a series with gaps.

lokm01:
> Looks good from the code perspective.
> 
> I'm wondering about the feature itself. Inferring report version from the publish path adds new requirements that we didn't have before
> 
> * All report version numbers must be without gaps. For automatic jobs that run every couple of hours, it is theoretically possible that a batch has no data. I'm wondering if a publish folder will be created in this case?
> * If a new raw data is landed before the previous one is published a wrong job might get triggered.

Alternatively if you guys see a better way of doing this, feel free to suggest.

yruslan:
> Alternatively if you guys see a better way of doing this, feel free to suggest.

Yeah, let's discuss requirements and what can we do.

Issue #517, opened 2019-06-20
Support cobol copybooks in menas


lokm01:
Allow users to upload schema as cobol copybooks.

This will ensure that the schemas between menas, spark and cobol are consistent.

Additionally the DAO can then retrieve the contents of the copybook from attachments.

PR #518, opened 2019-06-20
#517 Support cobol copybooks in menas

lokm01:
This allows the user to upload schema either as a Spark
StructType (default) or as a cobol copybook.

### DIFF
--- a/menas/pom.xml
+++ b/menas/pom.xml
@@ -165,6 +165,11 @@
             <version>${project.version}</version>
         </dependency>
         <dependency>
+            <groupId>za.co.absa.cobrix</groupId>
+            <artifactId>spark-cobol</artifactId>
+            <version>${cobrix.version}</version>
+        </dependency>
+        <dependency>
             <groupId>za.co.absa</groupId>
             <artifactId>enceladus-utils</artifactId>
             <version>${project.version}</version>
--- a/menas/pom.xml
+++ b/menas/pom.xml
@@ -165,11 +165,6 @@
             <version>${project.version}</version>
         </dependency>
         <dependency>
-            <groupId>za.co.absa.cobrix</groupId>
-            <artifactId>spark-cobol</artifactId>
-            <version>${cobrix.version}</version>
-        </dependency>
-        <dependency>
             <groupId>za.co.absa</groupId>
             <artifactId>enceladus-utils</artifactId>
             <version>${project.version}</version>
@@ -223,7 +218,6 @@
             <artifactId>spark-cobol</artifactId>
             <version>${cobrix.version}</version>
         </dependency>
-
         <!-- Test -->
         <dependency>
             <groupId>org.springframework.boot</groupId>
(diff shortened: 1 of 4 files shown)


lokm01:
[test1_copybook.txt](https://github.com/AbsaOSS/enceladus/files/3311090/test1_copybook.txt)

Zejnilovic:
I just saw that your `npm-shrinkwrap.json` differs from mine after the build. Does anyone else have this?

menas/pom.xml
```diff
@@ -164,6 +164,11 @@
             <artifactId>enceladus-migrations</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
```
yruslan: I'm pretty sure I've added the dependency in the previous PR. Please, be careful merging this. Rebasing against the current develop might be a good idea.

lokm01: Into menas pom?

yruslan: Yes, I've added it to check if it is compatible.

lokm01: Will check thx

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/SchemaController.scala
```diff
@@ -44,15 +48,26 @@ class SchemaController @Autowired() (
   @ResponseStatus(HttpStatus.CREATED)
   def handleFileUpload(@AuthenticationPrincipal principal: UserDetails,
                        @RequestParam file: MultipartFile,
-                       @RequestParam version: Int, @RequestParam name: String): CompletableFuture[_] = {
+                       @RequestParam version: Int, @RequestParam name: String, @RequestParam format: Optional[String]): CompletableFuture[_] = {
+    val origFormat: Option[String] = format
+    val fileContent = new String(file.getBytes)
+
+    val sparkStruct = origFormat match {
+      case Some("struct") | Some("") | None => sparkMenasConvertor.convertAnyToStructType(fileContent)
```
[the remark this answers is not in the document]
lokm01: Won't fix. This is arguable.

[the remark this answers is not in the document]
lokm01: No mention about this in the beloved scala style guide and some other best practices suggest this.
https://alvinalexander.com/scala/best-practice-use-match-expressions-pattern-matching-idioms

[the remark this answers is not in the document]
Zejnilovic: since you have already one multiline case, I am asking you to keep the style the same.

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/SchemaController.scala
```diff
@@ -44,15 +48,27 @@ class SchemaController @Autowired() (
...
+      case Some("copybook") =>
+        val parsedSchema = CopybookParser.parseTree(fileContent)
+        val cobolSchema = new CobolSchema(parsedSchema, SchemaRetentionPolicy.CollapseRoot, false)
+        cobolSchema.getSparkSchema
+      case Some(x) => throw new UnsupportedOperationException(s"$x is not a recognized schema format. Menas currently supports: 'struct' and 'copybook'")
+    }
```
benedeki: Doesn't this `match` miss the `case None` code branch? If it's an `Option[String]` I would expect it to be possible `None` would come in.

lokm01: It doesn't. Read again

benedeki: My bad, sorry. 😊

Issue #83, opened 2018-12-28
Conformance rule validation


GeorgiChochov:
When adding/editing conformance rules through the Menas UI, there should be front-end validation as well as back-end validation. This includes checking that required fields are populated and that input columns exist in the dataset's schema (as per #82).

Zejnilovic:
First validation. Output Column Name cannot have one of these  " ,;{}()\n\t=". Affected rules are:

- CastingConformanceRule
- ConcatenationConformanceRule
- LiteralConformanceRule
- MappingConformanceRule
- NegationConformanceRule
- SingleColumnConformanceRule
- SparkSessionConfConformanceRule
- UpperCaseConformanceRule

Zejnilovic:
Output column has to have the same parent

Zejnilovic:
There should be a warning that `UpperCaseConformanceRule` is going to turn the field into a string.

Zejnilovic:
`CastingConformanceRule` gives `null` on other default value if it cannot cast the source to the particular type.

We should warn/stop user if the current casting rule is going to produce a null.

GeorgiChochov:
I'll create additional tasks for the further validation and close this issue afterwards.

GeorgiChochov:
> There should be a warning that `UpperCaseConformanceRule` is going to turn the field into a string.

#554

GeorgiChochov:
> `CastingConformanceRule` gives `null` on other default value if it cannot cast the source to the particular type.
> 
> We should warn/stop user if the current casting rule is going to produce a null.

#555

Zejnilovic:
> Output column has to have the same parent

#552

GeorgiChochov:
Thanks for bringing these up, creating the tasks and helping break this down, @Zejnilovic !

PR #526, opened 2019-06-21
#83 Add Conformance Rule dialog validation

GeorgiChochov:
Validates:
* all required fields in the conformance rule forms
* minimum of 2 columns for concatenation
* concatenation column dialog
* minimum of 1 join condition for mapping
* join condition dialog
* output column that might override (of if inserting a rule in the middle of the list, be overridden by another rule later on)
* drop only columns which do not break the transitive schema (if adding a drop rule in the middle of the list)

### DIFF
--- a/menas/ui/components/dataset/conformanceRule/ConformanceRule.js
+++ b/menas/ui/components/dataset/conformanceRule/ConformanceRule.js
@@ -1,0 +1,295 @@
+/*
+ * Copyright 2018-2019 ABSA Group Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+class ConformanceRule {
+
+  constructor(rule) {
+    if (this.apply === undefined) {
+      throw new TypeError("Abstract function 'apply' not implemented.");
+    }
+
+    this._rule = rule;
+
+    const outputCol = rule.outputColumn;
+    const index = outputCol.lastIndexOf(".");
+    const name = (index === -1) ? outputCol : outputCol.slice(index + 1);
+    const path = (index === -1) ? "" : outputCol.slice(0, index);
+    this._outputCol = {name: name, path: path};
+  }
+
+  get rule() {
+    return this._rule;
+  }
+
+  get outputCol() {
+    return this._outputCol;
+  }
+
+  getInputCol(fields) {
+    return this.getCol(fields, this.rule.inputColumn);
+  }
+
+  getCol(fields, columnName) {
+    const splitPath = columnName.split(".");
+    return splitPath.reduce((fields, path, index) => {
+      const field = fields.find(field => field.name === path);
+      if (field === undefined) {
+        throw new UnknownFiledError(columnName, this.rule.order);
+      }
+      const children = field.children;
+      return (children && children.length > 0 && splitPath.length > index + 1) ? children : field;
+    }, fields);
+  }
+
+  addNewField(fields, newField) {
+    const splitPath = this.rule.outputColumn.split(".");
+    return splitPath.reduce((acc, path, index) => {
+      let element = acc.find(field => field.name === path);
+      if (!element) {
+        element = (index === splitPath.length - 1) ? newField : new SchemaField(path, splitPath.slice(0, index).join(","), "struct", true, []);
+        acc.push(element);
+      }
+      let ch = element.children;
+      if (!ch) {
+        element.children = [];
+      }
+      return element.children;
+    }, fields);
+  }
+
+  validate(fields, schemas) {
+    return this.apply(fields);
+  }
+
+}
+
+class CastingConformanceRule extends ConformanceRule {
+
+  constructor(rule) {
+    super(rule);
+  }
+
+  apply(fields) {
+    const inputCol = this.getInputCol(fields);
+    const newField = new SchemaField(this.outputCol.name, this.outputCol.path, this.rule.outputDataType, inputCol.nullable, []);
+    this.addNewField(fields, newField);
+    return fields;
+  }
+
+}
+
+class ConcatenationConformanceRule extends ConformanceRule {
+
+  constructor(rule) {
+    super(rule);
+  }
+
+  apply(fields) {
+    const isNullable = this.getInputCols(fields).some(field => field.nullable);
+    const newField = new SchemaField(this.outputCol.name, this.outputCol.path, "string", isNullable, []);
+    this.addNewField(fields, newField);
+    return fields;
+  }
+
+  getInputCols(fields) {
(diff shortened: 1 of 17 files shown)


Zejnilovic:
## 1 Unable to create nested structures
I am unable to create nested structures. For example, going to concatenation rule and in `OutputColumn` putting `alfa.beta`

image

menas/ui/components/dataset/datasetDetail.controller.js
```diff
@@ -325,16 +325,21 @@ sap.ui.define([
...
-
-      	const auditTable = this.byId("auditTrailTable");
-      	this._datasetService.getAuditTrail(currentDataset.name, auditTable);
+        this._transitiveSchemas = [];
+        const transitiveSchemas = this._transitiveSchemas;
+        new SchemaRestDAO().getByNameAndVersionSync(currentDataset.schemaName, currentDataset.schemaVersion).then((schema) => {
+          this._model.setProperty("/currentDataset/schema", schema)
+          this._schemaTable.model = schema;
+          transitiveSchemas.push(schema);
+          // var millisecondsToWait = 5000;
```
lokm01: Maybe clean these up?

Zejnilovic:
## 2 Input column alias in SingleColmn Rule allows `.`
Input column alias in SingleColmn Rule allows `.` which then does not translate to the embedded structure but as a name itself which can be misleading and lead to errors in the end. Haven't yet run 2e2 using this rule in this setting.

**Conformance rule view**
image

**Schema picker**
image

**Conformance rule edit**
image

menas/ui/components/dataset/conformanceRule/ConformanceRule.js
```diff
@@ -0,0 +1,295 @@
...
+}
+
+class ConcatenationConformanceRule extends ConformanceRule {
+
+  constructor(rule) {
+    super(rule);
+  }
+
+  apply(fields) {
+    const isNullable = this.getInputCols(fields).some(field => field.nullable);
```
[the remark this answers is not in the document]
GeorgiChochov: Nope, if one is nullable, the entire thing becomes nullable

menas/ui/components/dataset/conformanceRule/ConformanceRuleForm.js
```diff
@@ -0,0 +1,395 @@
...
+}
+
+class DropConformanceRuleForm extends ConformanceRuleForm {
+
+  constructor() {
+    super("DropConformanceRule", true)
+  }
+
+  get outputColumnControl() {
+    return sap.ui.getCore().byId("DropConformanceRule--schemaFieldSelector");
```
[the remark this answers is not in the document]
GeorgiChochov: Left over from when it had a reason.

menas/ui/components/dataset/conformanceRule/ConformanceRuleForm.js
```diff
@@ -0,0 +1,395 @@
...
+    return this.hasValidSchemaSelection(fieldValue, this.outputColumnControl);
+  }
+
+  reset() {
+    this.outputColumnControl
+      .getItems()
+      .forEach(item => item.setHighlight(sap.ui.core.ValueState.None));
+  }
+
+  hasValidTransitiveSchema(rule, schemas, rules) {
```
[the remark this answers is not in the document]
GeorgiChochov: I think we don't. Unless something uses `level1.level2` there's no reason to worry about it.

menas/ui/components/dataset/conformanceRule/ConformanceRuleForm.js
```diff
@@ -0,0 +1,395 @@
...
+  }
+
+  isValid(rule, schemas, rules) {
+    const hasValidOutputColumn = this.hasValidOutputColumn(rule.outputColumn)
+      && this.hasValidTransitiveSchema(rule, schemas, rules);
+    return hasValidOutputColumn & this.isCorrectlyConfigured(rule);
+  }
+
+  isCorrectlyConfigured(rule) {
+    return true;
```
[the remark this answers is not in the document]
GeorgiChochov: Good spot, this is left over from initial setup.
As for the abstract method, I already did that for `ConformanceRule::apply` a while ago, and will do it here as well.

menas/ui/service/GenericService.js
```diff
@@ -66,10 +66,18 @@ var GenericService = new function () {
     })
   };
 
+  this.isEmpty = function (str) {
+    return !str || str === "";
```
Zejnilovic: The second part is unneeded I think

image

Zejnilovic:
## 3 Uncaught UseCase
Steps:
1. Create a rule with output column `Alfa`
2. Create a drop rule for `Alfa`
3. Create another rule with output column `Alfa`
4. Move rule from step 3. above the drop rule.

This is something unexpected and should not happen as now I have two rules that create output column `Alfa`.

Zejnilovic:
## 1 Unable to create nested structures

Currently, I get the error below.
image

menas/ui/service/GenericService.js
GeorgiChochov: Holy crap, JS takes a lot of liberties in interpreting things as a boolean lol

GeorgiChochov:
> ## 1 Unable to create nested structures
> Currently, I get the error below.
> image

Pushed the fixed for that a minute ago, when I noticed it as well.

GeorgiChochov:
> ## 3 Uncaught UseCase
> Steps:
> 
> 1. Create a rule with output column `Alfa`
> 2. Create a drop rule for `Alfa`
> 3. Create another rule with output column `Alfa`
> 4. Move rule from step 3. above the drop rule.
> 
> This is something unexpected and should not happen as now I have two rules that create output column `Alfa`.

I remember specifically fixing this... Had it in mind since the rule reordering implementation.

menas/ui/components/dataset/conformanceRule/ConformanceRuleForm.js
```diff
@@ -0,0 +1,423 @@
...
+  }
+
+  hasValidInputColumn(fieldValue) {
+    return this.hasValidSchemaSelection(fieldValue, this.inputColumnControl);
+  }
+
+  hasValidSchemaSelection(fieldValue, inputControl) {
+    let isValid = true;
+
+    if (!fieldValue || fieldValue === "" || !GenericService.isValidColumnName(fieldValue)) {
```
Zejnilovic: Here is the `!fieldValue` enough as well.

menas/ui/components/dataset/conformanceRule/ConformanceRuleForm.js
```diff
@@ -0,0 +1,423 @@
...
+  isCorrectlyConfigured(rule) {
+    return this.hasValidInputColumn(rule.inputColumn)
+      & this.hasValidOutputDataType(rule.outputDataType);
+  }
+
+  hasValidOutputDataType(fieldValue) {
+    const inputControl = this.outputDataTypeControl;
+    let isValid = true;
+
+    if (!fieldValue || fieldValue === "") {
```
Zejnilovic: Here is the `!fieldValue` enough as well.

Zejnilovic:
## 1 Unable to create nested structures
**Fixed**

## 2 Input column alias in SingleColmn Rule allows `.`
**Fixed**

GeorgiChochov:
> ## 3 Uncaught UseCase
> Steps:
> 
> 1. Create a rule with output column `Alfa`
> 2. Create a drop rule for `Alfa`
> 3. Create another rule with output column `Alfa`
> 4. Move rule from step 3. above the drop rule.
> 
> This is something unexpected and should not happen as now I have two rules that create output column `Alfa`.

This PR is too big as it is, let's build on in separate tasks. Created #532 for this.

Issue #491, opened 2019-06-18
TypeError when session expires


Zejnilovic:
## Describe the bug
When a session expires and I try to create, edit or do something more than just view I get a `typeError`.

## Screenshots
![image](https://user-images.githubusercontent.com/17467768/59673606-d896fc00-91c1-11e9-9b1d-40b77f91df69.png)

## Desktop (please complete the following information):
 - OS: macOS 10.14.5
 - Browser: Brave Version 0.65.118 Chromium: 75.0.3770.80

benedeki:
I can confirm that after session expiration, _Menas_ behaves weirdly (Os: Windows; Browser: 75.0.3770.100 (Official Build) (64-bit))

Zejnilovic:
if I click on to the navigation menu I get a different kind of error

![image](https://user-images.githubusercontent.com/17467768/59852502-3a489900-936f-11e9-8ebc-2d0c4847502e.png)

Issue #525, opened 2019-06-21
Add Runs to the landing page tiles


Zejnilovic:
## Feature
Add Runs view to the landing page/home page

![image](https://user-images.githubusercontent.com/17467768/59919017-9de1cd80-9426-11e9-94db-d24ff24cfb2d.png)

PR #529, opened 2019-06-24
#525 Add runs to landing page

lokm01:


menas/src/main/scala/za/co/absa/enceladus/menas/repositories/RunMongoRepository.scala
```diff
@@ -53,6 +55,51 @@ class RunMongoRepository @Autowired()(mongoDb: MongoDatabase)
...
+
+  def getTodaysStdSuccessRuns(): Future[Int] = {
+    getTodayRunsCount(List(Filters.eq("runStatus.status", "stageSucceeded")))
+  }
+
+  def getTodaysRunningRuns(): Future[Int] = {
+    getTodayRunsCount(List(Filters.eq("runStatus.status", "running")))
+  }
+
+  def getTodaysSuccessWithErrors(): Future[Int] = {
```
DzMakatun: Having this another status is a great idea, as it will draw more attention to the errors. Maybe it would be better to implement it in std/cnfrm jobs, so that Menas UI and the database have the same set of statuses? Otherwise, it may be misleading when tracing the errors. Alternatively, "Completed jobs with data errors", could be displayed as a separate metrics, not to be confused with job execution status... What are your thoughts on this?

menas/src/main/scala/za/co/absa/enceladus/menas/repositories/RunMongoRepository.scala
```diff
@@ -53,6 +55,51 @@ class RunMongoRepository @Autowired()(mongoDb: MongoDatabase)
     excludeId()
   ))
 
+  private def getTodaysFilter() = {
+    val date = LocalDate.now().format(DateTimeFormatter.ofPattern("dd-MM-yyyy"))
```
Zejnilovic: This gives you local date where Menas is deployed right? Is the system-wide timezone set before this or used somehow?

menas/src/test/scala/za/co/absa/enceladus/menas/integration/repositories/RunRepositoryIntegrationSuite.scala
```diff
@@ -44,6 +46,8 @@ class RunRepositoryIntegrationSuite extends BaseRepositoryTest {
     runFixture.dropCollection()
   }
 
+  private val today = LocalDate.now().format(DateTimeFormatter.ofPattern("dd-MM-yyyy"))
```
Zejnilovic: This does not test the edge cases of what happens close to midnight

menas/ui/service/GenericService.js
```diff
@@ -33,9 +33,40 @@ var GenericService = new function () {
     })
   };
 
+  this.runStatusFormatter = function(sStatus) {
+    if(sStatus === "failed") return "Failed";
+    else if(sStatus === "successful") return "Successfull";
```
Zejnilovic:
```suggestion
    else if(sStatus === "successful") return "Successful";
```

menas/ui/service/GenericService.js
```diff
@@ -33,9 +33,40 @@ var GenericService = new function () {
     })
   };
 
+  this.runStatusFormatter = function(sStatus) {
```
Zejnilovic: JavaScript does have a switch statement, that could be better here and is supposedly faster

menas/ui/components/home/landingPage.view.xml
```diff
@@ -11,40 +12,64 @@
...
+                                </TileContent>
+                            </tileContent>
+                        </GenericTile>
+                        <GenericTile header="Runs" class="sapUiResponsiveMargin" press="masterNavigate">
+                            <customData>
+                                <core:CustomData key="target" value="runs"></core:CustomData>
+                            </customData>
+                            <tileContent>
+                                <TileContent>
+                                    <NumericContent icon="sap-icon://physical-activity" value="{/landingPageInfo/totalNumberRuns}"></NumericContent>
```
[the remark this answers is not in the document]
lokm01: Good point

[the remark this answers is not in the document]
lokm01: Short number format will do here.

menas/src/main/scala/za/co/absa/enceladus/menas/repositories/RunMongoRepository.scala
lokm01: @aaishaosman @yruslan 

So I guess we'd like to have as much metadata as possible. I believe in future we may need to show stuff like break down of errors etc.

lokm01:
> 1. Could we maybe sort the graph differently? This seems hard to read
> 2. When the session expires and I re-login, I get an error in the console and buttons don't work properly until refresh.
>    image

1) Feel free to suggest.
2) Seems like we're gonna have to be a bit more deffensive with ui5 wrapper for charts.js

menas/src/main/scala/za/co/absa/enceladus/menas/repositories/RunMongoRepository.scala
lokm01: Uses OS settings.

menas/src/test/scala/za/co/absa/enceladus/menas/integration/repositories/RunRepositoryIntegrationSuite.scala
lokm01: Feel free to contribute or file an issue

menas/ui/service/GenericService.js
lokm01: Sigh at some stage we should probably start using i18n

menas/ui/service/GenericService.js
lokm01: Which also bloats the code significantly (as much as I'd prefer to use a construct closer to pattern matching).

menas/src/main/scala/za/co/absa/enceladus/menas/repositories/RunMongoRepository.scala
Zejnilovic: But that is bad, isn't it? OS settings are who knows what and we have everything in Run model in UTC.

Zejnilovic:
> 1. Feel free to suggest.
> 2. Seems like we're gonna have to be a bit more deffensive with ui5 wrapper for charts.js

1. More like if @DzMakatun could tell you his style/logic so it corresponds
2. I am still getting these
image

menas/src/main/scala/za/co/absa/enceladus/menas/repositories/RunMongoRepository.scala
Zejnilovic: Created #541

menas/src/test/scala/za/co/absa/enceladus/menas/integration/repositories/RunRepositoryIntegrationSuite.scala
Zejnilovic: Created #541 then this can be updated. Currently, it would just fail

menas/ui/service/GenericService.js
Zejnilovic: I think there is less tbh. I know it is 2 lines longer, but in general less useless text.
```javascript
switch(sStatus) {
    case "failed":                  return "Failed";
    case "successful":              return "Successfull";
    case "successfulWithErrors":    return "Sucessful with errors";
    case "running":                 return "Running";
    case "stdSuccessful":           return "Standardization successful";
    default:                        return sStatus;
}
```

And you don't have to tab it if you don't want to.

Zejnilovic: Or check the approach in this [totaly unsuspicious link](https://ultimatecourses.com/blog/deprecating-the-switch-statement-for-object-literals#Object_Literal_lookups)

lokm01:
The latest changes should also fix #491

menas/src/main/scala/za/co/absa/enceladus/menas/repositories/RunMongoRepository.scala
```diff
@@ -53,6 +55,51 @@ class RunMongoRepository @Autowired()(mongoDb: MongoDatabase)
...
+    getTodayRunsCount(List(Filters.eq("runStatus.status", "running")))
+  }
+
+  def getTodaysSuccessWithErrors(): Future[Int] = {
+    getTodayRunsCount(List(
+        Filters.eq("runStatus.status", "allSucceeded"),
+        or(
+        and(Filters.exists("controlMeasure.metadata.additionalInfo.std_errors_count"),
+            Filters.notEqual("controlMeasure.metadata.additionalInfo.std_errors_count", "0")),
+        and(Filters.exists("controlMeasure.metadata.additionalInfo.std_errors_count"),
```
yruslan:
```suggestion
        and(Filters.exists("controlMeasure.metadata.additionalInfo.conform_errors_count"),
```

Zejnilovic: this will then require a change of the names of the columns in graphs.

lokm01: > this will then require a change of the names of the columns in graphs.

Why?

menas/src/main/scala/za/co/absa/enceladus/menas/repositories/RunMongoRepository.scala
yruslan: Yeah, breakdown of errors is something I'd expect in the future. I'm wondering what monitoring guys are going to say they want.

menas/src/main/scala/za/co/absa/enceladus/menas/repositories/RunMongoRepository.scala
Zejnilovic: We are speaking about standardization there. Aren't we? I don't have it built currently. 
Should we speak about conformance?

Zejnilovic:
> The latest changes should also fix #491

It does fix the number 2. But it does not fix #491 completely. 
image

lokm01:
> #491

Ok let's deal with that separately.

Issue #248, opened 2019-03-19
Log out button doesn't work if session is expired


Zejnilovic:
## Describe the bug
Log out button doesn't work if the session is expired

## To Reproduce
1. Log in
2. Get session expired
3. Press log out button

## Expected behaviour
I expect to get to the login page

## Desktop (please complete the following information):
 - OS: macOS 10.14.3
 - Browser Version 0.60.48 Chromium: 72.0.3626.121 (Official Build) (64-bit)

PR #546, opened 2019-06-25
#248 Log out button doesn't work if session is expired

lokm01:

### DIFF
--- a/menas/ui/service/GenericService.js
+++ b/menas/ui/service/GenericService.js
@@ -57,12 +57,8 @@ var GenericService = new function () {
   };
 
   this.logout = function (sLogoutMessage) {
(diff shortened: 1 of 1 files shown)


Zejnilovic:
When I click it, it throws 403, is that ok?

lokm01:
> When I click it, it throws 403, is that ok for now or you wanna fix it here?

I think that's expected. How can you logout server side, when it can't see who you are. That's why you're getting 403.

Zejnilovic:
> > When I click it, it throws 403, is that ok for now or you wanna fix it here?
> 
> I think that's expected. How can you logout server side, when it can't see who you are. That's why you're getting 403.

My mistake for doing things fast. Everywhere else I tried I get 401. Here I get 403. That is why I thought of it as an issue

Issue #551, opened 2019-06-26
Standardization/Conformance don't reauthenticate when Menas session expires


GeorgiChochov:
## Describe the bug
The default Spring session timeout is 30 minutes, which means that nay job running for longer than that will fail to update the run information for its checkpoints.

## To Reproduce
Steps to reproduce the behaviour:
1. Run a long-running standardization job (longer than 30 minutes) 
1. Wait for job to finish
1. See error in logs
1. See run status in Menas as still `Running`

## Expected behaviour
The Spark jobs should automatically reauthenticate and retry on 401/403 response code.

yruslan:
![Screenshot 2019-06-26 at 10 34 59](https://user-images.githubusercontent.com/4082463/60164675-4c5a8980-97fe-11e9-82ba-1723fceb3fad.png)

yruslan:
When a session is timed out Encaladus receives 403 (Forbidden) and it does not re-authenticate.

> 19/06/25 12:47:02 INFO Atum: Saving control framework checkpoints
19/06/25 12:47:02 INFO ControlFrameworkREST: URL: http://aaa.aaa.aaa.aaa:8080/menas/api/runs/updateControlMeasure/32814475-46ac-4459-af42-18c5e542e661 POST: {***}
19/06/25 12:47:02 INFO server.AbstractConnector: Stopped Spark@546668e1{HTTP/1.1,[http/1.1]}{0.0.0.0:4044}
19/06/25 12:47:02 INFO ui.SparkUI: Stopped Spark web UI at http://aaa:4044
**19/06/25 12:47:02 ERROR ControlFrameworkREST: RESPONSE: HTTP/1.1 403  - {"timestamp":"2019-06-25T10:49:08.716+0000","status":403,"error":"Forbidden","message":"Forbidden","path":"/menas/api/runs/updateControlMeasure/32814475-46ac-4459-af42-18c5e542e661"}**
19/06/25 12:47:02 ERROR EventListenerMenas: Unable to update control measurements for a Run object (32814475-46ac-4459-af42-18c5e542e661) in the database
19/06/25 12:47:02 INFO cluster.YarnClientSchedulerBackend: Interrupting monitor thread
19/06/25 12:47:02 INFO cluster.YarnClientSchedulerBackend: Shutting down all executors
19/06/25 12:47:02 INFO cluster.YarnSchedulerBackend$YarnDriverEndpoint: Asking each executor to shut down
[...]

PR #558, opened 2019-06-26
#551 Add automatic reauthentication on expired session from spark jobs

GeorgiChochov:


dao/src/main/scala/za/co/absa/enceladus/dao/EnceladusRestDAO.scala
```diff
@@ -45,15 +45,22 @@ import za.co.absa.enceladus.model.MappingTable
 import org.springframework.util.LinkedMultiValueMap
 
 object EnceladusRestDAO extends EnceladusDAO {
-  val conf = ConfigFactory.load()
-  val restBase = conf.getString("menas.rest.uri")
+  private val log = LogManager.getLogger("enceladus.conformance.EnceladusRestDAO")
```
Zejnilovic:
```suggestion
  private val log = LogManager.getLogger(this.getClass)
```

dao/src/main/scala/za/co/absa/enceladus/dao/MenasRestDAO.scala
```diff
@@ -136,17 +132,21 @@ object MenasRestDAO extends MenasDAO {
...
-        else {
-          val responseBody = getResponseBody(response)
-          log.error(s"RESPONSE: ${response.getStatusLine} - $responseBody")
+        if (unAuthorized) {
+          log.warn(s"Unauthorized POST request for Menas URL: $url")
+          log.warn(s"Expired session, reauthenticating")
+          enceladusLogin()
+          log.info(s"Retrying POST request for Menas URL: $url")
+          sendPostJson(url, json)
+        } else {
```
Zejnilovic: Is there a need for this else?

GeorgiChochov: Yeah, otherwise we get into the other case as well after a successful retry, which logs an error for the original failed try at `146` and returns `ok` at `148`

Zejnilovic: I meant to make it else if

dao/src/main/scala/za/co/absa/enceladus/dao/MenasRestDAO.scala
```diff
@@ -116,10 +113,9 @@ object MenasRestDAO extends MenasDAO {
     sendPostJson(url, json)
   }
 
-  private def sendPostJson(url: String, json: String): Boolean = {
```
yruslan: Sorry, I'm being nitpicky. But consider this. instead of passing "retryCount" as a "current number of retries" you can pass "maxRetry". And recursively call `sendPostJson(url, json, retryCount - 1)` if `retryCount >0`. This way you won't need a `maxRetries` constant and you let the client choose the number of retries. A reasonable default would still be `1`.

Issue #549, opened 2019-06-26
Provide generic tools for UI5 tables


lokm01:
It seems that for sorting, grouping, filtering etc, we would have to provide popups with custom sort/filter logic per UI5 table.

With a generic component, we could dynamically add the settings header, generate & initialize the pop up etc.

This could provide APIs for sort functionality initially, but we could easily extend to support grouping, filtering etc.

PR #559, opened 2019-06-26
#549 Provide generic tools for UI5 tables

lokm01:
- For testing, implemented this for dataset audit trails
- If happy, we can implement on other tables as separate PRs

### DIFF
--- a/menas/ui/components/app.controller.js
+++ b/menas/ui/components/app.controller.js
@@ -16,8 +16,9 @@
 sap.ui.define([
   "sap/ui/core/mvc/Controller",
   "sap/ui/core/Fragment",
-  "sap/m/MessageToast"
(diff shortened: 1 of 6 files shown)


GeorgiChochov:
Don't version and change time necessarily follow the same order?

Zejnilovic:
> Don't version and change time necessarily follow the same order?

Good spot. I would say yes, but then again I can imagine the questions from user, why can't they sort using one or the other. I think it doesn't block anything and I would vote to keep it.

lokm01:
1) Separate issue pls
2) You're truly killing me.... check now

lokm01:
> Don't version and change time necessarily follow the same order?

Will super simple to change these once we merge the TableUtils I guess.

Zejnilovic:
> 1. Separate issue pls

#571

Issue #271, opened 2019-03-21
Improve audit trails and runs tables


Zejnilovic:
## Background
Currently, audit trails and runs tables react to each click of a user, so if I want to copy something I must be careful to not click something. And in general, having more than 80% of the page clickable isn't a good strategy

## Proposition
Make the table plaintext and add a show button.

Zejnilovic:
Fixed by #560 before releasing 1.0.0

PR #560, opened 2019-06-26
#271 Improve Audit Trail Table

lokm01:
- Made the version link clickable instead of the whole row
- Wrapped the change details with expandable panel


lokm01:
> image
> I like how the linking is done, but not sure if the expandable panel makes things better. I think it now gets even harder to figure out what has changed (having to click through these, although there really was no information there for schemas to begin with).

Fair enough, lets drop the panel again.

Issue #235, opened 2019-03-12
Add schema export feature


Zejnilovic:
I would like to be able to export a schema from Menas. This schema should be in the same format as I uploaded it. Pretty printed if possible.

Zejnilovic:
Duplicate of #34

GeorgiChochov:
Let's break #34 down into smaller tasks.

PR #562, opened 2019-06-26
#235 Add schema export button

GeorgiChochov:


lokm01:
Awesome, this will be useful particularly for the cobol use cases ;)

Zejnilovic:
## 1 Cobol copybook has a `.json` file extension
When exported Cobol copybook is `SchemanName-Version.json` and then there are clashes with predefined tools for JSONs and need to rename.

GeorgiChochov:
It might make sense for me to upload a file and download it with a familiar name, but if somebody else downloads it that name might not make sense (or be particularly appropriate). I'll name the file based on the schema name and version, with an extension based on the MIME type.

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/SchemaController.scala
```diff
@@ -72,6 +74,19 @@ class SchemaController @Autowired() (
...
 
+  @GetMapping(path = Array("/export/{name}/{version}"))
+  @ResponseStatus(HttpStatus.OK)
+  def exportSchema(@AuthenticationPrincipal principal: UserDetails,
+                   @PathVariable name: String,
+                   @PathVariable version: Int,
+                   response: HttpServletResponse): CompletableFuture[Array[Byte]] = {
+    attachmentService.getSchemaByNameAndVersion(name, version).map { attachment =>
+      response.addHeader("filename", attachment.filename)
+      response.setContentType(attachment.fileMIMEType)
```
GeorgiChochov: This gets overridden by the fact that we are returning `Array[Byte]` and is always set to `application/octet-stream`.
I should write tests for this API endpoint.

menas/ui/components/schema/schemaDetail.controller.js
```diff
@@ -109,6 +110,17 @@ sap.ui.define([
         })
     },
 
+    _getFileExtension: function (mimeType) {
```
Zejnilovic: Seems like the cob isn't added there
image

I am uploading a copybook that is `NAMEXX.txt` and mongo says `{ "fileMIMEType": "text/plain" }`

Zejnilovic: Actually, even JSON now downloads a txt

GeorgiChochov: Mine have `.json` going in and `.json` coming out. Everything that has `.txt` comes out `.txt`. Things that have `.cob` or no extension going in, come out `.cob`. I'm not in control of the MIME type coming in...
What is your suggestion?

Zejnilovic: Not a suggestion but an issue 
Everything downloads as plain text file

menas/src/main/scala/za/co/absa/enceladus/menas/repositories/AttachmentMongoRepository.scala
```diff
@@ -15,17 +15,48 @@
...
+  }
+
+  private def getByCollectionAndNameAndVersion(refCollection: String,
+                                               name: String,
+                                               version: Int): Future[Option[MenasAttachment]] = {
+
+    val pipeline = Seq(
+      filter(equal("refCollection", refCollection)),
+      filter(equal("refName", name)),
+      project(Document(s"{diff: {$$subtract: [$version, '$$refVersion']}, doc: '$$$$ROOT'}")),
```
yruslan: 👍 Nice query

menas/ui/components/schema/schemaDetail.controller.js
GeorgiChochov: It happens, thanks for testing!

Issue #568, opened 2019-06-27
Cannot file format 'cobol' when running Standardization


yruslan:
## Describe the bug
> **Exception in thread "main" java.lang.ClassNotFoundException: Failed to find data source: cobol. Please find packages at http://spark.apache.org/third-party-projects.html**
	at org.apache.spark.sql.execution.datasources.DataSource$.lookupDataSource(DataSource.scala:657)
	at org.apache.spark.sql.DataFrameReader.load(DataFrameReader.scala:194)
	at org.apache.spark.sql.DataFrameReader.load(DataFrameReader.scala:178)
	at za.co.absa.enceladus.standardization.StandardizationJob$.prepareDataFrame(StandardizationJob.scala:171)
	at za.co.absa.enceladus.standardization.StandardizationJob$.main(StandardizationJob.scala:111)
	at za.co.absa.enceladus.standardization.StandardizationJob.main(StandardizationJob.scala)

## To Reproduce
Run a Standardization using 'cobol' as a file format
 
## Expected behaviour
Standardization should be able to recognize COBOL format.

## Desktop (please complete the following information):
 - OS: Linux cluster

## Additional context
Both `spark-xml` and `spark-cobol` register shot names for input formats in `META-inf.services/org.apache.spark.sql.sources.DataSourceRegister`. But when both are used in the same jar, one of formats configurations overwrites the others.

## Proposed solution
Use full class path format specifier for COBOL: `za.co.absa.cobrix.spark.cobol.source`

yruslan:
@lokm01 suggested to add an appender transformer for `META-inf.services/org.apache.spark.sql.sources.DataSourceRegister`. Looking into it...

PR #570, opened 2019-06-27
#568 Fix issues related to loading files in COBOL format

yruslan:
With these fixes in place the end-to-end standardization of a mainframe file ran successfully.


yruslan:
Sorry for the force push spam. Implemented more elegantly according to @lokm01 suggestion.

Issue #463, opened 2019-06-11
Add '--report-date' and '--report-version' to Run documents in MongoDB


DzMakatun:
## Background
1) Monitoring currently relies on informationDate and version (run->controlMeasure->metadata->informationDate/version) in documents of the run collection. These data are derived from _INFO files provided by users. The values may have arbitrary formats and errors. A more consistent way is needed.
2) Timestamps are currently stored as strings, as a result, monitoring queries require casting [dateFromString](https://docs.mongodb.com/manual/reference/operator/aggregation/dateFromString/). Storing timestamps in [ISODate](https://docs.mongodb.com/manual/core/shell-types/) format will improve performance. 

## Feature
Add the following attributes to the documents in the run mongodb collection:
a) **reportDate** :  matching the --report-date argument of enceladus job, the corresponding part of HDFS path, and enceladus_info_date= in splineRef. The format should be ISODate.  
b) **reportVersion** : matching the --report-version argument of enceladus job, the corresponding part of HDFS path, andenceladus_info_version= in splineRef. The format should be integer.  

## Example [Optional]
"reportDate" : ISODate("2019-05-23T00:00:00.000Z")
"reportVersion" : 1

yruslan:
Also, raw_record_count, std_record_count, conform_record_count.

PR #590, opened 2019-07-01
#463 Add new metrics to Atum's metadata

yruslan:
These metrics will help gethering statistics about data processed by Enceladus.
The list of new metrics: enceladus_info_date, enceladus_info_version,
raw_record_count, std_record_count, conform_record_count.

### DIFF
--- a/standardization/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
+++ b/standardization/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
@@ -341,7 +341,15 @@ object StandardizationJob {
     })
 
     measurement.flatMap(m =>
-      try Some(m.controlValue.toString.toLong)
+      try {
+        val rawCount = m.controlValue.toString.toLong
+        // Basic sanity check
+        if (rawCount >=0 ) {
+          Some(rawCount)
+        } else {
+          None
+        }
+      }
       catch {
         case NonFatal(_) => None
       }
--- a/standardization/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
+++ b/standardization/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
@@ -103,7 +103,7 @@ object StandardizationJob {
     // Enable Menas plugin for Control Framework
     MenasPlugin.enableMenas(cmd.datasetName, cmd.datasetVersion, isJobStageOnly = true)
 
-    // Add report date and version (aka Enceladus info data and version) to Atum's metadata
+    // Add report date and version (aka Enceladus info date and version) to Atum's metadata
     Atum.setAdditionalInfo(s"enceladus_info_date" -> cmd.reportDate)
     Atum.setAdditionalInfo(s"enceladus_info_version" -> reportVersion.toString)
 
@@ -324,7 +324,6 @@ object StandardizationJob {
     Atum.setAdditionalInfo(s"raw_record_count" -> rawRecordCount.toString)
   }
 
-
   /**
     * Gets the number of records in raw data by traversing Atum's checkpoints.
     *
@@ -348,6 +347,5 @@ object StandardizationJob {
       }
     )
   }
-
 
 }
--- a/standardization/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
+++ b/standardization/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
@@ -21,7 +21,6 @@ import java.text.MessageFormat
 import java.util.UUID
 
 import scala.util.control.NonFatal
-
 import org.apache.log4j.LogManager
 import org.apache.log4j.Logger
 import org.apache.spark.sql.Column
@@ -30,13 +29,11 @@ import org.apache.spark.sql.DataFrameReader
 import org.apache.spark.sql.SparkSession
 import org.apache.spark.sql.types.StructField
 import org.apache.spark.sql.types.StructType
-
 import com.typesafe.config.Config
 import com.typesafe.config.ConfigFactory
-
 import za.co.absa.atum.AtumImplicits
 import za.co.absa.atum.AtumImplicits.DataSetWrapper
-import za.co.absa.atum.core.Atum
+import za.co.absa.atum.core.{Atum, Constants}
 import za.co.absa.enceladus.dao.EnceladusRestDAO
 import za.co.absa.enceladus.dao.menasplugin.MenasPlugin
 import za.co.absa.enceladus.model.Dataset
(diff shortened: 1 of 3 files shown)


standardization/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
```diff
@@ -106,6 +103,10 @@ object StandardizationJob {
     // Enable Menas plugin for Control Framework
     MenasPlugin.enableMenas(cmd.datasetName, cmd.datasetVersion, isJobStageOnly = true)
 
+    // Add report date and version (aka Enceladus info data and version) to Atum's metadata
```
Zejnilovic:
```suggestion
    // Add report date and version (aka Enceladus info date and version) to Atum's metadata
```

standardization/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
```diff
@@ -303,4 +306,48 @@ object StandardizationJob {
...
+    val checkpointRawRecordCount = getRawRecordCountFromCheckpoints
+
+    val rawRecordCount = checkpointRawRecordCount match {
+      case Some(num) => num
+      case None      => df.count
+    }
+    Atum.setAdditionalInfo(s"raw_record_count" -> rawRecordCount.toString)
+  }
+
+
```
Zejnilovic: Needless new line

standardization/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
```diff
@@ -303,4 +306,48 @@ object StandardizationJob {
...
+    measurement.flatMap(m =>
+      try Some(m.controlValue.toString.toLong)
+      catch {
+        case NonFatal(_) => None
+      }
+    )
+  }
+
+
```
Zejnilovic: Needless new line

utils/src/main/scala/za/co/absa/enceladus/utils/performance/PerformanceMetricTools.scala
```diff
@@ -94,12 +95,13 @@ object PerformanceMetricTools {
...
   /** Returns the number of records failed, the number of records succeeded and the total number of errors encountered
     * when running a Standardization or a Dynamic Conformance job. */
-  private def getNumberOfErrors(spark: SparkSession, outputPath: String): (Long, Long, Long) = {
+  private def getNumberOfErrors(spark: SparkSession, outputPath: String): (Long, Long, Long, Long) = {
     val df = spark.read.parquet(outputPath)
     val errorCountColumn = SchemaUtils.getClosestUniqueName("enceladus_error_count", df.schema)
     val errCol = col(ErrorMessage.errorColumnName)
     val numRecordsFailed = df.filter(size(errCol) > 0).count
     val numRecordsSuccessful = df.filter(size(errCol) === 0).count
+    val numTotal = df.count
```
DzMakatun: As far as I understand, the 3 lines above are independent actions, which will cause Spark to traverse the RDD 3 times. Is it possible to combine them all into one action?

DzMakatun:
It is good you are using case insensitive comparison for controlType. Just out of curiosity, here is the statistics of how often different control types are used in one of our largest instances. 
{
    "_id" : "controlType.count",
    "count" : 420800.0
}
{
    "_id" : "controlType.aggregatedTotal",
    "count" : 331049.0
}
{
    "_id" : "controlType.Count",
    "count" : 134037.0
}
{
    "_id" : "controlType.absAggregatedTotal",
    "count" : 1654.0
}
{
    "_id" : "controlType.HashCrc32",
    "count" : 14.0
}

utils/src/main/scala/za/co/absa/enceladus/utils/performance/PerformanceMetricTools.scala
GeorgiChochov: Even if not pared down to a single action, isn't `numTotal == numRecordsFailed + numRecordsSuccessful `?

yruslan: Right now `numTotal == numRecordsFailed + numRecordsSuccessful` holds. But it might change the meaning according to user's request.

But after talking to Dima we concluded if users want to change a meaning of a metric a new metric should be added instead. So will combine calculate it like you said.

standardization/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
```diff
@@ -303,4 +306,46 @@ object StandardizationJob {
...
+    * Gets the number of records in raw data by traversing Atum's checkpoints.
+    *
+    * @return The number of records in a checkpoint corresponding to raw data (if available)
+    */
+  private def getRawRecordCountFromCheckpoints: Option[Long] = {
+    val controlMeasure = Atum.getControMeasure
+
+    val rawCheckpoint = controlMeasure
+      .checkpoints
+      .find(c => c.name.equalsIgnoreCase("raw") || c.workflowName.equalsIgnoreCase("raw"))
```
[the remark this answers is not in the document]
yruslan: - If Raw is not present I fallback to `df.count` on the initial dataframe. Although not perfect, isn't it the best thing we can do?
- If we write `raw_record_count` metric based on 'Source' checkpoint it will be confusing, won't it?
- Record count can be 0, can't it?
- I'm not sure we should do any validations and sanity checks at this stage. These mismatches can be detected in Statistics/Monitoring. That's what it does already - alerts on mismatches.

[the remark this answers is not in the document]
DzMakatun: Ok, since we are going to check controls of all of the checkpoints separately, the source value is probably not required here. Out of 88899 runs, the Source checkpoint is missing in 2841 instances, the Raw is missing in 3600. So the relative difference is not huge. 

I would still recommend to check the value from raw to be non-negative and fallback to count. This way the result will be more reliable and require less checks in monitoring/statistics.

utils/src/main/scala/za/co/absa/enceladus/utils/performance/PerformanceMetricTools.scala
DzMakatun: Maybe something like? 
(numRecordsFailed,  numRecordsSuccessful) = df
.map( row => (size(row(errCol)) > 0 ? (1,0) : (0,1) )
.reduce( (a,b) => (a._1 + b._1, a._2+b._2 ) )

standardization/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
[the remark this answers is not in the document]
yruslan: Added the sanity check you requested.

utils/src/main/scala/za/co/absa/enceladus/utils/performance/PerformanceMetricTools.scala
yruslan: The approach you specified does not compile. Tried a different one - with 2 aggregates.
```scala
    val result = df
      .agg(sum( when(size(errCol)>0, 1).otherwise(0) ).as("failed"),
        sum( when(size(errCol)>0, 0).otherwise(1) ).as("succeeded") )
      .collect()(0)

    val numRecordsFailed = result(0)
    val numRecordsSuccessful = result(1)
```
It works, but has caveats. Need to chcek against nulls if there are no records.

I'd prefer leaving this piece of code like it was for now.

PR #599, opened 2019-07-02
#598 Add local search to ui5 TableUtils

lokm01:


lokm01:
> I don't get what the search is searching for (`author`, content of `old value` and `new value`, maybe something else as well). I think we should build a proper filtering system where we can specify query parameters.

Its looking at author in this case and changes (object), which converted to json and then just text search in that.

100% agree with proper filtering functionality. These are meant to be utilities for small (local) tables without the need to integrate with backend for sort, group, search

Issue #458, opened 2019-06-10
Column enceladus_info_date does not respect timezones


Zejnilovic:
## Describe the bug
The new column that is written at the end of the DynamicConformance job does not respect timezones. My local timezone is set to Europe/Prague and when the date converts it always subtracts a day so it seems I ran it a day before the data even existed.

## To Reproduce
1. Change your timezone to Europe/Prague
2. Run conformance job
3. output folder date is correct
4. enceladus_info_date is mismatched by a day 

## Expected behaviour
I would expect timezone to be normalized or ignored in this part.

## Additional context
My timezone
```scala
scala> TimeZone.getDefault
res34: java.util.TimeZone = sun.util.calendar.ZoneInfo[id="Europe/Prague",offset=3600000,dstSavings=3600000,useDaylight=true,transitions=143,lastRule=java.util.SimpleTimeZone[id=Europe/Prague,offset=3600000,dstSavings=3600000,useDaylight=true,startYear=0,startMode=2,startMonth=2,startDay=-1,startDayOfWeek=1,startTime=3600000,startTimeMode=2,endMode=2,endMonth=9,endDay=-1,endDayOfWeek=1,endTime=3600000,endTimeMode=2]]
```

Error traced to _DynamicConformanceJob.scala_ line 110

benedeki:
The reason of the bug was that `SimpleDateFormat` object was created before the `TimeZoneNormalizer` executed its logic. Within the constructor of the `SimpleDateFormat` the current default time zone (at this time equal to system time zone) is bound in, and later used in some flows.

The main solution is in PR #604. 

There's an optional PR #605, which intent is to reduce the chance of a similar problems in future as much as possible. On the other hand, it contains some logic executed on the instantiation of the companion object, which might be considered undesired.

PR #604, opened 2019-07-03
#458: Column enceladus_info_date does not respect timezones

benedeki:
* simplifying the expression for enceladus_info_date, avoiding the source of problems (DynamicConformanceJob.processResult)
* Ensuring simpleDateFormat is created only after TimeZoneNormalizer logic is executed
* LoggerTestBase enhancement
* DynamicConformanceJob.main broke up into subroutines to make it shorter (hopefully clearer too)
* conformanceRuleShouldMatchExpected uses logger instead of println (not related to bug, spotted on bug investigation)
* Some stylistic adjustments

(Source of the bug - SimpleDateFormat was created before TimeZoneNormalizer logic, which bound that instance to the system time zone.

### DIFF
--- a/conformance/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
+++ b/conformance/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
@@ -88,7 +88,7 @@ object DynamicConformanceJob {
     // load data for input and mapping tables
     val inputData = DataSource.getData(stdPath, dateTokens(0), dateTokens(1), dateTokens(2), "")
 
-    val result = performTheConformance(conformance, inputData, enableCF)
+    val result = conform(conformance, inputData, enableCF)
 
     processResult(result, performance, publishPath, stdPath, reportVersion)
   }
@@ -172,8 +172,8 @@ object DynamicConformanceJob {
     performance
   }
 
-  private def performTheConformance(conformance: Dataset, inputData: sql.Dataset[Row], enableCF: Boolean)
-                                   (implicit spark: SparkSession, cmd: CmdConfig, fsUtils: FileSystemVersionUtils, dao: EnceladusDAO): DataFrame = {
+  private def conform(conformance: Dataset, inputData: sql.Dataset[Row], enableCF: Boolean)
+                     (implicit spark: SparkSession, cmd: CmdConfig, fsUtils: FileSystemVersionUtils, dao: EnceladusDAO): DataFrame = {
     try {
       DynamicInterpreter.interpret(conformance, inputData, isExperimentalRuleEnabled(), isCatalystWorkaroundEnabled(), enableCF)
     } catch {
--- a/conformance/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
+++ b/conformance/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
@@ -18,19 +18,15 @@ package za.co.absa.enceladus.conformance
 import java.io.PrintWriter
 import java.io.StringWriter
 import java.text.MessageFormat
-import java.text.SimpleDateFormat
 
 import scala.util.control.NonFatal
-
 import org.apache.log4j.LogManager
 import org.apache.log4j.Logger
-import org.apache.spark.sql.DataFrame
-import org.apache.spark.sql.SparkSession
-import org.apache.spark.sql.functions.lit
-
+import org.apache.spark.sql.{DataFrame, Row, SparkSession}
+import org.apache.spark.sql.functions.{lit, to_date}
 import com.typesafe.config.Config
 import com.typesafe.config.ConfigFactory
-
+import org.apache.spark.sql
 import za.co.absa.atum.AtumImplicits
 import za.co.absa.atum.AtumImplicits.DataSetWrapper
 import za.co.absa.atum.AtumImplicits.StringToPath
@@ -49,8 +45,8 @@ import za.co.absa.enceladus.utils.time.TimeZoneNormalizer
 
 object DynamicConformanceJob {
 
-  private val format = new SimpleDateFormat("yyyy-MM-dd")
   private val infoDateColumn = "enceladus_info_date"
+  private val reportDateFormat = "yyyy-MM-dd"
   private val infoVersionColumn = "enceladus_info_version"
 
   private val log: Logger = LogManager.getLogger(this.getClass)
@@ -59,10 +55,10 @@ object DynamicConformanceJob {
   def main(args: Array[String]) {
     implicit val spark: SparkSession = obtainSparkSession() // initialize spark
     implicit val cmd: CmdConfig = CmdConfig.getCmdLineArguments(args)
-    implicit val fsUtils = new FileSystemVersionUtils(spark.sparkContext.hadoopConfiguration)
-
+    implicit val fsUtils: FileSystemVersionUtils = new FileSystemVersionUtils(spark.sparkContext.hadoopConfiguration)
     implicit val dao: EnceladusDAO = EnceladusRestDAO // use REST DAO
-    implicit val enableCF: Boolean = true
+
+    val enableCF: Boolean = true
 
     EnceladusRestDAO.login = cmd.menasCredentials
     EnceladusRestDAO.enceladusLogin()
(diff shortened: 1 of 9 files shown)


standardization/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
```diff
@@ -152,7 +151,7 @@ object StandardizationJob {
   private def applyCobolOptions(dfReader: DataFrameReader,
                                 cmd: CmdConfig,
                                 dataset: Dataset): DataFrameReader = {
-    val isXcom = cmd.cobolOptions.map(_.isXcom).getOrElse(false)
+    val isXcom = cmd.cobolOptions.exists(_.isXcom)
```
yruslan: Yes, IDEA suggested this change. But I liked the original better since intent is more expressive there.

benedeki: Yes, it's from IDEA suggestion. But I admit, I  find the suggestion good (most of them I actually do 😉 ) - to me it's seems clear, clearer then the original.

PS
I touched the file during some debugging. Thought it good to keep. Can _revert_ in general preference is there. 😃

conformance/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
```diff
@@ -173,9 +143,55 @@ object DynamicConformanceJob {
...
+  private def initPerformanceMeasurer(stdPath: String)
+                                     (implicit spark: SparkSession, fsUtils: FileSystemVersionUtils): PerformanceMeasurer = {
+    // init performance measurer
+    val performance = new PerformanceMeasurer(spark.sparkContext.appName)
+    val stdDirSize = fsUtils.getDirectorySize(stdPath)
+    performance.startMeasurement(stdDirSize)
+    performance
+  }
+
+  private def performTheConformance(conformance: Dataset, inputData: sql.Dataset[Row], enableCF: Boolean)
```
GeorgiChochov: How about we call this `conform`?

conformance/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
```diff
@@ -173,9 +143,55 @@ object DynamicConformanceJob {
...
+        AtumImplicits.SparkSessionWrapper(spark).setControlMeasurementError("Conformance", e.getMessage, sw.toString)
+        throw e
+    }
+  }
+
   private def processResult(result: DataFrame, performance: PerformanceMeasurer, publishPath: String, stdPath: String, reportVersion: Int)
                            (implicit spark: SparkSession, cmd: CmdConfig, fsUtils: FileSystemVersionUtils): Unit = {
-    val withPartCols = result.withColumn(infoDateColumn, lit(new java.sql.Date(format.parse(cmd.reportDate).getTime)))
+    val withPartCols = result
+      .withColumn(infoDateColumn, to_date(lit(cmd.reportDate), reportDateFormat))
```
GeorgiChochov:
```
      to_date(lit(cmd.reportDate), reportDateFormat)
```
This is the only non-refactoring change in this file, right?

menas/src/main/scala/za/co/absa/enceladus/menas/repositories/OozieRepository.scala
```diff
@@ -363,3 +361,10 @@ class OozieRepository @Autowired() (oozieClientRes: Either[OozieConfigurationExc
     this.writeScheduleData(coordPath, content)
   }
 }
+
+object OozieRepository {
```
GeorgiChochov: The companion objects in the other `Repository` classes are at the top of the file. Please follow convention.

conformance/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
benedeki: Nope, the changes in _OozieRepository_ are also "functional"

menas/src/main/scala/za/co/absa/enceladus/menas/services/DatasetService.scala
```diff
@@ -21,31 +21,28 @@ import za.co.absa.enceladus.model.conformanceRule.ConformanceRule
...
-			.setSchemaVersion(dataset.schemaVersion)
-			.setHDFSPath(dataset.hdfsPath)
-			.setHDFSPublishPath(dataset.hdfsPublishPath)
-			.setConformance(dataset.conformance)
-			.setDescription(dataset.description).asInstanceOf[Dataset]
-			})
-			}
-	}
+  override def update(username: String, dataset: Dataset): Future[Option[Dataset]] = {
+      super.updateFuture(username, dataset.name, dataset.version) { latest =>
```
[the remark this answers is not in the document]
benedeki: Missed that one. :-)

utils/src/main/scala/za/co/absa/enceladus/utils/testUtils/LoggerTestBase.scala
```diff
@@ -21,16 +21,36 @@ import org.apache.spark.sql.DataFrame
...
       df.show(truncate = false)
     }
     val dfData = new String(outCapture.toByteArray).replace("\r\n", "\n")
-    logger.debug(dfData)
+    logFnc(dfData)
+  }
+}
+
+object LoggerTestBase {
+  object LogLevel extends Enumeration {
```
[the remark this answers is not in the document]
benedeki: Haha, my first _Scala_ `Enumeration` implantation used `sealed case objects`. I was told that in most cases it's an overkill, although for my particular case back then it was considered justified.

utils/src/main/scala/za/co/absa/enceladus/utils/testUtils/LoggerTestBase.scala
```diff
@@ -21,16 +21,36 @@ import org.apache.spark.sql.DataFrame
...
+    logLevel match {
+      case TRACE => logger.trace
+      case DEBUG => logger.debug
+      case INFO  => logger.info
+      case WARN  => logger.warn
+      case _     => logger.error
+    }
+  }
+
+  protected def logDataFrameContent(df: DataFrame, logLevel: LogLevel = DEBUG): Unit = {
```
yruslan: Consider having a function that gets a string out of a dataframe, say, `getDataFrameShowString(df)`.

Then, you won't need `logLevelToLogFunction()` and just use:
```scala
log.error(getDataFrameShowString(df))
```

Also, `getDataFrameShowString()` is easily testable, while `logDataFrameContent()` is no so easily testable.

benedeki: The testability is a point. On the other hand this is just a logging function, not really needs testing.
Also the point the function was created, was to just throw in the dataset and forget about it. The "string extracting" function would be a two step solution again.

yruslan: Testability is not the only point.
* Current implementation required you to create a sealed trait with all the case objects that replicate log levels of the underlying logging library. Looks a little bit redundant.
* Having contents of a DataFrame as a string is more reusable. For instance, if you want to `assert` differences of outputs returned by `.show()` you can use the same function for converting DataFrame contents to String.
* Fetching contents of a DataFrame and logging seems like 2 orthogonal responsibilities. When split they can be reused better.

standardization/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
yruslan: If you guys like it, I'm fine. Just think what you need to do to make the default value `true`.

utils/src/main/scala/za/co/absa/enceladus/utils/testUtils/LoggerTestBase.scala
```diff
@@ -19,18 +19,41 @@ import java.io.ByteArrayOutputStream
...
     }
     val dfData = new String(outCapture.toByteArray).replace("\r\n", "\n")
-    logger.debug(dfData)
+    logFnc(dfData)
   }
 }
+
+sealed trait LogLevel
+
+object LogLevel {
```
Zejnilovic: Is there a reason for not using [slf4j's log level enum](https://www.slf4j.org/apidocs/org/slf4j/event/Level.html)

utils/src/main/scala/za/co/absa/enceladus/utils/testUtils/LoggerTestBase.scala
benedeki: OK I created the function(s) to get the `DataFrame` data as string. But @yruslan I dare you I expect you to use it in at least one of your next 3 pull requests. 😉

utils/src/main/scala/za/co/absa/enceladus/utils/testUtils/LoggerTestBase.scala
benedeki: Now using it, Thanks for bringing this to my knowledge.

conformance/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
GeorgiChochov: I mean everything else in **this file** is refactoring

benedeki: Eh, sorry for misunderstanding. Yes, only in this file.

PR #605, opened 2019-07-03
#458: Column enceladus_info_date does not respect time zones (optional improvement)

benedeki:
* Changing TimeZoneNormalizer to normalize JVM time zone upon the object instantiation, in other words as soon as possible

### DIFF
--- a/utils/src/main/scala/za/co/absa/enceladus/utils/time/TimeZoneNormalizer.scala
+++ b/utils/src/main/scala/za/co/absa/enceladus/utils/time/TimeZoneNormalizer.scala
@@ -28,6 +28,8 @@ object TimeZoneNormalizer {
   private val log: Logger = LogManager.getLogger(this.getClass)
   private lazy val timeZone: String = getConf("timezone", "UTC")
 
+  normalizeJVMTimeZone()
+
   private def getConf(path: String, default: String): String = {
     val config: Config = ConfigFactory.load()
     if (config.hasPath(path)) {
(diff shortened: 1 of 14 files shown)


benedeki:
This PR is an optional improvement on #604. 
The intent is to reduce the chance of a problem similar to  #458 in future as much as possible. On the other hand, it contains some logic executed on the instantiation of the companion object, which might be considered undesired.

benedeki:
Eh, sorry omitted the _bugfix/_ prefix in the branch name 😊

utils/src/main/scala/za/co/absa/enceladus/utils/time/TimeZoneNormalizer.scala
```diff
@@ -38,6 +40,11 @@ object TimeZoneNormalizer {
     }
   }
 
+  /**
+    * This is just an empty function suitable to be called to ensure the object is created and therefore JVM time zone normalized
+    */
+  def activate(): Unit = {}
```
yruslan: Why not call `normalizeJVMTimeZone()` everywhere you call now `activate()`?

benedeki: The `normalizeJVMTimeZone()` is enough to execute only once. And at the same time it's a third party outside logic, which cost I don't know - therefore I try to avoid unnecessary multiple calls.

GeorgiChochov: I'd like to get CQC opinion on this.

benedeki: That's really a cool idea. It's fast to explain the background of this **optional** change. And I can than add my reasoning for the chosen solution. 👍 
And this is not really a hot issue, the original bug was fixed in the other PR.

benedeki:
Will not be merged.

Issue #626, opened 2019-07-08
In case Menas is running on https Spline on http protocol won't load


benedeki:
## Describe the bug
In case **Menas** page is is running on `https` protocol and **Spline** (Lineage) on `http` protocol, the page might not load

## To Reproduce
Steps to reproduce the behavior OR commands run:
1. Configure `za.co.absa.enceladus.spline.urlTemplate` using `http` protocol. E.g. `za.co.absa.enceladus.spline.urlTemplate=http//localhost:8080/spline/dataset/lineage/_search?path=%s&application_id=%s`
1. Run **Menas** using `https` protocol
3. Go to _Runs->Lineage_
4. IFrame won't load

## Expected behaviour
Avoid the error

PR #627, opened 2019-07-08
#626: In case Menas is running on https Spline on http protocol won't load

benedeki:
* changing the application.properties.template suggesting the solution for application.properties

### DIFF
--- a/menas/src/main/resources/application.properties.template
+++ b/menas/src/main/resources/application.properties.template
@@ -62,7 +62,9 @@ za.co.absa.enceladus.migrations.dump.filepath=/path/to/backups/menas
 
 za.co.absa.enceladus.menas.spark.master=local[1]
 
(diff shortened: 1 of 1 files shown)


yruslan:
Wow, that looks like a very nice solution!

I'm wondering out of curiosity if it is possible to be able to not to specify the hostname+port as well and just use the same hostname+port where Menas is running? So if we have several instance of Menas and Spline running it would automatically 'load balance' by directing to Spline instance located at the same host.

benedeki:
> 
> 
> Wow, that looks like a very nice solution!
> 
> I'm wondering out of curiosity if it is possible to be able to not to specify the hostname+port as well and just use the same hostname+port where Menas is running? So if we have several instance of Menas and Spline running it would automatically 'load balance' by directing to Spline instance located at the same host.

Unfortunately, I don't think it would work automatically out of the box for other parts of the URL. Already omitting the  two slashes at the beginning breaks the stuff. It would have to be solved within the code - e.g expanding before assigning to `iframes.src`.

Issue #637, opened 2019-07-09
Hamburger menu shows on login page on small width screen


Zejnilovic:
## Describe the bug
Some time ago we removed the hamburger menu from the login page. But there is still hamburger menu on a login page if the page's width is small enough to be viewed in the responsive mode. This button then doesn't do anything, just throws errors to console.

## To Reproduce
Steps to reproduce the behavior OR commands run:
1. Go to login page
2. Make the page 800px in width
3. See in the top left corner the hamburger menu

## Additional info
error in the console: `Uncaught TypeError: Cannot read property 'offsetHeight' of nul`

PR #638, opened 2019-07-09
#637 Hamburger menu shows on login page on small width screen

lokm01:
Also added a pubsub event for global resize.


lokm01:
Hmmm... Added 50ms timeout. Though it's weird, it was hiding properly on my side.

lokm01:
Well it seems that around 950 the App instance calls some code to reformat the master-detail pattern. I think we'd have to put a 200ms timeout to be safe. But I doubt this would be an issue

Issue #633, opened 2019-07-09
Unordered Mongo queries


GeorgiChochov:
## Describe the bug
Currently, most of our Mongo queries are not explicitly sorted, so they use natural order. Natural order can often coincide with insertion order, but not always. This causes results to have an unexpected order and occasionally to fail Mongo integration tests (which breaks CI).

## To Reproduce
Steps to reproduce the behavior OR commands run:
1. Go to Jenkins
2. See failed builds
3. Look for cause
4. See expected and actual results differ only by order

## Expected behaviour
Results should come in a predictable order.

PR #639, opened 2019-07-09
#633 Sort mongo queries for versioned entities

GeorgiChochov:


lokm01:
Seems that 2 are still fail though?

Issue #641, opened 2019-07-10
Implement TableUtils sort & search to audit trails in other menas entities


lokm01:
Currently this is only implemented on dataset audit trails.

PR #642, opened 2019-07-10
#641 Implement TableUtils to audit trails in other menas entities

lokm01:


menas/ui/components/AuditTrail.js
```diff
@@ -0,0 +1,34 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+sap.ui.define(["components/tables/TableUtils"],
+    function(TableUtils) {
+
+  const AuditTrail = sap.ui.base.Object.extend("components.AuditTrail", {
+    constructor: function(oControl) {
```
GeorgiChochov: How about we call this `auditTableUtils`, since that is what's required, not a generic `control`?

lokm01: Cool

Issue #592, opened 2019-07-01
Add run details to bar chart on click in Monitoring


DzMakatun:
For better usability, clicking on a bar chart should show details of the related run.

PR #645, opened 2019-07-10
#592 Add BarChat popover, change tooltips, add sun start and attemp c…

DzMakatun:
#592 Add run details to bar chart on clic
- Change chart tooltips to include more data
- Add run start time to monitoring
- Add run attempts counter (per infodate and version)
- Minor layout adjustments

### DIFF
--- a/menas/ui/components/dataset/datasetDetail.controller.js
+++ b/menas/ui/components/dataset/datasetDetail.controller.js
@@ -475,10 +475,8 @@ sap.ui.define([
     },
 
     onFindRunsPress: function (oEv){
-      // Check selected interval
       this.updateMonitoringData()
     },
-
     setDefaultMonitoringDateInterval: function () {
       let oEnd = new Date();
       let oStart = new Date(oEnd.getTime() - 14 * 24 * 60 * 60 * 1000); // Two weeks before today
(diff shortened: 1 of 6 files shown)


menas/ui/components/dataset/datasetDetail.controller.js
```diff
@@ -497,6 +495,40 @@ sap.ui.define([
...
+        this._oPopover.close();
+      }
+    },
+
+    onRecordsBarClick: function (oEvent) {
+      let oChart = oEvent.getSource().__chart;
+      let oActiveElement = oChart.getElementsAtEvent(oEvent.getParameters().event)[0];
+      if (!oActiveElement) return;
+      let runIndex = oActiveElement._index;
+      console.log(oEvent.getSource())
```
lokm01: Let's remove the logs pls

Zejnilovic:
## 1 Can we make the popup centred for the view?
Right now the pop up from bar chart moves as I scroll up and down. Can it be shown always at the same place?
image
image
image

Zejnilovic:
## 2 Close button for pop-up
I can already anticipate a question where is a close button. Can we add it?
Two possible placements
image

DzMakatun:
Thank you, @Zejnilovic 
1.  Changed position to always at the top of the page
2. Added close button

Issue #614, opened 2019-07-05
Schema search when linking to new dataset


aaishaosman:
## Background
when creating a new dataset, user needs to link a schema. Issue comes about when there are thousands of schemas, and it is then difficult for the user to go through the entire list to get to the required one.

## Feature
Adding in the ability to search for the required schema when linking would make it more user friendly.

## Example [Optional]
added a screenshot to show where the issue occurs

![image](https://user-images.githubusercontent.com/17737522/60722727-11f59880-9f32-11e9-8f8d-768609fd8d9e.png)

PR #652, opened 2019-07-12
#614 Schema search when linking to new dataset

lokm01:
Changed sap.m.Select with sap.m.SelectDialog

### DIFF
--- a/menas/ui/css/style.css
+++ b/menas/ui/css/style.css
@@ -26,6 +26,18 @@ html, body {
   min-width: 40rem !important;
 }
 
+.minW30rem {
+  min-width: 30rem !important;
+}
+
+.maxH1000 {
+  max-height: 90% !important;
+}
+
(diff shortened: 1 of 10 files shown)


Zejnilovic:
## 1 Schema numbers linger
If I previously had schema selected and cancelled or am just creating a new dataset, the schema versions are still there and selected.

image

menas/ui/css/style.css
```diff
@@ -26,6 +26,18 @@ html, body {
   min-width: 40rem !important;
 }
 
+.minW30rem {
+  min-width: 30rem !important;
+}
+
+.maxH1000 {
+  max-height: 90% !important;
```
Zejnilovic: I don't know if this is because I don't know CSS, but this seems confusing to me. I would read it as `max height 1000` and would infer that it is in `px` but this is `90%`

Issue #669, opened 2019-07-16
Init keytab tickets in the helper scripts


yruslan:
## Background
When running in a kerberized environment a user needs 
- a keytab file to authenticte to Menas
- a valid Kerberos ticket to authenticate to HDFS 

## Feature
Create a Kerberos ticket is a keyfile is provided to a helper scripts.

Issue #674, opened 2019-07-16
Helper scripts minor enhancements


yruslan:
## Background
A couple of tiny enhancements have been requested by the users.

## Feature
1. Make standardized path overridable. 
2. Forbid specifying 'local' deployment mode when running Standardization and Conformance from scripts.

PR #676, opened 2019-07-16
#669 Automatically initialize Kerberos ticket if a keytab is provided

yruslan:


scripts/bash/run_enceladus.sh
```diff
@@ -231,9 +231,26 @@ if [[ -z "$DRY_RUN" ]]; then
     DATE=`date +%Y_%m_%d-%H_%M_%S`
     NAME=`sed -e 's#.*\.\(\)#\1#' <<< $CLASS`
     TMP_PATH_NAME="$LOG_DIR/enceladus_${NAME}_${DATE}.log"
+
+    # Initializing Kerberos ticket
+    if [[ ! -z "$MENAS_AUTH_KEYTAB" ]]; then
+      # Get principle stored in the keyfile
+      PR=`printf "read_kt $MENAS_AUTH_KEYTAB\nlist" | ktutil | sed -n '5p' | awk '{print $3}' | cut -d '@' -f1`
```
lokm01: 😲

scripts/bash/run_enceladus.sh
```diff
@@ -170,6 +174,11 @@ validate "$DATASET_NAME" "--dataset-name"
 validate "$DATASET_VERSION" "--dataset-version"
 validate "$REPORT_DATE" "--report-date"
 
+if [[ "$DEPLOY_MODE" != "client" ]] && [[ "$DEPLOY_MODE" != "cluster" ]]; then
```
lokm01: Is this meant to be checked against master instead?

yruslan:
Added fixes to #674 to this PR as well since it affects the same file.

scripts/bash/run_enceladus.sh
yruslan: Indeed

yruslan: Thanks a lot for spotting this, btw!

scripts/bash/run_enceladus.sh
GeorgiChochov: Jesus...

scripts/bash/run_enceladus.sh
```diff
@@ -231,9 +240,26 @@ if [[ -z "$DRY_RUN" ]]; then
     DATE=`date +%Y_%m_%d-%H_%M_%S`
     NAME=`sed -e 's#.*\.\(\)#\1#' <<< $CLASS`
     TMP_PATH_NAME="$LOG_DIR/enceladus_${NAME}_${DATE}.log"
+
+    # Initializing Kerberos ticket
+    if [[ ! -z "$MENAS_AUTH_KEYTAB" ]]; then
+      # Get principle stored in the keyfile
+      PR=`printf "read_kt $MENAS_AUTH_KEYTAB\nlist" | ktutil | sed -n '5p' | awk '{print $3}' | cut -d '@' -f1`
```
Zejnilovic: Is the `sed -n '5p'` failsafe way to go? on my mac, it seems it is on the 4th line. 
But I am not sure if a failsafe way exists. Could the bellow work on a larger scale? 
```bash
printf "read_kt $MENAS_AUTH_KEYTAB\nlist" | grep -Pio "(?<=\ )[A-Za-z0-9]*?(?=@) | head -1" 
```

Next thing is, do we want to throw an error if PR is empty?

scripts/bash/run_enceladus.sh
```diff
@@ -170,6 +174,11 @@ validate "$DATASET_NAME" "--dataset-name"
 validate "$DATASET_VERSION" "--dataset-version"
 validate "$REPORT_DATE" "--report-date"
 
+if [[ "$MASTER" != "yarn" ]]; then
```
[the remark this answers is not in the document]
Zejnilovic: It has been explained to me that there is a default value.

Issue #677, opened 2019-07-17
If Timestamp/Date pattern contains both literal parts and fractions of second placeholders it's likely to cause error


benedeki:
## Describe the bug
If Timestamp/Date pattern contains both literal parts and fractions of second placeholders namely if any literal part is before any second fraction placeholder it will fail _Standardization_ of such filed or even crash with exception.

## To Reproduce
Steps to reproduce the behavior OR commands run:
1. Set the timestamp pattern to `"'at' HH:mm:ss.SSS"`
2. Try to _Standardize_ or just use 'EnceladusDateTimeParser.format'
3. Exception, wouldn't it be `SSS` be at the tail it would "just" fail the conversion in _Standardization_

## Expected behaviour
This needs to work properly

# Additional context
The reason is that the quote characters are counted in when position of second fractions are analyzed. But in the actual string to convert they are not present making the index out of place, even outside the boundaries.

Related to #466

dk1844:
May be solved by https://github.com/AbsaOSS/spark-data-standardization/issues/7

PR #679, opened 2019-07-17
Bugfix/466 standardization support for second fractions

benedeki:
* Class `Section` to define parts of the pattern string and string to convert
* Implicit classes additional logic to enable String and Columns to operate with `Section` class
* Actual implementation of the support for second fractions
* **S**, **i** and **n** placeholders in pattern
* `EnceladusDateTimeParser` support for the new placeholders
* Standardization  (`TypeParser`) support for the new placeholders
* Fixing and enhancing UTs for the new second fractions placeholders
* Validators to support the new second fractions placeholders
* `README.md` update

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,8 +223,8 @@ To enable processing of time entries from other systems **Standardization** offe
 string and even numeric values to timestamp or date types. It's done using Spark's ability to convert strings to 
 timestamp/date with some enhancements. The pattern placeholders and usage is described in Java's 
 [`SimpleDateFormat` class description](https://docs.oracle.com/javase/8/docs/api/java/text/SimpleDateFormat.html) with 
-the addition of recognizing two keywords `epoch` and `milliepoch` (case insensitive) to denote the number of 
-seconds/milliseconds since epoch (1970/01/01 00:00:00.000 UTC).
+the addition of recognizing some keywords (like `epoch` and `milliepoch` (case insensitive)) to denote the number of 
+seconds/milliseconds since epoch (1970/01/01 00:00:00.000 UTC) and some additional placeholders.
 It should be noted explicitly that `epoch` and `milliepoch` are considered a pattern including time zone.
  
 Summary:
@@ -254,13 +254,21 @@ Summary:
 | z | General time zone | Pacific Standard Time; PST; GMT-08:00 |
 | Z | RFC 822 time zone | -0800 |
 | X | ISO 8601 time zone | -08; -0800; -08:00 |
-| _epoch_ | Seconds since 1970/01/01 00:00:00 | 1557136493|
-| _milliepoch_ | Milliseconds since 1970/01/01 00:00:00.0000| 15571364938124 |
+| _epoch_ | Seconds since 1970/01/01 00:00:00 | 1557136493, 1557136493.136|
+| _epochmilli_ | Milliseconds since 1970/01/01 00:00:00.0000| 15571364938124, 15571364938124.001 |
+| _epochmicro_ | Milliseconds since 1970/01/01 00:00:00.0000| 15571364938124789, 15571364938124789.999 |
+| _epochnano_ | Milliseconds since 1970/01/01 00:00:00.0000| 15571364938124789101 |
+| i | Microsecond | 111, 321001 |
+| n | Nanosecond | 999, 542113879 |
+
 
 **NB!** Spark uses US Locale and because on-the-fly conversion would be complicated, at the moment we stick to this 
 hardcoded locale as well. E.g. `am/pm` for `a` placeholder, English names of days and months etc.
 
 **NB!** The keywords are case **insensitive**. Therefore, there is no difference between `epoch` and `EpoCH`.
+
+**NB!** While _nanoseconds_ designation is supported on input, it's not supported in storage or further usage. So any
+value behind microseconds precision will be truncated.
    
 ##### Time Zone support
 As it has been mentioned, it's highly recommended to use timestamps with the time zone. But it's not unlikely that the 
--- a/README.md
+++ b/README.md
@@ -229,37 +229,37 @@ It should be noted explicitly that `epoch` and `milliepoch` are considered a pat
  
 Summary:
 
-| placeholder | Description | Example |
-| --- | --- | --- |
-| G | Era designator | AD |
-| y | Year | 1996; 96 |
-| Y | Week year | 2009; 09 |
-| M | Month in year (context sensitive) |  July; Jul; 07 |
-| L | Month in year (standalone form) | July; Jul; 07 |
-| w | Week in year | 27 |
-| W | Week in month | 2 |
-| D | Day in year | 189 |
-| d | Day in month |  10 |
-| F | Day of week in month | 2 |
-| E | Day name in week | Tuesday; Tue |
-| u | Day number of week (1 = Monday, ..., 7 = Sunday) | 1 |
-| a | Am/pm marker | PM |
-| H | Hour in day (0-23) | 0 |
-| k | Hour in day (1-24) | 24 |
-| K | Hour in am/pm (0-11) |  0 |
-| h | Hour in am/pm (1-12) | 12 |
-| m | Minute in hour | 30 |
-| s | Second in minute | 55 |
-| S | Millisecond | 978 |
-| z | General time zone | Pacific Standard Time; PST; GMT-08:00 |
-| Z | RFC 822 time zone | -0800 |
-| X | ISO 8601 time zone | -08; -0800; -08:00 |
-| _epoch_ | Seconds since 1970/01/01 00:00:00 | 1557136493, 1557136493.136|
-| _epochmilli_ | Milliseconds since 1970/01/01 00:00:00.0000| 15571364938124, 15571364938124.001 |
-| _epochmicro_ | Milliseconds since 1970/01/01 00:00:00.0000| 15571364938124789, 15571364938124789.999 |
-| _epochnano_ | Milliseconds since 1970/01/01 00:00:00.0000| 15571364938124789101 |
-| i | Microsecond | 111, 321001 |
-| n | Nanosecond | 999, 542113879 |
+| placeholder | Description | Example | Note |
+| --- | --- | --- | --- |
+| G | Era designator | AD | |
+| y | Year | 1996; 96 | |
+| Y | Week year | 2009; 09 | |
+| M | Month in year (context sensitive) |  July; Jul; 07 | |
+| L | Month in year (standalone form) | July; Jul; 07 | |
+| w | Week in year | 27 | |
+| W | Week in month | 2 | |
+| D | Day in year | 189 | |
+| d | Day in month |  10 | |
+| F | Day of week in month | 2 | |
+| E | Day name in week | Tuesday; Tue | |
+| u | Day number of week (1 = Monday, ..., 7 = Sunday) | 1 | |
+| a | Am/pm marker | PM | |
+| H | Hour in day (0-23) | 0 | |
+| k | Hour in day (1-24) | 24 | |
+| K | Hour in am/pm (0-11) |  0 | |
+| h | Hour in am/pm (1-12) | 12 | |
+| m | Minute in hour | 30 | |
+| s | Second in minute | 55 | |
+| S | Millisecond | 978 | |
+| z | General time zone | Pacific Standard Time; PST; GMT-08:00 | |
+| Z | RFC 822 time zone | -0800 | |
+| X | ISO 8601 time zone | -08; -0800; -08:00 | |
+| _epoch_ | Seconds since 1970/01/01 00:00:00 | 1557136493, 1557136493.136| |
+| _epochmilli_ | Milliseconds since 1970/01/01 00:00:00.0000| 1557136493128, 1557136493128.001 | |
+| _epochmicro_ | Microseconds since 1970/01/01 00:00:00.0000| 1557136493128789, 1557136493128789.999 | |
+| _epochnano_ | Nanoseconds since 1970/01/01 00:00:00.0000| 1557136493128789101 | Seen the remark bellow regarding the loss of precision in _nanoseconds_ |
+| i | Microsecond | 111, 321001 | |
+| n | Nanosecond | 999, 542113879 | Seen the remark bellow regarding the loss of precision in _nanoseconds_ |
 
 
 **NB!** Spark uses US Locale and because on-the-fly conversion would be complicated, at the moment we stick to this 
--- a/standardization/src/test/scala/za/co/absa/enceladus/standardization/interpreter/StandardizationInterpreter_TimestampSuite.scala
+++ b/standardization/src/test/scala/za/co/absa/enceladus/standardization/interpreter/StandardizationInterpreter_TimestampSuite.scala
@@ -1,0 +1,347 @@
+package za.co.absa.enceladus.standardization.interpreter
+
+import java.sql.Timestamp
+
+import org.apache.spark.sql.types.{MetadataBuilder, StructField, StructType, TimestampType}
+import org.scalatest.FunSuite
+import za.co.absa.enceladus.utils.error.{ErrorMessage, UDFLibrary}
+import za.co.absa.enceladus.utils.testUtils.{LoggerTestBase, SparkTestBase}
+
+class StandardizationInterpreter_TimestampSuite extends FunSuite with SparkTestBase with LoggerTestBase {
+  import spark.implicits._
+
+  private implicit val udfLib: UDFLibrary = new UDFLibrary
+
+  private val fieldName = "tms"
+
+  test("epoch") {
+    val seq  = Seq(
+      0,
+      86400,
+      978307199,
+      1563288103
+    )
+    val desiredSchema = StructType(Seq(
+      StructField(fieldName, TimestampType, nullable = false,
+        new MetadataBuilder().putString("pattern", "epoch").build)
+    ))
+    val exp = Seq(
+      TimestampRow(Timestamp.valueOf("1970-01-01 00:00:00")),
+      TimestampRow(Timestamp.valueOf("1970-01-02 00:00:00")),
+      TimestampRow(Timestamp.valueOf("2000-12-31 23:59:59")),
+      TimestampRow(Timestamp.valueOf("2019-07-16 14:41:43"))
+    )
+
+    val src = seq.toDF(fieldName)
+
+    val std = StandardizationInterpreter.standardize(src, desiredSchema, "").cache()
+    logDataFrameContent(std)
+
+    assertResult(exp)(std.as[TimestampRow].collect().toList)
+  }
+
+  test("epochmilli") {
+    val seq  = Seq(
+      0.0,
+      86400000.5,
+      978307199999.05,
+      1563288103123.005
+    )
+    val desiredSchema = StructType(Seq(
+      StructField(fieldName, TimestampType, nullable = false,
+        new MetadataBuilder().putString("pattern", "epochmilli").build)
+    ))
+    val exp = Seq(
+      TimestampRow(Timestamp.valueOf("1970-01-01 00:00:00")),
+      TimestampRow(Timestamp.valueOf("1970-01-02 00:00:00.0005")),
+      TimestampRow(Timestamp.valueOf("2000-12-31 23:59:59.99905")),
+      TimestampRow(Timestamp.valueOf("2019-07-16 14:41:43.123005"))
+    )
+
+    val src = seq.toDF(fieldName)
+
+    val std = StandardizationInterpreter.standardize(src, desiredSchema, "").cache()
+    logDataFrameContent(std)
+
+    assertResult(exp)(std.as[TimestampRow].collect().toList)
+  }
+
+  test("epochmicro") {
+    val seq  = Seq(
+      0L,
+      86400000000L,
+      978307199999999L,
+      1563288103123456L
+    )
+    val desiredSchema = StructType(Seq(
+      StructField(fieldName, TimestampType, nullable = false,
+        new MetadataBuilder().putString("pattern", "epochmicro").build)
+    ))
+    val exp = Seq(
+      TimestampRow(Timestamp.valueOf("1970-01-01 00:00:00")),
+      TimestampRow(Timestamp.valueOf("1970-01-02 00:00:00")),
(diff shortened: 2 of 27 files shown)


benedeki:
I separated the logical chain of dependencies into separate commits, in hope it would help understanding this PR.

The basic logic of the solution is, that the the pattern is checked for the presence of milli-, micro- and/or nanoseconds. If not found, nothing special is needed.
If present their location is identified withing the _pattern_ (class `Section'), The parsing is then done in two steps:
1) The second fractions parts are removed both from the pattern and source and parsed - this is a timestamp with seconds precision
2) The fractions are extracted from the source and divided to be the appropriate decimal value of a second; then added to the timestamp.
Unfortunately there are complications, I failed to take into account and discovered during testing. See bug #677. Still this enhances the current possibilities of parsing, the errors are kind of corner-cases. Therefore I think this is worth to include, despite the known bug.

Btw, `epochmicro` and `epochnano` added as well. Furthermore, they don't have to be whole numbers only, can be decimal.

README.md
```diff
@@ -254,13 +254,21 @@ Summary:
 | z | General time zone | Pacific Standard Time; PST; GMT-08:00 |
 | Z | RFC 822 time zone | -0800 |
 | X | ISO 8601 time zone | -08; -0800; -08:00 |
-| _epoch_ | Seconds since 1970/01/01 00:00:00 | 1557136493|
-| _milliepoch_ | Milliseconds since 1970/01/01 00:00:00.0000| 15571364938124 |
+| _epoch_ | Seconds since 1970/01/01 00:00:00 | 1557136493, 1557136493.136|
+| _epochmilli_ | Milliseconds since 1970/01/01 00:00:00.0000| 15571364938124, 15571364938124.001 |
```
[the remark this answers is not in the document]
benedeki: Typo

README.md
```diff
@@ -254,13 +254,21 @@ Summary:
 | z | General time zone | Pacific Standard Time; PST; GMT-08:00 |
 | Z | RFC 822 time zone | -0800 |
 | X | ISO 8601 time zone | -08; -0800; -08:00 |
-| _epoch_ | Seconds since 1970/01/01 00:00:00 | 1557136493|
-| _milliepoch_ | Milliseconds since 1970/01/01 00:00:00.0000| 15571364938124 |
+| _epoch_ | Seconds since 1970/01/01 00:00:00 | 1557136493, 1557136493.136|
```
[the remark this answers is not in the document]
benedeki: Which would be a problem why? 😄

utils/src/main/scala/za/co/absa/enceladus/utils/general/Section.scala
```diff
@@ -0,0 +1,288 @@
...
+  * The class implements the `Ordered` trait
+  * Sections with negative start are BIGGER then the ones with positive start
+  * The section with smaller ABSOLUTE value of start is considered smaller, if starts equal the shorter section is the
+  * smaller
+  * NB! Interesting consequence of the ordering is, that if the sections are not overlapping and removal applied on
+  *  string from greatest to smallest one by one, the result is the same as removing all sections "at once"
+  * @param start the start position of the section, if negative the position is counted from the end
+  * @param length length of the section, cannot be negative
+  */
+case class Section(start: Int, length: Int) extends Ordered[Section] {
```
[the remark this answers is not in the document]
benedeki: Gladly.
A short intro before explanation in person:
The principle of this class is not that complicated I believe. Represents an (abstract) part of any string, defined by the _start_ index and _length_ of the part. Methods offer basic operations on a concrete `String`.
The somewhat non-trivial part is the ordering and the consequences of the chosen ordering.

utils/src/main/scala/za/co/absa/enceladus/utils/implicits/ColumnImplicits.scala
```diff
@@ -17,11 +17,57 @@ package za.co.absa.enceladus.utils.implicits
...
 import org.apache.spark.sql.functions._
+import za.co.absa.enceladus.utils.general.Section
 
 object ColumnImplicits {
   implicit class ColumnEnhancements(column: Column) {
     def isInfinite: Column = {
       column.isin(Double.PositiveInfinity, Double.NegativeInfinity)
     }
+
+    def zeroBasedSubstr(startPos: Int): Column = {
```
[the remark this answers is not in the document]
benedeki: Thanks, forgot to put them here.

README.md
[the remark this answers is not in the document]
Zejnilovic: Yep, I misread it. It's all good. 😊

utils/src/main/scala/za/co/absa/enceladus/utils/implicits/StringImplicits.scala
```diff
@@ -22,6 +22,21 @@ import scala.annotation.tailrec
 object StringImplicits {
   implicit class StringEnhancements(string: String) {
 
+    def injectWith(what: String, where: Int): String = {
```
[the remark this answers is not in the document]
benedeki: 👍 
I wonder if to keep it though. I created the methods, but turned out, I don't need it. Wondered if its useful enough to keep. Opinions?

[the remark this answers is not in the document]
GeorgiChochov: If you're not using it, I'd say remove it and consider putting it in `commons` instead.

standardization/src/test/scala/za/co/absa/enceladus/standardization/interpreter/StandardizationInterpreter_TimestampSuite.scala
```diff
@@ -0,0 +1,347 @@
...
+
+    val std = StandardizationInterpreter.standardize(src, desiredSchema, "").cache()
+    logDataFrameContent(std)
+
+    assertResult(exp)(std.as[TimestampRow].collect().toList)
+  }
+
+  test("epochmilli") {
+    val seq  = Seq(
+      0.0,
```
GeorgiChochov: How do these work with negative numbers? Or values that aren't numeric?

utils/src/main/scala/za/co/absa/enceladus/utils/general/Section.scala
```diff
@@ -0,0 +1,288 @@
...
+
+  /**
+    * Merges all touching (that includes overlaps too) sections and sort them
+    * @param sections the sections to merge
+    * @return         an ordered from greater to smaller sequence of distinct sections (their distance is at least 1 or undefined)
+    */
+  def mergeSections(sections: Seq[Section]): Seq[Section] = {
+    def fuse(into: Section, what: Section): Section = {
+      if (into.start + into.length >= what.start + what.length) {
+        //what is within into
```
GeorgiChochov: Rather than the comment, maybe extract the `if` clause as a function:
```
def containsSection(other: Section): Boolean = {
  this.start + this.length >= other.start + other.length
}
```
Also it's not clear to me that one section would be within another based on this comparison. You're just checking if one ends after the other. Suppose it also starts after the other one ends? They would not overlap and the second one would not be within the first.

utils/src/main/scala/za/co/absa/enceladus/utils/general/Section.scala
```diff
@@ -0,0 +1,288 @@
...
+  }
+
+  /**
+    * Converts the Section to actual indexes representing the section on the given string if used in the substring function
+    * for example, e.g. first item of the tuple inclusive, second exclusive
+    * The result respects the boundaries of the string, the indexes not to cause OutOfBound exception
+    * @param forString  the string which the section would be applied to
+    * @return           tuple representing the beginIndex and endIndex parameters for the substring function
+    */
+  def toSubstringParameters(forString: String): (Int, Int) = {
```
GeorgiChochov: Why is it called `forString`? I'd prefer a clearer name.

utils/src/test/scala/za/co/absa/enceladus/utils/general/SectionSuite.scala
```diff
@@ -0,0 +1,700 @@
...
+  }
+
+  test("removeMultiple: zero length") {
+    val sections = Seq(Section(1, 0), Section(-2, 0))
+    val result = Section.removeMultiple("abcdefghijklmnop", sections)
+    val expected = "abcdefghijklmnop"
+    assert(result == expected)
+  }
+
+  //mergeSections
```
GeorgiChochov: I cannot make sense of these tests.

utils/src/main/scala/za/co/absa/enceladus/utils/general/Section.scala
```diff
@@ -0,0 +1,288 @@
...
+    } else if (start.abs < that.start.abs) {
+      -1
+    } else {
+      1
+    }
+  }
+
+  /**
+    * Converts the Section to actual indexes representing the section on the given string if used in the substring function
+    * for example, e.g. first item of the tuple inclusive, second exclusive
```
GeorgiChochov: `for example` and `e.g.` mean the same thing, but I'm also confused by the example, shouldn't the example be more about where they can be used `e.g. as can be used in the substring function`, rather than explaining how `substring` works?

utils/src/main/scala/za/co/absa/enceladus/utils/general/Section.scala
```diff
@@ -0,0 +1,288 @@
...
+    * @param fromString the string to apply the section to
+    * @return           substring defined by this section
+    */
+  def extract(fromString: String): String = {
+    val (realStart, after) = toSubstringParameters(fromString)
+    fromString.substring(realStart, after)
+  }
+
+  /**
+    * Creates a string that is the remainder if the substring represented by this section is removed frm the provided string
```
GeorgiChochov:
```suggestion
    * Creates a string that is the remainder if the substring represented by this section is removed from the provided string
```

utils/src/main/scala/za/co/absa/enceladus/utils/general/Section.scala
```diff
@@ -0,0 +1,288 @@
...
+
+  /**
+    * Inverse function for `remove`, inserts the `what` string into the `into` string as defined by the `section`
+    * @param into the string to inject into
+    * @param what the string to inject
+    * @return     the newly created string
+    */
+  def inject(into: String, what: String): String = {
+
+    def invalidParameterExceptionMessage: String = {
```
GeorgiChochov:
```suggestion
    def throwParameterException(): String = {
```
Definitely needs parens as the exception it throws is a side effect and this is most certainly not a getter of an exception message

utils/src/main/scala/za/co/absa/enceladus/utils/general/Section.scala
```diff
@@ -0,0 +1,288 @@
...
+      throw new InvalidParameterException(
+        s"The length of the string to inject (${what.length}) doesn't match Section($start, $length) for string of length ${into.length}."
+      )
+    }
+
+    if (what.length > length) {
+      throw new InvalidParameterException(invalidParameterExceptionMessage)
```
GeorgiChochov: You are throwing an exception into an exception here lol. Some refactoring gone awry, I'm sure.

README.md
```diff
@@ -223,44 +223,52 @@ To enable processing of time entries from other systems **Standardization** offe
...
-| s | Second in minute | 55 |
-| S | Millisecond | 978 |
-| z | General time zone | Pacific Standard Time; PST; GMT-08:00 |
-| Z | RFC 822 time zone | -0800 |
-| X | ISO 8601 time zone | -08; -0800; -08:00 |
-| _epoch_ | Seconds since 1970/01/01 00:00:00 | 1557136493|
-| _milliepoch_ | Milliseconds since 1970/01/01 00:00:00.0000| 15571364938124 |
+| placeholder | Description | Example | Note |
+| --- | --- | --- | --- |
+| G | Era designator | AD | |
```
yruslan: I like that we have such a good description of pattern characters. I had to look up for this every time I needed to write a pattern. Now README is the place to look. :+1:

utils/src/main/scala/za/co/absa/enceladus/utils/general/Section.scala
```diff
@@ -0,0 +1,288 @@
...
+        case (`whatLengthDeficit`, `index`) =>
+          // at the beginning of the into string, maybe appropriately shorter if to be place "before" 0 index
+          what + into
+        case _ => fail()
+      }
+    }
+  }
+
+  /**
+    * Metrics defined on Section
```
Zejnilovic: Number of positions between two sections

utils/src/main/scala/za/co/absa/enceladus/utils/general/Section.scala
```diff
@@ -0,0 +1,288 @@
...
+    * For example Section(2,3) for string "Hello!" gives (2,5), for "abc" (2,3)
+    * @param forString  the string which the section would be applied to
+    * @return           tuple representing the beginIndex and endIndex parameters for the substring function
+    */
+  def toSubstringParameters(forString: String): (Int, Int) = {
+    val (realStart, after) = if (start >= 0) {
+        (start min forString.length, start + length)
+      } else {
+        val startIndex = forString.length + start
+        if (startIndex >= 0) { //
```
Zejnilovic:
```suggestion
        if (startIndex >= 0) {
```

utils/src/main/scala/za/co/absa/enceladus/utils/general/Section.scala
```diff
@@ -0,0 +1,288 @@
...
+          (startIndex, startIndex + length)
+        } else { // the distance from end is longer than the string itself
+          (0, length + startIndex max 0)
+        }
+      }
+    (realStart, after min forString.length)
```
Zejnilovic: change lines `72`, `75` from infix notation to prefix notation

utils/src/main/scala/za/co/absa/enceladus/utils/general/Section.scala
```diff
@@ -0,0 +1,288 @@
...
+  }
+
+  /**
+    * The substring represented by this Section within the provided string
+    * Complementary to `remove`
+    * @param fromString the string to apply the section to
+    * @return           substring defined by this section
+    */
+  def extract(fromString: String): String = {
```
Zejnilovic:
```suggestion
  def extractFrom(string: String): String = {
```

Do a similar change the rest of methods as consulted at the meeting

utils/src/main/scala/za/co/absa/enceladus/utils/general/Section.scala
```diff
@@ -0,0 +1,288 @@
...
+    }
+  }
+
+  /**
+    * Metrics defined on Section
+    * @param that the Section to compute the distance from/to
+    * @return     None - if one Section has a negative start and the other positive or zero
+    *             The end of the smaller section subtracted from the start of the greater one (see comparison), e.g. can be negative
+    */
+  def distance(that: Section): Option[Int] = {
```
Zejnilovic:
```suggestion
  def distance(secondSection: Section): Option[Int] = {
```

utils/src/main/scala/za/co/absa/enceladus/utils/general/Section.scala
```diff
@@ -0,0 +1,288 @@
...
+    def subtractLike(from: Section, what: Section): Int = {
```
Zejnilovic:
```suggestion
    def calculateDistance(first: Section, second: Section): Int = {
```

utils/src/main/scala/za/co/absa/enceladus/utils/general/Section.scala
```diff
@@ -0,0 +1,288 @@
...
+    * @return     true if the greater Section starts before the smaller one ends (see comparison) or right after it
+    *             false otherwise
+    */
+  def touches(that: Section): Boolean = {
+    distance(that).exists(_ <= 0)
+  }
+}
+
+object Section {
+  def fromIndexes(start: Int, end: Int): Section = {
```
Zejnilovic: please add comments to public methods

utils/src/main/scala/za/co/absa/enceladus/utils/general/Section.scala
```diff
@@ -0,0 +1,288 @@
...
+    Section(start, end - start + 1)
```
Zejnilovic: add a check that they are in the right order

utils/src/main/scala/za/co/absa/enceladus/utils/general/Section.scala
```diff
@@ -0,0 +1,288 @@
...
+  }
+
+  def ofSameChars(fromString: String, fromIndex: Int): Section = {
```
Zejnilovic:
```suggestion
  def ofSameChars(inputString: String, signedIndex: Int): Section = {
```

utils/src/main/scala/za/co/absa/enceladus/utils/general/Section.scala
```diff
@@ -0,0 +1,288 @@
...
+    val realFromIndex = if (fromIndex >= 0) {
```
Zejnilovic:
```suggestion
    val index = if (fromIndex >= 0) {
```

utils/src/main/scala/za/co/absa/enceladus/utils/general/Section.scala
```diff
@@ -0,0 +1,288 @@
...
+      }
+    }
+
+    def doMerge(input: Seq[Section]): Seq[Section] = {
+      if (input.isEmpty) {
+        input
+      } else {
+        val sorted = input.sorted
+        sorted.tail.foldLeft(List(sorted.head)) { (resultAcc, item) =>
+          if (item touches resultAcc.head) {
```
Zejnilovic: please change from infix notation to prefix notation
```suggestion
          if (item.touches(resultAcc.head)) {
```

utils/src/main/scala/za/co/absa/enceladus/utils/general/Section.scala
```diff
@@ -0,0 +1,288 @@
...
+      paring.collect{case (c, true) => c}.mkString
+    }
+  }
+
+  /**
+    * Merges all touching (that includes overlaps too) sections and sort them
+    * @param sections the sections to merge
+    * @return         an ordered from greater to smaller sequence of distinct sections (their distance is at least 1 or undefined)
+    */
+  def mergeSections(sections: Seq[Section]): Seq[Section] = {
```
Zejnilovic:
```suggestion
  def mergeTouchingSections(sections: Seq[Section]): Seq[Section] = {
```

utils/src/test/scala/za/co/absa/enceladus/utils/general/SectionSuite.scala
```diff
@@ -0,0 +1,700 @@
...
+    assert(result.isEmpty)
+  }
+
+  test("mergeSections: one item") {
+    val sections = Seq(Section(42, 7))
+    val result = Section.mergeSections(sections)
+    assert(result == sections)
+  }
+
+  test("mergeSections: touching") {
```
yruslan: Maybe illustrate an example of to with the explanation why for the given input we get a particular output?

For instance,
```
For a string:
1234567890ACDFEFGHIJKLMNOPQUSTUVWXYZ
 ^ ^                            ^^
 | | ^^                         |
 | | Section(5,1)               Section(-4,2)
 | Section(3,2)
 Section(1,1)

Output of the merge:
...
```
etc.

That is so we won't forget your explanation.

utils/src/main/scala/za/co/absa/enceladus/utils/time/DateTimePattern.scala
```diff
@@ -30,12 +31,19 @@ abstract sealed class DateTimePattern(pattern: String, isDefault: Boolean = fals
...
   val defaultTimeZone: Option[String]
   val isTimeZoned: Boolean
 
+  val millisecondsPosition: Option[Section]
+  val microsecondsPosition: Option[Section]
+  val nanosecondsPosition: Option[Section]
+
+  val secondFractionsSections: Seq[Section]
+  val patternWithoutSecondFractions: String
+  def containsSecondFractions: Boolean = secondFractionsSections.nonEmpty
```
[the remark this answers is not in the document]
benedeki: `lazy` having bigger overhead than this simple expression, which might very likely be even inligned by the compiler.

utils/src/main/scala/za/co/absa/enceladus/utils/general/Section.scala
```diff
@@ -0,0 +1,288 @@
...
+    * @param fromString the string to apply the section to
+    * @return concatenation of the string before and after the Section
+    */
+  def remove(fromString: String): String = {
+    val (realStart, after) = toSubstringParameters(fromString)
+    fromString.substring(0, realStart) + fromString.substring(after)
+  }
+
+  /**
+    * Inverse function for `remove`, inserts the `what` string into the `into` string as defined by the `section`
```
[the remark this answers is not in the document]
benedeki: Thanks, good idea 👍

utils/src/main/scala/za/co/absa/enceladus/utils/general/Section.scala
benedeki: I have to admit I really like the _inflix_ here. Makes it easy to read, no issues with precedence. You mind it much? 😉

utils/src/main/scala/za/co/absa/enceladus/utils/general/Section.scala
benedeki: Eh, sorry the name comes off as misleading IMO.
And sequence of Sections can be provided. Not just "touching" ones, distinct, overlapping, one included in other, same ...

utils/src/main/scala/za/co/absa/enceladus/utils/general/Section.scala
Zejnilovic: I don't _mind it_, I would just keep everything the "same".

utils/src/main/scala/za/co/absa/enceladus/utils/general/Section.scala
Zejnilovic: Merge sections seems like you are going to merge all of them and allow gaps in between them. Let's thing of something then.

benedeki:
Wanted to change the microsecond placeholder to _u_ (widely used shorthand for micro prefix). Unfortunately turns out, _u_ is already being used for day of the week. But it made me spot the bug #707

Issue #685, opened 2019-07-19
Schema file export does not work


GeorgiChochov:
## Describe the bug
The schema export button always says "no file found". The issue comes from the `SchemaRestDAO.js` ignoring `schema name` and `schema version` input params because the request URI has been overwritten with the `project name` and `project version` instead. 

Cause not clear and to be investigated further.

## To Reproduce
1. Deploy Menas built with `-PgenerateComponentPreload` in a running Tomcat
2. Import a schema file (to make sure there is one)
3. Try to export it
4. See file not found
5. Check request URI
6. See project name and version embedded in the URI string

## Expected behaviour
Export the schema requested.

## Additional context
This was observed in our UAT environment. I haven't been able to reproduce it locally yet, but the steps above should replicate it.

PR #686, opened 2019-07-19
#685 Fix broken schema file export

GeorgiChochov:
This fixes the bug observed, but does not address the underlying issue. Why did `${name}` and `${version}` get replaced with `project.name` and `project.version`. To be addressed in a follow-up PR for this task...

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -330,36 +330,6 @@
                 </executions>
             </plugin>
             <plugin>
-                <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-shade-plugin</artifactId>
-                <version>${maven.shade.version}</version>
-                <executions>
-                    <execution>
-                        <phase>package</phase>
-                        <goals>
-                            <goal>shade</goal>
-                        </goals>
-                    </execution>
-                </executions>
-                <configuration>
-                    <transformers>
-                        <transformer implementation="org.apache.maven.plugins.shade.resource.AppendingTransformer">
-                            <resource>application.conf</resource>
-                        </transformer>
-                    </transformers>
-                    <filters>
-                        <filter>
-                            <artifact>*:*</artifact>
-                            <excludes>
-                                <exclude>META-INF/*.SF</exclude>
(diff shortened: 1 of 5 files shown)


GeorgiChochov:
> Maven filtering? 🙄

My first thought as well, but we didn't implement that for the static resources, right? And couldn't find it at first glance.

lokm01:
I think there might be 2 plugins doing filtering (I think I recently stumbled upon one, which didn't seem to be configured in great detail).
I think the other applies it on properties only which what you'd expect.

GeorgiChochov:
Is this the culprit? Do you know why it's set to true?
https://github.com/AbsaOSS/enceladus/blob/develop/menas/pom.xml#L270

lokm01:
Could be... Probably a left over trial and error attempt from the very early days.

GeorgiChochov:
Fixed an additional related issue by removing the maven shade plugin from the parent `pom.xml` which bloated all modules (and build times to two times over) and caused war file deployability (fails to deploy on some environments when project version is `SNAPSHOT`,  deploys fine if not `SNAPSHOT`). All modules that needed shading already have their shade plugin configured.

Zejnilovic:
I love the speed optimization!! 

But I have noticed that after deploying to tomcat, it does not use collection "schema_v1" but "schema". Is there something to set or check in configuration? I have running previous build which uses the "v1" collections and this branch's build which uses plain "schema" collection.
image

GeorgiChochov:
> But I have noticed that after deploying to tomcat, it does not use collection "schema_v1" but "schema". Is there something to set or check in configuration? I have running previous build which uses the "v1" collections and this branch's build which uses plain "schema" collection.
> Did anyone notice anything similar?

I'm unable to reproduce this.

pom.xml
```diff
@@ -329,36 +329,6 @@
                     </execution>
                 </executions>
             </plugin>
-            <plugin>
```
yruslan: Good riddance!

Issue #687, opened 2019-07-22
Add index sort direction and uniqueness constraints to migration framework


GeorgiChochov:
## Background
* MongoDB compound indexes can have [a sort order](https://docs.mongodb.com/manual/core/index-compound/#index-ascending-and-descending).
Currently, the Migration framework allows to create compound indexes, but only with ascending order.

* MongoDB indexes can have [a uniqueness constraint](https://docs.mongodb.com/manual/core/index-unique/).
Currently, the Migration framework does not allow index uniqueness to be enforced.

## Feature
It would be very beneficial to expand the indexing functionality to include these features.

PR #689, opened 2019-07-22
#687 Expand index support in Migration Framework

GeorgiChochov:


migrations/src/main/scala/za/co/absa/enceladus/migrations/framework/migration/CollectionMigration.scala
```diff
@@ -114,12 +114,12 @@ trait CollectionMigration extends Migration {
     * @param collectionName A collection for setting up an index
     * @param fields         A list of fields that the index should contain
     */
-  def createIndex(collectionName: String, fields: Seq[IndexField]): Unit = {
+  def createIndex(collectionName: String, fields: Seq[IndexField], unique: Boolean = false): Unit = {
```
yruslan: Please, update the documentation for this change. Both
- for this function
- the one at the top of the file with an example usage.

migrations/src/test/scala/za/co/absa/enceladus/migrations/framework/fixture/UseCaseTestData.scala
```diff
@@ -89,34 +89,34 @@ class UseCaseTestData {
...
-        |				"pattern": "yyyy-MM-dd"
-        |			}
-        |		},
-        |		{
-        |			"name": "field4",
-        |			"type": "string",
-        |			"nullable": true,
-        |			"metadata": {}
-        |		}
+        | "type": "struct",
```
yruslan: Just a question. I'm wondering why the spacing is changed?

GeorgiChochov: It was tabbed, I changed it to spaces, since we use spaces everywhere. If you prefer I can switch it back.

yruslan: I prefer spaces as well.

benedeki: Yeah, we decided for spaces, so let's be consistent.

PR #699, opened 2019-07-24
Feature/688 Add more CSV options

yruslan:


standardization/src/test/scala/za/co/absa/enceladus/standardization/fixtures/CsvSpecialCharsFixture.scala
```diff
@@ -0,0 +1,75 @@
...
+
+import org.apache.spark.sql.types._
+import org.scalatest.{BeforeAndAfterAll, Suite}
+import za.co.absa.enceladus.model.Dataset
+
+trait CsvSpecialCharsFixture extends BeforeAndAfterAll {
+
+  this: Suite =>
+
+  val csvContent: String =
```
GeorgiChochov: I would put these (`csvContent`, `schema`, `inputFileName`, `dataSet`) with the test they belong to.
1. It's difficult to see what is being tested when I need to switch files.
2. It's not really the `Fixture`'s place to hold this data.

standardization/src/test/scala/za/co/absa/enceladus/standardization/interpreter/StandardizationCsvSuite.scala
```diff
@@ -0,0 +1,156 @@
...
+    val df = dfReader.load(inputFileName)
+
+    val actual = df.dataAsString(truncate = false)
+
+    assert(actual == expected)
+  }
+
+  test("Test standardizing a CSV file if a quote character is not specified") {
+    // This is a case where correct encoding and delimiter are specified.
+    // But one fields might contain an opening double quote character without a closing one.
```
GeorgiChochov:
```suggestion
    // But one field might contain an opening double quote character without a closing one.
```

standardization/src/test/scala/za/co/absa/enceladus/standardization/fixtures/CsvSpecialCharsFixture.scala
yruslan: Do you mean to put everything from this file to the suite itself?
The fixture can then be 'with temp file' accepting the files content and charset, right?

standardization/src/main/scala/za/co/absa/enceladus/standardization/CmdConfig.scala
```diff
@@ -134,6 +147,25 @@ object CmdConfig {
           success
         else
           failure("The --delimiter option is supported only for CSV raw data format"))
+
+    opt[String]("csv-quote").optional().action((value, config) =>
```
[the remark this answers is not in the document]
yruslan: Possibly. The value ends up as an argument to `.option(key, _)` which does not have a Char overload, just String. But it might be a good idea to have a Char anyway and convert it to String afterwards. Will check this possibility.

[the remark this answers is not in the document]
yruslan: It worked perfectly. Made `delimiter` to be `Char` as well.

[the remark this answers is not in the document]
yruslan: Sorry, it didn't work after all. Sometime it works, sometimes not. Reverting to String.

I think it does not work when UTF-8 encoding of a char is larger than 1 byte. 
```
Error: Option --delimiter failed when given '¡'. '[B@38ed139b' is not a char
```

standardization/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
```diff
@@ -128,45 +130,85 @@ object StandardizationJob {
...
+    }
+  }
+
+  private def getGenericOptions(cmd: CmdConfig): HashMap[String,Option[Any]] = {
+    HashMap("charset" -> cmd.charset)
+  }
+
+  private def getXmlOptions(cmd: CmdConfig): HashMap[String,Option[Any]] = {
+    if (cmd.rawFormat.equalsIgnoreCase("xml")) {
+      HashMap("rowtag" -> cmd.rowTag)
```
Zejnilovic: The default value for rowTag is `None` so isn't this if/else redundant when you have line logic on line 143- 151? + CmdConfig has these checks on its own. Same for others.

yruslan: Yeah, it is redundant. But I wanted for format-specific option getters to follow a certain pattern. So when new options are added the developer should not worry about this.

standardization/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
```diff
@@ -128,45 +130,86 @@ object StandardizationJob {
     spark
   }
 
-  private def readerFormatSpecific(dfReader: DataFrameReader, cmd: CmdConfig, dataset: Dataset): DataFrameReader = {
+  def getFormatSpecificReader(dfReader: DataFrameReader, cmd: CmdConfig, dataset: Dataset): DataFrameReader = {
```
Zejnilovic: 1. Please add docs for public methods
2. Why send in `dfReader` when we could create it as part of the method?

yruslan: Done (it does not show the code as `Outdated` because the documentation is added one line above)

Zejnilovic: What about number 2?

standardization/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
```diff
@@ -128,45 +130,94 @@ object StandardizationJob {
...
+    val options = getCobolOptions(cmd, dataset) ++
+      getGenericOptions(cmd) ++
+      getXmlOptions(cmd) ++
+      getCsvOptions(cmd) ++
+      getFixedWidthOptions(cmd)
+
+    // Applying all the options
+    options.foldLeft(dfReader) { (df, optionPair) =>
+      optionPair match {
+        case (key, Some(value)) =>
```
benedeki: What do you think about this code:
```
        case (key, Some(s: String)) => df.option(key, s)
        case (key, Some(b: Boolean)) => df.option(key, b)
        case (key, Some(l: Boolean)) => df.option(key, l)
        case (key, Some(d: Boolean)) => df.option(key, d)
        case _ => df
```
Also have the additional benefit, that it will not crash if any of the parameter values is not `String`, `Boolean`, `Long` nor `Double`. Which IMO the original code would.
I would still consider adding some type conversion Int->Long etc.

standardization/src/main/scala/za/co/absa/enceladus/standardization/CmdConfig.scala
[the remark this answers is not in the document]
benedeki: Just wondering where the problem is coming from. `Char` is a 16-bit unsigned integer, therefore it should have nor problem with UTF-8 🤔

standardization/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
```diff
@@ -128,45 +130,94 @@ object StandardizationJob {
...
-      } else {
-        dfReaderWithOptions
+    val options = getCobolOptions(cmd, dataset) ++
+      getGenericOptions(cmd) ++
+      getXmlOptions(cmd) ++
+      getCsvOptions(cmd) ++
+      getFixedWidthOptions(cmd)
+
+    // Applying all the options
+    options.foldLeft(dfReader) { (df, optionPair) =>
```
benedeki: Oh, so much better then the previous options application method. 😍

standardization/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
yruslan: The old code won't crash, true. But it would fail fail on a compilation error if a value of an unsupported type is passed into one of .option() overloads. We are trading compile time error versus runtime error. Still I want this to crash instead of silently ignoring an option of an unrecognized type.
Will add Int conversion though.

standardization/src/main/scala/za/co/absa/enceladus/standardization/CmdConfig.scala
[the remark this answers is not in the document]
yruslan: Yeah, I think it is a `scopt` specific issue.

standardization/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
yruslan: Thanks

standardization/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
yruslan: 👍 Reimplemented it using the strict typing approach you have suggested.

standardization/src/main/scala/za/co/absa/enceladus/standardization/CmdConfig.scala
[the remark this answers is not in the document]
Zejnilovic: We could log it in `scopt` and maybe do something about it on a later date. How about that?

standardization/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
benedeki: Thank you 🙇

README.md
```diff
@@ -201,8 +201,12 @@ The list of additional options available for running Standardization:
 |            Option          |                           Description |
 | -------------------------- |:----------------------------------------------------------------------------- |
 | --raw-format **format**    | A format for input data. Can be one of `parquet`, `json`, `csv`, `xml`, `cobol`, `fixed-width`. |
+| --charset **charset**      | Specifies a charset to use for `csv`, `json` on `xml`. Default is `UTF-8`.    |
```
GeorgiChochov:
```suggestion
| --charset **charset**      | Specifies a charset to use for `csv`, `json` or `xml`. Default is `UTF-8`.    |
```

standardization/src/main/scala/za/co/absa/enceladus/standardization/CmdConfig.scala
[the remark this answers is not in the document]
yruslan: https://github.com/scopt/scopt/issues/258

PR #702, opened 2019-07-25
#458: Early Timezone normalization

benedeki:
* TimeZoneNormalizer got rid of `lazy` property
* Calls to TimeZoneNormalizer, especially TimeZoneNormalizer.normalizeJVMTimeZone, moved to early stages of execution

### DIFF
--- a/examples/src/main/scala/za/co/absa/enceladus/examples/CustomRuleSample2.scala
+++ b/examples/src/main/scala/za/co/absa/enceladus/examples/CustomRuleSample2.scala
@@ -33,7 +33,7 @@ object CustomRuleSample2 {
     .appName("CustomRuleSample2")
     .config("spark.sql.codegen.wholeStage", value = false)
     .getOrCreate()
-  TimeZoneNormalizer.normalizeAll(Seq(spark))
+  TimeZoneNormalizer.normalizeAll(spark)) //normalize the timezone of JVM and the spark session
 
   def main(args: Array[String]) {
     // scalastyle:off magic.number
--- a/examples/src/main/scala/za/co/absa/enceladus/examples/CustomRuleSample2.scala
+++ b/examples/src/main/scala/za/co/absa/enceladus/examples/CustomRuleSample2.scala
@@ -33,7 +33,7 @@ object CustomRuleSample2 {
     .appName("CustomRuleSample2")
     .config("spark.sql.codegen.wholeStage", value = false)
     .getOrCreate()
-  TimeZoneNormalizer.normalizeAll(spark)) //normalize the timezone of JVM and the spark session
+  TimeZoneNormalizer.normalizeAll(spark) //normalize the timezone of JVM and the spark session
 
   def main(args: Array[String]) {
     // scalastyle:off magic.number
--- a/examples/src/main/scala/za/co/absa/enceladus/examples/CustomRuleSample2.scala
+++ b/examples/src/main/scala/za/co/absa/enceladus/examples/CustomRuleSample2.scala
@@ -33,10 +33,10 @@ object CustomRuleSample2 {
     .appName("CustomRuleSample2")
     .config("spark.sql.codegen.wholeStage", value = false)
     .getOrCreate()
+  TimeZoneNormalizer.normalizeAll(Seq(spark))
 
   def main(args: Array[String]) {
     // scalastyle:off magic.number
-    TimeZoneNormalizer.normalizeAll(Seq(spark))
     implicit val progArgs: CmdConfig = CmdConfig() // here we may need to specify some parameters (for certain rules)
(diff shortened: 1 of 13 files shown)


examples/src/main/scala/za/co/absa/enceladus/examples/CustomRuleSample4.scala
```diff
@@ -27,6 +26,8 @@ import za.co.absa.enceladus.model.Dataset
 import za.co.absa.enceladus.utils.time.TimeZoneNormalizer
 
 object CustomRuleSample4 {
+  TimeZoneNormalizer.normalizeJVMTimeZone()
```
GeorgiChochov: Why is this one different from the other Samples? I'd suggest making them more like this rather than using `normalizeAll`

benedeki: I see the _Samples_ as increasing in complexity. So early on just one call, later how it's done in rest of the project.
Also shows two possible ways how to call set the whole system to normalized time zone.

Zejnilovic: Overloading the `normalizeAll` method would be probably nice so you don't have to `Seq(X)` or call a different method.

```scala
def normalizeAll(sparks: Seq[SparkSession]): Unit = ???
def normalizeAll(sparks: SparkSession): Unit = ???
def normalizeAll(): Unit = ???
```

benedeki: Yeah, I was considering the second one. Third one is a synonym to `normalizeJVMTimeZone`

Zejnilovic: Yeah, having aliases. I still like that.

GeorgiChochov: > I see the _Samples_ as increasing in complexity. So early on just one call, later how it's done in rest of the project.
> Also shows two possible ways how to call set the whole system to normalized time zone.

Why even have `normalizeAll` if that is never really used? If using it can lead to unexpected problems that we avoid by eagerly doing `normalizeJVMTimeZone`, why give the option and provide misleading examples that we would not use ourselves?

benedeki: > 
> 
> Why even have `normalizeAll` if that is never really used? If using it can lead to unexpected problems that we avoid by eagerly doing `normalizeJVMTimeZone`, why give the option and provide misleading examples that we would not use ourselves?

It's actually used in _Menas_ as well.

examples/src/main/scala/za/co/absa/enceladus/examples/CustomRuleSample2.scala
```diff
@@ -33,10 +33,10 @@ object CustomRuleSample2 {
     .appName("CustomRuleSample2")
     .config("spark.sql.codegen.wholeStage", value = false)
     .getOrCreate()
+  TimeZoneNormalizer.normalizeAll(spark)) //normalize the timezone of JVM and the spark session
```
GeorgiChochov: Build is failing. Syntax error here.

benedeki: Thanks

examples/src/main/scala/za/co/absa/enceladus/examples/CustomRuleSample3.scala
```diff
@@ -29,11 +29,9 @@ object CustomRuleSample3 {
     .appName("CustomRuleSample3")
     .config("spark.sql.codegen.wholeStage", value = false)
     .getOrCreate()
-
-  spark.sparkContext.setLogLevel("WARN")
+  TimeZoneNormalizer.normalizeAll(Seq(spark)) //normalize the timezone of JVM and ALL the spark sessions (one in this case)
```
Zejnilovic: I think the `Seq` is obsolete now.
```suggestion
  TimeZoneNormalizer.normalizeAll(spark) //normalize the timezone of JVM and ALL the spark sessions (one in this case)
```

benedeki: It's a sample code. Shows options. I hoped the comment was explaining it.

Zejnilovic: I am not sure about that, it seems like it shows something that should not be used. But I don't have a strong preference for keeping it like this or not, so I am leaving it to you.

Issue #704, opened 2019-07-26
Lineage is often not shown after running Conformance


yruslan:
## Describe the bug
- After running Sandardization the lineage tab shows a lineage.
- After running Conformance the lineage tab displays 'No lineage found'
This not always happens but only on particular datasets.

## To Reproduce
Steps to reproduce:
1. Run a Standardization Job. Make sure the lineage is shown.
2. Run a Conformance Job.
3. Got to the Lineage tab of the run.
4. The 'Lineage not found' message will be displayed.

## Expected behaviour
A lineage should always be shown if any processing have been done.

## Additional context
This is a confirmed Spline issue: https://github.com/AbsaOSS/spline/issues/296

PR #705, opened 2019-07-26
#704 Update Spline version to 0.3.9 to fix lineage tracking failure

yruslan:
Confirmed on my local machine - the lineage is successfully saved to DB


benedeki:
Talked to Alex, according to him, the 0.3.8 `war` version should be OK to use. (Important for @hamc17 that he doesn't need to upgrade the dependency on all machines.)

PR #716, opened 2019-07-29
Bugfix/421 handle std column name clashes

yruslan:


utils/src/main/scala/za/co/absa/enceladus/utils/implicits/DataFrameImplicits.scala
```diff
@@ -55,6 +57,17 @@ object DataFrameImplicits {
       gatherData(showFnc)
     }
 
+    /**
+      * Adds a column to a dataframe if it does not exists
```
GeorgiChochov:
```suggestion
      * Adds a column to a dataframe if it does not exist
```

utils/src/main/scala/za/co/absa/enceladus/utils/schema/SparkUtils.scala
```diff
@@ -0,0 +1,44 @@
...
+
+
+/**
+  * General Spark utils
+  */
+object SparkUtils {
+  private val log: Logger = LogManager.getLogger(this.getClass)
+
+  /**
+    * Adds a column to a dataframe if it does not exists
```
GeorgiChochov:
```suggestion
    * Adds a column to a dataframe if it does not exist
```

utils/src/main/scala/za/co/absa/enceladus/utils/validation/SchemaValidator.scala
```diff
@@ -44,6 +46,36 @@ object SchemaValidator {
     errorsAccumulator.toList
   }
 
+  /**
+    * Validates the error column.
+    * Most of the time the error columns should not exist in the input schema. But uf it does exist, it should
```
GeorgiChochov:
```suggestion
    * Most of the time the error column should not exist in the input schema. But if it does exist, it should
```

utils/src/main/scala/za/co/absa/enceladus/utils/validation/SchemaValidator.scala
```diff
@@ -44,6 +46,36 @@ object SchemaValidator {
...
+    * conform to the expected type.
+    * Nullability of the error column is not an issue.
+    *
+    * @param schema A Spark schema
+    * @return A list of ValidationErrors objects, each containing a column name and the list of errors and warnings
```
GeorgiChochov:
```suggestion
    * @return A list of ValidationErrors, each containing a column name and the list of errors and warnings
```

utils/src/main/scala/za/co/absa/enceladus/utils/validation/SchemaValidator.scala
```diff
@@ -44,6 +46,36 @@ object SchemaValidator {
...
+    */
+  def validateErrorColumn(schema: StructType)
+                         (implicit spark: SparkSession)
+                         : List[FieldValidationFailure] = {
+    val expectedTypeNonNullable = ArrayType.apply(ErrorMessage.errorColSchema, containsNull = false)
```
GeorgiChochov:
```suggestion
    val expectedTypeNonNullable = ArrayType(ErrorMessage.errorColSchema, containsNull = false)
```

utils/src/main/scala/za/co/absa/enceladus/utils/validation/SchemaValidator.scala
```diff
@@ -44,6 +46,36 @@ object SchemaValidator {
...
+    val expectedTypeNullable = ArrayType.apply(ErrorMessage.errorColSchema, containsNull = true)
+    val errCol = schema.find(f => f.name == ErrorMessage.errorColumnName)
+    errCol match {
+      case Some(errColField) =>
+        if (errColField.dataType != expectedTypeNonNullable && errColField.dataType != expectedTypeNullable) {
+          val actualType = errColField.dataType
+          List(FieldValidationFailure(errColField.name, "",
+            ValidationError("The error column in the input data does not conform to the expected type. " +
+              s"Expected: $expectedTypeNonNullable, actual: $actualType") :: Nil))
```
Zejnilovic: Wouldn't we rather for them to see the nullable type as that is the one that we create.

standardization/src/test/scala/za/co/absa/enceladus/standardization/interpreter/StandardizationRerunSuite.scala
```diff
@@ -0,0 +1,165 @@
...
+        ||A1 |A2 |A3  |A4        |A5        |errCol|
+        |+---+---+----+----------+----------+------+
+        ||101|102|1   |2019-05-04|2019-05-04|[]    |
+        ||201|202|2   |2019-05-05|2019-05-05|[]    |
+        ||301|302|1   |2019-05-06|2019-05-06|[]    |
+        ||401|402|1   |2019-05-07|2019-05-07|[]    |
+        ||501|502|null|2019-05-08|2019-05-08|[]    |
+        |+---+---+----+----------+----------+------+
+        |
+        |""".stripMargin.replace("\r\n", "\n")
```
Zejnilovic: @benedeki please run these tests.
I am curious how it works if the `csvContent` file was created with system line endings while when read we replace CRLF for LF

utils/src/main/scala/za/co/absa/enceladus/utils/validation/SchemaValidator.scala
```diff
@@ -48,17 +48,17 @@ object SchemaValidator {
...
     * @param schema A Spark schema
-    * @return A list of ValidationErrors objects, each containing a column name and the list of errors and warnings
+    * @return A list of ValidationErrors, each containing a column name and the list of errors and warnings
     */
   def validateErrorColumn(schema: StructType)
                          (implicit spark: SparkSession)
                          : List[FieldValidationFailure] = {
-    val expectedTypeNonNullable = ArrayType.apply(ErrorMessage.errorColSchema, containsNull = false)
+    val expectedTypeNonNullable = ArrayType(ErrorMessage.errorColSchema, containsNull = false)
     val expectedTypeNullable = ArrayType.apply(ErrorMessage.errorColSchema, containsNull = true)
```
Zejnilovic: There is one more Apply, that could be removed

utils/src/main/scala/za/co/absa/enceladus/utils/validation/SchemaValidator.scala
yruslan: I'm not 100% sure, need to check. But I think we create a non-nullable error column field, but when it is saved to Parquet and then reopened for a second Standardization/Conformance, the field becomes nullable.

utils/src/main/scala/za/co/absa/enceladus/utils/schema/SparkUtils.scala
```diff
@@ -0,0 +1,44 @@
...
+    *
+    * @param df      A dataframe
+    * @param colName A column to add if it does not exist already
+    * @param col     An expression for the column to add
+    * @return true if the column is the only column in a struct
+    */
+  def withColumnIfDoesNotExist(df: DataFrame, colName: String, col: Column): DataFrame = {
+    if (df.schema.exists(field => field.name.equalsIgnoreCase(colName))) {
+      log.warn(s"Column '$colName' already exists. The content of the column won't be overwritten.")
```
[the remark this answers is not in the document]
yruslan: Good point! Will add a check and if the value currently in the column to add doesn't equal to the one already there, the program will append an error to `errCol` about it.

utils/src/main/scala/za/co/absa/enceladus/utils/schema/SparkUtils.scala
```diff
@@ -0,0 +1,85 @@
...
+      log.warn(s"Column '$colName' already exists. The content of the column will be overwritten.")
+      overwriteWithErrorColumn(df, colName, colExpr)
+    } else {
+      df.withColumn(colName, colExpr)
+    }
+  }
+
+  /**
+    * Overwrites a column with a value provided by an expression.
+    * If the value in the column does not match the one provided by the expression, ar error will be
```
GeorgiChochov:
```suggestion
    * If the value in the column does not match the one provided by the expression, an error will be
```

standardization/src/test/scala/za/co/absa/enceladus/standardization/interpreter/StandardizationRerunSuite.scala
benedeki: Looking into it...

utils/src/main/scala/za/co/absa/enceladus/utils/schema/SparkUtils.scala
```diff
@@ -0,0 +1,85 @@
...
+  */
+object SparkUtils {
+  private val log: Logger = LogManager.getLogger(this.getClass)
+
+  /**
+    * Adds a column to a dataframe if it does not exist
+    *
+    * @param df      A dataframe
+    * @param colName A column to add if it does not exist already
+    * @param colExpr     An expression for the column to add
```
Zejnilovic:
```suggestion
    * @param colExpr An expression for the column to add
```

Just nitpicking. Might be left alone if no further changes are going to be implemented.

Issue #719, opened 2019-07-30
Decrease the size of the build log - module Conformance


benedeki:
Decrease the size of the build log by eliminating warnings and other clutter generated from module _Conformance_.

PR #728, opened 2019-07-30
#718: Decrease the size of the build log - module Utils

benedeki:
* Adhere to Scalastyle
* Follow some best-practices:
** typed implicits
** types declared for public members
** limit implicit conversions and if mark them
* Parameterless function as type
* setup logging for tests - style, filtering
* remove direct output to console (println, Dataframe.show, etc)

PR #729, opened 2019-07-30
#719: Decrease the size of the build log - module Conformance

benedeki:
* Adhere to Scalastyle
* Follow some best-practices:
** typed implicits
** types declared for public members
* Parameterless function as type
* setup logging for tests - style, filtering
* remove direct output to console (println, Dataframe.show, etc)
* replaced deprecated class
* no call of Spark cache method in tests to decrease warnings of already cached data

### DIFF
--- a/conformance/src/main/scala/za/co/absa/enceladus/conformance/CmdConfig.scala
+++ b/conformance/src/main/scala/za/co/absa/enceladus/conformance/CmdConfig.scala
@@ -108,7 +108,7 @@ object CmdConfig {
 
     opt[String]("menas-auth-keytab").optional().action({ (file, config) =>
       keytabFile = Some(file)
-      if(!fsUtils.localExists(file) && fsUtils.hdfsExists(file)) {
+      if (!fsUtils.localExists(file) && fsUtils.hdfsExists(file)) {
         config.copy(menasCredentials = Some(Right(fsUtils.hdfsFileToLocalTempFile(file))))
       } else {
         config.copy(menasCredentials = Some(Right(file)))
@@ -116,7 +116,7 @@ object CmdConfig {
     }).text("Path to keytab file used for authenticating to menas").validate({ file =>
       if (credsFile.isDefined) {
         failure("Only one authentication method is allowed at a time")
-      } else if(fsUtils.exists(file)) {
+      } else if (fsUtils.exists(file)) {
         success
       } else {
         failure("Keytab file doesn't exist")
--- a/conformance/src/main/scala/za/co/absa/enceladus/conformance/CmdConfig.scala
+++ b/conformance/src/main/scala/za/co/absa/enceladus/conformance/CmdConfig.scala
@@ -98,12 +98,13 @@ object CmdConfig {
       credsFile = Some(path)
       config.copy(menasCredentials = Some(credential))
     }).text("Path to Menas credentials config file.").validate(path =>
-      // scalastyle:off if.brace
-      if (keytabFile.isDefined) failure("Only one authentication method is allow at a time")
-      else if (MenasCredentials.exists(MenasCredentials.replaceHome(path))) success
-      else failure("Credentials file not found."))
-      // scalastyle:on if.brace
-
+      if (keytabFile.isDefined) {
+        failure("Only one authentication method is allow at a time")
+      } else if (MenasCredentials.exists(MenasCredentials.replaceHome(path))) {
+        success
+      } else {
+        failure("Credentials file not found.")
+      })
 
     opt[String]("menas-auth-keytab").optional().action({ (file, config) =>
       keytabFile = Some(file)
@@ -113,11 +114,13 @@ object CmdConfig {
         config.copy(menasCredentials = Some(Right(file)))
       }
     }).text("Path to keytab file used for authenticating to menas").validate({ file =>
-      // scalastyle:off if.brace
-      if (credsFile.isDefined) failure("Only one authentication method is allowed at a time")
-      else if(fsUtils.exists(file)) success
-      else failure("Keytab file doesn't exist")
-      // scalastyle:on if.brace
+      if (credsFile.isDefined) {
+        failure("Only one authentication method is allowed at a time")
+      } else if(fsUtils.exists(file)) {
+        success
+      } else {
+        failure("Keytab file doesn't exist")
+      }
     })
 
     opt[String]("performance-file").optional().action((value, config) =>
--- a/conformance/src/main/scala/za/co/absa/enceladus/conformance/CmdConfig.scala
+++ b/conformance/src/main/scala/za/co/absa/enceladus/conformance/CmdConfig.scala
@@ -66,8 +66,11 @@ object CmdConfig {
     opt[Int]('d', "dataset-version").required().action((value, config) =>
       config.copy(datasetVersion = value)).text("Dataset version")
       .validate(value =>
-        if (value > 0) success
-        else failure("Option --dataset-version must be >0"))
+        if (value > 0) {
+          success
+        } else {
+          failure("Option --dataset-version must be >0")
+        })
 
     val reportDateMatcher: Regex = "^\\d{4}-\\d{2}-\\d{2}$".r
     opt[String]('R', "report-date").required().action((value, config) =>
@@ -82,8 +85,11 @@ object CmdConfig {
       config.copy(reportVersion = Some(value)))
       .text("Report version. If not provided, it is inferred based on the publish path (it's an EXPERIMENTAL feature)")
       .validate(value =>
-        if (value > 0) success
-        else failure("Option --report-version must be >0"))
+        if (value > 0) {
+          success
+        } else {
+          failure("Option --report-version must be >0")
+        })
 
     private var credsFile: Option[String] = None
     private var keytabFile: Option[String] = None
@@ -92,21 +98,26 @@ object CmdConfig {
       credsFile = Some(path)
       config.copy(menasCredentials = Some(credential))
     }).text("Path to Menas credentials config file.").validate(path =>
+      // scalastyle:off if.brace
       if (keytabFile.isDefined) failure("Only one authentication method is allow at a time")
       else if (MenasCredentials.exists(MenasCredentials.replaceHome(path))) success
       else failure("Credentials file not found."))
+      // scalastyle:on if.brace
+
 
     opt[String]("menas-auth-keytab").optional().action({ (file, config) =>
       keytabFile = Some(file)
       if(!fsUtils.localExists(file) && fsUtils.hdfsExists(file)) {
         config.copy(menasCredentials = Some(Right(fsUtils.hdfsFileToLocalTempFile(file))))
       } else {
-      config.copy(menasCredentials = Some(Right(file)))
+        config.copy(menasCredentials = Some(Right(file)))
       }
     }).text("Path to keytab file used for authenticating to menas").validate({ file =>
+      // scalastyle:off if.brace
       if (credsFile.isDefined) failure("Only one authentication method is allowed at a time")
       else if(fsUtils.exists(file)) success
       else failure("Keytab file doesn't exist")
+      // scalastyle:on if.brace
     })
 
     opt[String]("performance-file").optional().action((value, config) =>
--- a/conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/DynamicInterpreter.scala
+++ b/conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/DynamicInterpreter.scala
@@ -15,7 +15,7 @@
 
 package za.co.absa.enceladus.conformance.interpreter
 
-import org.apache.log4j.LogManager
+import org.slf4j.LoggerFactory
 import org.apache.spark.sql.execution.command.ExplainCommand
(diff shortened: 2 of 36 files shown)


conformance/src/main/scala/za/co/absa/enceladus/conformance/CmdConfig.scala
```diff
@@ -92,21 +98,29 @@ object CmdConfig {
...
-      config.copy(menasCredentials = Some(Right(file)))
+        config.copy(menasCredentials = Some(Right(file)))
       }
     }).text("Path to keytab file used for authenticating to menas").validate({ file =>
-      if (credsFile.isDefined) failure("Only one authentication method is allowed at a time")
-      else if(fsUtils.exists(file)) success
-      else failure("Keytab file doesn't exist")
+      if (credsFile.isDefined) {
+        failure("Only one authentication method is allowed at a time")
+      } else if(fsUtils.exists(file)) {
```
GeorgiChochov:
```suggestion
      } else if (fsUtils.exists(file)) {
```

conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/OptimizerTimeTracker.scala
```diff
@@ -33,7 +33,7 @@ class OptimizerTimeTracker(inputDf: DataFrame, isWorkaroundEnabled: Boolean)(imp
   private val idField1 = SchemaUtils.getUniqueName("tmpId", Option(inputDf.schema))
   private val idField2 = s"${idField1}_2"
   private val dfWithId = inputDf.withColumn(idField1, lit(1))
-  private val dfJustId = Seq(1).toDF(idField2).cache
+  private val dfJustId = Seq(1).toDF(idField2)
```
GeorgiChochov: Why remove `.cache`?

conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/DynamicInterpreter.scala
```diff
@@ -33,7 +33,7 @@ import za.co.absa.enceladus.utils.general.Algorithms
 import za.co.absa.enceladus.utils.schema.SchemaUtils
 
 object DynamicInterpreter {
-  private val log = LogManager.getLogger("enceladus.conformance.DynamicInterpreter")
+  private val log = LogManager.getLogger(this.getClass)
```
GeorgiChochov: Can we use `slf4j` interfaces rather than the underlying `log4j` implementation directly?

conformance/src/test/scala/za/co/absa/enceladus/conformance/interpreter/ChorusMockSuite.scala
```diff
@@ -30,7 +30,7 @@ case class MyMappingTableInner(description: String, name: String)
 case class MyData(id: Int, toJoin: Int)
 case class MyDataConfd(id: Int, toJoin: Int, confMapping: MyMappingTableInner)
 
-class ChorusMockSuite extends FunSuite with SparkTestBase {
+class ChorusMockSuite extends FunSuite with SparkTestBase with LoggerTestBase{
```
GeorgiChochov:
```suggestion
class ChorusMockSuite extends FunSuite with SparkTestBase with LoggerTestBase {
```

conformance/src/test/scala/za/co/absa/enceladus/conformance/interpreter/InterpreterSuite.scala
```diff
@@ -42,7 +42,7 @@ class InterpreterSuite extends FunSuite with SparkTestBase with BeforeAndAfterAl
   }
 
   def testEndToEndDynamicConformance(useExperimentalMappingRule: Boolean): Unit = {
-    // Enable Conformance Framweork
+    // Enable Conformance framework
```
GeorgiChochov:
```suggestion
    // Enable Conformance Framework
```
You've kept the capitalization elsewhere, let's keep it here as well.

conformance/src/test/scala/za/co/absa/enceladus/conformance/interpreter/rules/CastingRuleSuite.scala
```diff
@@ -51,16 +51,14 @@ class CastingRuleSuite extends FunSuite with SparkTestBase {
...
-      println(CastingRuleSamples.conformedOrdersJSON)
-      println("ACTUAL:")
-      println(conformedJSON)
-      println("DETAILS (Input):")
-      inputDf.printSchema()
-      inputDf.show
-      println("DETAILS (Conformed):")
-      conformed.printSchema()
-      conformed.show
+      logger.debug("EXPECTED:")
```
GeorgiChochov: These should probably be `.error`. I see you've already done that in [previous PRs].(https://github.com/AbsaOSS/enceladus/pull/728/files#diff-dc8efb232378bd7dcc7b766d4c264cb1R1020)

conformance/src/test/scala/za/co/absa/enceladus/conformance/interpreter/rules/NegationRuleSuite.scala
```diff
@@ -110,16 +110,14 @@ class NegationRuleSuite extends FunSuite with SparkTestBase {
...
-      println(expectedJSON)
-      println("ACTUAL:")
-      println(conformedJSON)
-      println("DETAILS (Input):")
-      inputDf.printSchema()
-      inputDf.show
-      println("DETAILS (Conformed):")
-      conformed.printSchema()
-      conformed.show
+      logger.debug("EXPECTED:")
```
GeorgiChochov: `.error` here as well

conformance/src/test/resources/log4j.properties
```diff
@@ -0,0 +1,65 @@
...
+# Set everything to be logged to the console
+log4j.rootCategory=WARN, general
+
+log4j.appender.general=org.apache.log4j.ConsoleAppender
+log4j.appender.general.target=System.err
+log4j.appender.general.layout=org.apache.log4j.PatternLayout
+log4j.appender.general.layout.ConversionPattern=[%p] %d{yy/MM/dd HH:mm:ss} %c{1}: %m%n
+
+# Suppress warnings logged within the code inside the tests
+log4j.logger.za.co.absa.enceladus.conformance.interpreter.OptimizerTimeTracker=ERROR
```
GeorgiChochov: I think you should remove the log suppression for `OptimizerTimeTracker` here, since it's in `conformance` [not in the `utils` module](https://github.com/AbsaOSS/enceladus/pull/728#pullrequestreview-268387229). But more importantly, these logger configurations shouldn't be added randomly regardless of whether what they are configuring is actually in the present module or not.

conformance/src/test/resources/log4j.properties
```diff
@@ -0,0 +1,65 @@
...
+log4j.appender.forschemachecker.target=System.err
+log4j.appender.forschemachecker.layout=org.apache.log4j.PatternLayout
+log4j.appender.forschemachecker.layout.ConversionPattern=[%p] %d{yy/MM/dd HH:mm:ss} %c{1}: %m%n
+log4j.appender.forschemachecker.filter.01=org.apache.log4j.varia.StringMatchFilter
+log4j.appender.forschemachecker.filter.01.StringToMatch=Validation error for column
+log4j.appender.forschemachecker.filter.01.AcceptOnMatch=false
+log4j.logger.za.co.absa.enceladus.standardization.interpreter.stages.SchemaChecker$=WARN, forschemachecker
+log4j.additivity.za.co.absa.enceladus.standardization.interpreter.stages.SchemaChecker$=false
+
+# Suppress validation error logs from KafkaParametersProcessor
```
GeorgiChochov: Is this actually logged in `conformance`?

conformance/src/test/scala/za/co/absa/enceladus/conformance/interpreter/NestedStructSuite.scala
```diff
@@ -27,48 +27,56 @@ import za.co.absa.enceladus.utils.testUtils.SparkTestBase
...
+      .setCatalystWorkaroundEnabled(true)
+      .setControlFrameworkEnabled(false)
+
     val conformed = DynamicInterpreter.interpret(
       nestedStructsCastingDS,
-      standardizedDf,
-      experimentalMappingRule = false,
-      isCatalystWorkaroundEnabled = true,
-      enableControlFramework = false
+      standardizedDf
```
GeorgiChochov: Might be better on a single line

conformance/src/test/scala/za/co/absa/enceladus/conformance/interpreter/NestedStructSuite.scala
```diff
@@ -27,48 +27,56 @@ import za.co.absa.enceladus.utils.testUtils.SparkTestBase
...
+      .setCatalystWorkaroundEnabled(true)
+      .setControlFrameworkEnabled(false)
+
     val conformed = DynamicInterpreter.interpret(
       nestedStructsNegationDS,
-      standardizedDf,
-      experimentalMappingRule = false,
-      isCatalystWorkaroundEnabled = true,
-      enableControlFramework = false
+      standardizedDf
```
GeorgiChochov: I've been going through the code bottom-up, I'll stop commenting the same thing now. Applies to all  of these. You get the picture :)

utils/src/main/scala/za/co/absa/enceladus/utils/testUtils/SparkTestBase.scala
```diff
@@ -36,6 +36,7 @@ trait SparkTestBase { self =>
     .master(sparkMaster)
     .appName(s"Enceladus test - ${self.getClass.getName}")
     .config("spark.ui.enabled", "false")
+    .config("spark.debug.maxToStringFields", 100)
```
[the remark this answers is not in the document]
benedeki: Another Warning I spotted after the clear up. The setting was in its text

conformance/src/test/scala/za/co/absa/enceladus/conformance/interpreter/NestedStructSuite.scala
benedeki: Yup 😉 , thanks for the spot.

conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/FeatureSwitches.scala
```diff
@@ -0,0 +1,55 @@
...
+
+package za.co.absa.enceladus.conformance.interpreter
+
+/**
+ * Class bundling together switched for features to be used during Conformance
+ * It's a sealed abstract case class to enforce the the provided constructor (apply). This way values can be
+ * changed only via the setters and in case of removal of any of them, compiler will complain whenever they were used.
+ * @param experimentalMappingRuleEnabled     If true the new explode-optimized conformance mapping rule interpreter will be used
+ * @param catalystWorkaroundEnabled If true the Catalyst optimizer workaround is enabled
+ * @param controlFrameworkEnabled     If true sets the checkpoints on the dataset upon conforming
```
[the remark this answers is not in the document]
benedeki: It's not. I originally added the names as they were. Then unified them and failed to notice the dis-alignment in the doc.

conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/FeatureSwitches.scala
```diff
@@ -0,0 +1,55 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.enceladus.conformance.interpreter
+
+/**
+ * Class bundling together switched for features to be used during Conformance
+ * It's a sealed abstract case class to enforce the the provided constructor (apply). This way values can be
+ * changed only via the setters and in case of removal of any of them, compiler will complain whenever they were used.
```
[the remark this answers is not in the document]
benedeki: The idea of the class to simplify adding similar switches without the need to rewrite the whole chain of functions using them. Or many of the tests, unless they use the new switch.
And by forcing the setting of the values via the setter only, it ensures that in future, when the particular switch is removed it will not cause miss-configuration. Imagine code:
`val fs = FeatureSwitches(false, true)` and now we don't need the `catalystWorkaroundEnabled` anymore so it's removed. The constructor approach would still pass the compiler but suddenly a wrong switch is set. The use of setters prevents this.
Will consider better doc. Suggestions are welcomed. 😃

utils/src/main/scala/za/co/absa/enceladus/utils/testUtils/SparkTestBase.scala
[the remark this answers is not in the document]
GeorgiChochov: Alright, nice

conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/FeatureSwitches.scala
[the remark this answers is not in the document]
GeorgiChochov: I see what you mean. If you declare them in the constructor as named params you get the same effect for free. But you can't enforce it so everyone will always remember to name the params, so it makes sense.

Issue #726, opened 2019-07-30
Decrease the size of the build log - module Datamodel


benedeki:
Decrease the size of the build log by eliminating warnings and other clutter generated from module _Datamodel_.

PR #739, opened 2019-07-30
#726: Decrease the size of the build log - module DataModel

benedeki:
* Adhere to Scalastyle
* Follow some best-practices:
** typed implicits
** types declared for public members
* setup logging for tests - style, filtering


dataModel/src/main/scala/za/co/absa/enceladus/model/Dataset.scala
```diff
@@ -27,26 +27,26 @@ case class Dataset(
...
+
   hdfsPath:        String,
   hdfsPublishPath: String,
 
   schemaName:    String,
   schemaVersion: Int,
 
   dateCreated: ZonedDateTime = ZonedDateTime.now(),
-  userCreated: String        = null,
+  userCreated: String        = null, // scalastyle:ignore null
```
Zejnilovic: Maybe here we will want to create a new issue for removing `null`s from the models and changing them for `""`. If my knowledge is correct then it should still work fine.

GeorgiChochov: `_` would almost certainly break things. 
`""` would also very likely break things.
It should eventually be refactored to a case class which holds a date and username. But that's the least of this model's problems.

dataModel/src/test/resources/log4j.properties
```diff
@@ -0,0 +1,65 @@
...
+# Set everything to be logged to the console
+log4j.rootCategory=WARN, general
+
+log4j.appender.general=org.apache.log4j.ConsoleAppender
+log4j.appender.general.target=System.err
+log4j.appender.general.layout=org.apache.log4j.PatternLayout
+log4j.appender.general.layout.ConversionPattern=[%p] %d{yy/MM/dd HH:mm:ss} %c{1}: %m%n
+
+# Suppress warnings logged within the code inside the tests
+log4j.logger.za.co.absa.enceladus.conformance.interpreter.OptimizerTimeTracker=ERROR
```
GeorgiChochov: Pretty sure most of these are irrelevant here.

Issue #727, opened 2019-07-30
Decrease the size of the build log - module Migrations


benedeki:
Decrease the size of the build log by eliminating warnings and other clutter generated from module _Migrations_.

PR #741, opened 2019-07-30
#727: Decrease the size of the build log - module Migrations

benedeki:
* Adhere to Scalastyle, white-spaces at the end of line


migrations/src/test/resources/log4j.properties
```diff
@@ -0,0 +1,65 @@
...
+# Set everything to be logged to the console
+log4j.rootCategory=WARN, general
+
+log4j.appender.general=org.apache.log4j.ConsoleAppender
+log4j.appender.general.target=System.err
+log4j.appender.general.layout=org.apache.log4j.PatternLayout
+log4j.appender.general.layout.ConversionPattern=[%p] %d{yy/MM/dd HH:mm:ss} %c{1}: %m%n
+
+# Suppress warnings logged within the code inside the tests
+log4j.logger.za.co.absa.enceladus.conformance.interpreter.OptimizerTimeTracker=ERROR
```
yruslan: The migrations module does not have these classes.

Why we are customizing the `log4j.properties` at all? It was good with the default settings.

GeorgiChochov: Same question

benedeki: Removed

Issue #761, opened 2019-08-05
Add support for a space character as a delimiter for CSV


yruslan:
## Describe the bug
This bug is reported by Mahlatse.
The helper scripts skip the argument to the '--delimiter' option if it is `' '` (e.g. space). So currently Standardization won't work if a space character is a delimiter for a CSV input file.

## To Reproduce
Run
```
run_standardization.sh \
--num-executors 2 \
--deploy-mode client \
--menas-auth-keytab my.keytab \
--dataset-name TestDataset \
--dataset-version 2 \
--report-date 2018-06-29 \
--report-version 1 \
--raw-format csv \
--delimiter ' ' \
--header true
```

## Expected behavior
Standardization should run normally for such kind of setup.

## Screenshots
```
Error: Unknown argument 'true'
```
## Desktop (please complete the following information):
Linux cluster

PR #762, opened 2019-08-05
#761 Add support for a space character delimiter for CSV.

yruslan:


scripts/bash/run_enceladus.sh
```diff
@@ -124,7 +124,11 @@ case $key in
     shift 2 # past argument and value
     ;;
     --delimiter)
-    DELIMITER="\\$2"
+    if [[ "$2" == " " ]]; then
+      DELIMITER="' '"
+    else
+      DELIMITER="\\$2"
```
DzMakatun: Not very common, but what if there are more than one (special) character in the delimiter, e.g. n spaces? Is it possible to use something like `DELIMITER=\'"${2}"\'`?

Issue #765, opened 2019-08-05
Spark jobs fail to re-login tom Menas if session expires


yruslan:
## Describe the bug
This bug is reported by @DzMakatun.

Currently, Enceladus Spark jobs use DAO for accessing Menas's REST API. The first step to access the API is authentication. If it is succeeded a session is established. But jobs can run longer than session expiration time. So DAO should re-authenticate if the session expires.

However, re-authentication happens only if 401 (Unauthorized) status code is received. Enceladus won't authenticate if 403 (Forbidden) is encountered.

## To Reproduce
Run a Standrdization or Conformance job that lasts more than half a hour.

## Expected behaviour
Standardization and Conformance jobs should re-authenticate if 403 status code is received.

## Additional context
```
19/08/05 12:48:25 ERROR dao.MenasRestDAO$: RESPONSE: HTTP/1.1 403  - {"timestamp":"2019-08-05T10:48:25.607+0000","status":403,"error":"Forbidden","message":"Forbidden","path":"/menas/api/runs/updateRunStatus/dea3e466-947d-4ed2-8b19-3f72cffcddff"}
19/08/05 12:48:25 ERROR EventListenerMenas: Unable to update status of a run object (dea3e466-947d-4ed2-8b19-3f72cffcddff) in the database
Exception in thread "main" org.apache.spark.SparkException: Job 1 cancelled because SparkContext was shut down
	at org.apache.spark.scheduler.DAGScheduler$$anonfun$cleanUpAfterSchedulerStop$1.apply(DAGScheduler.scala:820)
	at org.apache.spark.scheduler.DAGScheduler$$anonfun$cleanUpAfterSchedulerStop$1.apply(DAGScheduler.scala:818)
```

GeorgiChochov:
TIL: Don't rush things even under pressure. There is an additional underlying bug [here](https://github.com/AbsaOSS/enceladus/blob/develop/dao/src/main/scala/za/co/absa/enceladus/dao/EnceladusRestDAO.scala#L187) and [here](https://github.com/AbsaOSS/enceladus/blob/develop/dao/src/main/scala/za/co/absa/enceladus/dao/MenasRestDAO.scala#L136) that is unaddressed, which actually fails the retry even after successful reauthentication.
I'll open an additional PR, and we'll merge it after additional testing.

PR #771, opened 2019-08-05
#765 Fix request retry check in dao module

GeorgiChochov:


dao/src/main/scala/za/co/absa/enceladus/dao/EnceladusRestDAO.scala
```diff
@@ -182,11 +182,11 @@ object EnceladusRestDAO extends EnceladusDAO {
 
         if (isUnauthorized) {
           log.warn(s"Unauthorized GET request for Menas URL: $url")
+          if (retriesLeft == 0) {
```
Zejnilovic: would you mind putting `<= 0` in? Just o be safe if I don't know for what reason we decide in the future to manipulate the `restriesLeft`

Zejnilovic: Maybe add `else log.info("trying to log in")`

dao/src/main/scala/za/co/absa/enceladus/dao/EnceladusRestDAO.scala
```diff
@@ -182,11 +182,11 @@ object EnceladusRestDAO extends EnceladusDAO {
...
+            throw UnauthorizedException(s"Unable to reauthenticate after retries")
+          }
           log.warn(s"Expired session, reauthenticating")
           if (enceladusLogin()) {
```
Zejnilovic: if the login fails won't it go straight to else on line 193 which will throw an exception that will be caught by line 210 and there will be no retries? Meaning never more than 1 retrie for logging in.

dao/src/main/scala/za/co/absa/enceladus/dao/EnceladusRestDAO.scala
GeorgiChochov: 1. Sure.
2. Else is implied here and it does log it with a warning.

dao/src/main/scala/za/co/absa/enceladus/dao/EnceladusRestDAO.scala
GeorgiChochov: It's intended to only have 1 retry.

PR #776, opened 2019-08-06
Feature/165 enable standardization of csv with more than 20480 columns

DzMakatun:
If the number of columns in raw data exceeds the default SCVReader's limit of 20480 this increases the limit based on the schema. This was successfully tested on Asgard (dataset 25600cols): Standardization with 25600 columns takes approx. 11 minutes. The upper limit on the number of columns is not yet known and could be much higher. However, setting the conformance rules in web UI for such wide datasets is a bottleneck which should be addressed in case if further increase becomes necessary.

### DIFF
--- a/standardization/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
+++ b/standardization/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
@@ -47,6 +47,7 @@ object StandardizationJob {
 
   private val log: Logger = LogManager.getLogger(this.getClass)
   private val conf: Config = ConfigFactory.load()
+  private final val SparkCSVReaderMaxColumnsDefault: Int = 20480
 
   def main(args: Array[String]) {
     implicit val spark: SparkSession = obtainSparkSession()
@@ -133,15 +134,16 @@ object StandardizationJob {
     *
     * @param cmd      Command line parameters containing format-specific options
     * @param dataset  A dataset definition
+    * @param schemaOpt (Optional) Schema, enables reading CSV files with the number of columns larger than Spark default
     * @return The updated dataframe reader
     */
-  def getFormatSpecificReader(cmd: CmdConfig, dataset: Dataset, schema: Option[StructType] = None)(implicit spark: SparkSession): DataFrameReader = {
+  def getFormatSpecificReader(cmd: CmdConfig, dataset: Dataset, schemaOpt: Option[StructType] = None)(implicit spark: SparkSession): DataFrameReader = {
     val dfReader = spark.read.format(cmd.rawFormat)
     // applying format specific options
     val options = getCobolOptions(cmd, dataset) ++
       getGenericOptions(cmd) ++
       getXmlOptions(cmd) ++
-      getCsvOptions(cmd, schema) ++
+      getCsvOptions(cmd, schemaOpt) ++
       getFixedWidthOptions(cmd)
 
     // Applying all the options
@@ -172,9 +174,9 @@ object StandardizationJob {
     }
   }
 
-  private def getCsvOptions(cmd: CmdConfig, schema: Option[StructType] = None): HashMap[String,Option[RawFormatParameter]] = {
+  private def getCsvOptions(cmd: CmdConfig, schemaOpt: Option[StructType] = None): HashMap[String,Option[RawFormatParameter]] = {
     if (cmd.rawFormat.equalsIgnoreCase("csv")) {
-      val knownNumberOfFields = if (schema.isDefined) schema.get.fields.length else 0
+      val knownNumberOfFields = schemaOpt.map(_.fields.length).getOrElse(0)
       HashMap(
         "delimiter" -> cmd.csvDelimiter.map(StringParameter),
         "header" -> cmd.csvHeader.map(BooleanParameter),
@@ -182,7 +184,7 @@ object StandardizationJob {
         "escape" -> cmd.csvEscape.map(StringParameter),
         // increase the default limit on the number of columns if needed
         // default is set at org.apache.spark.sql.execution.datasources.csv.CSVOptions maxColumns
-        "maxColumns" -> Option(LongParameter(max(20480, knownNumberOfFields)))
+        "maxColumns" -> {if (knownNumberOfFields > SparkCSVReaderMaxColumnsDefault) Some(LongParameter(knownNumberOfFields)) else None}
       )
     } else {
       HashMap()
--- a/standardization/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
+++ b/standardization/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
@@ -135,7 +135,7 @@ object StandardizationJob {
     * @param dataset  A dataset definition
     * @return The updated dataframe reader
     */
-  def getFormatSpecificReader(cmd: CmdConfig, dataset: Dataset, schema: StructType)(implicit spark: SparkSession): DataFrameReader = {
+  def getFormatSpecificReader(cmd: CmdConfig, dataset: Dataset, schema: Option[StructType] = None)(implicit spark: SparkSession): DataFrameReader = {
     val dfReader = spark.read.format(cmd.rawFormat)
     // applying format specific options
     val options = getCobolOptions(cmd, dataset) ++
@@ -172,8 +172,9 @@ object StandardizationJob {
     }
   }
 
-  private def getCsvOptions(cmd: CmdConfig, schema: StructType): HashMap[String,Option[RawFormatParameter]] = {
+  private def getCsvOptions(cmd: CmdConfig, schema: Option[StructType] = None): HashMap[String,Option[RawFormatParameter]] = {
     if (cmd.rawFormat.equalsIgnoreCase("csv")) {
+      val knownNumberOfFields = if (schema.isDefined) schema.get.fields.length else 0
       HashMap(
         "delimiter" -> cmd.csvDelimiter.map(StringParameter),
         "header" -> cmd.csvHeader.map(BooleanParameter),
@@ -181,7 +182,7 @@ object StandardizationJob {
         "escape" -> cmd.csvEscape.map(StringParameter),
         // increase the default limit on the number of columns if needed
         // default is set at org.apache.spark.sql.execution.datasources.csv.CSVOptions maxColumns
-        "maxColumns" -> Option(LongParameter(max(20480, schema.fields.length)))
+        "maxColumns" -> Option(LongParameter(max(20480, knownNumberOfFields)))
       )
     } else {
       HashMap()
@@ -228,7 +229,7 @@ object StandardizationJob {
                                dataset: Dataset)
                               (implicit spark: SparkSession,
                                fsUtils: FileSystemVersionUtils): DataFrame = {
-    val dfReaderConfigured = getFormatSpecificReader(cmd, dataset, schema: StructType)
+    val dfReaderConfigured = getFormatSpecificReader(cmd, dataset, Some(schema))
     val dfWithSchema = (if (!cmd.rawFormat.equalsIgnoreCase("parquet")) {
       val inputSchema = PlainSchemaGenerator.generateInputSchema(schema).asInstanceOf[StructType]
       dfReaderConfigured.schema(inputSchema)
--- a/standardization/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
+++ b/standardization/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
@@ -134,16 +134,16 @@ object StandardizationJob {
     *
     * @param cmd      Command line parameters containing format-specific options
     * @param dataset  A dataset definition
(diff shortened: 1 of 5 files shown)


standardization/src/test/scala/za/co/absa/enceladus/standardization/interpreter/StandardizationCsvSuite.scala
```diff
@@ -65,7 +65,7 @@ class StandardizationCsvSuite extends fixture.FunSuite with SparkTestBase with T
   private def getTestDataFrame(tmpFileName: String, args: Array[String]): DataFrame = {
     val cmd: CmdConfig = CmdConfig.getCmdLineArguments(args)
     StandardizationJob
-      .getFormatSpecificReader(cmd, dataSet)
+      .getFormatSpecificReader(cmd, dataSet, Some(schema))
```
GeorgiChochov: Can you add a test case for when the schema passed in is `None`? I'm curious to see how it reacts.

standardization/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
```diff
@@ -171,13 +172,17 @@ object StandardizationJob {
     }
   }
 
-  private def getCsvOptions(cmd: CmdConfig): HashMap[String,Option[RawFormatParameter]] = {
+  private def getCsvOptions(cmd: CmdConfig, schema: Option[StructType] = None): HashMap[String,Option[RawFormatParameter]] = {
     if (cmd.rawFormat.equalsIgnoreCase("csv")) {
+      val knownNumberOfFields = if (schema.isDefined) schema.get.fields.length else 0
```
GeorgiChochov:
```suggestion
      val knownNumberOfFields = schema.map(_.fields.length).getOrElse(0)
```

standardization/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
```diff
@@ -171,13 +172,17 @@ object StandardizationJob {
     }
   }
 
-  private def getCsvOptions(cmd: CmdConfig): HashMap[String,Option[RawFormatParameter]] = {
+  private def getCsvOptions(cmd: CmdConfig, schema: Option[StructType] = None): HashMap[String,Option[RawFormatParameter]] = {
```
GeorgiChochov:
```suggestion
  private def getCsvOptions(cmd: CmdConfig, schemaOpt: Option[StructType] = None): HashMap[String,Option[RawFormatParameter]] = {
```
I think making it clear that this is an `Option` helps a bit.

standardization/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
```diff
@@ -134,13 +135,13 @@ object StandardizationJob {
     * @param dataset  A dataset definition
     * @return The updated dataframe reader
     */
-  def getFormatSpecificReader(cmd: CmdConfig, dataset: Dataset)(implicit spark: SparkSession): DataFrameReader = {
+  def getFormatSpecificReader(cmd: CmdConfig, dataset: Dataset, schema: Option[StructType] = None)(implicit spark: SparkSession): DataFrameReader = {
```
GeorgiChochov:
```suggestion
  def getFormatSpecificReader(cmd: CmdConfig, dataset: Dataset, schemaOpt: Option[StructType] = None)(implicit spark: SparkSession): DataFrameReader = {
```
To distinguish from the other non-optional parameters of the function.

Zejnilovic: please do add it to the documentation above

standardization/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
DzMakatun: ok

standardization/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
DzMakatun: thnx

standardization/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
DzMakatun: thnx

standardization/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
```diff
@@ -171,13 +172,17 @@ object StandardizationJob {
...
+      val knownNumberOfFields = if (schema.isDefined) schema.get.fields.length else 0
       HashMap(
         "delimiter" -> cmd.csvDelimiter.map(StringParameter),
         "header" -> cmd.csvHeader.map(BooleanParameter),
         "quote" -> cmd.csvQuote.map(StringParameter),
-        "escape" -> cmd.csvEscape.map(StringParameter)
+        "escape" -> cmd.csvEscape.map(StringParameter),
+        // increase the default limit on the number of columns if needed
+        // default is set at org.apache.spark.sql.execution.datasources.csv.CSVOptions maxColumns
+        "maxColumns" -> Option(LongParameter(max(20480, knownNumberOfFields)))
```
yruslan: Maybe set this option only if the limit is breached? We can put a lower limit in case it is changed in the future.
```scala
"maxColumns" -> if (knownNumberOfFields > 10000) Some(knownNumberOfFields) else None
```

Zejnilovic: Can we extract the number in case it changes over time or we realize something?

DzMakatun: Thank you. Indeed, it is better not to change the defaults unless needed. I would like to compare to 20480 for the same reason.

DzMakatun:
I was also thinking about the same, and prefer the current  solution
> Question about the architecture: How about we do not pass is the schema but just the number 
- We can only obtain the number of columns from the schema, which we already have and use at the point where we call the method (unless we plan some significant changes). Passing a number means one would need to get the number from the schema each time before the call. By doing the same thing inside the method we have one value less to think about and save a line of code in the external part.
>which would default not to none but to a "properties" number we set or get from spark which now would be 20480?

- When set to None the CSVReader uses Spark defaults. Unfortunately, the number 20480 is [hardcoded in Spark](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/csv/CSVOptions.scala) and cannot be easily accessed. 
- The meaning of the value is "Spark default limit for the number of columns", so I would not allow to adjust it as e.g. a job argument.
- Perhaps I can make 20480 a constant, but it only appears once in the code.

standardization/src/test/scala/za/co/absa/enceladus/standardization/interpreter/StandardizationCsvSuite.scala
DzMakatun: ok, sure

standardization/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
DzMakatun: thnx

standardization/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
DzMakatun: thnx, extracted as a constant. (see also discussion in the comment below)

standardization/src/test/scala/za/co/absa/enceladus/standardization/interpreter/StandardizationCsvSuite.scala
```diff
@@ -63,6 +63,14 @@ class StandardizationCsvSuite extends fixture.FunSuite with SparkTestBase with T
...
   /** Creates a dataframe from an input file name path and command line arguments to Standardization */
   private def getTestDataFrame(tmpFileName: String, args: Array[String]): DataFrame = {
+    val cmd: CmdConfig = CmdConfig.getCmdLineArguments(args)
+    StandardizationJob
+      .getFormatSpecificReader(cmd, dataSet, Some(schema))
+      .schema(schema)
+      .load(tmpFileName)
+  }
+
+  private def getTestDataFrameWOmaxColumsAdjustment(tmpFileName: String, args: Array[String]): DataFrame = {
```
GeorgiChochov: This ends up being a bit of duplicated code. How about we get rid of this function and add a param to the original one like so:
```
  private def getTestDataFrame(tmpFileName: String, args: Array[String], hasMaxColumns: Boolean = false): DataFrame = {
    val cmd = CmdConfig.getCmdLineArguments(args)
    val schemaOpt = if (hasMaxColumns) Some(schema) else None
    StandardizationJob
      .getFormatSpecificReader(cmd, dataSet, schemaOpt)
```

Zejnilovic:
> I was also thinking about the same, and prefer the current solution
> 
> * We can only obtain the number of columns from the schema, which we already have and use at the point where we call the method (unless we plan some significant changes). Passing a number means one would need to get the number from the schema each time before the call. By doing the same thing inside the method we have one value less to think about and save a line of code in the external part.
>
> * When set to None the CSVReader uses Spark defaults. Unfortunately, the number 20480 is [hardcoded in Spark](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/csv/CSVOptions.scala) and cannot be easily accessed.
> * The meaning of the value is "Spark default limit for the number of columns", so I would not allow to adjust it as e.g. a job argument.
> * Perhaps I can make 20480 a constant, but it only appears once in the code.

The method (`prepareDataFrame`) is called only once and should not be called more times. So there is no overhead of each time. Mainly CSVReader or getFormatSpecificReader using schema seems kinda hard to understand, why does format options need to know about the schema, but reader options getting a number of max columns is quite an easy principle. 
I think this might be the separation of concerns.

And while seeing the code now, I would set the default value to zero, not none. You save operations (`schemaOpt.map(_.fields.length).getOrElse(0)`) and you do end up with the number or the `0` anyway. Wrapping a value that doesn't have to receive null/None, ever, seems like overhead.

standardization/src/test/scala/za/co/absa/enceladus/standardization/interpreter/StandardizationCsvSuite.scala
DzMakatun: good spot, thank you

DzMakatun:
> > I was also thinking about the same, and prefer the current solution
> > 
> > * We can only obtain the number of columns from the schema, which we already have and use at the point where we call the method (unless we plan some significant changes). Passing a number means one would need to get the number from the schema each time before the call. By doing the same thing inside the method we have one value less to think about and save a line of code in the external part.
> > * When set to None the CSVReader uses Spark defaults. Unfortunately, the number 20480 is [hardcoded in Spark](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/csv/CSVOptions.scala) and cannot be easily accessed.
> > * The meaning of the value is "Spark default limit for the number of columns", so I would not allow to adjust it as e.g. a job argument.
> > * Perhaps I can make 20480 a constant, but it only appears once in the code.
> 
> The method (`prepareDataFrame`) is called only once and should not be called more times. So there is no overhead of each time. Mainly CSVReader or getFormatSpecificReader using schema seems kinda hard to understand, why does format options need to know about the schema, but reader options getting a number of max columns is quite an easy principle.
> I think this might be the separation of concerns.
> 
> And while seeing the code now, I would set the default value to zero, not none. You save operations (`schemaOpt.map(_.fields.length).getOrElse(0)`) and you do end up with the number or the `0` anyway. Wrapping a value that doesn't have to receive null/None, ever, seems like overhead.

Ok, agreed

standardization/src/test/scala/za/co/absa/enceladus/standardization/interpreter/StandardizationRerunSuite.scala
```diff
@@ -64,7 +64,7 @@ class StandardizationRerunSuite extends fixture.FunSuite with SparkTestBase with
 
     val cmd: CmdConfig = CmdConfig.getCmdLineArguments(args)
     StandardizationJob
-      .getFormatSpecificReader(cmd, dataSet)
+      .getFormatSpecificReader(cmd, dataSet, schemaStr.fields.length)
```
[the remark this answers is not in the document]
DzMakatun: > Whys is called `schemaStr`?
I don't know, it was before
 Can we rename t to `schema`?
Sure, renamed.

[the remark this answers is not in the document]
GeorgiChochov: @yruslan explained that it represents `a schema where all the fields are StringType` so let's give it a more precise name

PR #778, opened 2019-08-06
Feature/447 continuous migration

yruslan:
Continuous migration App that can help the old and the new Menas to run in parallel.

The migration app can be run periodically to transfer entities created in old Menas into the new Menas database. It supports two MongoDB databases to be running on different clusters.


migrations/src/main/scala/za/co/absa/enceladus/migrations/MongoContinuousMigratorApp.scala
```diff
@@ -0,0 +1,49 @@
...
+  * Syntax:
+  *   java -cp enceladus-migrations.jar za.co.absa.enceladus.migrations.MongoContinuousMigratorApp \
+  *     --src-mongodb-url <MongoDb URL>
+  *     --trg-mongodb-url <MongoDb URL>
+  *     --src-database <Database Name>
+  *     --trg-database <Database Name>
+  */
+object MongoContinuousMigratorApp {
+
+  def main(args: Array[String]) {
```
Zejnilovic: wouldn't main be usually defined as `def main(args: Array[String]): Unit = {`

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/EntityVersionMap.scala
```diff
@@ -0,0 +1,91 @@
...
+  * * When a conflict happens a new version of the imported entity is created and the mapping between the entity
+  *   in the old database and the new one is created.
+  * * When importing an entity that has a reference to another entity remap the 'foreign key' according to the map.
+  */
+abstract class EntityVersionMap {
+
+  /**
+    * Adds a 'name - version' mapping.
+    *
+    * @param collectionName The name of the collection that containg the entity
```
Zejnilovic:
```suggestion
    * @param collectionName The name of the collection that containing the entity
```

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/EntityVersionMap.scala
```diff
@@ -0,0 +1,91 @@
...
+          s"collection = '$collectionName', entity = '$entityName', oldVersion = '$oldVersion', " +
+            s"newVersion = '$newVersion'. Version stored = $storedVersion")
+        }
+      case None =>
+        addEntry(collectionName, entityName, oldVersion, newVersion)
+    }
+  }
+
+  /**
+    * An implementation if this abstract class should redefine this method. Clients of this class shoudl use add()
```
Zejnilovic:
```suggestion
    * An implementation if this abstract class should redefine this method. Clients of this class should use add()
```

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/EntityVersionMap.scala
```diff
@@ -0,0 +1,91 @@
...
+    * @param entityName     An entity name
+    * @param oldVersion     An version of the entity in the old version of the database
+    * @param newVersion     An version of the entity in the new version of the database
+    */
+  protected def addEntry(collectionName: String, entityName: String, oldVersion: Int, newVersion: Int): Unit
+
+  /**
+    * Gets a 'name - version' mapping.
+    *
+    * @param collectionName The name of the collection that containg the entity
```
Zejnilovic:
```suggestion
    * @param collectionName The name of the collection that containing the entity
```

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/EntityVersionMap.scala
```diff
@@ -0,0 +1,91 @@
...
+    * @param newVersion     An version of the entity in the new version of the database
+    */
+  @throws[IllegalStateException]
+  final def add(collectionName: String, entityName: String, oldVersion: Int, newVersion: Int): Unit = {
+    get(collectionName, entityName, oldVersion) match {
+      case Some(storedVersion) =>
+        if (storedVersion != newVersion) {
+          throw new IllegalStateException("Attempt to overwrite an entity-version mapping. " +
+          s"collection = '$collectionName', entity = '$entityName', oldVersion = '$oldVersion', " +
+            s"newVersion = '$newVersion'. Version stored = $storedVersion")
```
Zejnilovic: I am not sure I understand from this message which one is the "bad guy"

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/EntityVersionMapMongo.scala
```diff
@@ -0,0 +1,116 @@
...
+import org.mongodb.scala.bson.conversions.Bson
+import org.mongodb.scala.model.Filters._
+import org.mongodb.scala.{MongoCollection, MongoDatabase}
+import za.co.absa.enceladus.migrations.continuous.model.EntityVersionMapping
+import za.co.absa.enceladus.migrations.framework.dao.ScalaMongoImplicits
+
+import scala.reflect.ClassTag
+
+final class EntityVersionMapMongo(db: MongoDatabase) extends EntityVersionMap {
+  private val log: Logger = LogManager.getLogger(this.getClass)
```
Zejnilovic: This is actually never used

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/EntityVersionMapMongo.scala
```diff
@@ -0,0 +1,116 @@
...
+                                  newVersion: Int): Unit = {
+    getTypedCollection[EntityVersionMapping]
+      .insertOne(EntityVersionMapping(collectionName, entityName, oldVersion, newVersion))
+      .execute()
+  }
+
+  /**
+    * Gets a 'name - version' mapping.
+    *
+    * @param collectionName The name of the collection that containg the entity
```
Zejnilovic:
```suggestion
    * @param collectionName The name of the collection that containing the entity
```

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/EntityVersionMapMongo.scala
```diff
@@ -0,0 +1,116 @@
...
+  /**
+    * Returns a filter for the MongoDB query for finding entity version mapping.
+    */
+  private def getFilter(collectionName: String,
+                        entityName: String,
+                        oldVersion: Int): Bson = {
+    and(
+      and(
+        regex("collection", collectionName, "i"),
+        regex("entityName", entityName, "i")),
```
Zejnilovic: is there a reason to have two levels of `and`?

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/migrate01/EntityRepository.scala
```diff
@@ -0,0 +1,158 @@
...
+    * @param name A name of an entity.
+    * @return the latest version of an entity, or 0 if the collection is empty.
+    */
+  def getLatestVersion(name: String): Int = {
+    val documents = db.getCollection(collectionName)
+      .find(regex("name", name, "i"))
+      .sort(descending("version"))
+      .limit(1)
+      .execute()
+    documents.headOption.map(_.getInteger("version").toInt).getOrElse(0)
```
Zejnilovic: IntelliJ suggestion: `documents.headOption.fold(0)(_.getInteger("version").toInt)`

migrations/src/test/scala/za/co/absa/enceladus/migrations/framework/continuous/fixture/EntityVersionMapMock.scala
```diff
@@ -0,0 +1,63 @@
...
+
+import scala.collection.mutable
+import scala.collection.mutable.ListBuffer
+
+class EntityVersionMapMock extends EntityVersionMap {
+
+  /**
+    * Adds a 'name - version' mapping.
+    *
+    * @param collectionName The name of the collection that containg the entity
```
Zejnilovic:
```suggestion
    * @param collectionName The name of the collection that containing the entity
```

migrations/src/test/scala/za/co/absa/enceladus/migrations/framework/continuous/fixture/EntityVersionMapMock.scala
```diff
@@ -0,0 +1,63 @@
...
+    */
+  override def addEntry(collectionName: String, entityName: String, oldVersion: Int, newVersion: Int): Unit = {
+    db.put((collectionName, entityName, oldVersion), newVersion)
+    actionsExecuted += s"EntityVersionMapMock.put($collectionName,$entityName,$oldVersion,$newVersion)"
+  }
+
+  /**
+    * Gets a 'name - version' mapping.
+    *
+    * @param collectionName The name of the collection that containg the entity
```
Zejnilovic:
```suggestion
    * @param collectionName The name of the collection that containing the entity
```

migrations/src/main/scala/za/co/absa/enceladus/migrations/ContinuousMigrationCmdConfig.scala
```diff
@@ -0,0 +1,71 @@
...
+
+  private class CmdParser(programName: String) extends OptionParser[ContinuousMigrationCmdConfig](programName) {
+    head("\nStandardization", "")
+    var rawFormat: Option[String] = None
+
+    opt[String]("src-mongodb-url").required().action((value, config) =>
+      config.copy(mongoDbUrlSrc = value)).text("A source MongoDB URL")
+
+    opt[String]("trg-mongodb-url").required().action((value, config) =>
+      config.copy(mongoDbUrlTrg = value)).text("A target MongoDB URL")
```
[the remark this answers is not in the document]
yruslan: The Scala MongoDB driver will validate the URLs for me, maybe it is sufficient for the purpose.

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/migrate01/MigratorSchema.scala
```diff
@@ -0,0 +1,134 @@
...
+    * @param repo     An entity repository
+    */
+  def resolveConflict(schema: model1.Schema, objectId: String, repo: EntityRepository): Unit = {
+    var retriesLeft = EntityMigrator.NumberOfInsertRetries
+    var saved = false
+    while (retriesLeft > 0 && !saved) {
+      retriesLeft -= 1
+      val latestVersion = repo.getLatestVersion(schema.name)
+      val schemaToSave = schema.copy(version = latestVersion + 1)
+      val schema1Json = ObjectIdTools.putObjectIdIfNotPresent(Serializer1.serializeSchema(schemaToSave),
```
Zejnilovic: can these 3 `val`s be extracted above while?

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/migrate01/MigratorSchema.scala
```diff
@@ -0,0 +1,134 @@
...
+    * In case there is a conflict, resolve it by adding a new version of the schema
+    * and register the mapping between the old version and the new one in ahe entity version map.
+    *
+    * @param schema   A schema to save as an instance of Model 1 object
+    * @param objectId An Object Id if the schema
+    * @param repo     An entity repository
+    */
+  def resolveConflict(schema: model1.Schema, objectId: String, repo: EntityRepository): Unit = {
+    var retriesLeft = EntityMigrator.NumberOfInsertRetries
+    var saved = false
```
Zejnilovic: As always I don't like `var`s. If you have the will or time, we could eliminate them by having the `try` block in a different recursive method and using break instead of `saved`

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/migrate01/MigratorSchema.scala
```diff
@@ -0,0 +1,134 @@
...
+                     databaseOld: MongoDatabase,
+                     databaseNew: MongoDatabase) extends EntityMigrator {
+  private val log: Logger = LogManager.getLogger(this.getClass)
+
+  override protected val collectionBase: String = EntityMigrator.schemaCollection
+  override protected val collectionOld: String = MigrationUtils.getVersionedCollectionName(collectionBase, 0)
+  override protected val collectionNew: String = MigrationUtils.getVersionedCollectionName(collectionBase, 1)
+
+  protected val dbOld: MongoDatabase = databaseOld
+  protected val dbNew: MongoDatabase = databaseNew
```
Zejnilovic: these 4+ lines could be moved to the abstract class, couldn't they? Every MigratorEntity does this the same way.

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/migrate01/EntityRepository.scala
```diff
@@ -0,0 +1,158 @@
...
+    *
+    * @param name    A name of an entity.
+    * @param version A version of the entity.
+    * @return true if such a document exists, false otherwise.
+    */
+  def doesDocumentExist(name: String, version: Int): Boolean = {
+    val documents = db.getCollection(collectionName)
+      .find(
+        and(
+          regex("name", name, "i"),
```
Zejnilovic: Isn't this dangerous in case of having `Entity1` and `Entity123`? When you have a name as `Entity1` it would found even `Entity123`. Maybe surround it with `^` and `$`? 

This is not the only occurrence

migrations/src/main/scala/za/co/absa/enceladus/migrations/framework/ObjectIdTools.scala
```diff
@@ -79,4 +79,26 @@ object ObjectIdTools {
...
+    *
+    * the output of the method will be:
+    *
+    * `5b98eea5a43a28a6154a2453`
+    *
+    * @param objIdExpression An Object Id as a JSON key-value expression.
+    * @return A string value of the Object Id
+    */
+  def extractId(objIdExpression: String): String = {
+    val idRegexp = """.*\"\$oid\"\s*:\s*\"(.*)\".*""".r
```
Zejnilovic: I am thinking if `.*` is not too permissive, it's a hex and shouldn't be anything else.

migrations/src/main/scala/za/co/absa/enceladus/migrations/ContinuousMigrationCmdConfig.scala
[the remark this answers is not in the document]
Zejnilovic: True, closing

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/migrate01/MigratorSchema.scala
yruslan: Each time we can potentially have a different value returned from `repo.getLatestVersion(schema.name)`. And that changes values of the next vals.

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/EntityVersionMapMongo.scala
yruslan: Good spot! I have several instances of these.

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/migrate01/EntityRepository.scala
yruslan: Well spot! Discussed this with @GeorgiChochov.

* Not sure if `.regex()` behaves as you said. Need to check.
* But our entity names are case-sensitive (this is what I completely forgot). So will replace all `regexp` in migrations to `equal` anyway.

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/EntityVersionMapMongo.scala
Zejnilovic: This is actually not the only place if I remember correctly

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/migrate01/MigratorSchema.scala
yruslan: I'd prefer leaving it this way. The vars are confined in a method and the `while` loop approach seems easier to read for me.
If this still is a concert for you, please let me know, I'm going to rewrite it recursively.

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/migrate01/MigratorSchema.scala
Zejnilovic: So in case someone created a new version in db_v1 just now and it fails and we need to retry?

yruslan: Exactly

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/EntityVersionMap.scala
yruslan: This is going to throw if migration is trying to do something incorrectly. All entity version mappings should be immutable. Once a mapping is created it cannot be changed.

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/migrate01/MigratorSchema.scala
yruslan: Not sure if I can move these lines to the abstract class. The lines use objects passed to the final class constructor.

migrations/src/main/scala/za/co/absa/enceladus/migrations/ContinuousMigrationCmdConfig.scala
```diff
@@ -0,0 +1,71 @@
...
+
+
+object ContinuousMigrationCmdConfig {
+
+  def getCmdLineArguments(args: Array[String]): ContinuousMigrationCmdConfig = {
+    val parser = new CmdParser("java -cp MigrationsBundle.jar " +
+      "za.co.absa.enceladus.migrations.MongoMigratorApp " +
+      "--src-mongodb-url <MongoDb URL>" +
+      "--trg-mongodb-url <MongoDb URL>" +
+      "--src-database <Database Name> " +
```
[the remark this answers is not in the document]
yruslan: This is so the syntax is displayed like this when ran with `--help`:
```
Usage: java -cp MigrationsBundle.jar za.co.absa.enceladus.migrations.MongoContinuousMigratorApp --src-mongodb-url <MongoDb URL> --trg-mongodb-url <MongoDb URL> --src-database <Database Name> --trg-database <Database Name> [options]

  --src-mongodb-url <value>
                           A source MongoDB URL
  --trg-mongodb-url <value>
                           A target MongoDB URL
  --src-database <value>   A source database name
  --trg-database <value>   A target database name
  --help                   prints this usage text
```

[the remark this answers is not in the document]
yruslan: Renamed to `programSyntax`

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/EntityVersionMap.scala
yruslan: 👍 Will also add an index to enforce the condition on the database level.

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/migrate01/MigratorSchema.scala
Zejnilovic: that is the thing. Even the parameters `databaseOld` and `databaseNew` are always used. What I propose is:

## EntitytMigrator
```scala
abstract class EntityMigrator(evm: EntityVersionMap,
                              databaseOld: MongoDatabase,
                              databaseNew: MongoDatabase) {

  protected val collectionOld: String = MigrationUtils.getVersionedCollectionName(collectionBase, 0)
  protected val collectionNew: String = MigrationUtils.getVersionedCollectionName(collectionBase, 1)

  protected val dbOld: MongoDatabase = databaseOld
  protected val dbNew: MongoDatabase = databaseNew

  protected val migrationUserName = "c_migration"

//  /** A collection base name. E.g. 'schema' or 'dataset' */
  protected def collectionBase: String
```

## MigratorDataset
```scala
final class MigratorDataset(evm: EntityVersionMap,
                      databaseOld: MongoDatabase,
                      databaseNew: MongoDatabase) extends EntityMigrator(evm, databaseOld, databaseNew) {

  override protected val collectionBase: String = EntityMigrator.datasetCollection
```

Now I believe this should work properly.

yruslan: Yeah, makes sense. Thanks!

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/EntityVersionMap.scala
yruslan: Added the index

migrations/pom.xml
```diff
@@ -86,4 +86,36 @@
             </plugin>
         </plugins>
     </build>
+
+    <profiles>
+        <profile>
+            <id>bundleMigrationCmdTools</id>
```
yruslan: Added a profile that uses shade plugin for building migration command-line tools.

This is because `migrations` is both a dependency module and a set of [two] migration tools.

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/migrate01/EntityMigrator.scala
```diff
@@ -0,0 +1,73 @@
...
+  */
+abstract class EntityMigrator(databaseOld: MongoDatabase,
+                              databaseNew: MongoDatabase) {
+
+  protected val migrationUserName = "c_migration"
+
+  /** A collection base name. E.g. 'schema' or 'dataset' */
+  protected def collectionBase: String
+
+  /** A versioned collection name for the old version of the mode. E.g. 'schema_v1' or 'dataset_v1' */
```
DzMakatun: * for the NEW version? compare to line 47

DzMakatun: Probably the other way round )) 44 -old, schema_v0 and 47 - new, schema_v1

DzMakatun:
Did not went through the entire code yet, and, therefore, would like to ask two questions to make it a bit more explicit :
- Object updates: what happens when an object is updated in the old version? E.g. a new checkpoint is added to a run object?
- Concurrent updates: the same run starts in both versions simultaneously (both collections have the same latest runID just before the run).

yruslan:
@DzMakatun, the first point is completely valid. The current implementation does not handle mutable objects. So probably it should be running when there are no active jobs in the old Menas.
Need to think if we can handle this case better. Maybe it should skip moving runs that are not at their final state (e.g. `allSucceeded` or `Failed`).

The second point should be fine, if not for runs, but for other entities at least. If two entities are created at the same time and have the same version, once migrated the version of the entity created in the old database will be incremented.

migrations/pom.xml
GeorgiChochov: But it's still the same artifact, right? When we release it and it goes on Maven Central, wouldn't we want to have them as separate jars? One being the library, another being the CLI tool(s)?

yruslan: Yeah. Should we have the cli tools in a separate module?
Most of the time we need just a library. A normal migration we need only for testing or for manual migration. Continuous migration is needed, hopefully, only once. I was thinking only to compile CLI tools only when we need them. Let me know if you want to have a module for CLI tools.

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/EntityVersionMap.scala
```diff
@@ -0,0 +1,91 @@
...
+    * @param newVersion     An version of the entity in the new version of the database
+    */
+  protected def addEntry(collectionName: String, entityName: String, oldVersion: Int, newVersion: Int): Unit
+
+  /**
+    * Gets a 'name - version' mapping.
+    *
+    * @param collectionName The name of the collection that contains the entity
+    * @param entityName     An Entity name
+    * @param oldVersion     An version of the entity in the old version of the database
```
GeorgiChochov:
```suggestion
    * @param oldVersion     A version of the entity in the old version of the database
```

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/EntityVersionMap.scala
```diff
@@ -0,0 +1,91 @@
...
+    * @return An version of the entity in the new version of the database, None if the entity is not found
```
GeorgiChochov:
```suggestion
    * @return An option of the version of the entity in the new version of the database, None if the entity is not found
```

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/EntityVersionMap.scala
```diff
@@ -0,0 +1,91 @@
...
+  @throws[IllegalStateException]
+  def get(collectionName: String, entityName: String, oldVersion: Int): Option[Int]
+
+  /**
+    * Gets a safe 'name - version' mapping. If the mapping isn't found in the entity version map it is assumed
+    * the original version number can be used
+    *
+    * @param collectionName The name of the collection that contains the entity
+    * @param entityName     An Entity name
+    * @param oldVersion     An version of the entity in the old version of the database
```
GeorgiChochov:
```suggestion
    * @param oldVersion     A version of the entity in the old version of the database
```

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/EntityVersionMap.scala
```diff
@@ -0,0 +1,91 @@
...
+    * @return An version of the entity in the new version of the database, None if the entity is not found
```
GeorgiChochov:
```suggestion
    * @return An option of the version of the entity in the new version of the database, None if the entity is not found
```

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/migrate01/EntityMigrator.scala
```diff
@@ -0,0 +1,73 @@
...
+  val NumberOfInsertRetries = 3
+
+  val schemaCollection = "schema"
+  val mappingTableCollection = "mapping_table"
+  val datasetCollection = "dataset"
+  val runCollection = "run"
+}
+
+/**
+  * An base class for continuous migration providers.
```
GeorgiChochov:
```suggestion
  * A base class for continuous migration providers.
```

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/migrate01/EntityMigrator.scala
```diff
@@ -0,0 +1,73 @@
...
+  /** A versioned collection name for the old version of the model. E.g. 'schema' or 'dataset' */
+  protected lazy val collectionOld: String = MigrationUtils.getVersionedCollectionName(collectionBase, 0)
+
+  /** A versioned collection name for the new version of the model. E.g. 'schema_v1' or 'dataset_v1' */
+  protected lazy val collectionNew: String = MigrationUtils.getVersionedCollectionName(collectionBase, 1)
+
+  protected val dbOld: MongoDatabase = databaseOld
+  protected val dbNew: MongoDatabase = databaseNew
+
+  /** Migrate a specific entity. This should be overridden an implemented in concrete classes */
```
GeorgiChochov:
```suggestion
  /** Migrate a specific entity. This should be implemented in concrete subclasses */
```

migrations/src/test/scala/za/co/absa/enceladus/migrations/framework/continuous/fixture/EntityVersionMapMock.scala
```diff
@@ -0,0 +1,63 @@
...
+import scala.collection.mutable.ListBuffer
+
+class EntityVersionMapMock extends EntityVersionMap {
+
+  /**
+    * Adds a 'name - version' mapping.
+    *
+    * @param collectionName The name of the collection that contains the entity
+    * @param entityName     An Entity name
+    * @param oldVersion     An version of the entity in the old version of the database
```
GeorgiChochov:
```suggestion
    * @param oldVersion     A version of the entity in the old version of the database
```

migrations/src/test/scala/za/co/absa/enceladus/migrations/framework/continuous/fixture/EntityVersionMapMock.scala
```diff
@@ -0,0 +1,63 @@
...
+    * @param newVersion     An version of the entity in the new version of the database
```
GeorgiChochov:
```suggestion
    * @param newVersion     A version of the entity in the new version of the database
```

migrations/src/test/scala/za/co/absa/enceladus/migrations/framework/continuous/fixture/EntityVersionMapMock.scala
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  def getActionsExecuted: List[String] = actionsExecuted.toList
+
+  def resetExecutedActions(): Unit = {
+    actionsExecuted.clear()
+  }
+
+  type LookupType = (String, String, Int)
+  private val db = new mutable.HashMap[LookupType, Int]()
```
GeorgiChochov: Can we keep the fields at the top of the class?

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/EntityVersionMapMongo.scala
```diff
@@ -0,0 +1,121 @@
...
+    DEFAULT_CODEC_REGISTRY)
+
+  import ScalaMongoImplicits._
+
+  private val entityMapCollectionName = "entity_version_map"
+
+  ensureCollectionExists()
+
+  /**
+    * An implementation if this abstract class should redefine this method. Clients of this class should use add()
```
GeorgiChochov: This is not abstract class. I'm starting to wonder if copying the scaladocs to subclasses has any value.

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/EntityVersionMap.scala
```diff
@@ -0,0 +1,91 @@
...
+          s"collection = '$collectionName', entity = '$entityName', oldVersion = '$oldVersion', " +
+            s"newVersion = '$newVersion'. Version stored = $storedVersion")
+        }
+      case None =>
+        addEntry(collectionName, entityName, oldVersion, newVersion)
+    }
+  }
+
+  /**
+    * An implementation if this abstract class should redefine this method. Clients of this class should use add()
```
GeorgiChochov:
```suggestion
    * An implementation of this abstract class should redefine this method. Clients of this class should use add()
```

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/EntityVersionMapMongo.scala
```diff
@@ -0,0 +1,121 @@
...
+    * as it does additional checks.
+    *
+    * @param collectionName The name of the collection that contains the entity
+    * @param entityName     An entity name
+    * @param oldVersion     An version of the entity in the old version of the database
```
GeorgiChochov:
```suggestion
    * @param oldVersion     A version of the entity in the old version of the database
```

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/EntityVersionMapMongo.scala
```diff
@@ -0,0 +1,121 @@
...
+    * @param newVersion     An version of the entity in the new version of the database
```
GeorgiChochov:
```suggestion
    * @param newVersion     A version of the entity in the new version of the database
```

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/EntityVersionMapMongo.scala
```diff
@@ -0,0 +1,121 @@
...
+      .insertOne(EntityVersionMapping(collectionName, entityName, oldVersion, newVersion))
+      .execute()
+  }
+
+  /**
+    * Gets a 'name - version' mapping.
+    *
+    * @param collectionName The name of the collection that contains the entity
+    * @param entityName     An entity name
+    * @param oldVersion     An version of the entity in the old version of the database
```
GeorgiChochov:
```suggestion
    * @param oldVersion     A version of the entity in the old version of the database
```

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/EntityVersionMapMongo.scala
```diff
@@ -0,0 +1,121 @@
...
+    * @return An version of the entity in the new version of the database, None if the entity is not found
```
GeorgiChochov:
```suggestion
    * @return A version of the entity in the new version of the database, None if the entity is not found
```

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/EntityVersionMapMongo.scala
```diff
@@ -0,0 +1,121 @@
...
+    * Ensures that the entity version mapping collections exists
+    */
+  private def ensureCollectionExists(): Unit = {
+    val collections = db.listCollectionNames().execute()
+    if (!collections.contains(entityMapCollectionName)) {
+      db.createCollection(entityMapCollectionName).execute()
+      db.getCollection(entityMapCollectionName)
+        .createIndex(
+          Document(
+            Seq("collection" -> 1, "entityName" -> 2, "oldVersion" -> 3)
```
GeorgiChochov:
```suggestion
            Seq("collection" -> 1, "entityName" -> 1, "oldVersion" -> 1)
```
It's standard practice to specify `1 (ASC)` and `-1 (DESC)`, let's not use other values.

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/migrate01/EntityRepository.scala
```diff
@@ -0,0 +1,163 @@
...
+      .find(Filters.equal("_id", new ObjectId(id)))
+      .execute()
+    documents.nonEmpty
+  }
+
+  /**
+    * Returns true if a document with the particular name and version exists in the collection.
+    *
+    * @param name    A name of an entity.
+    * @param version A version of the entity.
```
GeorgiChochov:
```suggestion
    * @param version A version of an entity.
```
It's odd to see `the` there, especially after the previous line is analogous but has `an entity` in it. You have it like this in all following cases as well.

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/migrate01/MigratorSchema.scala
```diff
@@ -0,0 +1,129 @@
...
+import scala.util.control.NonFatal
+
+/**
+  * The class provides schemas continuous migration from model version 0 to model version 1.
+  *
+  * @param evm         An entity mapper for tracking the mapping between old versions of entities and new ones.
+  * @param databaseOld An instance of a MongoDB database connection containing old model documents.
+  * @param databaseNew An instance of a MongoDB database connection containing new model documents.
+  */
+final class MigratorSchema(evm: EntityVersionMap,
```
GeorgiChochov:
```suggestion
final class SchemaMigrator(evm: EntityVersionMap,
```
Seems like a more appropriate name, it follows the pattern of its superclass `EntityMigrator` where the `entity` is a ` schema`.

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/migrate01/MigratorSchema.scala
```diff
@@ -0,0 +1,129 @@
...
+import za.co.absa.enceladus.migrations.continuous.EntityVersionMap
+import za.co.absa.enceladus.migrations.framework.ObjectIdTools
+import za.co.absa.enceladus.migrations.migrations.model0.Serializer0
+import za.co.absa.enceladus.migrations.migrations.model1.Serializer1
+import za.co.absa.enceladus.migrations.migrations.{MigrationToV1, model1}
+
+import scala.util.control.NonFatal
+
+/**
+  * The class provides schemas continuous migration from model version 0 to model version 1.
```
GeorgiChochov:
```suggestion
  * The class provides continuous migration for schemas from model version 0 to model version 1.
```

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/migrate01/MigratorSchema.scala
```diff
@@ -0,0 +1,129 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.enceladus.migrations.continuous.migrate01
+
+import org.apache.log4j.{LogManager, Logger}
```
GeorgiChochov: Can we use `slf4j` instead? I suspect we are inheriting a somewhat outdated version of `log4j` from somewhere and explicitly using a transitive dependency.

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/migrate01/MigratorSchema.scala
```diff
@@ -0,0 +1,129 @@
...
+                           databaseNew: MongoDatabase) extends EntityMigrator(databaseOld, databaseNew) {
+  private val log: Logger = LogManager.getLogger(this.getClass)
+
+  override protected val collectionBase: String = EntityMigrator.schemaCollection
+
+  /**
+    * Migrates a single instance of schema.
+    *
+    * @param srcSchemaJson A JSON representation of a schema in the model 0 format.
+    * @param objectId      An Object Id if the schema.
```
GeorgiChochov:
```suggestion
    * @param objectId      An Object Id of the schema.
```

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/migrate01/MigratorSchema.scala
```diff
@@ -0,0 +1,129 @@
...
+    * @param repo          An entity repository.
+    */
+  def migrateEntity(srcSchemaJson: String, objectId: String, repo: EntityRepository): Unit = {
+    val schema1Opt = try {
```
GeorgiChochov: I read this as `schema10pt` and thought, what's `10 pt`?
Maybe we can change this to a `Try` and user 
```
Try {
 ...
} match {
  case Success(schema1) =>
  case Failure(e)       =>
}
```

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/migrate01/MigratorSchema.scala
```diff
@@ -0,0 +1,129 @@
...
+    } catch {
+      case NonFatal(e) =>
+        log.warn(s"Unable to append a document for 'schema': ${e.getMessage}. Retrying...")
+        resolveConflict(schema, objectId, repo)
+    }
+  }
+
+  /**
+    * In case there is a conflict, resolve it by adding a new version of the schema
+    * and register the mapping between the old version and the new one in ahe entity version map.
```
GeorgiChochov:
```suggestion
    * and register the mapping between the old version and the new one in the entity version map.
```

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/migrate01/MigratorSchema.scala
```diff
@@ -0,0 +1,129 @@
...
+        Option(objectId))
+
+      try {
+        repo.insertDocument(schema1Json)
+        evm.add(collectionBase, schema.name, schema.version, schemaToSave.version)
+        saved = true
+      } catch {
+        case NonFatal(e) =>
+          if (retriesLeft > 0) {
+            log.warn(s"Unable to append a document for '$collectionBase': ${e.getMessage}")
```
GeorgiChochov: Might be worth mentioning that a retry will follow

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/migrate01/MigratorDataset.scala
```diff
@@ -0,0 +1,152 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.enceladus.migrations.continuous.migrate01
+
+import java.time.ZonedDateTime
+
+import org.apache.log4j.{LogManager, Logger}
```
GeorgiChochov: Let's use `slf4j` all around.

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/migrate01/MigratorDataset.scala
```diff
@@ -0,0 +1,152 @@
...
+import scala.util.control.NonFatal
+
+/**
+  * The class provides datasets continuous migration from model version 0 to model version 1.
+  *
+  * @param evm         An entity mapper for tracking the mapping between old versions of entities and new ones.
+  * @param databaseOld An instance of a MongoDB database connection containing old model documents.
+  * @param databaseNew An instance of a MongoDB database connection containing new model documents.
+  */
+final class MigratorDataset(evm: EntityVersionMap,
```
GeorgiChochov:
```suggestion
final class DatasetMigrator(evm: EntityVersionMap,
```
Same as for `SchemaMigrator`, probably for `MappngTable` and `Run` as well

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/migrate01/MigratorDataset.scala
```diff
@@ -0,0 +1,152 @@
...
+        migrationUserName,
+        conformance = dataset0.conformance.map(migrateConformanceRule)
+      ))
+    } catch {
+      case NonFatal(e) =>
+        log.warn(s"Encountered a serialization error for '$collectionBase': ${e.getMessage}")
+        None
+    }
+
+    dataset1Opt.foreach(dataset1 => {
```
GeorgiChochov: Same idea with `Try` as for the schema one, probably in `MappingTable` and `Run` as well.

migrations/pom.xml
GeorgiChochov: Yeah, I think that's a good idea.
Thanks for already having done it 😄!

DzMakatun:
@yruslan Use case: a user updates a dataset in the old instance, the same version already exists in the new instance, therefore, after migration, the dataset version has a different number in the new instance. The schema referred by the dataset also has different versions in the old and new instances.
Question: How can a user identify corresponding versions of the objects between old and new version? E.g. how can a user reliably find in the new instance the version which was recently created in the old? Does it necessary require a detailed comparison of the object contents?

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/migrate01/MigratorDataset.scala
yruslan: Done for migrations. Raised #802 for other modules.

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/migrate01/MigratorDataset.scala
yruslan: Usually, I like common (more general) name bits to go at the beginning of a name so files form a natural grouping. E.g. all migrator subclasses are grouped together in the list of files or the list of classes. But it might be just a habit from C++ where tooling for managing classes is not so great. Agree that reversing the bits sounds more natural.

yruslan:
@DzMakatun, sent you the link to the design doc. Let me know if you have any questions on it. We can discuss the design and the implementation in more detail if you want.

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/migrate01/MigratorSchema.scala
yruslan: Nice! Much cleaner approach indeed.

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/migrate01/EntityRepository.scala
```diff
@@ -0,0 +1,163 @@
...
+  import za.co.absa.enceladus.migrations.framework.dao.ScalaMongoImplicits._
+
+  /**
+    * Returns iterator to a sorted collection of documents. Documents are sorted by name and version.
+    *
+    * @return An iterator to a JSON representation of the documents.
+    */
+  def getSortedDocuments: Iterator[String] = {
+    db.getCollection(collectionName)
+      .find()
```
DzMakatun: Here we sort and iterate an entire collection on each call of continuous migration. Is it possible to iterate over the new documents which were added since the last run? As far as I understand, by default, [ObjectId](https://docs.mongodb.com/manual/reference/method/ObjectId/) in mongo includes a timestamp of when the document was created. Can we use it to filter the documents (ObjectId.getTimestamp() >= timestampOfTheLastMigratedDocumentInThePreviousRun) in the old instance? 

This could require special treatment in case of updatable objects, like runs. Possibly, we can have a moving timeHorizon: all the documents created before timeHorizon are safe to consider migrated and unchanged. The timeHorizon increases as we migrate documents, until we encounter an unfinished run (runs recognized to be in a Zombie state can be ignored).

What do you think?

yruslan: This is a good idea. Just wondering if it worth the effort. We can experiment how much time a continuous migration cycle takes in UAT on prod-like volume of data and decide if we need to optimize.

The last immutable state timestamp can be derived for sure. And we can store it in a nearby collection. But can we query MongoDb for the records arrived since the last timestamp?

DzMakatun: > Just wondering if it worth the effort. We can experiment how much time a continuous migration cycle takes in UAT on prod-like volume of data and decide if we need to optimize.

Right, it is a good idea to check on prod db dump. If it is not an issue right now, we can log it as a separate feature for future, when the database grows larger.  

>But can we query MongoDb for the records arrived since the last timestamp?

Yes. For example in mongo shell you can find (and count) all the objects created after June 1st 2019:

```
db.getCollection('dataset_v1').find(
  {_id:
    {$gt: ObjectId("5cf1a3e00000000000000000")}
  }).count()
```
Here I used this [calculator](https://steveridout.github.io/mongo-object-time/) to encode the timestamp (June 1st 2019) into ObjectId string.

yruslan: Thanks! Good to know. Going to check how much time it takes to run a continuous migration cycle, and then decide if we should implement the timestamp-based optimization.

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/migrate01/DatasetMigrator.scala
```diff
@@ -0,0 +1,157 @@
...
+      )
+    } match {
+      case Success(dataset1) =>
+        if (repo.doesDocumentExist(dataset1.name, dataset1.version)) {
+          resolveConflict(dataset1, objectId, repo)
+        } else {
+          normalInsert(dataset1, objectId, repo)
+        }
+      case Failure(e) =>
+        log.warn(s"Encountered a serialization error for '$collectionBase': ${e.getMessage}")
```
benedeki: Warning or error here?

migrations-cli/pom.xml
```diff
@@ -0,0 +1,81 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <artifactId>enceladus-migrations-cli</artifactId>
+    <packaging>jar</packaging>
+
+    <parent>
+        <groupId>za.co.absa</groupId>
+        <artifactId>enceladus-parent</artifactId>
+        <version>1.1.0-SNAPSHOT</version>
```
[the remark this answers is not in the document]
Zejnilovic: Doesn't maven inherit version from parent?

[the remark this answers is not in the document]
yruslan: Not sure

migrations-cli/pom.xml
```diff
@@ -0,0 +1,81 @@
...
+            <groupId>com.github.scopt</groupId>
+            <artifactId>scopt_${scala.compat.version}</artifactId>
+            <version>${scopt.version}</version>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <sourceDirectory>src/main/scala</sourceDirectory>
+        <plugins>
+            <plugin>
```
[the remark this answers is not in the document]
yruslan: Isn't it in the parent pom?

migrations-cli/src/main/scala/za/co/absa/enceladus/migrationscli/cmd/MigratorCmdConfig.scala
```diff
@@ -13,36 +13,36 @@
...
-  * This is a class for configuration provided by the command line parameters to the command like Mgration Tool
+  * This is a class for configuration provided by the command line parameters to the command line Migration Tool
   *
   * Note: scopt requires all fields to have default values.
   * Even if a field is mandatory it needs a default value.
   */
-case class CmdConfig (mongoDbURL: String = "",
-                      database: String = "",
-                      targetVersion: Int = -1)
+case class MigratorCmdConfig(mongoDbURL: String = "",
```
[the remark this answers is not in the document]
yruslan: Thanks

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/migrate01/DatasetMigrator.scala
yruslan: Agree, having `error` here is better.

Normal migration encountered several documents that were just bogus, so it was logging these instances as warnings. The continuous migration should be a little stricter maybe.

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/migrate01/DatasetMigrator.scala
```diff
@@ -0,0 +1,157 @@
...
+    var saved = false
+    while (retriesLeft > 0 && !saved) {
+      retriesLeft -= 1
+      val latestVersion = repo.getLatestVersion(dataset.name)
+      val datasetToSave = dataset.copy(version = latestVersion + 1)
+      val dataset1Json = ObjectIdTools
+        .putObjectIdIfNotPresent(Serializer1.serializeDataset(datasetToSave), Option(objectId))
+
+      try {
+        repo.insertDocument(dataset1Json)
```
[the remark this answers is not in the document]
yruslan: It can cause duplicates only if a document is inserted successfully, but a mapping cannot be added for some bizzare reason. Results of this will be a version of the document that is not used anywhere. Not a big deal for such a rare ephemeral event.

migrations/src/main/scala/za/co/absa/enceladus/migrations/framework/ObjectIdTools.scala
```diff
@@ -79,4 +79,26 @@ object ObjectIdTools {
...
+    *
+    * the output of the method will be:
+    *
+    * `5b98eea5a43a28a6154a2453`
+    *
+    * @param objIdExpression An Object Id as a JSON key-value expression.
+    * @return A string value of the Object Id
+    */
+  def extractId(objIdExpression: String): String = {
+    val idRegexp = """.*\"\$oid\"\s*:\s*\"([0-9a-fA-F]*)\".*""".r
```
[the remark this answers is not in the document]
yruslan: 1. Yes, the `\` symbols are used for escaping characters that have special meaning in regex.
2. Maybe, but I'd rather not.

migrations-cli/pom.xml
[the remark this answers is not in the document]
benedeki: Yeah, when I wrote it, i thought the PR with unification of poms would be already merged. Now waiting for Jan's reaction there..

migrations/src/main/scala/za/co/absa/enceladus/migrations/framework/ObjectIdTools.scala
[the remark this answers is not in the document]
yruslan: Sorry, misunderstood the first question. Yeah, quotes not need to be escaped. Will remove redundant `\`.

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/migrate01/EntityRepository.scala
GeorgiChochov: @yruslan Any update on this?

migrations-cli/pom.xml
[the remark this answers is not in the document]
GeorgiChochov: No, you have to specify the version of the parent pom. It can't be inherited or extracted to a variable.

migrations-cli/pom.xml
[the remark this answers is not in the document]
GeorgiChochov: That #782 is merged. Let's rebase and add the plugin.

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/migrate01/EntityRepository.scala
yruslan: Haven't checked it on a real database yet.

migrations-cli/pom.xml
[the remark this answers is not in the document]
benedeki: I feared it might be the case.

migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/migrate01/EntityRepository.scala
yruslan: Just ran on my local copy of a database. The migration runs very quickly (a couple of seconds).
The only thing to keep in mind is that indexes must exist for the old database since `sort` relies on it. If `sort` requires more than 32 MB the process will fail.

GeorgiChochov: Great to know! Thanks!

Issue #420, opened 2019-05-27
In case the column cannot be standardized as expected it gets the default value even if nullable


benedeki:
## Describe the bug
In case the _Standardization_ fails on a value of a column the `default` value is the output (as defined by the desired schema meta data or global default). This happens even in case the column is `nullable`. This seem a little weird behavior, let's decide if it's the desired outcome and document the resolution.

## To Reproduce
Steps to reproduce the behavior OR commands run:
```
    val desiredSchema = StructType(Seq(
           StructField("glasses", BooleanType, nullable = true)
      ))

    val src = Seq(
      ("true"),
      ("false"),
      (null),
      ("not anymore")
    )

    val std = StandardizationInterpreter.standardize(src, desiredSchema, "").cache()
```

Ouput:
| true  |
| false |
| null   |
| false |

## Expected behaviour
To be decided

PR #784, opened 2019-08-07
#420 - Nulls allowed as default values

benedeki:
* Nulls are now possible to use as field's explicit default value (if it's nullable)
* Nulls are also used as global default for nullable fields
* TypedStructField class to envelope StructField adding extra capabilities
* EnceladusDateTimeParser renamed to DateTimeParser and moved
* Several new parser for other types
* Validation closer bound to StructField usage, now part of TypedStructField
* TypeParser uses TypedStructField instead of StructField
* Defaults simplified
* Used metadata keys defined as constants
* DateTimePattern has not dependency on StructField, it was unsystematic. Logic moved to TypedStructField
* Existing tests modified to follow new Nulls appearances


benedeki:
I marked the PR `work in progress`. There's one more thing I want to look into (marked with TODO) and considering adding some more UTs.
But the new functionality is already covered by some of the older tests. I don't expect any changes in the core code.

utils/src/main/scala/za/co/absa/enceladus/utils/validation/SchemaValidator.scala
```diff
@@ -37,9 +38,9 @@ object SchemaValidator {
     val flatSchema = flattenSchema(schema)
     for {s <- flatSchema} {
       val fieldWithPath = if (s.structPath.isEmpty) s.field else s.field.copy(name = s.structPath + "." + s.field.name)
-      val issues = validateColumnName(s.field.name, s.structPath) ++ FieldValidator.validate(fieldWithPath)
+      val issues = validateColumnName(s.field.name, s.structPath) ++ TypedStructField(fieldWithPath).validate()
       if (issues.nonEmpty) {
-        val pattern = if (s.field.metadata contains "pattern") s.field.metadata.getString("pattern") else ""
+        val pattern = if (s.field.metadata contains "pattern") s.field.metadata.getString("pattern") else "" //TODO look into this
```
benedeki: This is the TODO mentioned in my the initial comment. Created Issue #787 about this. TODO has been removed, just not committed yet.

Zejnilovic:
Ran test locally. Seems good

standardization/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
```diff
@@ -92,29 +93,29 @@ object TypeParser extends StandardizationCommon {
...
+      case _: StructType => StructParser(TypedStructField.asStructTypeStructField(field), _, _, _)
+      case _: ByteType => IntegralParser(TypedStructField(field), _, _, _,Set(ShortType, IntegerType, LongType), Byte.MinValue, Byte.MaxValue)
+      case _: ShortType => IntegralParser(TypedStructField(field), _, _, _, Set(IntegerType, LongType), Short.MinValue, Short.MaxValue)
+      case _: IntegerType => IntegralParser(TypedStructField(field), _, _, _, Set(LongType), Int.MinValue, Int.MaxValue)
+      case _: LongType => IntegralParser(TypedStructField(field), _, _, _, Set.empty, Long.MinValue, Long.MaxValue)
+      case _: FloatType | _: DoubleType => FractionalParser(TypedStructField.asFractionalTypeStructField(field), _, _, _)
+      case _: DecimalType => DecimalParser(TypedStructField(field), _, _, _)
+      case _: StringType => StringParser(TypedStructField(field), _, _, _)
+      case _: BooleanType => BooleanParser(TypedStructField(field), _, _, _)
+      case _: DateType => DateParser(TypedStructField.asDateTimeTypeStructField(field), _, _, _)
```
Zejnilovic: Did we already have a chat if these should be on the new line or not?
Like `...Type =>\n Array...`

utils/src/main/scala/za/co/absa/enceladus/utils/types/TypedStructField.scala
```diff
@@ -0,0 +1,231 @@
...
+      case _: IntegerType => IntegralTypeStructField(structField)
+      case _: LongType => IntegralTypeStructField(structField)
+      case _: FloatType => FractionalTypeStructField(structField)
+      case _: DoubleType => FractionalTypeStructField(structField)
+      case dt: DecimalType => DecimalTypeStructField(structField, dt)
+      case _: TimestampType => DateTimeTypeStructField(structField, TimestampFieldValidator)
+      case _: DateType => DateTimeTypeStructField(structField, DateFieldValidator)
+      case at: ArrayType => ArrayTypeStructField(structField, at)
+      case st: StructType => StructTypeStructField(structField, st)
+      case _ => GeneralTypeStructField(structField)
```
Zejnilovic: 1. Just the indentations
```scala
  def apply(structField: StructField): TypedStructField = {
    structField.dataType match {
      case _: StringType      => StringTypeStructField(structField)
      case _: BooleanType     => BooleanTypeStructField(structField)
      case _: ByteType        => IntegralTypeStructField(structField)
      case _: ShortType       => IntegralTypeStructField(structField)
      case _: IntegerType     => IntegralTypeStructField(structField)
      case _: LongType        => IntegralTypeStructField(structField)
      case _: FloatType       => FractionalTypeStructField(structField)
      case _: DoubleType      => FractionalTypeStructField(structField)
      case dt: DecimalType    => DecimalTypeStructField(structField, dt)
      case _: TimestampType   => DateTimeTypeStructField(structField, TimestampFieldValidator)
      case _: DateType        => DateTimeTypeStructField(structField, DateFieldValidator)
      case at: ArrayType      => ArrayTypeStructField(structField, at)
      case st: StructType     => StructTypeStructField(structField, st)
      case _                  => GeneralTypeStructField(structField)
    }
  }
```
2. I tried putting all integral together and looks supper ugly ...
3. By this time, we should already have our own dataType called integral and fractional.

utils/src/main/scala/za/co/absa/enceladus/utils/schema/MetadataKeys.scala
```diff
@@ -0,0 +1,27 @@
...
+
+package za.co.absa.enceladus.utils.schema
+
+object MetadataKeys {
+  // all
+  val SourceColumn = "sourcecolumn"
+  val DefaultValue = "default"
+  // date & timestamp
+  val Pattern = "pattern"
+  val DefaultTimeZone= "timezone"
```
Zejnilovic:
```suggestion
  val DefaultTimeZone = "timezone"
```

utils/src/main/scala/za/co/absa/enceladus/utils/types/Defaults.scala
```diff
@@ -48,47 +40,24 @@ object Defaults {
...
+  /** A function which defines default values for primitive types, allowing possible Null*/
+  def getGlobalDefaultWithNull(dt: DataType, nullable: Boolean): Try[Option[Any]] = {
+    val result = if (nullable) {
+      Success(None)
+    } else {
+      Try{
+        getGlobalDefault(dt)
+      }.map(Some(_))
+    }
+    result
```
Zejnilovic: The result val is probably unnecessary

utils/src/main/scala/za/co/absa/enceladus/utils/types/TypedStructField.scala
```diff
@@ -0,0 +1,231 @@
...
+import za.co.absa.enceladus.utils.types.parsers._
+import za.co.absa.enceladus.utils.validation.ValidationIssue
+import za.co.absa.enceladus.utils.validation.field.{DateFieldValidator, DateTimeFieldValidator, FieldValidator, ScalarFieldValidator, TimestampFieldValidator}
+
+import scala.util.{Failure, Success, Try}
+
+sealed abstract class TypedStructField(structField: StructField) extends StructFieldEnhancements(structField) {
+  protected def convertString(string: String): Try[Any]
+
+  def validate():Seq[ValidationIssue]
```
Zejnilovic:
```suggestion
  def validate(): Seq[ValidationIssue]
```

utils/src/main/scala/za/co/absa/enceladus/utils/types/TypedStructField.scala
```diff
@@ -0,0 +1,231 @@
...
+      }
+    } else {
+      convertString(string).map(Some(_))
+    }
+  }
+
+  def ownDefaultValue: Try[Option[Option[Any]]] = {
+    if (hasMetadataKey(MetadataKeys.DefaultValue)) {
+      for {
+        defeaultValueString <- Try{structField.metadata.getString(MetadataKeys.DefaultValue)}
```
Zejnilovic:
```suggestion
        defaultValueString <- Try{structField.metadata.getString(MetadataKeys.DefaultValue)}
```

utils/src/main/scala/za/co/absa/enceladus/utils/types/TypedStructField.scala
```diff
@@ -0,0 +1,231 @@
...
+        Success(None)
+      } else {
+        Failure(new IllegalArgumentException(s"null is not a valid value for field '${structField.name}'"))
+      }
+    } else {
+      convertString(string).map(Some(_))
+    }
+  }
+
+  def ownDefaultValue: Try[Option[Option[Any]]] = {
```
Zejnilovic: Is there maybe a way to make the return value less complicated?

utils/src/main/scala/za/co/absa/enceladus/utils/types/TypedStructField.scala
```diff
@@ -0,0 +1,231 @@
...
+      }
+    }
+
+    override def validate(): Seq[ValidationIssue] = {
+      ScalarFieldValidator.validate(this)
+    }
+  }
+
+  sealed case class DecimalTypeStructField(override val structField: StructField, override val dataType: DecimalType) extends TypedStructField(structField) {
+    val parser = DecimalParser(precision, scale)
```
Zejnilovic: Someone is getting ready for the changeable precision decimal. I like that 👍

utils/src/main/scala/za/co/absa/enceladus/utils/types/TypedStructField.scala
```diff
@@ -0,0 +1,231 @@
...
+
+    override def validate(): Seq[ValidationIssue] = {
+      ScalarFieldValidator.validate(this)
+    }
+
+    def precision: Int = dataType.precision
+    def scale: Int = dataType.scale
+  }
+
+  sealed case class DateTimeTypeStructField(override val structField: StructField, validator: DateTimeFieldValidator) extends TypedStructField(structField) {
```
Zejnilovic: Move the parameters on the new line

utils/src/main/scala/za/co/absa/enceladus/utils/types/parsers/DecimalParser.scala
```diff
@@ -0,0 +1,36 @@
...
+ * limitations under the License.
+ */
+
+package za.co.absa.enceladus.utils.types.parsers
+
+class DecimalParser(val precision: Int, val scale: Int) {
+
+  def parseDecimal(value: String): BigDecimal = {
+    val result = BigDecimal(value)
+    if (result.precision > precision || result.scale > scale) {
```
Zejnilovic: Not sure about this. When I expect the scale to be 10. I expect exactly 10 numbers. This check only if it is bigger. Is that enough? What about smaller?

utils/src/main/scala/za/co/absa/enceladus/utils/types/parsers/FractionalParser.scala
```diff
@@ -0,0 +1,49 @@
...
+    if (result.isInfinity && (!allowInfinity)) {
+      throw new NumberFormatException(s"The Float value '$value' is infinite or out of range")
+    }
+    result
+  }
+
+  def parseDouble(value: String): Double = {
+    val result = value.toDouble
+    if (result == Double.NaN) {
+      throw new NumberFormatException(s"The Float value '$value' is NaN")
```
Zejnilovic:
```suggestion
      throw new NumberFormatException(s"The Double value '$value' is NaN")
```

utils/src/main/scala/za/co/absa/enceladus/utils/types/parsers/FractionalParser.scala
```diff
@@ -0,0 +1,49 @@
...
+    }
+    if (result.isInfinity && (!allowInfinity)) {
+      throw new NumberFormatException(s"The Float value '$value' is infinite or out of range")
```
Zejnilovic:
```suggestion
      throw new NumberFormatException(s"The Double value '$value' is infinite or out of range")
```

utils/src/main/scala/za/co/absa/enceladus/utils/validation/field/FieldValidator.scala
```diff
@@ -15,35 +15,25 @@
...
-        Option(FieldValidatorTimestamp)
-      case _: BooleanType =>
-        Option(FieldValidatorScalar)
-      case _: NumericType =>
-        Option(FieldValidatorScalar)
-      case _: StringType =>
-        Option(FieldValidatorScalar)
-      case _ => None
+class FieldValidator {
+  protected def tryToValidationIssues(tryValue: Try[Any]): Seq[ValidationIssue] = {
```
Zejnilovic: This name is a bit weird to me

utils/src/main/scala/za/co/absa/enceladus/utils/validation/field/FieldValidator.scala
```diff
@@ -15,35 +15,25 @@
...
-      case _: TimestampType =>
-        Option(FieldValidatorTimestamp)
-      case _: BooleanType =>
-        Option(FieldValidatorScalar)
-      case _: NumericType =>
-        Option(FieldValidatorScalar)
-      case _: StringType =>
-        Option(FieldValidatorScalar)
-      case _ => None
+class FieldValidator {
```
Zejnilovic: Could be abstract?

Zejnilovic:
Created #790 for the issue with Menas not showing null when it is a default value.

utils/src/main/scala/za/co/absa/enceladus/utils/validation/field/FieldValidator.scala
benedeki: Plan was and is to use it in `trait WeakSupport`, Added it will be in commit.

standardization/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
benedeki: Honestly I don't remember if we did.
Tried, when multiline  the cases drag, decreasing the clarity. But I aligned the => arrows the way how you like 😉

utils/src/main/scala/za/co/absa/enceladus/utils/validation/field/FieldValidator.scala
benedeki: Naming ala `stringToInt`, try is `Try` the noun.

Zejnilovic: Document it please

benedeki:
I removed the 'Work in progress' label, as I've added the stuff I wanted.
With the commit I also addressed the comments so far.

utils/src/test/scala/za/co/absa/enceladus/utils/validation/field/DateFieldValidatorSuite.scala
```diff
@@ -18,78 +18,76 @@ package za.co.absa.enceladus.utils.validation.field
...
+    assert(DateFieldValidator.validate(field("yyMMdd_zz", Option("190301_EST!!!!"))).toSet == expected)
     //timestamp with offset time zone
-    assert(FieldValidatorDate.validateStructField(field("yyyy/MM/dd XXX", Option("2019/01/31 -11:00"))).toSet == expected)
+    assert(DateFieldValidator.validate(field("yyyy/MM/dd XXX", Option("2019/01/31 -11:00"))).toSet == expected)
   }
 
   test("invalid pattern") {
     val expected1 = Set(
-      ValidationError("Illegal pattern character 'f'"),
-      ValidationWarning("No year placeholder 'yyyy' found."),
```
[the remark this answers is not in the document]
GeorgiChochov: Good to know!

utils/src/main/scala/za/co/absa/enceladus/utils/validation/field/FieldValidator.scala
```diff
@@ -15,35 +15,25 @@
...
+      case Failure(e)                      => Seq(ValidationError(e.getMessage))
+      case Success(seq: Seq[_])            => seq.collect{case x:ValidationIssue => x} //have to use collect because of type erasure
+      case Success(opt: Option[_])         => opt.collect{case x:ValidationIssue => x}.toSeq
+      case Success(issue: ValidationIssue) => Seq(issue)
+      case _                               => Nil
     }
-    validator.map(_.validateStructField(field)).getOrElse(Nil)
+  }
+
+  def validate(field: TypedStructField): Seq[ValidationIssue] = {
```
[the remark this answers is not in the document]
GeorgiChochov: Still don't understand why that's the case.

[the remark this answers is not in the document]
GeorgiChochov: Oh, is it that only `timestamp` and `date` that have something to validate? (eventually numerics as well, I guess?)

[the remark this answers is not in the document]
benedeki: General fields have nothing.
Scalars (numerics, Boolean and string) validate local `default` in case it is defined (with numerics eventually having `pattern` as well)
Date and time validate `default`, `pattern` and `timezone`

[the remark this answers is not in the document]
GeorgiChochov: Perfect, thanks!

utils/src/main/scala/za/co/absa/enceladus/utils/types/TypedStructField.scala
```diff
@@ -0,0 +1,231 @@
...
+  def asBooleanTypeStructField(structField: StructField): BooleanTypeStructField = TypedStructField(structField).asInstanceOf[BooleanTypeStructField]
+  def asIntegralTypeStructField(structField: StructField): IntegralTypeStructField = TypedStructField(structField).asInstanceOf[IntegralTypeStructField]
+  def asFractionalTypeStructField(structField: StructField): FractionalTypeStructField = TypedStructField(structField).asInstanceOf[FractionalTypeStructField]
+  def asDecimalTypeStructField(structField: StructField): DecimalTypeStructField = TypedStructField(structField).asInstanceOf[DecimalTypeStructField]
+  def asDateTimeTypeStructField(structField: StructField): DateTimeTypeStructField = TypedStructField(structField).asInstanceOf[DateTimeTypeStructField]
+  def asArrayTypeStructField(structField: StructField): ArrayTypeStructField = TypedStructField(structField).asInstanceOf[ArrayTypeStructField]
+  def asStructTypeStructField(structField: StructField): StructTypeStructField = TypedStructField(structField).asInstanceOf[StructTypeStructField]
+
+  sealed case class StringTypeStructField(override val structField: StructField) extends TypedStructField(structField) {
```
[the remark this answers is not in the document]
GeorgiChochov: Isn't that what access modifiers are for? `sealed` has a different meaning and doesn't change the fact that the class in `public` by default.

utils/src/main/scala/za/co/absa/enceladus/utils/types/TypedStructField.scala
GeorgiChochov: Very interesting

utils/src/main/scala/za/co/absa/enceladus/utils/types/TypedStructField.scala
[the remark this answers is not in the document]
benedeki: I don't think the access modifiers are of much use here. It's OK to use these classes all around (I would say encouraged). 
The `sealed` server two purposes here, both of which are IMHO the intended purposes of  the modifier.
1. Allow exhaustive match checking - that perfectly works here.
2. Control the classes creation while allowing to use the classes and their extra abilities freely. (That's also reason for the reason of the typed "constructors" offered)
The `private final` would make sense, if all the subtypes would have the same public interface. But they offer and require too diverse capabilities to make one common interface set feasible.
There is actually a design flaw (IMO) in Sparks Data class structure we talked about with @sasa few days ago. There are  ancestors `IntegralType`, `NumericType`, `FractionalType` - but they are set `private` which totally hinders their usage.
Also I found out that `final sealed` is not allowed combination (when I originally wanted to mark the classes that way).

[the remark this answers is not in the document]
GeorgiChochov: 1. You get that by making their super class `sealed` not the subclasses`, unless I'm missing that you also have subclasses of these subclasses (i.e. `... extends StringTypeStructField`). In which case I'd have a bigger problem with them `case` classes that you are extending, but I don't think that's the case.
2 Looking at your previous comment I was left witht he impression that you
> Don't want them to be created "wildly" but via the TypedStructField.aplly method

making them sealed has nothing to do with this. This is an `access modifier` issue. In which case fine keep them `public`, but I can't see how your previous comment. And of course you can't make them `sealed final`, that would be contradictory. `sealed` expresses that it is extended by a limited set of classes, while `final` expresses that nothing can extend it.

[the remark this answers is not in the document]
benedeki: Let's discuss in person on Monday. I am not exactly sure, where the gist of this dispute is.

utils/src/main/scala/za/co/absa/enceladus/utils/types/TypedStructField.scala
Zejnilovic: I am marking this as resolved. I presume no one has any more comments here.

utils/src/main/scala/za/co/absa/enceladus/utils/types/TypedStructField.scala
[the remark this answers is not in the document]
benedeki: OK, after thinking it through I hope I got to the key point of the comment. That with `sealed` I was not achieving what I claimed, and what was my main goal:. preventing the `TypedStructField` sub-classes creation outside of the `TypedStructFiled.apply` method. The latest commit is addressing that.

yruslan:
Was this tested on an actual cluster?

Zejnilovic:
> Was this tested on an actual cluster?

Not by me.

benedeki:
> 
> 
> Was this tested on an actual cluster?

I tested it locally, and as far as I know @Zejnilovic tested it also (by his words locally only too).
What risk/difference you see here compared to run on cluster?

yruslan:
During the lifetime of the project, I've encountered a couple of time a situation when a Spark job runs fine locally but fails on a cluster. Most of the time it was related to serialization issues and dependency conflicts.

This particular PR should not affect the runnability on a cluster so I don't insist or doing that. But it is generally a good practice if a Spark job is changed to try it locally and on a cluster.

benedeki:
> 
> 
> During the lifetime of the project, I've encountered a couple of time a situation when a Spark job runs fine locally but fails on a cluster. Most of the time it was related to serialization issues and dependency conflicts.
> 
> This particular PR should not affect the runnability on a cluster so I don't insist or doing that. But it is generally a good practice if a Spark job is changed to try it locally and on a cluster.

In that case we desperately need automated cluster testing environment.
I am a noob with _Spark_, but it sounds it's little unpredictable regarding the difference between local and cluster runs. Re-testing manually only when we _think_ the PR might hit the difference is unfeasible in longer run.

utils/src/main/scala/za/co/absa/enceladus/utils/types/TypedStructField.scala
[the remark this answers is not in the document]
GeorgiChochov: I totally see what you're trying to do now. Keep the definition of the class public while making the constructor private. `sealed` here applies to the anonymous subclass making it `final`. It's a bit of weird construct `sealed abstract case class` but it does what it's supposed to here. 

You could make the constructor `private` but with a `case class` you'd also need to override the `apply` method or have it as a regular class and implement the `unapply` method yourself. But either would be a bloated alternative which I don't like too much.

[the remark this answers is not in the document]
yruslan: Looks like a bit of overengineering. Why use case classes for purposes other than just POJOs/PODs/etc.? 
Also, the usage of anonymous subclasses raises a lot of 'why's.
Why you need classes to be public if they cannot be instantiated, and only anonymous subclasses are used? Let's come up with a simpler practice for this. Please. Currently what is achieved is why is hard for me to comprehend.

utils/src/main/scala/za/co/absa/enceladus/utils/validation/field/FieldValidator.scala
benedeki: Done

utils/src/main/scala/za/co/absa/enceladus/utils/types/TypedStructField.scala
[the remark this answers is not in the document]
benedeki: Added explanatory comment to help understand why the thing was done it has been.

Zejnilovic:
> In that case we desperately need automated cluster testing environment.
Hopefully, DevOps and AWS can help us with this. Still, a lot to set up, still in the PoC phase. 

I will set up a few scripts on ****** server when its up.

utils/src/main/scala/za/co/absa/enceladus/utils/types/TypedStructField.scala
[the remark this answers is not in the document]
GeorgiChochov: Let's have a discussion about this in CQC. For now I can live with this oddity but I see quite a bit of contention.

Issue #783, opened 2019-08-07
Address the several remaing scattered build warnings


benedeki:
## Background
There are still some warnings remaining in the build log that can be addressed

## Feature
Refactoring for better build log

PR #801, opened 2019-08-10
#783: Addressing some remaining scattered build warnings

benedeki:
* Deprecated
* Really long lines
* Tags on prohibited packages and rare language features
* Scalastyle

### DIFF
--- a/dao/src/main/scala/za/co/absa/enceladus/dao/EnceladusRestDAO.scala
+++ b/dao/src/main/scala/za/co/absa/enceladus/dao/EnceladusRestDAO.scala
@@ -35,7 +35,7 @@ import com.fasterxml.jackson.databind.SerializationFeature
 import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule
 import com.fasterxml.jackson.module.scala.DefaultScalaModule
 import com.typesafe.config.ConfigFactory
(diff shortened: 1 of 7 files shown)


dao/src/main/scala/za/co/absa/enceladus/dao/EnceladusRestDAO.scala
```diff
@@ -35,7 +35,7 @@ import com.fasterxml.jackson.databind.SerializationFeature
 import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule
 import com.fasterxml.jackson.module.scala.DefaultScalaModule
 import com.typesafe.config.ConfigFactory
-import sun.security.krb5.internal.ktab.KeyTab
+import sun.security.krb5.internal.ktab.KeyTab //scalastyle:ignore illegal.imports
```
yruslan: Why ignoring an illegal import?

benedeki: As it's the key part of the `KeyTab` support and imported with clear intent.
Having _warnings_ that have no consequence just adds noise. If the import is a worry, let's create a ticket for it, that might be eventually addressed.

Issue #797, opened 2019-08-09
Include raw format in run metadata


yruslan:
## Background
Currently, there is no convenient way to determine what was the original raw format of the published data.

## Feature
Include raw format in run metadata. Keeping information on raw format there has an additonal benefit of it being stored in _INFO files.

## Proposed Solution [Optional]
`"raw_format", "xml"`

PR #803, opened 2019-08-12
#797 Add raw format to _INFO file metadata.

yruslan:


standardization/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
```diff
@@ -106,6 +106,9 @@ object StandardizationJob {
     Atum.setAdditionalInfo(s"enceladus_info_date" -> cmd.reportDate)
     Atum.setAdditionalInfo(s"enceladus_info_version" -> reportVersion.toString)
 
+    // Add the raw format of the input file(s) to Atum's metadta as well
+    Atum.setAdditionalInfo(s"raw_format" -> cmd.rawFormat)
```
Zejnilovic: Why is there `s` in front of the string?

yruslan: I knew something is wrong with this PR.

PR #804, opened 2019-08-12
#731: DeepArrayTransformations.nestedAddColumn function parameter cor…

benedeki:
…rect typing

* simple `Column` is enough


Zejnilovic:
Re-running Jenkins. Seemed like a yum glitch.

PR #809, opened 2019-08-13
Feature/446 add control features to scheduling

lokm01:
This PR adds
 - Run now
 - Suspend coordinator
 - Resume coordinator


menas/src/main/scala/za/co/absa/enceladus/menas/controllers/OozieController.scala
```diff
@@ -32,6 +34,25 @@ class OozieController @Autowired() (oozieService: OozieService) extends BaseCont
...
   @GetMapping(path = Array("/coordinatorStatus/{id}"))
-  def getCoordinatorStatus(@PathVariable id: String): CompletableFuture[OozieCoordinatorStatus] = oozieService.getCoordinatorStatus(id)  
-  
+  def getCoordinatorStatus(@PathVariable id: String): CompletableFuture[OozieCoordinatorStatus] = {
+    oozieService.getCoordinatorStatus(id)
+  }
+
+  @PostMapping(path = Array("/runNow"))
+  def runNow(@RequestBody schedule: OozieSchedule, @RequestParam reportDate: Optional[String]): CompletableFuture[String] = {
+    val rd: Option[String] = reportDate
```
Zejnilovic: Is this step really needed?

menas/src/main/scala/za/co/absa/enceladus/menas/repositories/OozieRepository.scala
```diff
@@ -43,6 +43,7 @@ import za.co.absa.enceladus.model.menas.scheduler.RuntimeConfig
 import za.co.absa.enceladus.menas.exceptions.EntityAlreadyExistsException
 import za.co.absa.enceladus.utils.time.TimeZoneNormalizer
 import OozieRepository._
+import scala.util.Try
```
Zejnilovic: This import became obsolete 
```scala
import org.apache.oozie.client.WorkflowJob.{Status => WorkflowStatus}
```

menas/src/main/scala/za/co/absa/enceladus/menas/repositories/OozieRepository.scala
```diff
@@ -322,11 +316,40 @@ class OozieRepository @Autowired() (oozieClientRes: Either[OozieConfigurationExc
...
     }
   }
 
+  /**
+   * Run a workflow now
+   */
+  def runWorkflow(wfPath: String, runtimeParams: RuntimeConfig, reportDate: String): Future[String] = {
+    getOozieClientWrap { oozieClient: OozieClient =>
+      val conf = getOozieConf(oozieClient, runtimeParams)
+      conf.setProperty(OozieClient.APP_PATH, s"$wfPath");
```
Zejnilovic:
```suggestion
      conf.setProperty(OozieClient.APP_PATH, wfPath)
```

menas/src/main/scala/za/co/absa/enceladus/menas/services/OozieService.scala
```diff
@@ -22,13 +22,41 @@ import org.springframework.stereotype.Component
...
   }
+
+  def runNow(oozieSchedule: OozieSchedule, reportDate: Option[String]): Future[String] = {
+    val wfPath = oozieSchedule.activeInstance match {
+      case Some(instance) => instance.workflowPath
+      case None           => throw new IllegalArgumentException("Cannot run a non-active schedule.")
+    }
+    val reportDateString = reportDate match {
+      case Some(date) => date
+      case None => {
```
Zejnilovic: the brackets don't have to be here

menas/ui/css/style.css
```diff
@@ -80,11 +76,31 @@ html, body {
...
+  font-size: 0.9rem !important;
+  line-height: 1rem !important;
+}
+
+.scheduleActionButton > .sapMBtnInner > .sapMBtnIcon {
+  height: 1.25rem !important;
+  font-size: 0.9rem !important;
+  line-height: 1rem !important;
+  padding-top: 0.1rem !important;
+}
```
Zejnilovic: add a newline

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/OozieController.scala
lokm01: Yeah true... it infers the target type from the call

Zejnilovic:
## 1 App directory does not exist
if I try run now and the app directory does not exist I get 500 in the console and a network response with the message but nothing visible for the user. Can the error `E0504 : E0504: App directory [hdfs:&#47;&#47;localhost:8020&#47;tmp&#47;menas-oozie-schedule-wf-AAA-20&#47;workflow.xml] does not exist` be propagated to the user somehow?

Zejnilovic:
## 2 add text to the action buttons
I think these are not self-explanatory and could use a text bellow them and they might be even a bit bigger.
image

menas/ui/service/OozieService.js
```diff
@@ -23,9 +23,37 @@ var OozieService = new function () {
...
+        model.setProperty("/currentDataset/schedule/activeInstance/status", oData);
+      })
+    }
+  };
+
+  this.runNow = function() {
+    const oSchedule = model.getProperty("/currentDataset/schedule")
+    if(oSchedule) {
+      RestClient.post("api/oozie/runNow", oSchedule).then((oData) => {
+        sap.m.MessageToast.show(`Schedule has been submitted with ID: ${oData}`, {duration: 10000});
```
Zejnilovic: This is not a schedule but a job, is it not?

Zejnilovic:
## 3 Editing of a suspended schedule
If I have a schedule that is suspended. I am able to still click edit, but I am unable to fill anything. 
_Suggestion:_ Be able to edit and by saving you re-enable the schedule with the new params.

Zejnilovic:
**4 Request**: Can the run now have a not gonna run button have a delay of 2 seconds before it registers the click again?

menas/src/main/scala/za/co/absa/enceladus/menas/services/OozieService.scala
```diff
@@ -22,13 +22,41 @@ import org.springframework.stereotype.Component
...
     oozieRepository.getCoordinatorStatus(coordinatorId)
   }
+
+  def runNow(oozieSchedule: OozieSchedule, reportDate: Option[String]): Future[String] = {
+    val wfPath = oozieSchedule.activeInstance match {
+      case Some(instance) => instance.workflowPath
+      case None           => throw new IllegalArgumentException("Cannot run a non-active schedule.")
+    }
+    val reportDateString = reportDate match {
+      case Some(date) => date
```
benedeki: Question: What if date is of wrong format/pattern or generic nonsense. Isn't it an request incoming value?

Zejnilovic: Oozie will still create a workflow job and that job will then fail as your spark-submit would fail after being unable to parse the `--report-date` arg. This will be then seen by users on the history server with an appropriate message. 
`Error: Match error in 'XXXX'. Option --report-date expects a date in 'yyyy-MM-dd' format`

benedeki: Ok, that's a non-damaging outcome. Still I would consider checking the input.
Might be a separate task of course.

benedeki: Btw, do we know if that format is hard-coded in Oozie or they use system Locale? 🤔

lokm01: This is parsed in Std/Conf jobs I believe. So the report date format is fixed.

Zejnilovic:
## 1 App directory does not exist
I would ask that only the message is shown to the user. Currently, it seems like too much
image

Zejnilovic:
## 4 Resume schedule without app directory
I am able to resume a schedule that does not have an app directory. I this might be problematic

Zejnilovic:
## 2, 3, 4 fixed

## Code
I have no comments for the code

lokm01:
> ## 5 Resume schedule without app directory
> I am able to resume a schedule that does not have an app directory. I this might be problematic

Separate issue pls

Zejnilovic:
> > ## 5 Resume schedule without app directory
> > I am able to resume a schedule that does not have an app directory. I this might be problematic
> 
> Separate issue pls

#827

menas/ui/service/RestDAO.js
```diff
@@ -17,13 +17,29 @@ class RestClient {
 
   static cache = _.memoize($.ajax, ({url}) => url);
 
-  static get(url, shouldUseCache = false) {
+  static _setBusy(oCtl) {
```
GeorgiChochov: I'd like to maintain separation of concerns and keep `RestClient` and `RestDAO` unaware of any `ui5` controls.

menas/src/main/scala/za/co/absa/enceladus/menas/repositories/OozieRepository.scala
```diff
@@ -321,12 +315,47 @@ class OozieRepository @Autowired() (oozieClientRes: Either[OozieConfigurationExc
...
+      conf.setProperty(OozieClient.COORDINATOR_APP_PATH, s"$coordPath")
       oozieClient.submit(conf)
     }
   }
 
+  /**
+   * Run a workflow now
+   */
+  def runWorkflow(wfPath: String, runtimeParams: RuntimeConfig, reportDate: String): Future[String] = {
+    getOozieClient { oozieClient: OozieClient =>
```
GeorgiChochov:
```suggestion
    getOozieClient { oozieClient =>
```
Why do all of these have an explicit type?

menas/src/main/scala/za/co/absa/enceladus/menas/repositories/OozieRepository.scala
```diff
@@ -321,12 +315,47 @@ class OozieRepository @Autowired() (oozieClientRes: Either[OozieConfigurationExc
...
+      val conf = getOozieConf(oozieClient, runtimeParams)
+      conf.setProperty(OozieClient.APP_PATH, wfPath)
+      conf.setProperty("reportDate", reportDate)
+      val res = Try {
+        oozieClient.run(conf)
+      }
+      res match {
```
GeorgiChochov: You can just do 
```
Try {
...
} match {
```
no need for the local variable `res`

menas/src/main/scala/za/co/absa/enceladus/menas/MenasRestExceptionHandler.scala
```diff
@@ -0,0 +1,38 @@
...
+import org.springframework.web.bind.annotation.ControllerAdvice
+import org.springframework.web.bind.annotation.ExceptionHandler
+import org.springframework.web.servlet.mvc.method.annotation.ResponseEntityExceptionHandler
+
+import za.co.absa.enceladus.menas.models.RestError
+import za.co.absa.enceladus.menas.exceptions.OozieActionException
+
+@Order(org.springframework.core.Ordered.HIGHEST_PRECEDENCE)
+@ControllerAdvice
+class MenasRestExceptionHandler extends ResponseEntityExceptionHandler {
```
GeorgiChochov: Let's use `za.co.absa.enceladus.menas.controllers.RestExceptionHandler` instead.

menas/src/main/scala/za/co/absa/enceladus/menas/MenasRestExceptionHandler.scala
```diff
@@ -0,0 +1,38 @@
...
+
+  @ExceptionHandler(Array(classOf[OozieActionException]))
+   def handleOozieActionException(ex: RuntimeException): ResponseEntity[Object] = {
+    val err = RestError(ex.getMessage)
+    logger.error(s"Exception: $err\n${ex.getStackTrace.map(trace => s"\t\t${trace.toString}").mkString("\n")}")
```
GeorgiChochov: With `logger::error` you can pass in the exception as a second argument and it will log the stack trace for you.

menas/src/main/scala/za/co/absa/enceladus/menas/services/OozieService.scala
```diff
@@ -22,13 +22,41 @@ import org.springframework.stereotype.Component
...
   def isOozieEnabled: Boolean = oozieRepository.isOozieEnabled()
 
   def getCoordinatorStatus(coordinatorId: String): Future[OozieCoordinatorStatus] = {
     oozieRepository.getCoordinatorStatus(coordinatorId)
   }
+
+  def runNow(oozieSchedule: OozieSchedule, reportDate: Option[String]): Future[String] = {
+    val wfPath = oozieSchedule.activeInstance match {
+      case Some(instance) => instance.workflowPath
+      case None           => throw new IllegalArgumentException("Cannot run a non-active schedule.")
```
GeorgiChochov: Can we make this a specific exception for the `ExceptionHandler` to return a proper message to the user?

menas/ui/service/OozieService.js
```diff
@@ -13,19 +13,64 @@
...
   const eventBus = sap.ui.getCore().getEventBus();
 
   this.getCoordinatorStatus = function () {
-    const coordinatorId = model.getProperty("/currentDataset/schedule/activeInstance/coordinatorId")
+    const coordinatorId = model().getProperty("/currentDataset/schedule/activeInstance/coordinatorId")
     if(coordinatorId) {
-      Functions.ajax(`api/oozie/coordinatorStatus/${coordinatorId}`, "GET", {}, function (oData) {
-        model.setProperty("/currentDataset/schedule/activeInstance/status", oData)
-      }, function () {
+      RestClient.get(`api/oozie/coordinatorStatus/${coordinatorId}`).then((oData) => {
```
GeorgiChochov: I'd like the `RestClient` to be abstracted away, as much as possible as well as the URIs so that they can be changed in a single location.
Let's extract these into an `OozieRestDAO`:
- `RestClient.get(api/oozie/coordinatorStatus/${coordinatorId})` => `OozieRestDAO.getStatus(coordinatorId)`
- `RestClient.post("api/oozie/runNow", oSchedule)` => `OozieRestDAO.postSchedule(oSchedule)`
... etc

menas/src/main/scala/za/co/absa/enceladus/menas/MenasRestExceptionHandler.scala
lokm01: Yep, missed that.

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/RestExceptionHandler.scala
```diff
@@ -45,4 +49,11 @@ class RestExceptionHandler {
     ResponseEntity.notFound().build[Any]()
   }
 
+  @ExceptionHandler(Array(classOf[OozieActionException]))
+   def handleOozieActionException(ex: RuntimeException): ResponseEntity[Object] = {
```
GeorgiChochov:
```suggestion
   def handleOozieActionException(ex: OozieActionException): ResponseEntity[RestError] = {
```

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/RestExceptionHandler.scala
```diff
@@ -45,4 +49,11 @@ class RestExceptionHandler {
...
+    val err = RestError(ex.getMessage)
+    logger.error(s"Exception: $err", ex.getCause)
```
GeorgiChochov: Why log only the cause? Don't you lose some of the stack trace if you only do `ex.getCause`?

menas/src/main/scala/za/co/absa/enceladus/menas/repositories/OozieRepository.scala
```diff
@@ -222,21 +223,14 @@ class OozieRepository @Autowired() (oozieClientRes: Either[OozieConfigurationExc
...
-  }
-
   /**
    * Kill a running coordinator
    */
   def killCoordinator(coordId: String): Future[Unit] = {
-    getOozieClientWrap({ oozieClient: OozieClient =>
-      oozieClient.kill(coordId)
+    getOozieClientWrap({ oozieClient =>
+      Try {
```
GeorgiChochov: Why add a `Try` here? `getOozieClientWrap` returns a Future, if an exception is thrown it will be caught by `Try` and swallowed. The `Future` will succeed without a trace.

menas/ui/components/dataset/datasetDetail.controller.js
```diff
@@ -439,6 +440,47 @@ sap.ui.define([
...
+      }
+    },
+
+    _clearBusy: function(oCtl) {
+      if(oCtl && oCtl.setBusy) {
+        oCtl.setBusy(false);
+      }
+    },
+
+    _clearBusyPromise: function(oCtl, oPromise) {
```
GeorgiChochov: I like these a lot!

menas/ui/components/dataset/datasetDetail.controller.js
```diff
@@ -439,6 +440,47 @@ sap.ui.define([
...
+        oPromise.always(() => {
+          this._clearBusy(oCtl);
+        })
+      } else {
+        this._clearBusy(oCtl);
+      }
+    },
+    
+    scheduleRunNow: function() {
+      this._setBusy(this._scheduleActionMenu);
```
GeorgiChochov: This pattern is repeated 3 times. Maybe we should extract a function for it (along with the private `set/clearBusy` funcitons) in some generic location so we can reuse it across controllers.

menas/src/main/scala/za/co/absa/enceladus/menas/services/OozieService.scala
```diff
@@ -15,20 +15,50 @@
...
   def isOozieEnabled: Boolean = oozieRepository.isOozieEnabled()
 
   def getCoordinatorStatus(coordinatorId: String): Future[OozieCoordinatorStatus] = {
     oozieRepository.getCoordinatorStatus(coordinatorId)
   }
+
+  def runNow(oozieSchedule: OozieSchedule, reportDate: Option[String]): Future[String] = {
+    val wfPath = oozieSchedule.activeInstance match {
+      case Some(instance) => instance.workflowPath
+      case None           => throw OozieActionException("Cannot run a job without an active schedule. Click 'Edit Schedule' to configure schedule first.")
```
GeorgiChochov: This is a response from a REST API, shouldn't mention or even know about any UI things like clicking buttons.

menas/ui/components/dataset/datasetDetail.controller.js
lokm01: Gonna extract into a function now.

RE generic location, I would like to create a layer or two of generic (UI5) controllers which the entity ones could then extend. But that's for a separate issue.

menas/src/main/scala/za/co/absa/enceladus/menas/services/OozieService.scala
lokm01: Makes sense

menas/ui/components/dataset/datasetDetail.controller.js
GeorgiChochov: Good idea!

benedeki:
For the record: My comments were addressed, I just didn't want to approve the whole PR while discussion on other parts of the code were/are ongoing.

Issue #672, opened 2019-07-16
Option to automatically clean up standardized output


lokm01:
(probably) Dataset should have an option to clean the standardized output automatically after successful conformance.

PR #811, opened 2019-08-14
#672 Add an option to automatically clean up standardized folder

yruslan:
A standardized folder is only cleaned if a Conformance job has finished successfully.

### DIFF
--- a/conformance/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
+++ b/conformance/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
@@ -242,7 +242,9 @@ object DynamicConformanceJob {
       }
     })
 
-    autocleanStandardizedHdfsPath(pathCfg)
+    if (isAutocleanStdFolderEnabled()) {
+      fsUtils.deleteDirectoryRecursively(pathCfg.stdPath)
+    }
   }
 
   def buildPublishPath(infoDateCol: String,
@@ -257,17 +259,6 @@ object DynamicConformanceJob {
         s"${ds.hdfsPublishPath}/$folderPrefix/$infoDateCol=${cmd.reportDate}/$infoVersionCol=$reportVersion"
       case (Some(publishPathOverride), _) =>
         publishPathOverride
-    }
-  }
-
-  /**
-    * Cleans up standardized data by deleting the standardized directory if it is enabled by the configuration
-    * or by a command line parameter.
-    */
-  private def autocleanStandardizedHdfsPath(pathCfg: PathCfg)
-                                   (implicit cmd: CmdConfig, fsUtils: FileSystemVersionUtils): Unit = {
-    if (isAutocleanStdFolderEnabled()) {
-      fsUtils.deleteDirectoryRecursively(pathCfg.stdPath)
     }
   }
 
--- a/conformance/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
+++ b/conformance/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
@@ -108,6 +108,14 @@ object DynamicConformanceJob {
     enabled
   }
 
+  private def isAutocleanStdFolderEnabled()(implicit cmd: CmdConfig): Boolean = {
+    val enabled = getCmdOrConfigBoolean(cmd.autocleanStandardizedFolder,
+      "conformance.autoclean.standardized.hdfs.folder",
+      defaultValue = false)
+    log.info(s"Autoclean standardized HDFS folder = $enabled")
+    enabled
+  }
+
   /**
     * Returns an effective value of a parameter according to the following priorities:
     * - Command line arguments [highest]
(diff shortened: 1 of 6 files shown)


utils/src/main/scala/za/co/absa/enceladus/utils/fs/FileSystemVersionUtils.scala
```diff
@@ -164,6 +164,15 @@ class FileSystemVersionUtils(conf: Configuration) {
     fs.listStatus(hdfsPath).map(_.getPath.toString)
   }
 
+  /**
+    * Deletes a directory and all its contents recursively
+    */
+  def deleteDirectoryRecursively(path: String): Unit = {
+    log.warn(s"Deleting '$path' recursively...")
```
benedeki: Isn't the warning little late here?
If it's potentially problematic, perhaps it should be announced upon start the user to have soemthing do about it? Or it's just a normal step of the processing... 🤔

conformance/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
```diff
@@ -108,6 +108,14 @@ object DynamicConformanceJob {
     enabled
   }
 
+  private def isAutocleanStdFolderEnabled()(implicit cmd: CmdConfig): Boolean = {
```
benedeki: This is somewhat unfitting place of the function. Why not to make it a method of `CmdConfig`

conformance/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
```diff
@@ -108,6 +108,14 @@ object DynamicConformanceJob {
...
+    val enabled = getCmdOrConfigBoolean(cmd.autocleanStandardizedFolder,
+      "conformance.autoclean.standardized.hdfs.folder",
+      defaultValue = false)
+    log.info(s"Autoclean standardized HDFS folder = $enabled")
```
benedeki: Also I find it weird to log this value upon querying the function. Imagine it would be queried twice for some reason.
IMHO the logging would be better suited in the creation of `CmdConfig` (Actually why not to log all switches there? 🤔

utils/src/main/scala/za/co/absa/enceladus/utils/fs/FileSystemVersionUtils.scala
lokm01: This is more for further log analytics and monitoring in my opinion. Remember the idea is that these are automated, nobody's gonna watch potentially thousands of jobs run every day :)

conformance/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
yruslan: This is done for consistency with `isExperimentalRuleEnabled()` and `isCatalystWorkaroundEnabled()`.
Maybe we can demote these defs to lazy vals as a part of refactoring (some other task), that way we can [potentially] save both query time and logging it twice.

conformance/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
yruslan: It checks both Cmd and Config. Currently, Cmd knows nothing about Config.

utils/src/main/scala/za/co/absa/enceladus/utils/fs/FileSystemVersionUtils.scala
benedeki: In that case, if it's a part of "normal" processing, I would think INFO severity might be more suitable. But not a hard request, just an opinion :-)

conformance/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
benedeki: Ah true.

conformance/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
benedeki: OK, let's have it this way now. In refactoring having a Configuration object - like you have in Migration nad Continuous migration - would be a nice to group the setup into.

utils/src/main/scala/za/co/absa/enceladus/utils/fs/FileSystemVersionUtils.scala
DzMakatun: It also looks to me that INFO level is more appropriate here

GeorgiChochov: I agree, `INFO` seems more appropriate.

GeorgiChochov:
@benedeki, @DzMakatun - please don't approve the PR while still asking for changes
@lokm01 - please don't approve while changes are being requested. It's good that you engaged with the comment, just don't approve until it's been resolved

Issue #817, opened 2019-08-16
File handle leak


benedeki:
## Describe the bug
`Source.fromFile("src/test/resources/json_output").getLines`
Source is not closed 
Inspection info: 
_Using a function like .mkString or . getLines directly on a Source opened by fromFiIe, fromURL, fromURI will leak the underlying file handle._ 
_Consider using . close() with try / finally to manage the resource's lifetime._
 
## To Reproduce
Files:
conformance/src/test/scala/za/co/absa/enceladus/conformance/interpreter/InterpreterSuite.scala
standardization/src/test/scala/za/co/absa/enceladus/standardization/interpreter/SampleDataSuite.scala
standardization/src/test/scala/za/co/absa/enceladus/standardization/interpreter/StandardizationInterpreterSuite.scala
testutils/src/main/scala/za/co/absa/enceladus/testutils/e2eSparkRunner/CmdConfig.scala
testutils/src/test/scala/za/co/absa/enceladus/testutils/datasetComparison/ComparisonJobTest.scala
utils/src/main/scala/za/co/absa/enceladus/utils/testUtils/SparkTestBase.scala

## Expected behaviour
Close the resource.

PR #818, opened 2019-08-16
#817: File handle leak

benedeki:
* new trait providing functionality to read files and ensuring the file handle is closed


utils/src/main/scala/za/co/absa/enceladus/utils/general/FileReader.scala
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  def readFileAsString(filename: String, lineSeparator: String = "\n"): String = {
+    val sourceFile = Source.fromFile(filename)
+    try {
+      sourceFile.getLines().mkString(lineSeparator)
+    } finally {
+      sourceFile.close()
+    }
+    readFileAsLines(filename).mkString(lineSeparator)
```
Zejnilovic: you call `readFileAsLines` here again after reading it in the try?

benedeki: Thanks a lot!
Forgotten there after some copy & paste. (basically means reading the file twice). Will fix,

utils/src/main/scala/za/co/absa/enceladus/utils/general/FileReader.scala
```diff
@@ -0,0 +1,41 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.enceladus.utils.general
+
+import scala.io.Source
+
+trait FileReader {
+  def readFileAsLines(filename: String): Seq[String] = {
```
yruslan: Maybe use `List[String]` as a return type?
The usage like this
```
readFileAsLines("src/test/resources/json_output").toList
```
makes it convert to a List anyway.

utils/src/main/scala/za/co/absa/enceladus/utils/general/FileReader.scala
```diff
@@ -0,0 +1,41 @@
...
+    val sourceFile = Source.fromFile(filename)
+    try {
+      sourceFile.getLines().toList.map(x => x)
+    } finally {
+      sourceFile.close()
+    }
+  }
+
+  def readFileAsString(filename: String, lineSeparator: String = "\n"): String = {
```
yruslan: I feel like these methods belong to FsUtils object rather it being inside a trait. Is there a particular reason for a trait?

benedeki: Moving to `fs` folder (more appropriate) but keeping as separate `trait` as it's lightweight.

utils/src/main/scala/za/co/absa/enceladus/utils/general/FileReader.scala
benedeki: Then perhaps a `Vector`. But `Seq` is just simple and more general class

utils/src/main/scala/za/co/absa/enceladus/utils/fs/FileReader.scala
```diff
@@ -0,0 +1,40 @@
...
+
+package za.co.absa.enceladus.utils.fs
+
+import scala.io.Source
+
+trait FileReader {
+  def readFileAsLines(filename: String): Seq[String] = {
+    val sourceFile = Source.fromFile(filename)
+    try {
+      sourceFile.getLines().toVector.map(x => x) // need to the actual copy before the stream is closed
```
yruslan: ... or `sourceFile.getLines().toList` which will do copying automatically. I thought `Vector` does it as well since it is a very sophisticated collection.

benedeki: You were right. When turning to vector the explicit copy via `map` is not needed anymore. 👍

utils/src/main/scala/za/co/absa/enceladus/utils/fs/FileReader.scala
```diff
@@ -0,0 +1,40 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.enceladus.utils.fs
+
+import scala.io.Source
+
+trait FileReader {
+  def readFileAsLines(filename: String): Seq[String] = {
```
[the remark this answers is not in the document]
benedeki: Hehe, we discussed this with @yruslan on Friday.
`Vector` is an oddly specific type to return from a function such this. `Seq` feels much more natural. But it was necessary to copy the file content to a new structure, and `vector` is generally thought as the more performant (than `list` for example) .

utils/src/main/scala/za/co/absa/enceladus/utils/general/FileReader.scala
yruslan: Still Why `trait` and not an `object`? What are the benefits of requiring anyone who wants to use a function from FsUtils to mixin the trait?

utils/src/main/scala/za/co/absa/enceladus/utils/fs/FileReader.scala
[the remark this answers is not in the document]
yruslan: Vector is faster on accessing elements by index. When processing elements sequentially starting from the beginning, `List` and `Vector` have the same performance profile.
From my personal perspective, I slightly prefer `List` since it is more lightweight. But I'm fine with `Vector` here as well.

As for return type, it might be something to discuss on CQC. We might come up with the rules when to return an interface (e.g. `Seq`), and when to return a concrete type.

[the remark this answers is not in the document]
Zejnilovic: [Vector performance](https://docs.scala-lang.org/overviews/collections/performance-characteristics.html) by scala-lang.org.

I would go for `List` as well, we don't gain much by using `Vector` and I think `List` is expected from `getLines` in general. So keeping it together with scala.

utils/src/main/scala/za/co/absa/enceladus/utils/general/FileReader.scala
benedeki: It crossed my mind originally then i dismissed it. 
Rethinking under the inspiration of the comment, it probably is a good adept for a static object instead of making it a class' "own" function.

utils/src/main/scala/za/co/absa/enceladus/utils/fs/FileReader.scala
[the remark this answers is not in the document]
benedeki: List it is then. I can even rename the function and then go with `List` as return type...

Issue #788, opened 2019-08-07
Conformance unit tests fail nondeterministically


yruslan:
## Describe the bug

A unit test in Conformance fails non-deterministically.

```
[2019-08-07T12:02:50.585Z] - End to end array dynamic conformance test (experimental optimized mapping rule) *** FAILED ***

[2019-08-07T12:02:50.585Z]   "{"id":1,"[errCol":[],"legs":{"conditions":{"checks":[{"checkNums":["1","2","3b","4","5c","6"]}],"country":"CZE","currency":"Kc","product":"Stock","conformed_country":"Czech Republic","conformed_currency":"CZK","conformed_product":"STK","conformed_product_new":"STK2"},"legid":100,"ConformedLegId":{"legId":100}},"MyLiteral":"abcdef","MyUpperLiteral":"ABCDEF","legs_id":0,"legs_size":1,"legs_idx":0,"legs_conditions_id":0,"legs_conditions_size":1,"legs_conditions_idx":0,"Concatenated":"abcdefABCDEF"}

[2019-08-07T12:02:50.585Z]   {"id":2,"errCol":[],"legs":{"conditions":{"checks":[{"checkNums":["8","9","10b","11","12c","13"]}],"country":"SA","currency":"Rand","product":"Stock","conformed_country":"South Africa","conformed_currency":"ZAR","conformed_product":"STK","conformed_product_new":"STK2"},"legid":200,"ConformedLegId":{"legId":200}},"MyLiteral":"abcdef","MyUpperLiteral":"ABCDEF","legs_id":8589934592,"legs_size":1,"legs_idx":0,"legs_conditions_id":8589934592,"legs_conditions_size":1,"legs_conditions_idx":0,"Concatenated":"abcdefABCDEF"}

[...]

PR #819, opened 2019-08-16
#788 Fix experimental mapping rule non-deterministic bug

yruslan:
This is caused by wrong explodeContext being used when a Catalyst
workaround is applied. This led to an additional array explosion
inside an explosion context, which was not actually applied.
That lead to actual explosions not being collapsed.

### DIFF
--- a/conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/DynamicInterpreter.scala
+++ b/conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/DynamicInterpreter.scala
@@ -84,19 +84,19 @@ object DynamicInterpreter {
     val conformedDf = steps.foldLeft(dfInputWithIdForWorkaround)({
       case (df, rule) =>
         val (ruleAppliedDf, ec) = applyConformanceRule(df, rule, explodeContext)
-        explodeContext = ec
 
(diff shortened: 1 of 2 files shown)


conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/DynamicInterpreter.scala
```diff
@@ -84,19 +84,19 @@ object DynamicInterpreter {
     val conformedDf = steps.foldLeft(dfInputWithIdForWorkaround)({
       case (df, rule) =>
         val (ruleAppliedDf, ec) = applyConformanceRule(df, rule, explodeContext)
-        explodeContext = ec
 
         val conformedDf = if (explodeContext.explosions.isEmpty &&
           optimizerTimeTracker.isCatalystWorkaroundRequired(ruleAppliedDf, rulesApplied)) {
           // Apply a workaround BEFORE applying the rule so that the execution plan generation still runs fast
-          val (workAroundDf, ec) = applyConformanceRule(
+          val (workAroundDf, ecInner) = applyConformanceRule(
```
benedeki: Wow, for figuring this out! 😎

yruslan: Thanks!

Issue #826, opened 2019-08-22
Add an environment name in the page title


yruslan:
## Background
When Menas is deployed to different environments it is hard to track which hostname corresponds to each environment, e.g. dev, uat, pre-prod, prod.

## Feature
It could be helpful if an environment name can be configured for each environment and such a name is displayed on the log in page and in the page title.

## Example [Optional]
These are just ideas. It is up to the implemeter to decide the actual placement of an environment name. 
`Login screen`
![Screenshot 2019-08-22 at 10 54 14](https://user-images.githubusercontent.com/4082463/63502549-166c1580-c4ce-11e9-8718-9d8a162a2516.png)

`Page Title`
![Screenshot 2019-08-22 at 10 59 37](https://user-images.githubusercontent.com/4082463/63502556-19670600-c4ce-11e9-819f-ff97408b2d09.png)

Issue #510, opened 2019-06-20
Make menas properties configurable at runtime


Zejnilovic:
## Background
Currently, you need to build menas each time you want to use it, so it always builds for the correct URLs. This is a hassle and should be switched for property configuration on the fly.

## Feature
Make menas properties configurable at runtime

PR #828, opened 2019-08-22
#826 Add environment name to login and home page

Zejnilovic:
With this PR and with advice from @GeorgiChochov (BTW many thanks!!) I created a controller `ConfigurationController`. This controller could, later on, hold more than just the environment name, but we could even move `menasVersion` to it from `UserInfoController`.  `ConfigurationController` is for the API calls that do not require authentication.

### DIFF
--- a/menas/src/main/resources/application.properties.template
+++ b/menas/src/main/resources/application.properties.template
@@ -60,6 +60,7 @@ za.co.absa.enceladus.menas.mongo.connection.database=menas
 za.co.absa.enceladus.migration.mongo.query.timeout.seconds=300
 
 za.co.absa.enceladus.menas.version=@project.version@
+za.co.absa.enceladus.menas.environment=localhost
 
 za.co.absa.enceladus.menas.spark.master=local[1]
 
--- a/menas/src/main/scala/za/co/absa/enceladus/menas/controllers/ConfigurationController.scala
+++ b/menas/src/main/scala/za/co/absa/enceladus/menas/controllers/ConfigurationController.scala
@@ -1,0 +1,31 @@
+/*
+ * Copyright 2018-2019 ABSA Group Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.enceladus.menas.controllers
+
+import org.springframework.beans.factory.annotation.Value
+import org.springframework.web.bind.annotation.{GetMapping, RequestMapping, RestController}
+
+@RestController
+@RequestMapping(Array("/api/configuration"))
+class ConfigurationController extends BaseController {
+  @Value("${za.co.absa.enceladus.menas.environment}")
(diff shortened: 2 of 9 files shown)


yruslan:
Looks great!

menas/ui/components/login/loginDetail.controller.js
```diff
@@ -45,6 +45,10 @@ sap.ui.define([
         this.handleMaster();
       }, this);
 
+      ConfigRestClient.getEnvironmentName()
+        .then( sEnvironmentName => sap.ui.getCore().getModel().setProperty("/menasEnvironment", sEnvironmentName) )
+        .fail(console.log("Failed to get Environment variable"));
```
GeorgiChochov:
```suggestion
        .fail(console.log("Failed to get Environment name"));
```
lol, same as in the other controller

Zejnilovic: Well, I forgot I have two of those 😑

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/ConfigurationController.scala
```diff
@@ -0,0 +1,31 @@
...
+
+package za.co.absa.enceladus.menas.controllers
+
+import org.springframework.beans.factory.annotation.Value
+import org.springframework.web.bind.annotation.{GetMapping, RequestMapping, RestController}
+
+@RestController
+@RequestMapping(Array("/api/configuration"))
+class ConfigurationController extends BaseController {
+  @Value("${za.co.absa.enceladus.menas.environment}")
```
lokm01: Is this a required parameter?

Zejnilovic: Yes, build should fail without it.

menas/src/main/resources/application.properties.template
```diff
@@ -60,6 +60,7 @@ za.co.absa.enceladus.menas.mongo.connection.database=menas
 za.co.absa.enceladus.migration.mongo.query.timeout.seconds=300
 
 za.co.absa.enceladus.menas.version=@project.version@
+za.co.absa.enceladus.menas.environment=localhost
```
benedeki: I am thinking how much it makes sense to have this in the properties file. 🤔

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/ConfigurationController.scala
benedeki: This seems little constricting. (In conjunction with my previous comment, if `properties` file is the proper location for this info.)

menas/src/main/resources/application.properties.template
Zejnilovic: It does not. And it does not for many more things. I hope this issue will solve it #510
I am not going against the flow for a small issue.

Zejnilovic:
> I really like the idea to show the environment name.
> But isn't gathering the info via a separate request against our effort to speed up the start up of Menas? Couldn't it be grouped together with other requests? The one gathering the version comes to mind right away. 🤔

Yes, when we move other things to the ConfigurationController we will make one big call. But not a scope of this issue as I do not know what are all the things needed to move. Here is a ticket #830

benedeki:
> 
> 
> > I really like the idea to show the environment name.
> > But isn't gathering the info via a separate request against our effort to speed up the start up of Menas? Couldn't it be grouped together with other requests? The one gathering the version comes to mind right away. 🤔
> 
> Yes, when we move other things to the ConfigurationController we will make one big call. But not a scope of this issue as I do not know what are all the things needed to move. Here is a ticket #830
> 
> I can use cache for the environment name.

OK, then

hamc17:
Looks like this is the only PR failing after the Jenkins fix, please take a look at the logs and review!

Issue #796, opened 2019-08-08
Upgrade to latest version of the mongo-scala driver


GeorgiChochov:
## Background
Currently, the mongo scala driver does not support polymorphic serDe of classes extending abstract types. We have a work around with a custom mongo-scala driver version, while a PR  is under review - [mongo-scala-driver:49](https://github.com/mongodb/mongo-scala-driver/pull/49).

## Feature
When the PR is merged and a version containing this feature is released (i.e., 2.7) we should switch over to the official driver.

PR #829, opened 2019-08-23
#796 Upgrade mongo-scala-driver to 2.7.0

GeorgiChochov:


menas/pom.xml
```diff
@@ -109,11 +109,6 @@
             <artifactId>mongo-scala-driver_${scala.compat.version}</artifactId>
             <version>${mongo.scala.driver.version}</version>
         </dependency>
-        <dependency>
```
yruslan: The same dependency is in `migrations`. Maybe remove it from these as well?

GeorgiChochov: Smart! I might have merged without doing this if the reviews didn't go stale.

Issue #770, opened 2019-08-05
Unsuccessful casting rule to timestamp or date quits jobs with non zero exit


Zejnilovic:
## Describe the bug
If casting rule for timestamp or date fail, spark job quits with non zero exit immediately
If other casting rules fail, they are written to errCol and the job continues

## Expected behaviour
All casting rules should mainly write to errCol and not quite jobs in the middle.

yruslan:
It happens when converting from `String` to `Timestamp`, right?

Zejnilovic:
> It happens when converting from `String` to `Timestamp`, right?

Any type that fails. String, int, boolean, etc.

yruslan:
Should we support casting to `timestamp` at all? Since it requires many a timezone and a pattern. And this conversion should be done in Standardization.

How about adding a validation to prevent casting to `timestamp` and `date`? 
If we have a use case for such a conversion - we can add it in the future.

benedeki:
I definitely see a use-case for that.  E.g. two string columns one with date one with  time to make the timestamp.

yruslan:
Cool. This task is about making the rule not fail the job, the full support for timestamps and dates are described in #555

yruslan:
@Zejnilovic , I couldn't reproduce the issue in a unit test. Could you please help me reproduce this.

Zejnilovic:
@yruslan as requested 

## Conformance Rule 
```json
{
  "_t" : "CastingConformanceRule",
  "order" : 0,
  "outputColumn" : "BoolToDate",
  "controlCheckpoint" : true,
  "inputColumn" : "Boolean",
  "outputDataType" : "date"
}
```

## StackTrace
```
Exception in thread "main" org.apache.spark.sql.AnalysisException: cannot resolve 'CAST(`Boolean` AS DATE)' due to data type mismatch: cannot cast boolean to date;;
'Project [Boolean#0, Byte#1, Date#2, Decimal#3, Double#4, Float#5, Integer#6, Long#7L, ManipulationBox#8, MappingBox#9, Short#10, String#11, Timestamp#12, errCol#13, tmpId_6968118448880013757#225, array(CASE WHEN (isnotnull(Boolean#0) && isnull(cast(Boolean#0 as date))) THEN UDF:confCastErr(BoolToDate, cast(Boolean#0 as string)) ELSE null END) AS errorList_7596960675371077137#252, cast(Boolean#0 as date) AS BoolToDate#251]
[...]

PR #834, opened 2019-08-26
#770 Casting conformance rule: Add type casting compatibility validation

yruslan:


conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/RuleValidators.scala
```diff
@@ -54,4 +58,42 @@ object RuleValidators {
...
+    * @param inputColumnName A name of an input column to be used as part of an exception error message
+    * @param inputType       A type of an input field
+    * @param outputType      A type to cast to
+    * @param spark           (implicit) A Spark Session
+    */
+  @throws[ValidationException]
+  def validateTypeCompatibility(ruleName: String,
+                                inputColumnName: String,
+                                inputType: DataType,
+                                outputType: String)
```
benedeki:
```suggestion
                                outputTypeName: String)
```
`outputType` is little misleading compared to `inputType'

conformance/src/test/scala/za/co/absa/enceladus/conformance/interpreter/rules/CastingRuleSuite.scala
```diff
@@ -125,4 +127,60 @@ class CastingRuleSuite extends FunSuite with SparkTestBase with LoggerTestBase{
...
+  test("Test Casting rule ok handling conversion from 'string' to 'date'") {
+    RuleValidators.validateTypeCompatibility(ruleName, columnName, StringType, "date")
+  }
+
+  test("Test Casting rule ok handling conversion from 'boolean' to 'decimal(10,2)'") {
+    RuleValidators.validateTypeCompatibility(ruleName, columnName, BooleanType, "decimal(10,2)")
+  }
+
+  test("Test Casting rule ok handling conversion from 'long' to 'timestamp'") {
+    RuleValidators.validateTypeCompatibility(ruleName, columnName, LongType, "timestamp")
```
benedeki: Perhaps a comment explaining how the conversion works.

conformance/src/test/scala/za/co/absa/enceladus/conformance/interpreter/rules/CastingRuleSuite.scala
```diff
@@ -125,4 +127,60 @@ class CastingRuleSuite extends FunSuite with SparkTestBase with LoggerTestBase{
...
+  }
+
+  test("Test Casting rule ok handling conversion from 'timestamp' to 'boolean'") {
+    RuleValidators.validateTypeCompatibility(ruleName, columnName, TimestampType, "boolean")
```
benedeki: Wow, this converts how? :-)

conformance/src/test/scala/za/co/absa/enceladus/conformance/interpreter/rules/CastingRuleSuite.scala
```diff
@@ -125,4 +127,60 @@ class CastingRuleSuite extends FunSuite with SparkTestBase with LoggerTestBase{
...
+    assert(e.message.contains("conversion from 'decimal(10,2)' to 'date' is not supported"))
+  }
+
+  test("Test Casting rule failure handling conversion from 'boolean' to 'timestamp'") {
+    val e = intercept[ValidationException] {
+      RuleValidators.validateTypeCompatibility(ruleName, columnName, BooleanType, "timestamp")
+    }
+    assert(e.message.contains("conversion from 'boolean' to 'timestamp' is not supported"))
+  }
+
```
benedeki: Perhaps some `Double` to `Timestamp`, `Date` and `Long` and visa versa conversion tests.

conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/RuleValidators.scala
```diff
@@ -54,4 +58,42 @@ object RuleValidators {
...
+
+    // This generates a dataframe we can use to check cast()
+    // Various values need to be here, otherwise Spark may accept incompatible types
+    val df = List("", "true", "false", "0", "1", "-999", "2.2", "a",
+      "2019-01-01", "2019-01-01 00:00:00", "1970-01-01 00:00:00").toDF("dummy")
+
+    // Check if inputType can be cast to outputType
+    // If types are completely incompatible Spark throws an exception
+    try {
+      // The initial type is 'string'. Convert it to 'inputType' first (conversion from string never throws), and than
```
benedeki: Note to myself - conversion from `String` never throws. Interesting. I guess it then returns `null` in the column value,
Cool trick 😎

conformance/src/test/scala/za/co/absa/enceladus/conformance/interpreter/rules/CastingRuleSuite.scala
yruslan: The beginning of the epoch is `false`, otherwise `true`. Don't ask why 😄

conformance/src/test/scala/za/co/absa/enceladus/conformance/interpreter/rules/CastingRuleSuite.scala
yruslan: Ok, will add

conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/RuleValidators.scala
yruslan: Yes, exactly

conformance/src/test/scala/za/co/absa/enceladus/conformance/interpreter/rules/CastingRuleSuite.scala
yruslan: Do you mean to explain that long is expected to represent seconds since epoch?

yruslan: Added

benedeki: Yup, thank you. 👍

yruslan:
1. Logging REST request/responses is helpful for diagnosing prod issues. What's your concern?

2. Yeah, this validation might take a couple of seconds since it constructs a dataframe. The alternative maybe is to manually compare type pairs against ones that are allowed. Might be 1. more complicated 2. if Spark implementation changes type conversion compatibility it might require changes in our code as well. On the other hand, we might be more strict than Spark and always deny strange conversions such as 'boolean to timestamp'.

Zejnilovic:
> 1. Logging REST request/responses is helpful for diagnosing prod issues. What's your concern?

As with all the REST request/responses. Too much data in the terminal for standard run
 
> On the other hand, we might be more strict than Spark and always deny strange conversions such as 'boolean to timestamp'.

I would be for it. Being more strict then spark. Gives us more security. Doesn't it?

yruslan:
Both approaches have pros and cons. The explicit type conversion white list might be a little cumbersome. Personally, like the current approach as it delegates all boring work to Spark. 

I've marked this PR as `work in progress`? But could you please change your review to `changes requested` so that it won't be accidentally merged.

Issue #835, opened 2019-08-26
Comparison job doesn't read in XML


Zejnilovic:
## Describe the bug
test utils are missing spark-xml in pom.xml

PR #836, opened 2019-08-26
#835 Add spark-xml dependency to testUtils

Zejnilovic:


Zejnilovic:
> Is this needed for something particular in testUtils?

ComparisonJob of XML files

PR #837, opened 2019-08-27
573 Fix error message for bad entity names

Zejnilovic:
Tested on Datasets creation and Mapping Table creation. 
Schemas not properly tested because there is a Hanging PR (#832)

### DIFF
--- a/menas/ui/service/EntityValidationService.js
+++ b/menas/ui/service/EntityValidationService.js
@@ -52,7 +52,7 @@ var EntityValidationService = new function () {
     } else if (GenericService.hasWhitespace(oEntity.name)) {
       oInput.setValueState(sap.ui.core.ValueState.Error);
       oInput.setValueStateText(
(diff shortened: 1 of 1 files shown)


menas/ui/service/EntityValidationService.js
```diff
@@ -52,7 +52,7 @@ var EntityValidationService = new function () {
     } else if (GenericService.hasWhitespace(oEntity.name)) {
       oInput.setValueState(sap.ui.core.ValueState.Error);
       oInput.setValueStateText(
-        sEntityType + " name '" + oEntity.name + "' should not have whitespaces. Please remove spaces and retry");
+        sEntityType + " name can only contain alphanumeric characters and underscores");
```
benedeki: Why removing the info of the offending entry (`oEntity.name`)?

Zejnilovic: Because it tells you nothing. It is what you have written in the field above it.
and
This is the same solution as in Conformance Rules

Issue #838, opened 2019-08-27
Dataframe comparison tool fails on specific data


yruslan:
## Describe the bug
The dataframe comparison tool fails on a specific dataframe. The data source for that dataframe is `xml`, not sure if it is related to the failure.

```
Exception in thread "main" org.apache.spark.sql.AnalysisException: cannot resolve
 'concat(CASE WHEN (`actual_field` = `expected_field`) THEN [] ELSE array('_field') END, `tmp`)' 
due to data type mismatch: argument 1 requires string type, however,
 'CASE WHEN (`actual_field` = `expected_field`) THEN [] ELSE array('_field') END'
 is of array<string> type. argument 2 requires string type, however,
 '`tmp`' is of array<string> type.;;
```

## To Reproduce
Run the dataframe comparison tool with the following command line parameters:
```
spark-submit \
--class za.co.absa.enceladus.testutils.datasetComparison.ComparisonJob \
--master yarn \
--deploy-mode client \
--executor-memory 12g \
--name compare \
--conf "spark.app.id=compare" \
enceladus-testUtils-1.1.0-SNAPSHOT.jar \
--raw-format xml \
--row-tag account \
--new-path a.xml \
--ref-path b.xml \
--out-path c \
--keys x, y, z.x
```

## Expected behaviour
The dataframe comparison tool should handle schemas of any complexity.

yruslan:
It turned out this is not an issue. But still, a unit test for it might be nice to have.

PR #840, opened 2019-08-28
#838 Add a TestUtils unit test for XML comparison.

yruslan:


testutils/src/test/resources/xml_examples/example12_diff.txt
```diff
@@ -0,0 +1,10 @@
...
+|1.0.0                  |[a, b2]      |1          |2           |BBB          |11.2          |1.0.0                |[a, b2]    |1        |2         |BBB        |11.1        |[value]                                                     |
+|1.0.0                  |[a, b4]      |2          |1           |DDD          |1000000.1     |1.0.0                |[a, b4]    |2        |1         |DDD1       |1000000.1   |[name]                                                      |
+|1.0.0                  |[a, b5]      |2          |2           |EEE          |-9123.0       |1.0.0                |[a, b5]    |2        |2         |EEE2       |-9123.0     |[name]                                                      |
+|1.0.0                  |[a, b9]      |3          |1           |GGG          |812.0         |null                 |null       |null     |null      |null       |null        |[value, name, id2, id, ckey_value, ckey_key, _schemaVersion]|
++-----------------------+-------------+-----------+------------+-------------+--------------+---------------------+-----------+---------+----------+-----------+------------+------------------------------------------------------------+
```
lokm01: Just wondering... would it be safer to use Datasets/ case classes for the actual comparison?

This seems like it could be fragile should the format of this change in diff versions of spark etc.

yruslan: Yeah, it is fragile to `.show()` formatting. Case classes might be better. But now that you've mentioned it, I'm thinking about using conversion to JSON for comparison. It is format independent and the code won't be bloated with additional case classes. What do you think?

lokm01: Definitely better than show :)

Strongly typed way would probably be my preference, but json is fine by me, will be easier to read diff.

Issue #777, opened 2019-08-06
Report copybook syntax errors when uploading a schema


yruslan:
## Background
Sometimes it happens that copybooks provided by our users do not pass syntax check. This can happen, for example, if lines are not properly spaced.

## Feature
Cobrix generates a detailed syntax error explanation as a message of the syntax error exception. It would be nice if Menas could display that message on schema upload if the provided copybook generates a syntax error exception.

Otherwise, in order to determine what's wrong with the copybook one needs to write a small App and use Cobrix as a library.

yruslan:
The new error message box will look like this if there are no objections:
![Screenshot 2019-08-28 at 13 21 02](https://user-images.githubusercontent.com/4082463/63851788-1caf3580-c997-11e9-9685-b47191674391.png)

PR #841, opened 2019-08-28
#777 Show detailed error message on failure to parse a schema.

yruslan:


menas/ui/components/schema/schemaDetail.controller.js
```diff
@@ -172,8 +172,11 @@ sap.ui.define([
 
       if (status === 500) {
         const sSchemaType = this.byId("schemaFormatSelect").getSelectedItem().getText();
+        let oRawResponse = JSON.parse(oParams.getParameter("responseRaw"));
+        const errorMessage = oRawResponse.message;
```
benedeki: Will `oRawResponse` always defined and guaranteed to have `message` parameter?

benedeki: Why `const`?

yruslan: To this API call, yes.

lokm01: Var vs Val :)

I'd say oRawResponse should be const too...

benedeki: OK, it seemed little against our current style, when basically everywhere we use `let`. Also especially for a local variable in JS, it seems uncommon, but if it's the new way, I have no issues going with it 😉

benedeki: I dunno. Especially for status 500, to rely that the server will return an expected format of response seems little unsafe. 🤔

GeorgiChochov: It's true it shouldn't be returning `500 Internal Server Error`. We have to catch and handle the error and return `400 Bad Request`

GeorgiChochov: `const` is completely valid in a local scope. And the reason our style so far has been mostly `let` is because I was slow to adopt `const`.

GeorgiChochov: 1. Let's create a task to change the response code.
2. On `500` let's have an additional check here that the response is valid JSON that contains a `message` field.

menas/ui/components/schema/schemaDetail.controller.js
```diff
@@ -173,7 +173,16 @@ sap.ui.define([
       if (status === 500) {
         const sSchemaType = this.byId("schemaFormatSelect").getSelectedItem().getText();
 
-        MessageBox.error(`Failed to upload new schema. Ensure that the file is a valid ${sSchemaType} schema and try again.`)
+        const oRawResponse = JSON.parse(oParams.getParameter("responseRaw"));
```
GeorgiChochov: Maybe this needs to be wrapped in a `try/catch`. Pretty sure that it would toss an exception if the response is not JSON.

GeorgiChochov: The default 500 response from tomcat is html

menas/ui/components/schema/schemaDetail.controller.js
```diff
@@ -173,7 +173,16 @@ sap.ui.define([
...
         const sSchemaType = this.byId("schemaFormatSelect").getSelectedItem().getText();
 
-        MessageBox.error(`Failed to upload new schema. Ensure that the file is a valid ${sSchemaType} schema and try again.`)
+        let errorMessageDetails = "";
+        try {
+          const oRawResponse = JSON.parse(oParams.getParameter("responseRaw"));
+          if (oRawResponse.message) {
+            errorMessageDetails = `\n\nDetails:\n${oRawResponse.message}`;
+          }
+        } catch (unused) {}
```
GeorgiChochov: Maybe logging the error would be helpful here: `console.error()`

yruslan: 👍  Thanks for going through tese steps! Good to know what is the common pattern for such caes.

yruslan: Added logging. Simulated an HTML response - worked well.

PR #843, opened 2019-08-28
427 Add a warning message if column to drop does not exist

Zejnilovic:
- added a check for the presence of a column to be dropped
- fixed previous UTs of drop rule

### DIFF
--- a/conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/rules/DropRuleInterpreter.scala
+++ b/conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/rules/DropRuleInterpreter.scala
@@ -19,21 +19,20 @@ import org.apache.spark.sql.{Dataset, Row, SparkSession}
 import za.co.absa.enceladus.conformance.CmdConfig
 import za.co.absa.enceladus.dao.EnceladusDAO
 import za.co.absa.enceladus.model.conformanceRule.DropConformanceRule
+import za.co.absa.enceladus.utils.schema.SchemaUtils
 import za.co.absa.enceladus.utils.transformations.DeepArrayTransformations
-
-import scala.util.{Failure, Success, Try}
 
 case class DropRuleInterpreter(rule: DropConformanceRule) extends RuleInterpreter {
 
   def conform(df: Dataset[Row])(implicit spark: SparkSession, dao: EnceladusDAO, progArgs: CmdConfig): Dataset[Row] = {
-    Try(df(rule.outputColumn)) match {
-      case Success(_) =>
+    SchemaUtils.getFieldType(rule.outputColumn, df.schema) match {
+      case Some(_) =>
         if (rule.outputColumn.contains('.')) {
           conformNestedField(df)
         } else {
           conformRootField(df)
         }
-      case Failure(_) =>
+      case None =>
         log.warn(s"Could not drop ${rule.outputColumn}. Column is not present in the dataset")
         df
     }
--- a/conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/rules/DropRuleInterpreter.scala
+++ b/conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/rules/DropRuleInterpreter.scala
@@ -15,21 +15,27 @@
 
 package za.co.absa.enceladus.conformance.interpreter.rules
 
-import za.co.absa.enceladus.model.conformanceRule.DropConformanceRule
-import org.apache.spark.sql.Dataset
-import org.apache.spark.sql.Row
-import org.apache.spark.sql.SparkSession
-import za.co.absa.enceladus.dao.EnceladusDAO
+import org.apache.spark.sql.{Dataset, Row, SparkSession}
 import za.co.absa.enceladus.conformance.CmdConfig
-import za.co.absa.enceladus.utils.transformations.{ArrayTransformations, DeepArrayTransformations}
+import za.co.absa.enceladus.dao.EnceladusDAO
+import za.co.absa.enceladus.model.conformanceRule.DropConformanceRule
+import za.co.absa.enceladus.utils.transformations.DeepArrayTransformations
+
+import scala.util.{Failure, Success, Try}
 
 case class DropRuleInterpreter(rule: DropConformanceRule) extends RuleInterpreter {
 
   def conform(df: Dataset[Row])(implicit spark: SparkSession, dao: EnceladusDAO, progArgs: CmdConfig): Dataset[Row] = {
-    if (rule.outputColumn.contains('.')) {
-      conformNestedField(df)
-    } else {
-      conformRootField(df)
+    Try(df(rule.outputColumn)) match {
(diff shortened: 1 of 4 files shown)


conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/rules/DropRuleInterpreter.scala
```diff
@@ -15,21 +15,27 @@
...
+import scala.util.{Failure, Success, Try}
 
 case class DropRuleInterpreter(rule: DropConformanceRule) extends RuleInterpreter {
 
   def conform(df: Dataset[Row])(implicit spark: SparkSession, dao: EnceladusDAO, progArgs: CmdConfig): Dataset[Row] = {
-    if (rule.outputColumn.contains('.')) {
-      conformNestedField(df)
-    } else {
-      conformRootField(df)
+    Try(df(rule.outputColumn)) match {
```
lokm01: It would probably better to inspect the schema.

See something like:
SchemaUtils.getFieldType(path: String, schema: StructType): Option[DataType]

Zejnilovic: Did not know we had something for it. Like this?

lokm01: Yeah seems like a better approach to me

GeorgiChochov: I'm not sure I get what the benefit here is. @lokm01 ?

GeorgiChochov: Thanks, @lokm01 explained that it spares the overhead of throwing an exception and more importantly, catalyst optimizations can take place when checking if the field exists proactively.

conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/rules/DropRuleInterpreter.scala
```diff
@@ -15,21 +15,26 @@
...
+import za.co.absa.enceladus.utils.transformations.DeepArrayTransformations
 
 case class DropRuleInterpreter(rule: DropConformanceRule) extends RuleInterpreter {
 
   def conform(df: Dataset[Row])(implicit spark: SparkSession, dao: EnceladusDAO, progArgs: CmdConfig): Dataset[Row] = {
-    if (rule.outputColumn.contains('.')) {
-      conformNestedField(df)
-    } else {
-      conformRootField(df)
+    SchemaUtils.getFieldType(rule.outputColumn, df.schema) match {
```
GeorgiChochov: I thought there might already be a function to check if the path exists, but can't find it. It might be worth extracting an `exists` function.

Zejnilovic: Will be done by 17:00 today.

Zejnilovic: I am not sure if this is what you wanted. We basically keep everything the same. But it is true you don't have to convert to boolean in the end

conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/rules/DropRuleInterpreter.scala
```diff
@@ -15,21 +15,25 @@
...
-      conformNestedField(df)
+    if (SchemaUtils.fieldExists(rule.outputColumn, df.schema)) {
+      if (rule.outputColumn.contains('.')) {
+        conformNestedField(df)
+      } else {
+        conformRootField(df)
+      }
     } else {
-      conformRootField(df)
+      log.warn(s"Could not drop ${rule.outputColumn}. Column is not present in the dataset")
```
yruslan: Sorry for the late comment. Didn't notice when looked at it the first time. Are you ok with the warning here? Should it be a validation exception? True, that if a column does not exists we might ignore the 'drop'.  But shouldn't we disallow it anyway? No strong opinion on this, just a suggestion.

Found `validateFieldExistence(...)` method used for similar checks for other conformance rules. Of course, there the existence of an input field is critical.

utils/src/main/scala/za/co/absa/enceladus/utils/schema/SchemaUtils.scala
```diff
@@ -50,6 +50,27 @@ object SchemaUtils {
...
+  def fieldExists(path: String, schema: StructType): Boolean = {
+    def existsHelper(pt: List[String], st: DataType): Boolean = {
+      if (pt.isEmpty) {
+        true
+      } else {
+        st match {
+          case str: StructType => Try {
+            existsHelper(pt.tail, str(pt.head).dataType)
+          }.getOrElse(false)
+          case ArrayType(el: StructType, _) => Try {
```
yruslan: I'm pretty sure this implementation won't work on arrays of arrays or arrays of primitives. It might return `false` if a field actually exists. This is because the implementation relies on the element type of an array to be a StructType.

yruslan: Might be better to have a separate issue/PR to fix this here and in the above `getFieldType ()` method.

Zejnilovic: image

I have an array full of structs and then struct with array of arrays of primitives. I would say it works ok. Or did I miss your concern?

conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/rules/DropRuleInterpreter.scala
Zejnilovic: I have thought about it and came to a conclusion to follow spark api for now. Spark api does not throw anything in case of dropping column that does not exist. 

I am more than happy to change it if needed/wanted.

utils/src/main/scala/za/co/absa/enceladus/utils/schema/SchemaUtils.scala
benedeki: Yes. `Array` of `Array` of `Struct`
The function how they were build don't merge into `Array` unless it's `Array` of `Struct`
I needed some changes in the `SchemaUitls` myself. After @yruslan  pointed this issue out, I enhanced my changes So now it can handle nested `Arrays`. And I also added `fieldExist` function and copied your tests. It's all part of  PR #859.
Sorry for messing a little into your code...

Zejnilovic:
I am not going to overwrite public history by force pushing, so if @GeorgiChochov and @yruslan are happy, let me merge. As to the better version of `fieldExists` good job. Just overwrite it later.

Issue #842, opened 2019-08-28
Migtations module fails to build


GeorgiChochov:
## Describe the bug
After merging #829 and #778 the build failed on Jenkins with a test in the `migrations` module. This seems to be a mongo-scala-driver serialization issue caused by the upgrade of the driver version.

## To Reproduce
Run `mvn clean package`

## Expected behaviour
The build should pass.

## Error message
```
- Test schema migrates properly and conflicts are resolved *** FAILED ***
  ContinuousMigrationIntegrationSuite.this.datasetExists("TestDataset", 2, "TestSchema1", 2, "TestMT1", 2) was false (ContinuousMigrationIntegrationSuite.scala:63)
```

## Desktop:
 - Version 1.1.0-SNAPSHOT

Zejnilovic:
in `src/main/scala/za/co/absa/enceladus/migrations/continuous/migrate01/DatasetMigrator.scala`
We have a bit of a workaround for serialization:
```scala
      val fixJson = srcDatasetJson.replaceAll("\"jsonClass\" :", "\"_t\" :")
```
for the new mongo we just need to delete the spaces as now the JSONs in string are represented with `"key": "value"` instead of `"key" :"value"`
```scala
      val fixJson = srcDatasetJson.replaceAll("\"jsonClass\":", "\"_t\":")
```

I was trying to solve my failing Jenkins build and ended here. I also found #844 which could be fixed in the same PR.

This is not a full solution imho. We are depending on something unstable.

GeorgiChochov:
Thanks @Zejnilovic

GeorgiChochov:
We have a better regex option, thanks @Zejnilovic

PR #846, opened 2019-08-29
#842 Fix type hint replacement issue in Migrations

GeorgiChochov:
Thanks for the help tracking it down @Zejnilovic @yruslan and not letting me push a hacky fix.

### DIFF
--- a/migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/migrate01/DatasetMigrator.scala
+++ b/migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/migrate01/DatasetMigrator.scala
@@ -52,7 +52,7 @@ final class DatasetMigrator(evm: EntityVersionMap,
     */
   def migrateEntity(srcDatasetJson: String, objectId: String, repo: EntityRepository): Unit = {
     Try {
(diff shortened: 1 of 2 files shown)


migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/migrate01/DatasetMigrator.scala
```diff
@@ -52,7 +52,7 @@ final class DatasetMigrator(evm: EntityVersionMap,
     */
   def migrateEntity(srcDatasetJson: String, objectId: String, repo: EntityRepository): Unit = {
     Try {
-      val fixJson = srcDatasetJson.replaceAll("\"jsonClass\" :", "\"_t\" :")
+      val fixJson = srcDatasetJson.replaceAll("\"jsonClass\"\\w*:", "\"_t\" :")
```
Zejnilovic: small `w` is `[a-zA-Z0-9_]` which should never occur between the `"jsonClass"` key and `:`

GeorgiChochov: lol, it's what I get for being hasty.
Thanks!

PR #853, opened 2019-08-30
#491 Type error when session expires

lokm01:


menas/ui/components/dataset/conformanceRule/SchemaManager.js
```diff
@@ -16,11 +16,13 @@
 class SchemaManager {
 
   static getTransitiveSchemas(schemas, rules) {
-    rules.map(RuleFactory.createRule).forEach((rule, index) => {
-      const schema = $.extend(true, [], schemas[index]);
-      rule.apply(schema.fields);
-      schemas.push(schema);
-    })
+    if(rules) {
```
Zejnilovic: I would use the return early pattern. This is JS and JS doesn't mind using `if (!rules) { return }` and then next code

Issue #798, opened 2019-08-09
Default value of a field is accepted only when it's a string


benedeki:
## Describe the bug
In Standardization default value of a field is accepted only when it's a string (or null). Even for fields of different type it has to be entered as string.

## To Reproduce
1. Defined and Integer field
2. In it's Metadata defined "default": 1
3. See error

Same with Boolean, float etc.

## Expected behaviour
Values of appropriate type are accepted

benedeki:
It has been agreed in the team not to support non-string metadata values and instead fix and enhance checks and validation of the input data, which would include verification they are in the form of `Map(String, String)`.

Issue #799, opened 2019-08-10
Uploading schema with default value having integer breaks the schema


Zejnilovic:
## Describe the bug
Having a schema with a value that has a default of an int `{ "default" : 1 }` tells the user something went wrong and that it is not a valid spark schema. On the next refresh of the page, there is a new version which is broken and won't allow the user to even open the schema to get to the audit trails.

## To Reproduce
Steps to reproduce the behavior OR commands run:
1. Have schema with an integer default value
2. Upload schema
3. Get error
4. Refresh
5. Try to go to the schema uploaded.

## Expected behaviour
Either fail or don't. When the upload fails, rollback everything.

Zejnilovic:
Part of the schema from mongo

```json
"userUpdated" : "user",
    "disabled" : false,
    "dateDisabled" : null,
    "userDisabled" : null,
    "fields" : [ 
        {
            "name" : "Boolean",
            "type" : "boolean",
            "path" : "",
            "elementType" : null,
            "containsNull" : null,
            "nullable" : true,
            "metadata" : {
                "default" : 1
            },
            "children" : []
        }, 
        {
            "name" : "Byte",
            "type" : "byte",
            "path" : "",
            "elementType" : null,
            "containsNull" : null,
            "nullable" : true,
            "metadata" : {},
            "children" : []
        }, 
        {
            "name" : "Date",
            "type" : "date",
            "path" : "",
            "elementType" : null,
            "containsNull" : null,
            "nullable" : true,
            "metadata" : {
                "pattern" : "yyyy-MM-dd"
            },
            "children" : []
        }, 
        {
            "name" : "Decimal",
            "type" : "decimal(38,18)",
            "path" : "",
            "elementType" : null,
            "containsNull" : null,
            "nullable" : true,
            "metadata" : {},
            "children" : []
        }, 
```

benedeki:
This actually breaks for any non-string metadata

PR #859, opened 2019-09-01
#798: Default value of a field is accepted only when it's a string

benedeki:
* Using recover to try read default value in appropriate type
* Added UTs

### DIFF
--- a/utils/src/main/scala/za/co/absa/enceladus/utils/types/TypedStructField.scala
+++ b/utils/src/main/scala/za/co/absa/enceladus/utils/types/TypedStructField.scala
@@ -22,7 +22,6 @@ import za.co.absa.enceladus.utils.time.DateTimePattern
 import za.co.absa.enceladus.utils.types.parsers._
 import za.co.absa.enceladus.utils.validation.ValidationIssue
 import za.co.absa.enceladus.utils.validation.field.{DateFieldValidator, DateTimeFieldValidator, FieldValidator, ScalarFieldValidator, TimestampFieldValidator}
-
 import scala.util.{Failure, Success, Try}
 
 sealed abstract class TypedStructField(structField: StructField) extends StructFieldEnhancements(structField) {
@@ -68,10 +67,7 @@ sealed abstract class TypedStructField(structField: StructField) extends StructF
    */
   def ownDefaultValue: Try[Option[Option[Any]]] = {
     if (hasMetadataKey(MetadataKeys.DefaultValue)) {
-      for {
-        defaultValueString <- Try{structField.metadata.getString(MetadataKeys.DefaultValue)}
-        defaultValueTyped <- stringToTyped(defaultValueString)
-      } yield Some(defaultValueTyped)
+      defaultValueInMetadata.map(Some(_))
     } else {
       Success(None)
     }
@@ -113,6 +109,13 @@ sealed abstract class TypedStructField(structField: StructField) extends StructF
     var result = 1
     result = prime * result + (if (structField == null) 0 else structField.hashCode)
     result
+  }
+
+  protected def defaultValueInMetadata: Try[Option[Any]] = {
+    for {
+      defaultValueString <- Try{structField.metadata.getString(MetadataKeys.DefaultValue)}
+      defaultValueTyped <- stringToTyped(defaultValueString)
+    } yield defaultValueTyped
   }
 }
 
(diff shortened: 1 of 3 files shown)


utils/src/main/scala/za/co/absa/enceladus/utils/types/TypedStructField.scala
```diff
@@ -212,6 +227,14 @@ object TypedStructField {
     override def validate(): Seq[ValidationIssue] = {
       ScalarFieldValidator.validate(this)
     }
+
+    override protected def defaultValueInMetadata: Try[Option[Any]] = {
+      super.defaultValueInMetadata.recover {
+        case _: ClassCastException => Some(structField.metadata.getDouble(MetadataKeys.DefaultValue))
+      }.recover {
+        case _: ClassCastException => Some(structField.metadata.getLong(MetadataKeys.DefaultValue))
```
Zejnilovic: Why is there a long here? Even on line 258

utils/src/main/scala/za/co/absa/enceladus/utils/types/TypedStructField.scala
```diff
@@ -227,6 +250,14 @@ object TypedStructField {
 
     def precision: Int = dataType.precision
     def scale: Int = dataType.scale
+
+    override protected def defaultValueInMetadata: Try[Option[Any]] = {
+      super.defaultValueInMetadata.recover {
+        case _: ClassCastException => Some(structField.metadata.getDouble(MetadataKeys.DefaultValue))
```
Zejnilovic: Isn't the decimal almost twice as big as double? Will this work?

utils/src/test/scala/za/co/absa/enceladus/utils/types/TypedStructFieldSuite.scala
```diff
@@ -179,5 +177,37 @@ class TypedStructFieldSuite extends FunSuite {
...
+    val nullable = false
+    val field = createField(fieldType, nullable, Some(true))
+    val typed = TypedStructField(field)
+    checkField(typed, fieldType, Success(Some(Some(true))), Success(Some(true)), nullable)
+  }
+
+  test("Float type nullable, with default defined as Long") {
+    val fieldType = FloatType
+    val nullable = false
+    val field = createField(fieldType, nullable, Some(1000L))
```
yruslan: Not sure how this works. Yes, you can put anything into `metadata` inside a schema when you already have it as a Scala object. But `StructType` schema is received as a JSON from Menas. It will lose the difference between `Int`, `Long` and `Double`, won't it?

Maybe better to test metadata processing by having examples of JSONs that could be received from Menas and check what types can be recognized after deserialization.

Zejnilovic:
So when I marked this issue #798 as blocking the issue #799 it seems it is vice versa. Menas is not able to handle Integers in default values even if I put them there using Mongo Shell. So the #799 must be solved before this for me to be sure enough in this.

utils/src/test/scala/za/co/absa/enceladus/utils/types/TypedStructFieldSuite.scala
benedeki: From the JSON it's read in the form how JSON interpreter understands it. E.g. 3.14 as Double, -1 as Integer.
I run a standardization task with these defaults seemed working fine.
Now want to check if Menas accepts such schema JSON

benedeki:
> 
> 
> So when I marked this issue #798 as blocking the issue #799 it seems it is vice versa. Menas is not able to handle Integers in default values even if I put them there using Mongo Shell. So the #799 must be solved before this for me to be sure enough in this.

I think they are complimentary. :-) One doesn't make much sense without the other. Definitely should go in the same release, if we add this feature.

utils/src/test/scala/za/co/absa/enceladus/utils/types/TypedStructFieldSuite.scala
benedeki: As actually tested/observed by @Zejnilovic in Menas it doesn't work well (Issue #799)

benedeki:
It has been agreed in the team not to support non-string metadata values and instead fix and enhance checks and validation of the input data, which would include verification they are in the form of `Map(String, String)`.

Issue #822, opened 2019-08-20
Use defaultWithNull in Conformance


benedeki:
## Background
#420 introduces global default which is null if the column is nullable.

## Feature
The logic is used only in Standardization, for consistency it should be used in Conformance too.
Swutch `Defaults.getGlobalDefault` for `Defaults.getGlobalDefaultWithNull` where appropriate.

PR #860, opened 2019-09-01
#822: Use defaultWithNull in Conformance

benedeki:
* Negation Rule uses Defaults.getGlobalDefaultWithNull instead of Defaults.getGlobalDefault
* Support for Arrays in Arrays in SchemaUtils
* Unified similar code in SchemaUtils
* added getStructField and fieldExists into SchemaUtils
* Updated and enhanced UTs

### DIFF
--- a/conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/rules/NegationRuleInterpreter.scala
+++ b/conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/rules/NegationRuleInterpreter.scala
@@ -32,7 +32,7 @@ case class NegationRuleInterpreter(rule: NegationConformanceRule) extends RuleIn
   override def conform(df: Dataset[Row])(implicit spark: SparkSession, dao: EnceladusDAO, progArgs: CmdConfig): Dataset[Row] = {
     NegationRuleInterpreter.validateInputField(progArgs.datasetName, df.schema, rule.inputColumn)
 
-    val field = SchemaUtils.getStructField(rule.inputColumn, df.schema).get
+    val field = SchemaUtils.getField(rule.inputColumn, df.schema).get
 
     val negationErrUdfCall = callUDF("confNegErr", lit(rule.outputColumn), col(rule.inputColumn))
     val errCol = "errCol"
@@ -48,10 +48,10 @@ case class NegationRuleInterpreter(rule: NegationConformanceRule) extends RuleIn
         // Negating floating point numbers cannot fail, but we need to account
         // for signed zeros (see the note for getNegator()).
         DeepArrayTransformations.nestedWithColumnMap(df, rule.inputColumn, rule.outputColumn, c => getNegator(c, field))
-      case _ =>
+      case dt =>
         // The generic negation with checking for error conditions
         DeepArrayTransformations.nestedWithColumnAndErrorMap(df, rule.inputColumn, rule.outputColumn, errCol,
-          c => getNegator(c, field), c => getError(c, negationErrUdfCall, field.dataType))
+          c => getNegator(c, field), c => getError(c, negationErrUdfCall, dt))
     }
   }
 
--- a/conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/rules/NegationRuleInterpreter.scala
+++ b/conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/rules/NegationRuleInterpreter.scala
@@ -32,28 +32,30 @@ case class NegationRuleInterpreter(rule: NegationConformanceRule) extends RuleIn
   override def conform(df: Dataset[Row])(implicit spark: SparkSession, dao: EnceladusDAO, progArgs: CmdConfig): Dataset[Row] = {
     NegationRuleInterpreter.validateInputField(progArgs.datasetName, df.schema, rule.inputColumn)
 
-    val fieldType = SchemaUtils.getFieldType(rule.inputColumn, df.schema).get
+    val field = SchemaUtils.getStructField(rule.inputColumn, df.schema).get
+
     val negationErrUdfCall = callUDF("confNegErr", lit(rule.outputColumn), col(rule.inputColumn))
     val errCol = "errCol"
 
-    fieldType match {
+    field.dataType match {
       case _: DecimalType =>
         // Negating decimal cannot fail
         DeepArrayTransformations.nestedWithColumnMap(df, rule.inputColumn, rule.outputColumn, c => negate(c))
       case _: BooleanType =>
+        // Negating Boolean cannot fail
         DeepArrayTransformations.nestedWithColumnMap(df, rule.inputColumn, rule.outputColumn, c => not(c))
       case _: DoubleType | _: FloatType =>
         // Negating floating point numbers cannot fail, but we need to account
         // for signed zeros (see the note for getNegator()).
-        DeepArrayTransformations.nestedWithColumnMap(df, rule.inputColumn, rule.outputColumn, c => getNegator(c, fieldType))
+        DeepArrayTransformations.nestedWithColumnMap(df, rule.inputColumn, rule.outputColumn, c => getNegator(c, field))
       case _ =>
         // The generic negation with checking for error conditions
         DeepArrayTransformations.nestedWithColumnAndErrorMap(df, rule.inputColumn, rule.outputColumn, errCol,
-          c => getNegator(c, fieldType), c => getError(c, negationErrUdfCall, fieldType))
+          c => getNegator(c, field), c => getError(c, negationErrUdfCall, field.dataType))
     }
   }
 
-  private def getNegator(inputColumn: Column, inputDataType: DataType): Column = {
+  private def getNegator(inputColumn: Column, field: StructField): Column = {
     // Just a couple JVM things:
     // 1. Beware of silent integer overflow, -Int.MinValue == Int.MaxValue + 1 == Int.MinValue
     //    Proof:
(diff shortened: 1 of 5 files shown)


utils/src/main/scala/za/co/absa/enceladus/utils/schema/SchemaUtils.scala
```diff
@@ -22,6 +22,45 @@ import scala.util.{Random, Try}
...
+        Option(structField)
+      } else {
+        structField.dataType match {
+          case str: StructType              => diveInto(names.tail, str(names.head))
+          case ArrayType(el: DataType, _) =>
+            goThroughArrays(el) match {
+              case struct: StructType => diveInto(names.tail, struct(names.head))
+              case _ => None
+            }
+          case _                            => None
```
yruslan: Arrows (`=>`) in this function are oddly aligned.

utils/src/main/scala/za/co/absa/enceladus/utils/schema/SchemaUtils.scala
```diff
@@ -253,8 +269,28 @@ object SchemaUtils {
     * @param fieldName Name of the field to be appended to the path
     * @return The path with the new field appended or the field itself if path is empty
     */
-  private[enceladus] def appendPath(path: String, fieldName: String) = {
-    if (path.isEmpty) fieldName else s"$path.$fieldName"
+  def appendPath(path: String, fieldName: String): String = {
```
yruslan: Since this is a public method, it might also validate input values. Make sure the input field name does not contain whitespaces or dots, for instance.
There are rules how one can specify a field name with whitespaces. I think '\`' is used (e.g. MY_PATH.\`MY FIELD\`) in such cases. But the conformance engine does not support such fields. I'm pretty sure all nested arrays mapping methods won't work properly with field names contain whitespaces. So maybe just throw an exception if an input column contains one.

utils/src/main/scala/za/co/absa/enceladus/utils/schema/SchemaUtils.scala
```diff
@@ -253,8 +269,28 @@ object SchemaUtils {
...
+    }
+  }
+
+  /**
+   * Combines path and field name using dot notation
+   * @param path      The dot-separated existing path
+   * @param fieldName Name of the field to be appended to the path
+   * @return          The combined path and field name
+   */
+  def fullPath(path: String, fieldName: String): String = {
```
yruslan: Method name as a verb? `appendPath()`? (not sure my suggestion is the best, I'm not super good at naming as well)

utils/src/main/scala/za/co/absa/enceladus/utils/schema/SchemaUtils.scala
```diff
@@ -22,6 +22,45 @@ import scala.util.{Random, Try}
 
 object SchemaUtils {
 
+  /**
+   * Get a field from a text path and a given schema
+   * @param path   The dot-separated path to the field
+   * @param schema The schema which should contain the specified path
+   * @return       Some(the requested field) or None if the field does not exist
+   */
+  def getStructField(path: String, schema: StructType): Option[StructField] = {
```
yruslan: Maybe just `getField`? The 'Struct' part is missing from the other similar methods.

utils/src/main/scala/za/co/absa/enceladus/utils/schema/SchemaUtils.scala
```diff
@@ -22,6 +22,45 @@ import scala.util.{Random, Try}
...
+    @tailrec
+    def goThroughArrays(dataType: DataType): DataType = {
+      dataType match {
+        case ArrayType(dt, _) => goThroughArrays(dt)
+        case result => result
+      }
+    }
+
+    @tailrec
+    def diveInto(names: List[String], structField: StructField): Option[StructField] = {
```
yruslan: `goThroughArrays` and `diveInto` seem not consistent. Why not `goThroughStruct` and `diveIntoArrays`, for instance? (Althlough aestetically I like it, reads like a literature)
Maybe `processStruct` / `processArray`?

conformance/src/test/scala/za/co/absa/enceladus/conformance/interpreter/rules/NegationRuleSuite.scala
```diff
@@ -92,8 +101,13 @@ class NegationRuleSuite extends FunSuite with SparkTestBase with LoggerTestBase{
     testRule(inputDataset, expectedDataset, expectedJSON)
   }
 
-  private def testRule(inputDataset: Dataset[String], enceladusDataset: ConfDataset, expectedJSON: String): Unit = {
-    val inputDf = spark.read.schema(NegationRuleSamples.schema).json(inputDataset)
+  private def testRule(inputDataset: Dataset[String], enceladusDataset: ConfDataset, expectedJSON: String, nullableSchema: Boolean = true): Unit = {
```
yruslan: My personal taste is for `nullableSchema` not to have a default value, but require tests explicitly specify if they want to deal with a nullable or not nullable schema.

conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/rules/NegationRuleInterpreter.scala
```diff
@@ -63,13 +65,18 @@ case class NegationRuleInterpreter(rule: NegationConformanceRule) extends RuleIn
     // 2. Beware negative floating-point zeroes, i.e. 0.0 * -1 == -0.0
     //    Equality (0.0 == -0.0) holds true, but Spark SQL considers 0.0 and -0.0 distinct and fails joins
     // The above is true not only for JVM, but for the most of the CPU/hardware implementations of numeric data types
+
+    def defaultValue(dt: DataType, nullable: Boolean): Any = {
```
yruslan: Why `Any`? Shouldn't it always be a `Column`?

conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/rules/NegationRuleInterpreter.scala
```diff
@@ -32,28 +32,30 @@ case class NegationRuleInterpreter(rule: NegationConformanceRule) extends RuleIn
...
         DeepArrayTransformations.nestedWithColumnMap(df, rule.inputColumn, rule.outputColumn, c => negate(c))
       case _: BooleanType =>
+        // Negating Boolean cannot fail
         DeepArrayTransformations.nestedWithColumnMap(df, rule.inputColumn, rule.outputColumn, c => not(c))
       case _: DoubleType | _: FloatType =>
         // Negating floating point numbers cannot fail, but we need to account
         // for signed zeros (see the note for getNegator()).
-        DeepArrayTransformations.nestedWithColumnMap(df, rule.inputColumn, rule.outputColumn, c => getNegator(c, fieldType))
+        DeepArrayTransformations.nestedWithColumnMap(df, rule.inputColumn, rule.outputColumn, c => getNegator(c, field))
       case _ =>
```
Zejnilovic: Since there are already changes I would go a step further and on line `51` do `case dataType =>` and then use the `dataType` value on line `54`

utils/src/main/scala/za/co/absa/enceladus/utils/schema/SchemaUtils.scala
Zejnilovic: Since we are trying to solve for Arrays of Arrays. How would this handle if I wanted to send an array index to it? If I remember correctly using `arryName[0]` is the first element of arryName. Do we support this?

utils/src/main/scala/za/co/absa/enceladus/utils/schema/SchemaUtils.scala
Zejnilovic: What exactly is the difference between append path and full path? I see one more `else if` but why couldn't they be one?

utils/src/main/scala/za/co/absa/enceladus/utils/schema/SchemaUtils.scala
benedeki: > 
> 
> Since we are trying to solve for Arrays of Arrays. How would this handle if I wanted to send an array index to it? If I remember correctly using `arryName[0]` is the first element of arryName. Do we support this?

This working on schema, not the actual fields. I believe there is no indexing there. Ans also we are after element types not element values, and those are same withing one array type.

conformance/src/test/scala/za/co/absa/enceladus/conformance/interpreter/rules/NegationRuleSuite.scala
benedeki: OK. I was lazy 😉

conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/rules/NegationRuleInterpreter.scala
benedeki: OK

utils/src/main/scala/za/co/absa/enceladus/utils/schema/SchemaUtils.scala
Zejnilovic: > Difference - one add adds a dot at the end of the `path` in case the `fieldName` is empty. The other not. I haven't dug that deep to verify empty string is actually a real scenario...
Main question is, why they couldn't be merged as one method. They do the same.

benedeki: OK, you convinced me. 😉 
Thinking about it, the empty `fieldName` case should not be happening and definitely not relied on.

utils/src/main/scala/za/co/absa/enceladus/utils/schema/SchemaUtils.scala
```diff
@@ -22,6 +22,45 @@ import scala.util.{Random, Try}
...
+      if (names.isEmpty) {
+        Option(structField)
+      } else {
+        structField.dataType match {
+          case str: StructType              => diveInto(names.tail, str(names.head))
```
GeorgiChochov:
```suggestion
          case struct: StructType              => diveInto(names.tail, struct(names.head))
```
`str` usually means `String`, and 3 lines below this you have the exact same thing but called `struct`, so let's keep it consistent.

benedeki: Last remnants of the original code. Will fix.

PR #861, opened 2019-09-02
Feature/849 status code from 500 to 400

yruslan:


menas/src/main/scala/za/co/absa/enceladus/menas/exceptions/BadRequestException.scala
```diff
@@ -0,0 +1,18 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.enceladus.menas.exceptions
+
+case class BadRequestException(message: String, cause: Throwable = None.orNull) extends Exception(message, cause)
```
benedeki: Avoiding the `null` warning? 😉 
While smart, doesn't it more obscure the code, then help?

menas/src/main/scala/za/co/absa/enceladus/menas/models/rest/RestResponse.scala
```diff
@@ -0,0 +1,40 @@
+/*
```
lokm01: Wouldn't you rather extend the existing RestError class?

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/SchemaController.scala
```diff
@@ -57,16 +63,35 @@ class SchemaController @Autowired() (
     val fileContent = new String(file.getBytes)
 
     val sparkStruct = origFormat match {
-      case Some("struct") | Some("") | None => sparkMenasConvertor.convertAnyToStructType(fileContent)
-      case Some("copybook") =>
-        val parsedSchema = CopybookParser.parseTree(fileContent)
-        val cobolSchema = new CobolSchema(parsedSchema, SchemaRetentionPolicy.CollapseRoot, false)
-        cobolSchema.getSparkSchema
-      case Some(x) => throw new UnsupportedOperationException(s"$x is not a recognized schema format. Menas currently supports: 'struct' and 'copybook'")
+      case Some(SchemaTypeStruct) | Some("") | None =>
```
Zejnilovic: I think `Some("") | None` should have their own case. From this, it seems like it is something appropriate and I think would never then produce the `SchemaFormatError`.

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/SchemaController.scala
```diff
@@ -57,16 +63,35 @@ class SchemaController @Autowired() (
...
+        try {
+          sparkMenasConvertor.convertAnyToStructType(fileContent)
+        } catch {
+          case NonFatal(e) => throw SchemaParsingException(SchemaTypeStruct, e.getMessage)
+        }
+      case Some(SchemaTypeCopybook) =>
+        try {
```
Zejnilovic: Maybe for eas of reading extract these to stand-alone methods

menas/src/test/scala/za/co/absa/enceladus/menas/integration/controllers/SchemaApiIntegrationSuite.scala
```diff
@@ -282,6 +286,105 @@ class SchemaApiIntegrationSuite extends BaseRestApiTest {
     }
   }
 
+  s"POST $apiUrl/upload" should {
+    "return 201" when {
+      "a copybook has no errors" should {
+        "return a new version of the schema" in {
+          val schema = SchemaFactory.getDummySchema()
+          val responseCreated = sendPost[Schema, Schema](s"$apiUrl/create", bodyOpt = Some(schema))
```
GeorgiChochov: How about we use `SchemaFixtureService.add(schema)` to keep the test cases simple?

menas/src/main/scala/za/co/absa/enceladus/menas/models/rest/RestResponse.scala
yruslan: 1. Too afraid to touch Oozie stuff.
2. `RestError` is a case class. To extend it i'd probably need to replace it with a class hierarchy, which I'm not comfortable to do in Menas [yet].
3. Wanted for responses to have a certain structure, where `error` is an optional response field. 

But yes, ideas from `RestError` implementation were used to track issues by id and timestamp. 😄 

In the future, `RestError` and `RestResponse + ResponseError` might get combined (after refactoring) and a uniform approach is used.

menas/src/main/scala/za/co/absa/enceladus/menas/models/rest/ResponseError.scala
```diff
@@ -0,0 +1,33 @@
...
+  * This abstract class is used as a parent for all REST errors.
+  */
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "_t")
+@JsonSubTypes(Array(
+  new Type(value = classOf[SchemaFormatError], name = "SchemaFormatError"),
+  new Type(value = classOf[SchemaParsingError], name = "SchemaParsingError")
+))
+abstract class ResponseError {
+  /** An error type. Should be as concise as possible to be parsable by a service. */
+  val errorType: String
```
GeorgiChochov: I have a feeling that this is duplicating the `_t` field Jackson adds. Maybe it's not necessary. 
1. You can change the `name` parameter in the `@JsonSubTypes` annotation to be `schema_format` and `schema_parsing`, rather than the class names.
2. You can change the `@JsonTypeInfo` annotation's `property` param to be `errorType` rather than `_t`.

You would lose this information wen deserialized to a Scala object but you have the class type available, which should be enough. 

You could also add a `serDe` test (look at the tests in `za.co.absa.enceladus.model.conformanceRule`)

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/SchemaController.scala
```diff
@@ -57,16 +63,35 @@ class SchemaController @Autowired() (
...
+          val parsedSchema = CopybookParser.parseTree(fileContent)
+          val cobolSchema = new CobolSchema(parsedSchema, SchemaRetentionPolicy.CollapseRoot, false)
+          cobolSchema.getSparkSchema
+        } catch {
+          case e: SyntaxErrorException =>
+            throw SchemaParsingException(SchemaTypeCopybook, e.getMessage, Some(e.lineNumber), None, Some(e.field), e)
+          case NonFatal(e) =>
+            throw SchemaParsingException(SchemaTypeCopybook, e.getMessage)
```
GeorgiChochov:
```suggestion
            throw SchemaParsingException(SchemaTypeCopybook, e.getMessage, e)
```
Let's add the cause here.

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/SchemaController.scala
```diff
@@ -57,16 +63,35 @@ class SchemaController @Autowired() (
...
+        }
+      case Some(schemaType) =>
+        throw SchemaFormatException(schemaType, s"$schemaType is not a recognized schema format. " +
+          s"Menas currently supports: '$SchemaTypeStruct' and '$SchemaTypeCopybook'.")
```
GeorgiChochov:
```suggestion
          s"Menas currently supports: '$SchemaTypeStruct' and '$SchemaTypeCopybook'.", e)
```
Let's add the cause here.

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/SchemaController.scala
```diff
@@ -57,16 +63,35 @@ class SchemaController @Autowired() (
...
-      case Some("copybook") =>
-        val parsedSchema = CopybookParser.parseTree(fileContent)
-        val cobolSchema = new CobolSchema(parsedSchema, SchemaRetentionPolicy.CollapseRoot, false)
-        cobolSchema.getSparkSchema
-      case Some(x) => throw new UnsupportedOperationException(s"$x is not a recognized schema format. Menas currently supports: 'struct' and 'copybook'")
+      case Some(SchemaTypeStruct) | Some("") | None =>
+        try {
+          sparkMenasConvertor.convertAnyToStructType(fileContent)
+        } catch {
+          case NonFatal(e) => throw SchemaParsingException(SchemaTypeStruct, e.getMessage)
```
GeorgiChochov:
```suggestion
          case NonFatal(e) => throw SchemaParsingException(SchemaTypeStruct, e.getMessage, e)
```
Let's add the cause here.

menas/src/main/scala/za/co/absa/enceladus/menas/models/rest/ResponseError.scala
yruslan: I imagined `errorType` to be more specific than the class used for storing error details. It is not used like this yet, but for the future might be useful.
Let's say we have `AuthenticationError` class that combines errors related to authentication. In this case, `errorType` can be `ldap_server_unreachable`, `kdc_server_timeout`, `invalid_keyfile`, etc. There might be much more causes than classes.

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/SchemaController.scala
yruslan: This is an initial exception.

GeorgiChochov: lol, I was on a roll from the other ones

menas/src/test/scala/za/co/absa/enceladus/menas/integration/controllers/SchemaApiIntegrationSuite.scala
```diff
@@ -282,6 +286,105 @@ class SchemaApiIntegrationSuite extends BaseRestApiTest {
...
+          assertCreated(responseCreated)
+
+          val schemaParams = HashMap[String, Any] (
+            "name" -> schema.name, "version" -> schema.version, "format" -> "copybook")
```
Zejnilovic: I don't see tests for `None` or `""` format.

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/SchemaController.scala
Zejnilovic: Ok so the default value for upload is nothing which indicates struct for compatibility reasons. These options are ok

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/SchemaController.scala
```diff
@@ -57,16 +63,35 @@ class SchemaController @Autowired() (
...
+        }
+      case Some(SchemaTypeCopybook) =>
+        try {
+          val parsedSchema = CopybookParser.parseTree(fileContent)
+          val cobolSchema = new CobolSchema(parsedSchema, SchemaRetentionPolicy.CollapseRoot, false)
+          cobolSchema.getSparkSchema
+        } catch {
+          case e: SyntaxErrorException =>
+            throw SchemaParsingException(SchemaTypeCopybook, e.getMessage, Some(e.lineNumber), None, Some(e.field), e)
+          case NonFatal(e) =>
```
yruslan: Catch only relevant exceptions (`IllegatStateException`, etc.) to return 400. Pass other exceptions downstream to be handled in `RestExceptionHandling` for returning 500.

menas/src/test/scala/za/co/absa/enceladus/menas/integration/controllers/SchemaApiIntegrationSuite.scala
```diff
@@ -282,6 +286,103 @@ class SchemaApiIntegrationSuite extends BaseRestApiTest {
...
+          val body = response.getBody
+
+          assertBadRequest(response)
+          body.error match {
+            case Some(e: SchemaFormatError) =>
+              assert(e.errorType == "schema_format")
+              assert(e.schemaType == "ERROR")
+              assert(body.message.contains("ERROR is not a recognized schema format."))
+            case e => fail(s"Expected an instance of SchemaFormatError, got $e.")
+          }
```
yruslan: Add a test for 404

menas/src/main/scala/za/co/absa/enceladus/menas/models/rest/errors/SchemaParsingError.scala
```diff
@@ -0,0 +1,37 @@
...
+import za.co.absa.enceladus.menas.models.rest.ResponseError
+import za.co.absa.enceladus.menas.models.rest.exceptions.SchemaParsingException
+
+/**
+  * This error is produced when a parsing error is occurred when uploading a schema.
+  */
+case class SchemaParsingError(
+                               errorType: String,
+                               schemaType: String,
+                               line: Option[Int],
```
yruslan: Add an optional `field` parameter that exists in `SchemaParsingException`.

yruslan:
Thanks for the thorough PR review to the code quality meetup guys! 
(@jozefbakus, @kevinwallimann, @cerveada, @Zejnilovic, @benedeki, @GeorgiChochov)

menas/src/test/scala/za/co/absa/enceladus/menas/integration/controllers/SchemaApiIntegrationSuite.scala
yruslan: Added below.

PR #880, opened 2019-09-05
879 Remove request label from feature request template defaults

GeorgiChochov:


yruslan:
I was wondering, do we need a template for epics?

GeorgiChochov:
> I was wondering, do we need a template for epics?

Maybe, I'll have a think about it.

PR #882, opened 2019-09-05
220 Disable editing of old versions

Zejnilovic:
Thought about it and instead of throwing errors around after edit, I disabled the edit button for older versions of entities.

### DIFF
--- a/menas/src/main/scala/za/co/absa/enceladus/menas/controllers/VersionedModelController.scala
+++ b/menas/src/main/scala/za/co/absa/enceladus/menas/controllers/VersionedModelController.scala
@@ -69,6 +69,12 @@ abstract class VersionedModelController[C <: VersionedModel with Product with Au
     }
   }
 
+  @GetMapping(Array("/detail/{name}/latestVersion"))
+  @ResponseStatus(HttpStatus.OK)
+  def getLatestVersionNumber(@PathVariable name: String): CompletableFuture[Int] = {
+    versionedModelService.getLatestVersionNumber(name)
+  }
+
   @GetMapping(Array("/detail/{name}/audit"))
   @ResponseStatus(HttpStatus.OK)
   def getAuditTrail(@PathVariable name: String): CompletableFuture[AuditTrail] = {
--- a/menas/src/main/scala/za/co/absa/enceladus/menas/services/VersionedModelService.scala
+++ b/menas/src/main/scala/za/co/absa/enceladus/menas/services/VersionedModelService.scala
@@ -62,6 +62,13 @@ abstract class VersionedModelService[C <: VersionedModel with Product with Audit
 
   }
 
+  def getLatestVersionNumber(name: String): Future[Int] = {
+    versionedMongoRepository.getLatestVersionValue(name).flatMap({
+      case Some(version) => Future(version)
+      case _ => throw NotFoundException()
+    })
+  }
+
   def getLatestVersionValue(name: String): Future[Option[Int]] = {
     versionedMongoRepository.getLatestVersionValue(name)
   }
--- a/menas/src/main/scala/za/co/absa/enceladus/menas/services/VersionedModelService.scala
+++ b/menas/src/main/scala/za/co/absa/enceladus/menas/services/VersionedModelService.scala
@@ -146,11 +146,10 @@ abstract class VersionedModelService[C <: VersionedModel with Product with Audit
 
   private[services] def updateFuture(username: String, itemName: String, itemVersion: Int)(transform: C => Future[C]): Future[Option[C]] = {
     for {
-      versionToUpdate <- getVersion(itemName, itemVersion)
-      latestVersion <- getLatestVersionNumber(itemName)
+      versionToUpdate <- getLatestVersion(itemName)
       transformed <- if (versionToUpdate.isEmpty) {
         Future.failed(NotFoundException(s"Version $itemVersion of $itemName not found"))
-      } else if (versionToUpdate.get.version != latestVersion) {
+      } else if (versionToUpdate.get.version != itemVersion) {
         Future.failed(ValidationException(Validation().withError("version", s"Version $itemVersion of $itemName is not the latest version, therefore cannot be edited")))
       }
       else {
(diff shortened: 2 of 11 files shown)


menas/ui/components/dataset/datasetDetail.controller.js
```diff
@@ -350,12 +350,16 @@ sap.ui.define([
...
-          this._model.setProperty("/currentDataset/schema", schema)
+          this._model.setProperty("/currentDataset/schema", schema);
           this._schemaTable.model = schema;
           transitiveSchemas.push(schema);
           SchemaManager.getTransitiveSchemas(transitiveSchemas, currentDataset.conformance)
         });
 
+        this._datasetRestDAO = new DatasetRestDAO();
+        this._datasetRestDAO.getLatestVersionByName(currentDataset.name)
+          .then(version => sap.ui.getCore().getModel().setProperty("/editingEnabled", currentDataset.version === version));
```
GeorgiChochov:
```suggestion
          .then(version => this._model.setProperty("/editingEnabled", currentDataset.version === version));
```

menas/ui/components/mappingTable/mappingTableDetail.controller.js
```diff
@@ -285,10 +285,16 @@ sap.ui.define([
...
-      this._mappingTableService.getAuditTrail(currentMT.name, auditTable);
+      this.byId("info").setModel(new sap.ui.model.json.JSONModel(currentMT), "mappingTable");
+      if (currentMT) {
+        this.fetchSchema();
+        const auditTable = this.byId("auditTrailTable");
+        this._mappingTableService.getAuditTrail(currentMT.name, auditTable);
+
+        this._mtRestDAO = new MappingTableRestDAO();
+        this._mtRestDAO.getLatestVersionByName(currentMT.name)
+          .then(version => sap.ui.getCore().getModel().setProperty("/editingEnabled", currentMT.version === version));
```
GeorgiChochov:
```suggestion
          .then(version => this._model.setProperty("/editingEnabled", currentMT.version === version));
```

menas/ui/components/dataset/datasetDetail.view.xml
```diff
@@ -65,8 +65,9 @@
             <Bar>
                 <contentRight>
                     <Button id="Delete" text="Delete" icon="sap-icon://delete" press="onRemovePress" />
-                    <Button id="AddRule" text="Add Conformance Rule" icon="sap-icon://add-equipment" press="onAddConformanceRulePress"></Button>
-                    <Button id="Edit" text="Edit" icon="sap-icon://edit" press="onEditPress" />
+                    <Button id="AddRule" text="Add Conformance Rule" icon="sap-icon://add-equipment" press="onAddConformanceRulePress" />
+                    <Button id="Edit" text="Edit" icon="sap-icon://edit" press="onEditPress" enabled="{/editingEnabled}"
+                            tooltip="{= ${/editingEnabled} ? 'Editing enabled' : 'Editing old version is disabled'}" />
```
GeorgiChochov: There is something odd about `Editing enabled`, it's confusing to see when I expect to be able to edit. Makes sense to tell me why it's disabled, because it's an unexpected condition. 
For when it's enabled I'd expect a tooltip that tells me something like `Edit current dataset`.

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/VersionedModelController.scala
```diff
@@ -69,6 +69,12 @@ abstract class VersionedModelController[C <: VersionedModel with Product with Au
     }
   }
 
+  @GetMapping(Array("/detail/{name}/latestVersion"))
+  @ResponseStatus(HttpStatus.OK)
+  def getLatestVersionNumber(@PathVariable name: String): CompletableFuture[Int] = {
```
GeorgiChochov: Let's get an API test for this. (see `za.co.absa.enceladus.menas.integration.controllers.SchemaApiIntegrationSuite:43` for inspiration.

Zejnilovic: Have you seen the line 98? Seems like we will have to have the second issue opened to handle it with rest as well.

Zejnilovic:
This issue is now being fixed on two fronts.
1st is the REST API returns proper Bad Request
2nd is that users using UI can't even press edit if they are on the bad version

Please review, comment, etc.

menas/src/main/scala/za/co/absa/enceladus/menas/services/VersionedModelService.scala
```diff
@@ -139,11 +146,15 @@ abstract class VersionedModelService[C <: VersionedModel with Product with Audit
 
   private[services] def updateFuture(username: String, itemName: String, itemVersion: Int)(transform: C => Future[C]): Future[Option[C]] = {
     for {
-      version <- getVersion(itemName, itemVersion)
-      transformed <- if (version.isEmpty) {
+      versionToUpdate <- getVersion(itemName, itemVersion)
+      latestVersion <- getLatestVersionNumber(itemName)
```
GeorgiChochov: The original idea for this function was to perform updates based on any version, so `getVersion` was used. Now that we are enforcing updates only based on the latest version, we can change the `getVersion` call to `getLatestVersion` and skip the call to `getLatestVersionNumber`. 

Since we get the latest version's number from `getLatestVersion` we would just need to compare it against the `itemVersion` parameter, to make sure that the user is actually updating the intended entity version. 

(e.g. **A** starts updating when the latest version is 1. **B** updates the entity to version 2 before **A** submits. **A** will send the update request with `itemVersion = 1`, the call to `getLatestVersion` will return an entity with latest version 2, and **A** will see that their update is outdated.)

menas/src/test/scala/za/co/absa/enceladus/menas/integration/controllers/SchemaApiIntegrationSuite.scala
```diff
@@ -99,29 +99,54 @@ class SchemaApiIntegrationSuite extends BaseRestApiTest {
...
+        val schemaV2 = SchemaFactory.getDummySchema(name = "schema1",
+                                                    version = 2,
+                                                    parent = Some(SchemaFactory.toParent(schemaV1)))
+        schemaFixture.add(schemaV1)
+        schemaFixture.add(schemaV2)
+
+        val response = sendGet[String](s"$apiUrl/detail/schema1/latestVersion")
+
+        assert("2" == response.getBody)
+        assertOk(response)
```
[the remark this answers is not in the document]
Zejnilovic: I don't think it is pedantic. This is actually a great catch because if it starts failing on status 4XX or something like that I get faster to the root of the problem.

Issue #892, opened 2019-09-12
Mapping rule with literal in the join condition (may) throw a cross join error


lokm01:
## Describe the bug
Since Spark 2.3.1, joins (in our case left_outer) throw a cartesian product exception if one part of the join condition is a column previously produced by a literal rule.

The behaviour of this is questionable as it seems that catalyst isn't able to "reduce" the logical plan in certain cases.

lokm01:
Seems related to https://issues.apache.org/jira/browse/SPARK-24839

GeorgiChochov:
For release notes:
Mapping rules can now safely join on columns created by a literal rule. (currently just a workaround which will be refined in #895)

PR #894, opened 2019-09-12
#892 Mapping rule with literal in join condition may throw error

lokm01:


conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/rules/MappingRuleInterpreter.scala
```diff
@@ -46,6 +46,10 @@ case class MappingRuleInterpreter(rule: MappingConformanceRule, conformance: Con
     log.info(s"Processing mapping rule to conform ${rule.outputColumn}...")
     import spark.implicits._
 
+    //A fix for cases, where the join condition only uses columns previously created by a literal rule
+    //see https://github.com/AbsaOSS/enceladus/issues/892
+    spark.conf.set("spark.sql.crossJoin.enabled", "true")
```
benedeki: Well this works. But wouldn't it be better/nicer to do this on higher level, when the_Conformance_ rules are loaded?
* We would be setting the configuration only once.
* There would be no duplicate code (here and in `MappingRuleInterpreterGroupExplode.scala`
* The class `FeatureSwitches` could be leveraged to quickly and efficiently switch on/off the settting.

conformance/src/test/scala/za/co/absa/enceladus/conformance/interpreter/LiteralJoinMappingRuleTest.scala
```diff
@@ -0,0 +1,93 @@
...
+import za.co.absa.enceladus.dao.EnceladusDAO
+import za.co.absa.enceladus.model.{ Dataset => ConfDataset }
+import za.co.absa.enceladus.model.MappingTable
+import za.co.absa.enceladus.model.conformanceRule.DropConformanceRule
+import za.co.absa.enceladus.model.conformanceRule.LiteralConformanceRule
+import za.co.absa.enceladus.model.conformanceRule.MappingConformanceRule
+import za.co.absa.enceladus.utils.testUtils.LoggerTestBase
+import za.co.absa.enceladus.utils.testUtils.SparkTestBase
+
+class LiteralJoinMappingRuleTest extends FunSuite with SparkTestBase with LoggerTestBase {
```
benedeki: So were you able to replicate the issue on small scale?

lokm01: Yes this fails locally when removing the config fix

conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/rules/MappingRuleInterpreter.scala
lokm01: Yes good point, by all means would be cleaner and nicer if configurable.

I would very much like to prioritize #895 though, since that would represent a proper fix for long term.

Issue #893, opened 2019-09-12
The convenience scripts should always propagate spark's status codes


lokm01:
## Describe the bug
The convenience scripts always return successful status code regardless of spark-submit's status.

## To Reproduce
Run convenience scripts for a job which fails

## Expected behaviour
The script should return spark-submit's status code

GeorgiChochov:
Convenience now propagate the spark job's exit code.

PR #896, opened 2019-09-13
Issue #893: The convenience scripts should always propagate spark's status codes

DavidMatusik:
- Added statement to return exit status of the last failed command in the subshell pipeline running spark submit
- Spark submit subshell exit status is propagated to the scripts final exit status and is printed to the user in case of client deploy mode

### DIFF
--- a/scripts/bash/run_enceladus.sh
+++ b/scripts/bash/run_enceladus.sh
@@ -307,7 +307,7 @@ if [[ -z "$DRY_RUN" ]]; then
         sleep 10
       fi
     fi
-    # Log the log locationd
+    # Log the log location
     echo "$CMD_LINE" >> "$TMP_PATH_NAME"
     echo "The log will be saved to $TMP_PATH_NAME"
     # Run the job and return exit status of the last failed command in the subshell pipeline (Issue #893)
@@ -316,18 +316,15 @@ if [[ -z "$DRY_RUN" ]]; then
 	EXIT_STATUS="$?"
 	# Test if the command executed successfully
 	if [ $EXIT_STATUS -eq 0 ]; then
-		# Report the log location
-		echo
-		echo "Job has finished. The logs are saved to $TMP_PATH_NAME"
+      RESULT="passed"
 	else
-		# Report the failed exit status and log location
-		echo
-		echo "Job failed with exit status $EXIT_STATUS. Refer to logs at $TMP_PATH_NAME"
-		exit $EXIT_STATUS
+	  RESULT="failed"
 	fi
+	# Report the result and log location
+	echo ""
+	echo "Job $RESULT with exit status $EXIT STATUS. Refer to logs at $TMP_PATH_NAME"
+	exit $EXIT_STATUS
   else
     bash -c "$CMD_LINE"
-	# Exit with the same status as spark submit
-	exit $?
   fi
 fi
--- a/scripts/bash/run_enceladus.sh
+++ b/scripts/bash/run_enceladus.sh
@@ -322,7 +322,7 @@ if [[ -z "$DRY_RUN" ]]; then
 	fi
 	# Report the result and log location
 	echo ""
-	echo "Job $RESULT with exit status $EXIT STATUS. Refer to logs at $TMP_PATH_NAME"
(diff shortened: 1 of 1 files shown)


scripts/bash/run_enceladus.sh
```diff
@@ -308,18 +307,27 @@ if [[ -z "$DRY_RUN" ]]; then
...
-    # Run the job
-    bash -c "$CMD_LINE 2>&1 | tee -a $TMP_PATH_NAME"
-
-    # Report the log location
-    echo
-    echo "Job has finished. The logs are saved to $TMP_PATH_NAME"
+    # Run the job and return exit status of the last failed command in the subshell pipeline (Issue #893)
+    bash -c "set -o pipefail; $CMD_LINE 2>&1 | tee -a $TMP_PATH_NAME"
+	# Save the exit status of spark submit subshell run
+	EXIT_STATUS="$?"
```
benedeki: Question: Why the `EXIT_STATUS` have to be assigned this indirect way? Cannot it be done directly from the `bash` call?

DavidMatusik: When assigned directly in bash call the exit status returns empty value. Will test further to make sure I don't miss the more elegant solution you suggest

DavidMatusik:
@benedeki yes, the `run_conformance.sh` and `run_standardization.sh` exit status is propagated from `source` of `run_enceladus.sh` properly

scripts/bash/run_enceladus.sh
```diff
@@ -308,18 +307,27 @@ if [[ -z "$DRY_RUN" ]]; then
...
+	# Test if the command executed successfully
+	if [ $EXIT_STATUS -eq 0 ]; then
+		# Report the log location
+		echo
+		echo "Job has finished. The logs are saved to $TMP_PATH_NAME"
```
Zejnilovic: Logs are saved either way, so I would have `if else` for echo `Passed/Failed` and print the rest the same. And either way call `exit $EXIT_STATUS`

scripts/bash/run_enceladus.sh
Zejnilovic: We can skip it if we use `$?` directly in the `if` statement. Then again, if you add a print or a call in between you broke the code. I like this.

scripts/bash/run_enceladus.sh
```diff
@@ -308,18 +307,27 @@ if [[ -z "$DRY_RUN" ]]; then
...
+	else
+		# Report the failed exit status and log location
+		echo
+		echo "Job failed with exit status $EXIT_STATUS. Refer to logs at $TMP_PATH_NAME"
+		exit $EXIT_STATUS
+	fi
   else
     bash -c "$CMD_LINE"
+	# Exit with the same status as spark submit
+	exit $?
```
Zejnilovic: This is redundant

scripts/bash/run_enceladus.sh
benedeki: If you tried and didn't work, keep this solution. No need to spend more time on it IMO.
Maybe add a comment explaining, to avoid future enquirers.

scripts/bash/run_enceladus.sh
benedeki: If the other call above to `bash -c "set -o pipefail; $CMD_LINE 2>&1 | tee -a $TMP_PATH_NAME"` is an indication, it might not work propagating the exit code. 🤔

scripts/bash/run_enceladus.sh
```diff
@@ -308,18 +307,27 @@ if [[ -z "$DRY_RUN" ]]; then
         sleep 10
       fi
     fi
-
-    # Log the log location
+    # Log the log locationd
```
GeorgiChochov:
```suggestion
    # Log the log location
```

scripts/bash/run_enceladus.sh
Zejnilovic: The previous call is propagating the exit status. But it is the exit status of the `tee` command. That is where the problem was. So if we just run something without the pipe behind it, it should be fine.

GeorgiChochov: Can we consider this resolved?

Issue #898, opened 2019-09-16
Allow additional environment settings in convenience scripts


GeorgiChochov:
## Background
Currently, our convenience scripts require specific environment settings to be configured in `enceladus_env.sh` but there might be use cases which require further environment settings to be provided. When users need to add those, they need to change other scripts as well. 

## Feature
It would be easier to understand if we provide a way to add environment specific configurations in one centralized location along with the rest of the  environment settings.

## Example [Optional]
Running conformance from a Docker container in `client` mode requires the `spark.driver.host` property to be set to the host machine's IP, not the container's, so yarn can refer to the containerized driver:
```bash
ADDITIONAL_SPARK_CONF="spark.driver.host=some.host.example.com spark.driver.port=5050"
```

## Proposed Solution [Optional]
Add a variable for the `non-enceladus` spark settings in `enceladus_env.sh`.

PR #901, opened 2019-09-17
#898 Add environment-specific Spark options to the convenience scripts.

yruslan:


scripts/bash/enceladus_env.template.sh
```diff
@@ -34,3 +34,6 @@ CONF_CLASS="za.co.absa.enceladus.conformance.DynamicConformanceJob"
 CONF_JAR="enceladus-conformance.jar"
 
 LOG_DIR="/tmp"
+
+# Additional environment-specific Spark options, e.g. "spark.driver.host=myhost"
+ADDITIONAL_SPARK_CONF=""
```
lokm01: This is a dummy question, but this can be set by export right? The empty string won't overwrite that?

Excuse my "bashgnorance" 😄

yruslan: Double checked - the empty string will overwrite the variable even if it is already set by `export`. So this should be safe against users trying to inject arbitrary options outside of `enceladus_env.sh` 😄

yruslan: Thanks for the concern - every concern should be checked.

PR #906, opened 2019-09-17
484 Make scalastyle less strict

GeorgiChochov:


pom.xml
```diff
@@ -292,6 +293,7 @@
                     <artifactId>scalastyle-maven-plugin</artifactId>
                     <version>1.0.0</version>
                     <configuration>
+                        <quiet>${quiet.scalastyle.check}</quiet>
```
GeorgiChochov: With `quiet` set to true by default, `scalastyle` still runs, fails on error, saves the output to a file and only prints this:
image
If it fails the build you can find the cause of failure in the output file or rerun it with the `-Pscalastyle` profile.
It really is very lightweight, disabling it would shave no more than a single-digit seconds of build time, so I'd rather have it run always.

GeorgiChochov:
> I was thinking perhaps making `org.scalastyle.scalariform.ReturnChecker` an **error**.
> At the moment we have it only at one place, marked with check exclusion and a ticket to fix. I don't think we should add any `return` in the code and if _really_ needed still can be marked.
> 
> `org.scalastyle.scalariform.NullChecker` is another candidate for **error** but there were some cases where the exclusion was not added because not clear if justified

@lokm01, @yruslan, @Zejnilovic, @DzMakatun, opinions?

yruslan:
👍  on `org.scalastyle.scalariform.ReturnChecker`.
`org.scalastyle.scalariform.NullChecker` - not sure. Since nulls are used with Spring, the suppression might be needed in too many places.

benedeki:
> 
> 
> 👍 on `org.scalastyle.scalariform.ReturnChecker`.
> `org.scalastyle.scalariform.NullChecker` - not sure. Since nulls are used with Spring, the suppression might be needed in too many places.

Even if with the _Warning_ setting I would be strongly for putting the suppressors there for required _Spring_ and _Spark_ usage.

Zejnilovic:
I am for both return and null checker to throw errors. Suppressions need to be well thought through.

lokm01:
Let's keep nulls as warning. Agree on return

GeorgiChochov:
Alright, we're keeping `NullChecker` as a warning for now. We might revisit in the future.

Issue #887, opened 2019-09-06
allowinfinity metadata key should be a string


benedeki:
## Describe the bug
Menas accepts only metadata where the value is type `String`. While `allowinfinity` is expected to be `Boolean`

## To Reproduce
Use `allowinfinity` with a field

## Expected behaviour
`allowinfinity` can be used and it's working.
Document the feature too.

benedeki:
**Release notes**
_allowinfinity_ metadata key for `Float` and `Double` type fields operational now within Standardization.
The possible values for the key are _"True"_ and _"False"_ (Boolean in String format), when not specified _False_ is considered default. If set to _True_ it allows the column to be marked as _infinity/-infinity_ for too big values.

PR #908, opened 2019-09-18
#887: allowinfinity metadata key should be a string

benedeki:
* Read the value of allowinfinity metadata key as key and then convert to Boolean
* FieldValidator to check this metadata for fractional types
* Added UTs

### DIFF
--- a/utils/src/main/scala/za/co/absa/enceladus/utils/implicits/StructFieldImplicits.scala
+++ b/utils/src/main/scala/za/co/absa/enceladus/utils/implicits/StructFieldImplicits.scala
@@ -24,8 +24,8 @@ object StructFieldImplicits {
       Try(structField.metadata.getString(key)).toOption
     }
 
-    def getMetadataBoolean(key: String): Option[Boolean] = {
-      Try(structField.metadata.getBoolean(key)).toOption
+    def getMetadataStringAsBoolean(key: String): Option[Boolean] = {
+      Try(structField.metadata.getString(key).toBoolean).toOption
     }
 
     def hasMetadataKey(key: String): Boolean = {
--- a/utils/src/main/scala/za/co/absa/enceladus/utils/validation/field/FractionalFieldValidator.scala
+++ b/utils/src/main/scala/za/co/absa/enceladus/utils/validation/field/FractionalFieldValidator.scala
@@ -18,9 +18,7 @@ import za.co.absa.enceladus.utils.schema.MetadataKeys
 import za.co.absa.enceladus.utils.types.TypedStructField
 import za.co.absa.enceladus.utils.validation.{ValidationError, ValidationIssue}
 
-object FractionalFieldValidator extends FractionalFieldValidator
-
-class FractionalFieldValidator extends ScalarFieldValidator {
+object FractionalFieldValidator extends ScalarFieldValidator {
   private def validateAllowInfinity(field: TypedStructField): Seq[ValidationIssue] = {
     if (field.hasMetadataKey(MetadataKeys.allowInfinity) && field.getMetadataStringAsBoolean(MetadataKeys.allowInfinity).isEmpty) {
       Seq(ValidationError(s"allowInfinity metadata value of field '${field.name}' is not Boolean in String format"))
--- a/utils/src/main/scala/za/co/absa/enceladus/utils/validation/field/FractionalFieldValidator.scala
+++ b/utils/src/main/scala/za/co/absa/enceladus/utils/validation/field/FractionalFieldValidator.scala
@@ -14,25 +14,23 @@
  */
 
 package za.co.absa.enceladus.utils.validation.field
-
+import za.co.absa.enceladus.utils.schema.MetadataKeys
 import za.co.absa.enceladus.utils.types.TypedStructField
-import za.co.absa.enceladus.utils.validation.ValidationIssue
+import za.co.absa.enceladus.utils.validation.{ValidationError, ValidationIssue}
 
-import scala.util.Try
+object FractionalFieldValidator extends FractionalFieldValidator
 
-
-/**
-  * Scalar types schema validation against default value
-  */
-object ScalarFieldValidator extends ScalarFieldValidator
-
-class ScalarFieldValidator extends FieldValidator {
-
-  private def validateDefaultValue(field: TypedStructField): Try[Any] = {
-    field.defaultValueWithGlobal
+class FractionalFieldValidator extends ScalarFieldValidator {
+  private def validateAllowInfinity(field: TypedStructField): Seq[ValidationIssue] = {
+    if (field.hasMetadataKey(MetadataKeys.allowInfinity) && field.getMetadataStringAsBoolean(MetadataKeys.allowInfinity).isEmpty) {
+      Seq(ValidationError(s"allowInfinity metadata value of field '${field.name}' is not Boolean in String format"))
+    } else {
+      Nil
+    }
   }
 
   override def validate(field: TypedStructField): Seq[ValidationIssue] = {
-    tryToValidationIssues(validateDefaultValue(field))
+    validateAllowInfinity(field) ++ super.validate(field)
   }
 }
+
--- a/utils/src/test/scala/za/co/absa/enceladus/utils/types/TypedStructFieldSuite.scala
+++ b/utils/src/test/scala/za/co/absa/enceladus/utils/types/TypedStructFieldSuite.scala
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2018-2019 ABSA Group Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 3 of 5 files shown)


utils/src/main/scala/za/co/absa/enceladus/utils/validation/field/ScalarFieldValidator.scala
```diff
@@ -24,7 +24,9 @@ import scala.util.Try
 /**
   * Scalar types schema validation against default value
   */
-object ScalarFieldValidator extends FieldValidator {
+object ScalarFieldValidator extends ScalarFieldValidator
+
+class ScalarFieldValidator extends FieldValidator {
```
GeorgiChochov: I see that you turned this into a class so that it could be inherited by `FractionalFieldValidator`, shouldn't `ScalarFieldValidator` be declared `abstract` since you want the companion `object` to be the only instance?

utils/src/main/scala/za/co/absa/enceladus/utils/validation/field/FractionalFieldValidator.scala
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+package za.co.absa.enceladus.utils.validation.field
+import za.co.absa.enceladus.utils.schema.MetadataKeys
+import za.co.absa.enceladus.utils.types.TypedStructField
+import za.co.absa.enceladus.utils.validation.{ValidationError, ValidationIssue}
+
+object FractionalFieldValidator extends FractionalFieldValidator
+
+class FractionalFieldValidator extends ScalarFieldValidator {
```
GeorgiChochov:
```suggestion
object FractionalFieldValidator extends ScalarFieldValidator {
```
Can't this simply be the way you had it `ScalarFieldValidator`? I don't understand the pattern with the companion object extending it's own class. This might be a topic for CQC.

utils/src/main/scala/za/co/absa/enceladus/utils/implicits/StructFieldImplicits.scala
```diff
@@ -24,8 +24,8 @@ object StructFieldImplicits {
       Try(structField.metadata.getString(key)).toOption
     }
 
-    def getMetadataBoolean(key: String): Option[Boolean] = {
-      Try(structField.metadata.getBoolean(key)).toOption
+    def getMetadataStringAsBoolean(key: String): Option[Boolean] = {
```
GeorgiChochov: Renaming the method seems to leak too much implementation details. I'd keep it as it was.

utils/src/test/scala/za/co/absa/enceladus/utils/types/TypedStructFieldSuite.scala
```diff
@@ -179,5 +182,13 @@ class TypedStructFieldSuite extends FunSuite {
...
 
-
+  test("Float type nullable, with default defined as Long and allowInfinity as binary Boolean") {
+    val fieldType = FloatType
+    val nullable = false
+    val field = createField(fieldType, nullable, Some(1000L), Map( MetadataKeys.allowInfinity->false ))
+    val typed = TypedStructField(field)
+    val errMsg = "java.lang.Long cannot be cast to java.lang.String"
+    val fail = Failure(new ClassCastException(errMsg))
+    checkField(typed, fieldType, fail, fail, nullable, Seq(ValidationError("allowInfinity metadata value of field 'test_field' is not Boolean in String format"), ValidationError(errMsg)))
```
Zejnilovic: this is actually over 180 chars.

utils/src/test/scala/za/co/absa/enceladus/utils/types/TypedStructFieldSuite.scala
```diff
@@ -15,21 +15,24 @@ class TypedStructFieldSuite extends FunSuite {
   private def createField(dataType: DataType,
                           nullable: Boolean = false,
                           default: Option[Any] = None,
-                          otherMetadata: Map[String, String] = Map.empty
+                          otherMetadata: Map[String, Any] = Map.empty
```
Zejnilovic: This guy is missing the copyright

utils/src/test/scala/za/co/absa/enceladus/utils/types/TypedStructFieldSuite.scala
```diff
@@ -143,17 +146,17 @@ class TypedStructFieldSuite extends FunSuite {
...
     val fail = Failure(new ClassCastException(errMsg))
     checkField(typed, fieldType, fail, fail, nullable, Seq(ValidationError(errMsg)))
   }
 
-  test("Float type nullable, with default defined in exponential notation") {
+  test("Float type nullable, with default defined in exponential notation, allowInfinity is set to true") {
     val fieldType = FloatType
     val nullable = true
-    val field = createField(fieldType, nullable, Some("314e-2")) // scalastyle:ignore null
+    val field = createField(fieldType, nullable, Some("314e-2"), Map( MetadataKeys.allowInfinity->"true" ))
```
GeorgiChochov:
```suggestion
    val field = createField(fieldType, nullable, Some("314e-2"), Map(MetadataKeys.allowInfinity -> "true"))
```
This looks like some sort of reverse spacing, lol

utils/src/test/scala/za/co/absa/enceladus/utils/types/TypedStructFieldSuite.scala
benedeki: Weird. Sometimes my template doesn't kick in, seems. Thanks for noticing, will add it.

utils/src/test/scala/za/co/absa/enceladus/utils/types/TypedStructFieldSuite.scala
benedeki: Even in tests? 😉

utils/src/main/scala/za/co/absa/enceladus/utils/validation/field/FractionalFieldValidator.scala
benedeki: Here it definitely can. Not expected to be inherited. (Unlike `ScalarFieldValidator` where I had to convert it to a class and add the object exactly for this reason)

utils/src/main/scala/za/co/absa/enceladus/utils/validation/field/ScalarFieldValidator.scala
benedeki: Class can be freely instantiated, not that there is a much point of it, the companion object implementing the class doess all the class offers.
But the restriction via `abstract` declaration seems unnecessarily restrictive.

utils/src/test/scala/za/co/absa/enceladus/utils/types/TypedStructFieldSuite.scala
benedeki: Will fix 😉

utils/src/main/scala/za/co/absa/enceladus/utils/implicits/StructFieldImplicits.scala
benedeki: Would `getMetadataBooleanString` serve better? I want to distinguish from reading an actual Boolean metadata value.

GeorgiChochov: No, the intent is clearer with `getMetadataStringAsBoolean`, let's keep it that way.

utils/src/main/scala/za/co/absa/enceladus/utils/implicits/StructFieldImplicits.scala
```diff
@@ -24,8 +24,8 @@ object StructFieldImplicits {
...
+      Try(structField.metadata.getString(key).toBoolean).toOption
```
yruslan: Why converting to `Option`? Returning a `Try` is a good practice when errors are expected.

utils/src/test/scala/za/co/absa/enceladus/utils/types/TypedStructFieldSuite.scala
```diff
@@ -15,21 +30,24 @@ class TypedStructFieldSuite extends FunSuite {
...
+        case Some(i: Int) => builder.putLong(key, i)
+        case Some(l: Long) => builder.putLong(key, l)
+        case Some(b: Boolean) => builder.putBoolean(key, b)
+        case Some(f: Float) => builder.putDouble(key, f)
+        case Some(d: Double) => builder.putDouble(key, d)
+        case Some(x) => builder.putString(key, x.toString)
+      }
+    }
+    val metadataBuilder: MetadataBuilder = otherMetadata.foldLeft(new MetadataBuilder()) (
+      (builder, data) => addMetadata(builder, data._1, Some(data._2)))
```
yruslan: Consider replacing `._1` and `._2` with matching (unapplying) the pair and give names to its elements.

benedeki: Hope the solution with named parameters address the issue well enough. I don't like using the numeric designation of tuple members, but here I considered it good trade-off for more concise code.

Issue #483, opened 2019-06-17
Conformance and standardisation share too much code/logic


Zejnilovic:
## Description
Standardization and conformance share too much code and logic. These could be extracted to some general methods and/or classes. This would also prompt clean up.

Now, when we want to change something we have to duplicate our code.

Zejnilovic:
@yruslan do you think this was solved by #1149? Or was that mostly combination of the two and deduplication would come later?

benedeki:
IMHO not by far. I actually think this should be turned into an Epic.
The merge was just putting the two module into one. Not mch shared code extraction has been done so far.
And then there is the whole Utils module.

yruslan:
Yeah, there is still a lot of duplicate code around Standardization and Conformance main functions that we can combine.

AdrianOlosutean:
Can this be closed?

PR #911, opened 2019-09-19
909 Allow processing empty datasets

Zejnilovic:
Run E2E locally. Works fine. Found out performance.PerformanceMetricTool also checks this.
```
19/09/19 13:47:25 ERROR performance.PerformanceMetricTools$: No successful records after running the Spark Application. Possibly the schema is incorrectly defined for the dataset.
```

I know code duplicates itself but there is #483 already for it. I would not like to set some weird standard that would later "bully" the person doing #483.

### DIFF
--- a/conformance/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
+++ b/conformance/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
@@ -36,6 +36,7 @@ import za.co.absa.enceladus.utils.fs.FileSystemVersionUtils
 import za.co.absa.enceladus.utils.performance.{PerformanceMeasurer, PerformanceMetricTools}
 import za.co.absa.enceladus.utils.time.TimeZoneNormalizer
 
+import scala.util.Try
 import scala.util.control.NonFatal
 
 object DynamicConformanceJob {
@@ -217,12 +218,8 @@ object DynamicConformanceJob {
       case None    => withPartCols.count
       case Some(p) => p
     }
+    if (recordCount == 0) { handleEmptyOutputAfterConformance }
 
-    if (recordCount == 0) {
-      val errMsg = "Empty output after running Dynamic Conformance."
-      AtumImplicits.SparkSessionWrapper(spark).setControlMeasurementError("Conformance", errMsg, "")
-      throw new IllegalStateException(errMsg)
-    }
     // ensure the whole path but version exists
     fsUtils.createAllButLastSubDir(pathCfg.publishPath)
 
@@ -244,6 +241,23 @@ object DynamicConformanceJob {
 
     if (isAutocleanStdFolderEnabled()) {
       fsUtils.deleteDirectoryRecursively(pathCfg.stdPath)
+    }
+  }
+
+  private def handleEmptyOutputAfterConformance(implicit spark: SparkSession): Unit = {
+    val areCountMeasurementsAllZero = Atum.getControMeasure.checkpoints
+      .flatMap(checkpoint =>
+        checkpoint.controls.filter(control =>
+          control.controlName.equalsIgnoreCase(za.co.absa.atum.core.Constants.controlTypeRecordCount)))
+      .forall(m => Try(m.controlValue.toString.toLong).getOrElse(1L) == 0L)
+
+    if (areCountMeasurementsAllZero) {
+      log.warn("Empty output after running Dynamic Conformance. Previous checkpoint show this is correct.")
+    } else {
+      val errMsg = "Empty output after running Dynamic Conformance, " +
+        "while previous checkpoints show non zero record count"
+      AtumImplicits.SparkSessionWrapper(spark).setControlMeasurementError("Standardization", errMsg, "")
+      throw new IllegalStateException(errMsg)
     }
   }
 
--- a/conformance/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
+++ b/conformance/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
@@ -251,7 +251,7 @@ object DynamicConformanceJob {
       .flatMap(checkpoint =>
         checkpoint.controls.filter(control =>
           control.controlName.equalsIgnoreCase(controlTypeRecordCount)))
-      .forall(m => Try(m.controlValue.toString.toLong).getOrElse(1L) == 0L)
+      .forall(m => Try(m.controlValue.toString.toDouble).toOption.contains(0D))
 
     if (areCountMeasurementsAllZero) {
       log.warn("Empty output after running Dynamic Conformance. Previous checkpoints show this is correct.")
--- a/conformance/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
+++ b/conformance/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
@@ -218,7 +218,7 @@ object DynamicConformanceJob {
       case None    => withPartCols.count
       case Some(p) => p
     }
-    if (recordCount == 0) { handleEmptyOutputAfterConformance }
+    if (recordCount == 0) { handleEmptyOutputAfterConformance() }
 
     // ensure the whole path but version exists
     fsUtils.createAllButLastSubDir(pathCfg.publishPath)
@@ -244,15 +244,17 @@ object DynamicConformanceJob {
     }
   }
 
-  private def handleEmptyOutputAfterConformance(implicit spark: SparkSession): Unit = {
+  private def handleEmptyOutputAfterConformance()(implicit spark: SparkSession): Unit = {
+    import za.co.absa.atum.core.Constants._
+
     val areCountMeasurementsAllZero = Atum.getControMeasure.checkpoints
       .flatMap(checkpoint =>
         checkpoint.controls.filter(control =>
-          control.controlName.equalsIgnoreCase(za.co.absa.atum.core.Constants.controlTypeRecordCount)))
+          control.controlName.equalsIgnoreCase(controlTypeRecordCount)))
       .forall(m => Try(m.controlValue.toString.toLong).getOrElse(1L) == 0L)
 
     if (areCountMeasurementsAllZero) {
-      log.warn("Empty output after running Dynamic Conformance. Previous checkpoint show this is correct.")
+      log.warn("Empty output after running Dynamic Conformance. Previous checkpoints show this is correct.")
     } else {
       val errMsg = "Empty output after running Dynamic Conformance, " +
         "while previous checkpoints show non zero record count"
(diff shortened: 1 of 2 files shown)


conformance/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
```diff
@@ -247,6 +244,23 @@ object DynamicConformanceJob {
     }
   }
 
+  private def handleEmptyOutputAfterConformance(implicit spark: SparkSession): Unit = {
```
GeorgiChochov:
```suggestion
  private def handleEmptyOutputAfterConformance()(implicit spark: SparkSession): Unit = {
```

conformance/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
```diff
@@ -217,12 +218,8 @@ object DynamicConformanceJob {
       case None    => withPartCols.count
       case Some(p) => p
     }
+    if (recordCount == 0) { handleEmptyOutputAfterConformance }
```
GeorgiChochov:
```suggestion
    if (recordCount == 0) { handleEmptyOutputAfterConformance() }
```
Needs the parens as it is not an accessor or a field, but a method with side effects (throwing an exception) on top.

conformance/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
```diff
@@ -247,6 +244,23 @@ object DynamicConformanceJob {
     }
   }
 
+  private def handleEmptyOutputAfterConformance(implicit spark: SparkSession): Unit = {
+    val areCountMeasurementsAllZero = Atum.getControMeasure.checkpoints
+      .flatMap(checkpoint =>
+        checkpoint.controls.filter(control =>
+          control.controlName.equalsIgnoreCase(za.co.absa.atum.core.Constants.controlTypeRecordCount)))
```
lokm01: Maybe import constants?

conformance/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
```diff
@@ -247,6 +244,23 @@ object DynamicConformanceJob {
...
+      .forall(m => Try(m.controlValue.toString.toLong).getOrElse(1L) == 0L)
```
lokm01: .getOrElse(1L) == 0L
This could probably be a bit more intuitive ;)

Otherwise interesting idea.

conformance/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
```diff
@@ -247,6 +244,23 @@ object DynamicConformanceJob {
...
+
+    if (areCountMeasurementsAllZero) {
+      log.warn("Empty output after running Dynamic Conformance. Previous checkpoint show this is correct.")
```
lokm01: checkpoint**S**

standardization/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
```diff
@@ -299,15 +297,33 @@ object StandardizationJob {
     stdRenameSourceColumns.writeInfoFile(pathCfg.outputPath)
   }
 
-  private def ensureSplittable(df: DataFrame, path: String, schema: StructType)(implicit spark: SparkSession, fsUtils: FileSystemVersionUtils) = {
+  private def handleEmptyOutputAfterStandardization(implicit spark: SparkSession): Unit = {
+    val areCountMeasurementsAllZero = Atum.getControMeasure.checkpoints
+      .flatMap(checkpoint =>
+        checkpoint.controls.filter(control =>
+          control.controlName.equalsIgnoreCase(za.co.absa.atum.core.Constants.controlTypeRecordCount)))
```
lokm01: Same here

conformance/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
GeorgiChochov: Can `toString.toLong` be replaced by pattern matching?

conformance/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
lokm01: I think it reads better too

conformance/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
yruslan: Might be a lot of things to match since the input is `Any`. `toString()` is safer here in my opinion.

yruslan: Well, not 'safer', more like 'simpler' while covering all cases something can go wrong I can imagine.

yruslan: I like this approach since it is very short. The alternative is a pattern match against `Success` / `Failure` which might be too verbose for this case.

Zejnilovic: Yeah, this is idea of the one and only (mostly because names on GitHub are unique) @yruslan. But I like it, it is simple and covers all the corners.

conformance/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
Zejnilovic: sure

conformance/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
GeorgiChochov: I agree that it's not the most intuitive thing but it's probably alright.

GeorgiChochov:
Seems good to me, just need come to an agreement about `.getOrElse(1L) == 0L`.

conformance/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
Zejnilovic: Add a comment?

```
Anything that cannot be converted to Long can be count as a positive count and therefore not 0.
```

conformance/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
```diff
@@ -217,12 +218,8 @@ object DynamicConformanceJob {
       case None    => withPartCols.count
       case Some(p) => p
     }
+    if (recordCount == 0) { handleEmptyOutputAfterConformance() }
```
benedeki: Tiny: I find the one-liner withing curly braces weird.
I would prefer the full:
```
if (recordCount == 0) {
  handleEmptyOutputAfterConformance()
}
```
or if to keep short because simple then
`if (recordCount == 0) handleEmptyOutputAfterConformance()`
The current state is a strange mixture, IMO 😉

conformance/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
```diff
@@ -247,6 +244,25 @@ object DynamicConformanceJob {
     }
   }
 
+  private def handleEmptyOutputAfterConformance()(implicit spark: SparkSession): Unit = {
```
benedeki: Tiny: Wouldn't be a `checkEmptyOutputAfterConformance` be a better suiting name?

conformance/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
benedeki: What about 0.0?

benedeki:
```suggestion
      .forall(m => Try(m.controlValue.toString.toLong).toOption.contains(0L))
```

Zejnilovic: Seems like I will need to switch it to double first
```scala
scala> "0.0".toLong
java.lang.NumberFormatException: For input string: "0.0"
  at java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
  at java.lang.Long.parseLong(Long.java:589)
  at java.lang.Long.parseLong(Long.java:631)
  at scala.collection.immutable.StringLike$class.toLong(StringLike.scala:277)
  at scala.collection.immutable.StringOps.toLong(StringOps.scala:29)
  ... 51 elided

scala> "0.0".toDouble.toLong
res12: Long = 0
```

conformance/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
Zejnilovic: No, because we know it is empty. That is why we are calling this method.

conformance/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
benedeki: No need to bother converting to `Long` after that, I think.
```
scala> Option("-0.0".toDouble).contains(0D)
res5: Boolean = true
```

Zejnilovic: I used your option. In the end, we only care about the zero. Did not know about the contains functionality.

benedeki:
Well, I have approved this. But came up with a possibility - I am not sure it can happen and how should be evaluated if it would.
What if the control metric would be `null`?

conformance/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
GeorgiChochov: Can record count even be anything other than a natural number?

yruslan: Control metrics come from source systems, better be defensive I think.

Replying to @benedeki 's question here as well. A control value shouldn't be `null` (never happened yet anyway). If it happens and the job throws a `NullPointerException`, I'd say it is ok. The raw info file needs to be fixed then.

GeorgiChochov: I think we can consider this resolved then.

Issue #912, opened 2019-09-20
Migrations CLI tools 


yruslan:
## Describe the bug

Migrations CLI tools do not log what they are doing since no SLF4J API implementation is provided. Migrations work fine in Menas since embedded Spark provides a SLF4J API implementation.

```
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
```

## To Reproduce
Run a migration from CLI

## Expected behaviour
CLI tools should log their actions to the terminal.

PR #913, opened 2019-09-20
#912 Fix logging issues in the migration CLI tool.

yruslan:


migrations/src/main/scala/za/co/absa/enceladus/migrations/continuous/migrate01/EntityMigrator.scala
```diff
@@ -57,17 +60,23 @@ abstract class EntityMigrator(databaseOld: MongoDatabase,
...
       objectId.foreach(id => {
+        count += 1
         if (!repoNew.doesDocumentExist(id)) {
+          migratedCount += 1
           migrateEntity(entityOld, id, repoNew)
         }
       })
     })
+
+    log.info(s"Migrated $migratedCount of $count entities from $collectionOld to $collectionNew.")
```
Zejnilovic: Getting message Migrated 10 out of 100 entities might scare a person. But in this case, it means we found just 10 new entities in the old DB, right?

PR #919, opened 2019-09-20
918 Add missing licenses

GeorgiChochov:


pom.xml
```diff
@@ -461,6 +461,7 @@
                                 <exclude>**/*.MF</exclude>
                                 <exclude>**/*.crc</exclude>
                                 <exclude>**/*.parquet</exclude>
+                                <exclude>**/example*.xml</exclude>
```
GeorgiChochov: There are some `*.xml` files in `testutils`:
- `testutils/src/test/resources/xml_examples/example1.xml`
- `testutils/src/test/resources/xml_examples/example2.xml`

These shouldn't need a license header, but we don't want to be too permissive and exclude all `*.xml` files or we might end up with `pom.xml` without a header.

utils/src/test/scala/za/co/absa/enceladus/utils/types/TypedStructFieldSuite.scala
```diff
@@ -1,3 +1,18 @@
+/*
```
benedeki: This has been added as part of #908.
Can you exclude it please, to avoid merge conflict/need for merge commit?

GeorgiChochov: How about we merge yours first and I'll rebase this branch? I'm not scared of merge conflicts.

benedeki: I thought you would be changing the `xml` failing to notice, it's already specific.

PR #930, opened 2019-09-26
Bugfix/921 migrate data by chunks

yruslan:


migrations/src/main/scala/za/co/absa/enceladus/migrations/framework/dao/ScalaMongoImplicits.scala
```diff
@@ -41,4 +41,23 @@ object ScalaMongoImplicits {
   implicit class FindObservableExecutor[T](observable: FindObservable[T]) {
     def execute(): Seq[T] = Await.result(observable.toFuture, executionTimeout)
   }
+
+  implicit class FindObservableTraversable[T](observable: FindObservable[T]) {
+    def foreach(transform: T => Unit): Unit = {
```
yruslan: When implementing the same for the migration framework I've found that MongoDB Scala Driver actually has `foreach()` for observables. And it does the same thing, lol.
Removing this 'fancy' piece of code keeps the migration method do the same. 😄

yruslan: It seems that it is not that simple. The driver's `foreach()` is not synchronous so that the observable still needs to be handled (btw, this is not what I'd expect from `foreach()`).

Keeping this implementation. Maybe the method worth renaming to `syncForeach()` so it won't be confused with the one provided by the driver?

Issue #398, opened 2019-05-15
sourcecolumn metadata might not work as expected


benedeki:
## Describe the bug
`sourcecolumn` metadata entry doesn't seem to result in renaming of the column. The column stays the same name as the specified in `sourcecolumn` instead of the name of field in the metadata belongs to. Verify.
Also this can cause column names duplication.

## To Reproduce
Use schema:
```
{
    "name": "name",
    "type": "string",
    "nullable": false,
    "metadata": {
        "sourcolumn": "source_field"
    }
}
```
Output column is _source_field_

## Expected behaviour
The output field should be named _name_

benedeki:
Bug confirmed!

benedeki:
Well, the whole thing kind of works. But there are several renames there and back, sometimes skipping in case of inner structures, with a possible change of name clashes. Also, suspition is that due to that some renames are not registered with ATUM. Worth refactoring and clean up.

benedeki:
Proper fix tasks:
- [x] `TypeParser` proper renames
- [ ] No renames elsewhere
- [ ] Error columns aggregation in `StandardizationInterpreter`
- [ ] Registering renames with ATUM
- [ ] Field name validations in `FieldValidator`
- [ ] Unsplitable files issue

benedeki:
**For Release Noted**
* Renaming in Standardization (using `sourcecolumn` metadata key) got increased robustness
  * Rename does not run the danger of name clashes anymore
  * _ErrorCol_ name clash danger removed 
  * Even embeded fields renames are registered with ATUM
  * One field can be a source for multiple standardized fields, if needed for whatever reason

PR #938, opened 2019-09-30
#398: sourcecolumn metadata value doesn't work

benedeki:
* to produce the output data frame the desired schema for renaming is used, which contains the metadata
* to prevent error column from being dropped it's added explicitly

### DIFF
--- a/conformance/pom.xml
+++ b/conformance/pom.xml
@@ -20,7 +20,7 @@
     <parent>
         <groupId>za.co.absa</groupId>
         <artifactId>enceladus-parent</artifactId>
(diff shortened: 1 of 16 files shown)


benedeki:
> 
> 
> Looks good to me, but what about the tests in `StandardizationInterpreterSuite`?

The tests are not affected in any way by this change. It's done on the top level (basically withing the `main` function) - something that the proper fix should also address. It's weird to do such basic low level data operation within the main script, not mentioning that it leads to lack of tests/testability.
Or you meant the note in a different way?

PR #946, opened 2019-10-01
945 Handle request body serialization explictly

GeorgiChochov:
WIP until I have Menas handle empy Runs.


menas/src/main/scala/za/co/absa/enceladus/menas/services/RunService.scala
```diff
@@ -160,15 +161,38 @@ class RunService @Autowired()(runMongoRepository: RunMongoRepository)
...
+      validation.withError("datasetVersion", NotSpecified)
+    } else {
+      validation
+    }
+  }
+
+  private def validateDatasetName(run: Run, validation: Validation) = {
+    if (Option(run.dataset).isEmpty) {
+      validation.withError("dataset", NotSpecified)
+    } else {
```
Zejnilovic: Why is there not more name specific validations? Some regexp?

Issue #941, opened 2019-09-30
Useful metadata is not available if a job has failed


yruslan:
## Describe the bug
Standardization and Conformance jobs add metadata to Run objects. It is helpful for debugging and auditing runs. For instance, command line parameters passed to a job are added as metadata.

However, if a job has failed, it does not reach the point when it adds command line parameters and other useful metadata fields.

## Expected behaviour
Add as much metadata as possible at the beginning of jobs so if a job fails the metadata is available for investigating issues.

PR #948, opened 2019-10-02
#941 Add generic job information at the beginning of a job.

yruslan:
Also add a couple of additional metadata fields that might be helpful in certain circumstances.

### DIFF
--- a/utils/src/main/scala/za/co/absa/enceladus/utils/performance/PerformanceMetricTools.scala
+++ b/utils/src/main/scala/za/co/absa/enceladus/utils/performance/PerformanceMetricTools.scala
@@ -28,15 +28,62 @@ object PerformanceMetricTools {
 
   private val logger: Logger = LoggerFactory.getLogger(this.getClass)
 
+  /**
+    * Adds general job information to Atum's metadata that will end up in an info file.
+    *
+    * The method should be ran at the beginning of a job so that this general information is available for debugging.
+    *
+    * @param spark         A Spark Session
+    * @param optionPrefix  A prefix for all performance metrics options, e.g. 'std' for Standardization and 'conform' for Conformance
+    * @param inputPath     A path to an input directory of the job
+    * @param outputPath    A path to an output directory of the job
+    * @param loginUserName A login user name who performed the job
+    *
+    **/
+  def addJobInfoToAtumMetadata(optionPrefix: String,
+                                inputPath: String,
+                                outputPath: String,
+                                loginUserName: String,
+                                cmdLineArgs: String)
+                               (implicit spark: SparkSession): Unit = {
+    // Spark job configuration
+    val sc = spark.sparkContext
+
+    // The number of executors minus the driver
+    val numberOfExecutors = sc.getExecutorMemoryStatus.keys.size - 1
+
+    val fsUtils = new FileSystemVersionUtils(spark.sparkContext.hadoopConfiguration)
+    // Directory sizes and size ratio
+    val inputDirSize = fsUtils.getDirectorySize(inputPath)
+
+    addSparkConfig(optionPrefix, "spark.driver.memory", "driver_memory")
+    addSparkConfig(optionPrefix, "spark.driver.cores", "driver_cores")
+    addSparkConfig(optionPrefix, "spark.driver.memoryOverhead", "driver_memory_overhead")
+    addSparkConfig(optionPrefix, "spark.executor.memory", "executor_memory")
+    addSparkConfig(optionPrefix, "spark.executor.cores", "executor_cores")
+    addSparkConfig(optionPrefix, "spark.executor.memoryOverhead", "executor_memory_overhead")
+    addSparkConfig(optionPrefix, "spark.submit.deployMode", "yarn_deploy_mode")
+    addSparkConfig(optionPrefix, "spark.master", "spark_master")
+
+    Atum.setAdditionalInfo(s"${optionPrefix}_cmd_line_args" -> cmdLineArgs)
+    Atum.setAdditionalInfo(s"${optionPrefix}_input_dir" -> inputPath)
+    Atum.setAdditionalInfo(s"${optionPrefix}_output_dir" -> outputPath)
+    Atum.setAdditionalInfo(s"${optionPrefix}_input_dir_size" -> inputDirSize.toString)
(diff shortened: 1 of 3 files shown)


utils/src/main/scala/za/co/absa/enceladus/utils/performance/PerformanceMetricTools.scala
```diff
@@ -89,6 +119,19 @@ object PerformanceMetricTools {
     }
   }
 
+  /**
+    * Adds a Spark config key-value to Atum metadata if such key is present in Spark runtime config.
+    *
+    * @param optionPrefix A prefix for a job (e.g. "std", "conf", etc.)
+    * @param sparkKey A Spark configuration key
+    * @param atumKey An Atum metadata key
```
Zejnilovic: If you are aligning other comments, why not these?

utils/src/main/scala/za/co/absa/enceladus/utils/performance/PerformanceMetricTools.scala
```diff
@@ -28,15 +28,62 @@ object PerformanceMetricTools {
...
+    addSparkConfig(optionPrefix, "spark.master", "spark_master")
+
+    Atum.setAdditionalInfo(s"${optionPrefix}_cmd_line_args" -> cmdLineArgs)
+    Atum.setAdditionalInfo(s"${optionPrefix}_input_dir" -> inputPath)
+    Atum.setAdditionalInfo(s"${optionPrefix}_output_dir" -> outputPath)
+    Atum.setAdditionalInfo(s"${optionPrefix}_input_dir_size" -> inputDirSize.toString)
+    Atum.setAdditionalInfo(s"${optionPrefix}_enceladus_version" -> ProjectMetadataTools.getEnceladusVersion)
+    Atum.setAdditionalInfo(s"${optionPrefix}_application_id" -> spark.sparkContext.applicationId)
+    Atum.setAdditionalInfo(s"${optionPrefix}_username" -> loginUserName)
+    Atum.setAdditionalInfo(s"${optionPrefix}_executors_num" -> s"$numberOfExecutors")
```
Zejnilovic: Isn't a number of executors also a `spakrConfig` option?

utils/src/main/scala/za/co/absa/enceladus/utils/performance/PerformanceMetricTools.scala
```diff
@@ -28,15 +28,62 @@ object PerformanceMetricTools {
...
+    val inputDirSize = fsUtils.getDirectorySize(inputPath)
+
+    addSparkConfig(optionPrefix, "spark.driver.memory", "driver_memory")
+    addSparkConfig(optionPrefix, "spark.driver.cores", "driver_cores")
+    addSparkConfig(optionPrefix, "spark.driver.memoryOverhead", "driver_memory_overhead")
+    addSparkConfig(optionPrefix, "spark.executor.memory", "executor_memory")
+    addSparkConfig(optionPrefix, "spark.executor.cores", "executor_cores")
+    addSparkConfig(optionPrefix, "spark.executor.memoryOverhead", "executor_memory_overhead")
+    addSparkConfig(optionPrefix, "spark.submit.deployMode", "yarn_deploy_mode")
+    addSparkConfig(optionPrefix, "spark.master", "spark_master")
```
Zejnilovic: can we somehow pull the `--conf` data as well?

utils/src/main/scala/za/co/absa/enceladus/utils/performance/PerformanceMetricTools.scala
yruslan: It's not always there. The way it gets the number of executors now is a "safer way".

Most of the listed options are absent at runtime unless explicitly specified in `spark-submit`. For instance, if the number of cores is not specified, this configuration key will be missing: `spark.executor.cores`. I assume that Spark will use the default value which is `1` for this setting.

utils/src/main/scala/za/co/absa/enceladus/utils/performance/PerformanceMetricTools.scala
yruslan: Not quite following. Do you mean values such as Spline URL etc?

Zejnilovic: Yes, those. They can have things in that affect the job, right?

utils/src/main/scala/za/co/absa/enceladus/utils/performance/PerformanceMetricTools.scala
Zejnilovic: Cool

utils/src/main/scala/za/co/absa/enceladus/utils/performance/PerformanceMetricTools.scala
yruslan: There are not many config options available at runtime. I've included these that we already use and know they affect performance.
Specifically, JVM options I'm reluctant to include due to security reasons.

utils/src/main/scala/za/co/absa/enceladus/utils/performance/PerformanceMetricTools.scala
```diff
@@ -28,15 +28,62 @@ object PerformanceMetricTools {
...
+
+    val fsUtils = new FileSystemVersionUtils(spark.sparkContext.hadoopConfiguration)
+    // Directory sizes and size ratio
+    val inputDirSize = fsUtils.getDirectorySize(inputPath)
+
+    addSparkConfig(optionPrefix, "spark.driver.memory", "driver_memory")
+    addSparkConfig(optionPrefix, "spark.driver.cores", "driver_cores")
+    addSparkConfig(optionPrefix, "spark.driver.memoryOverhead", "driver_memory_overhead")
+    addSparkConfig(optionPrefix, "spark.executor.memory", "executor_memory")
+    addSparkConfig(optionPrefix, "spark.executor.cores", "executor_cores")
```
DzMakatun: Very happy to see the executor cores included! 👍

yruslan:
### For release notes:
* Run metadata is added at the beginning of a Spark job so if the job fails users have additional information for debugging
* Added additional Spark configuration parameters to run's metadata.

Issue #951, opened 2019-10-03
Users getting unexpectedly shited values in timestamp fields after standardization


Zejnilovic:
## Describe the bug
Users getting unexpectedly shited values in timestamp fields after standardization

## To Reproduce
Steps to reproduce the behavior OR commands run:
1. Set up standardization task that includes a timestamp in the schema
2. the timestamp should have no time zone specified (not in _pattern_ nor in metadata field _timezone_)
3. Run standardization
4. Check output in a session of local time-zone. Timestamp is shifted, while it's not expected

benedeki:
The hotfix #954 addresses the imminent issue, the users are facing, that their dates are shifted incorrectly unless they specify the time zone for all timestamp fields in all schemas.

It **doesn't** address the issue, that if the `timezone` configuration is changed, weird things my start to happen with timestamp standardizations.

benedeki:
The original bug was cloned into #963 and this one rewritten to correspond to the fix that was actually provided.

PR #954, opened 2019-10-03
#951: Global time zone setting for timestamps and dated

benedeki:
* Definition of the two new settings for Standardization and their usage in TypeParser
* Increased project version

### DIFF
--- a/standardization/src/main/resources/application.conf.template
+++ b/standardization/src/main/resources/application.conf.template
@@ -18,7 +18,7 @@ standardized.hdfs.path="/tmp/conformance-output/standardized-{0}-{1}-{2}-{3}"
 #system-wide time zone
 timezone="UTC"
 #time zone used for timestamp fields standardization if none specified in the schema
-#defaultTimestampTimeZone
+#defaultTimestampTimeZone="Africa/Johannesburg"
 #time zone used for date fields standardization if none specified in the schema; recommended not to define
-#defaultDateTimeZone
+#defaultDateTimeZone="UTC"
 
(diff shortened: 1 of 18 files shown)


benedeki:
**For Release Notes**
Definition of the two new settings for Standardization (`application.conf`)
`defaultTimestampTimeZone` - time zone used for timestamp fields standardization if none specified in the schema
`defaultDateTimeZone` - time zone used for date fields standardization if none specified in the schema; recommended not to define

benedeki:
_Hotfix notes:_
I am aware that the ` private val defaultTimeZone: Option[String]` is defined both in `DateParser` and `TimestampParser` while it could be in their common ancestor.
Also the way the setting is done, it's hard to test other then with an E2E test.

standardization/src/main/resources/application.conf.template
```diff
@@ -17,3 +17,8 @@ standardized.hdfs.path="/tmp/conformance-output/standardized-{0}-{1}-{2}-{3}"
 
 #system-wide time zone
 timezone="UTC"
+#time zone used for timestamp fields standardization if none specified in the schema
+#defaultTimestampTimeZone
```
Zejnilovic: Does this equal nothing?

Zejnilovic: Default is UTC, set by the code in TimeZoneNormalizer, right?

benedeki: > 
> 
> Does this equal nothing?

Sorry. Added a better example.

benedeki: > 
> 
> Default is UTC, set by the code in TimeZoneNormalizer, right?

Default is the value of `timezone` setting. Which in the current state of code should not deviate from _UTC_.

Issue #953, opened 2019-10-03
Split between JVM and Spark additional options in run scripts


yruslan:
## Describe the bug
Currently, the `ADDITIONAL_SPARK_CONF` variable in an environment settings file allows setting only JVM options (e.g. `-D...`). 

## To Reproduce
* Set `ADDITIONAL_SPARK_CONF` to a non-empty value and run `run_standardization.sh` with `--dry-run` option.
* See that the value is passed to the JVM section of the command line

## Expected behaviour
While setting JVM options is helpful we need to be able to set Spark options as well. The solution to that is to split this variable into 2:
*  `ADDITIONAL_SPARK_CONF` for setting Spark config options to spark-submit (e.g. `--conf spark.something=something`)
* `ADDITIONAL_JVM_CONF` for providing options in form of `-D=...`

yruslan:
For release notes: Fixed handling of environment-specific additional Spark configuration options by distinguishing between Spark configuration options and JVM configuration options.

Issue #929, opened 2019-09-25
Validate mandatory options in helper scripts


yruslan:
## Background
Currently, helper scripts do not validate that all mandatory command line parameters are supplied. Validation happens only from inside the job. This means that all nodes are already allocated for executors. It would be nice if scripts can do basic validation as well so that a job that does not provide mandatory arguments won't run at all.

This is especially useful for checking if either a keytab or a menas credentials file is provided.

## Feature
Helper scripts should validate that all mandatory command line parameters are provided.

yruslan:
If neither of the validation parameters is specified the script prints the following:
```
Either --menas-credentials-file or --menas-auth-keytab should be specified
```

yruslan:
For release notes: Added validation to run scripts to ensure authentication credentials are provided.

PR #964, opened 2019-10-07
Feature/922 add more options to scripts

### LINKED ISSUE #922: Add more options to control resources to helper scripts
## Background
Running Standardization and Conformance on big files revealed options that are essential to run these jobs efficiently.

## Feature
Need to expend `spark-submit` resource request options. Add support for the ollowing:
- `--executor-cores`
- `--conf spark.executor.memoryOverhead=2048`
- `--conf spark.memory.fraction=0.1`

yruslan:
Fixes #922, #929, #953.

### DIFF
--- a/scripts/bash/run_enceladus.sh
+++ b/scripts/bash/run_enceladus.sh
@@ -219,25 +219,25 @@ set -- "${POSITIONAL[@]}" # restore positional parameters
 # Validation
 VALID="1"
 validate() {
-    if [ -z "$1" ]; then
-        echo "Missing mandatory option $2"
+    if [[ -z "$2" ]]; then
+        echo "Missing mandatory option $1"
         VALID="0"
     fi
 }
 
 validate_either() {
-    if [[ -z "$1" && -z "$3" ]]; then
-        echo "Either $2 or $4 should be specified"
+    if [[ -z "$2" && -z "$4" ]]; then
+        echo "Either $1 or $3 should be specified"
         VALID="0"
     fi
 }
 
-validate "$NUM_EXECUTORS" "--num-executors"
-validate "$DATASET_NAME" "--dataset-name"
-validate "$DATASET_VERSION" "--dataset-version"
-validate "$REPORT_DATE" "--report-date"
+validate "--num-executors" "$NUM_EXECUTORS"
+validate "--dataset-name" "$DATASET_NAME"
+validate "--dataset-version" "$DATASET_VERSION"
+validate "--report-date" "$REPORT_DATE"
 
-validate_either "$MENAS_CREDENTIALS_FILE" "--menas-credentials-file" "$MENAS_AUTH_KEYTAB" "--menas-auth-keytab"
+validate_either "--menas-credentials-file" "$MENAS_CREDENTIALS_FILE" "--menas-auth-keytab" "$MENAS_AUTH_KEYTAB"
 
 if [[ "$MASTER" != "yarn" ]]; then
   echo "Master '$MASTER' is not allowed. The only allowed master is 'yarn'."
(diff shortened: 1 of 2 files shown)


scripts/bash/run_enceladus.sh
```diff
@@ -208,11 +225,20 @@ validate() {
     fi
 }
 
+validate_either() {
```
Zejnilovic: This is a bit weird. Because usually, 1st is the name of parameter and then the value. What lead you to switch the order?

scripts/bash/run_enceladus.sh
```diff
@@ -231,6 +257,13 @@ add_to_cmd_line() {
     fi
 }
 
+# Puts Spark configuration properties to the command line
+add_spark_conf_cmd() {
```
Zejnilovic: In some `ifs` you use the single bracket in some you use double ones. There is a difference between them. [Bash manual for more](https://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html#Conditional-Constructs)

scripts/bash/run_enceladus.sh
```diff
@@ -254,11 +287,16 @@ add_to_cmd_line "--master" ${MASTER}
...
 add_to_cmd_line "--driver-cores" ${DRIVER_CORES}
 add_to_cmd_line "--driver-memory" ${DRIVER_MEMORY}
 
+# Adding Spark config options
+add_spark_conf_cmd "spark.executor.memoryOverhead" ${CONF_SPARK_EXECUTOR_MEMORY_OVERHEAD}
+add_spark_conf_cmd "spark.memory.fraction" ${CONF_SPARK_MEMORY_FRACTION}
+
 # Adding JVM configuration, entry point class name and the jar file
-CMD_LINE="${CMD_LINE} --conf \"${CONF} ${ADDITIONAL_SPARK_CONF}\" --class ${CLASS} ${JAR}"
+CMD_LINE="${CMD_LINE} ${ADDITIONAL_SPARK_CONF} --conf \"${JVM_CONF} ${ADDITIONAL_JVM_CONF}\" --class ${CLASS} ${JAR}"
```
Zejnilovic: This seems a bit weird to me. On one hand, you have `ADDITIONAL_SPARK_CONF` where you ask for spark config options and then you have the lines `295` and `296` where you ask for specific spark config options. Can these be written so that they overwrite themselves in case of both of them being specified? Maybe add a check in `add_spark_conf_cmd` with a regex that would overwrite the value?

yruslan: * Specific Spark options are used to fine-tune each individual job (e.g. specify memory overhead for super big files).
* Additional Spark configuration options are used for additional environment configuration, e.g. to use scripts from containers (`spark.driver.host`, for example)

Zejnilovic: I understand that, but I think you are forcing users to understand the innards of the script, otherwise, they could specify the `--conf spark.executor.memoryOverhead` twice.  I am suggesting we "handle" that.

yruslan: I see now, thanks. Probably we need to handle it so that options provided by the command line override ones specified in `enceladus_env.sh`. I'm going to check if it behaves like this in this implementation. 

In this implementation command line provided options go after the ones specified in `enceladus_env.sh`. I'm going to check which option becomes effective.

scripts/bash/run_enceladus.sh
yruslan: Yeah, it makes sense to revert the order in `valudate()` and `validate_either()`.

scripts/bash/run_enceladus.sh
yruslan: Yeah, I usually go with single brackets is it is sufficient for the task (since it is 'more portable'), and go for double ones when what I want is easier to achieve that way.

Would you prefer it to be consistent across our scripts? If yes, I'd use double ones.
Maybe in different PR...

scripts/bash/run_enceladus.sh
yruslan: Thanks for spotting this! These are valid points. I confirm that the current logic is reverted.

When you specify the same config key twice with different values, Spark sets the second (latter) value to be the effective configuration value.
E.g. for
```
spark-submit --conf spark.executor.memoryOverhead=500M --conf spark.executor.memoryOverhead=600M
```
Spark going to set
```
spark.executor.memoryOverhead=600M
```
Will change the script so that the command line provided options take precedence over environment defaults.

yruslan:
@Zejnilovic, thanks for the tip on `--conf spark.logConf=true`. Helped to debug script config issues. Does it make sense to include it in the script for every job run so that all config key-values are also in the log? Seems very helpful.

Zejnilovic:
> @Zejnilovic, thanks for the tip on `--conf spark.logConf=true`. Helped to debug script config issues. Does it make sense to include it in the script for every job run so that all config key-values are also in the log? Seems very helpful.

I would add it. Seems helpful to me. And you see what is actually running and not what we hope is running.

Issue #962, opened 2019-10-05
ScalaMongoImplicits.FindObservableTraversable is found obsolete by the compiler


benedeki:
## Describe the bug
```
Warning:(55, 52) Adaptation of argument list by inserting () has been deprecated: this is unlikely to be what you want.
        signature: Promise.success(value: T): Promise.this.type
  given arguments: <none>
 after adaptation: Promise.success((): Unit)
        override def onComplete(): Unit = p.success()
```
## To Reproduce
Compile with Deprecation Warnings on

## Expected behaviour
Get rid of the warning.

PR #967, opened 2019-10-08
#962 Fix a deprecation warning.

yruslan:
Adaptation of argument list by inserting () has been deprecated: this is unlikely to be what you want.
        signature: Promise.success(value: T): Promise.this.type
  given arguments: <none>
 after adaptation: Promise.success((): Unit)


benedeki:
Checked, the warning is gone 👍

Issue #968, opened 2019-10-09
Exclude metadata when measuring input/output sizes


DzMakatun:
## Background [Optional]
Aside from data, the directories contain the following files:
RAW: 
    _INFO       \~1kB

STD:  
   _LINEAGE \~120kB
   _INFO       \~2kB 
   _SUCESS =0B

This introduces a significant error when small files are processed (e.g. 100kB - 1 MB, which is quite often). It skews performance analysis for small files.  For example, std_size_ratio may become more than 100%. 

## Question
Is it possible to exclude metadata files when measuring input/output sizes?

DzMakatun:
@yruslan

yruslan:
Definitely. Good spot!

yruslan:
For release notes: add metadata for input and output data sizes. These are similar to input and output directory sizes but do not include hidden files (_INFO, _LINEAGE, etc).

PR #970, opened 2019-10-10
#968 Ignore hidden files when calculating output directory size

yruslan:
This adds additional metadata fields '*_data_size'.

### DIFF
--- a/utils/src/main/scala/za/co/absa/enceladus/utils/fs/FileSystemVersionUtils.scala
+++ b/utils/src/main/scala/za/co/absa/enceladus/utils/fs/FileSystemVersionUtils.scala
@@ -153,28 +153,31 @@ class FileSystemVersionUtils(conf: Configuration) {
     */
   def getDirectorySizeNoHidden(path: String): Long = {
     def getDirSizeHelper(f: Path): Long = {
-      val status = fs.getFileStatus(f)
-      if (status.isFile) {
-        status.getLen
-      } else {
-        var totalLength = 0L
-        for (fileStatus <- fs.listStatus(f)) {
-          val fileName = fileStatus.getPath.getName
-          if (!fileName.startsWith("_") && !fileName.startsWith(".")) {
-            val length = if (fileStatus.isDirectory) {
-              getDirSizeHelper(fileStatus.getPath)
-            }
-            else {
-              fileStatus.getLen
-            }
-            totalLength += length
+      var totalLength = 0L
+      for (fileStatus <- fs.listStatus(f)) {
+        val fileName = fileStatus.getPath.getName
+        if (!fileName.startsWith("_") && !fileName.startsWith(".")) {
+          val length = if (fileStatus.isDirectory) {
+            getDirSizeHelper(fileStatus.getPath)
           }
+          else {
+            fileStatus.getLen
+          }
+          totalLength += length
         }
-        totalLength
       }
+      totalLength
     }
 
-    getDirSizeHelper(new Path(path))
+    val fsPath = new Path(path)
+    val status = fs.getFileStatus(fsPath)
+
+    if (status.isFile) {
+      // If a specific file is provided return its length even if this file is hidden.
+      status.getLen
+    } else {
+      getDirSizeHelper(new Path(path))
+    }
   }
 
   /**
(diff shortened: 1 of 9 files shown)


utils/src/test/scala/za/co/absa/enceladus/FsUtilsSpec.scala
```diff
@@ -49,4 +49,25 @@ class FsUtilsSpec extends FlatSpec with Matchers with SparkTestBase {
...
+    val dirSize = fsUtils.getDirectorySize("src/test/resources/test_data/test_dir2")
+    assert(dirSize == 87L)
+  }
+
+  "getDirectorySizeNoHidden" should "return the size of all non-hidden files in a directory" in {
+    val dirSize = fsUtils.getDirectorySizeNoHidden("src/test/resources/test_data/test_dir")
+    assert(dirSize == 20L)
+  }
+
+  "getDirectorySizeNoHidden" should "return the size of all non-hidden files recursively" in {
```
GeorgiChochov: I'm not sure if this test counts the non-hidden file in the hidden directory. 
Maybe something like `return the size of all non-hidden files recursively along non-hidden paths` would be more explicit?

utils/src/main/scala/za/co/absa/enceladus/utils/fs/FileSystemVersionUtils.scala
```diff
@@ -144,6 +145,38 @@ class FileSystemVersionUtils(conf: Configuration) {
...
+  /**
+    * Returns directory size in bytes, skipping hidden files and directories (starting from '_' or '.').
+    *
+    * @param path A path to a directory or a file.
+    * @return Directory size in bytes
+    */
+  def getDirectorySizeNoHidden(path: String): Long = {
+    def getDirSizeHelper(f: Path): Long = {
+      val status = fs.getFileStatus(f)
+      if (status.isFile) {
```
Zejnilovic: I have a feeling this `if` can be skipped.

utils/src/test/scala/za/co/absa/enceladus/FsUtilsSpec.scala
```diff
@@ -49,4 +49,25 @@ class FsUtilsSpec extends FlatSpec with Matchers with SparkTestBase {
...
+  }
+
+  "getDirectorySizeNoHidden" should "return the size of all non-hidden files in a directory" in {
+    val dirSize = fsUtils.getDirectorySizeNoHidden("src/test/resources/test_data/test_dir")
+    assert(dirSize == 20L)
+  }
+
+  "getDirectorySizeNoHidden" should "return the size of all non-hidden files recursively along non-hidden paths" in {
+    val dirSize = fsUtils.getDirectorySizeNoHidden("src/test/resources/test_data/test_dir2")
+    assert(dirSize == 40L)
```
Zejnilovic: We could use a test for a single file as well.

```scala
val dirSize = fsUtils.getDirectorySizeNoHidden("src/test/resources/test_data/test_dir2/dummy.txt")
assert(dirSize == 20L)
```

utils/src/main/scala/za/co/absa/enceladus/utils/fs/FileSystemVersionUtils.scala
```diff
@@ -144,6 +145,38 @@ class FileSystemVersionUtils(conf: Configuration) {
...
+        status.getLen
+      } else {
```
GeorgiChochov: This is only applicable for the outer function call. Otherwise we end up checking if a dir (which we already know from line 164) is a file with each recursive call.

utils/src/test/scala/za/co/absa/enceladus/FsUtilsSpec.scala
yruslan: 👍 Good idea! Both a single file and a single hidden file case should be tested.

utils/src/main/scala/za/co/absa/enceladus/utils/fs/FileSystemVersionUtils.scala
yruslan: Moved this out of the helper function.

PR #983, opened 2019-10-18
Feature/553 searchable schema fieldselector

lokm01:


menas/ui/components/AuditTrail.js
```diff
@@ -24,10 +24,10 @@ sap.ui.define(["components/tables/TableUtils"],
 
   AuditTrail.prototype.applyTableUtils = function() {
     const auditTableUtils = new TableUtils(this._oAuditTable, "Audit Trail");
+    auditTableUtils.makeSearchable(["updatedBy", "changes"]);
```
Zejnilovic: Just curious if there is any reason for moving this?

Will test the FrontEnd later. Code seems fine.

lokm01: Yeah there's no explicit ordering of the items in the toolbar. So basically want the search bar to appear before group and filter buttons.

Zejnilovic:
## 1 
The search field persists the value searched for across the datasets. Might be confusing.

## 2 
_((No strong preference))_
Add a label displaying name of what is currently selected.

## 3 
Mapping tables join conditions do not have the search, why?

## 4
Does not support Schema paths in it. For example `AnyStruct.AnyNumber` won't find anything even though it is a valid path. 
And in general searching for stuff in some structures seems tedious. You search for `Number` and then you have to click through all and unfold all structures.

lokm01:
1) Fixed
2) Let's create an issue and discuss please
3) Fixed
4) Fixed search of absolute path, please create an issue for automatic expansion of the tree component to the results - this is a bigger task.

PR #990, opened 2019-10-22
989 Fix control measure standardization failure

GeorgiChochov:


GeorgiChochov:
> If you wanna use it there is a class (MediaType)[https://yuanjiang.space/spring-resttemplate-post-json-with-utf-8]. Anyways, looks good

I remember it was there, but couldn't find it just now. Thanks!

GeorgiChochov:
Looking at it now, not really liking the way the MediaType is created... Let's stick to this implementation for now.

PR #997, opened 2019-10-25
Bugfix/748 empty mapping table err msg

yruslan:


conformance/src/main/scala/za/co/absa/enceladus/conformance/datasource/DataSource.scala
```diff
@@ -61,4 +67,17 @@ object DataSource {
   private[conformance] def setData(path: String, data: Dataset[Row]) {
     dfs += (path -> data)
   }
+
+  private def loadMappingTable(path: String)
+                              (implicit spark: SparkSession): DataFrame = {
+    try {
+      spark.read.parquet(path)
+    } catch {
+      case ex: AnalysisException if ex.getMessage.contains("Unable to infer schema for Parquet") =>
```
GeorgiChochov: Why check the message? This would make us dependent on the message not changing. 
In what other case would an `AnalysisException` by thrown by `spark.read.parquet`?

conformance/src/test/scala/za/co/absa/enceladus/conformance/interpreter/rules/testcasefactories/SimpleTestCaseFactory.scala
```diff
@@ -0,0 +1,162 @@
...
+    * @param experimentalMappingRule If true, the experimental mapping rule will be used.
+    * @param conformanceRules        Zero or more conformance rules to be applied as the part of conformance.
+    * @return A dataframe, a dataset, a Menas DAO, a Cmd Config and feature switches prepared to run conformance interpreter
+    */
+  def getTestCase(experimentalMappingRule: Boolean,
+                  conformanceRules: ConformanceRule*): (DataFrame, Dataset, MenasDAO, CmdConfig, FeatureSwitches) = {
+    import spark.implicits._
+    val inputDf = spark.read.schema(testCaseSchema).json(testCaseDataJson.toDS)
+    val dataset = getDataSetWithConformanceRules(testCaseDataset, conformanceRules: _*)
+    val cmdConfig: CmdConfig = CmdConfig(reportDate = reportDate)
```
GeorgiChochov:
```suggestion
    val cmdConfig = CmdConfig(reportDate = reportDate)
```

conformance/src/test/scala/za/co/absa/enceladus/conformance/interpreter/rules/testcasefactories/SimpleTestCaseFactory.scala
```diff
@@ -0,0 +1,162 @@
...
+
+    val dao: MenasDAO = mock(classOf[MenasDAO])
```
GeorgiChochov:
```suggestion
    val dao = mock(classOf[MenasDAO])
```

conformance/src/test/scala/za/co/absa/enceladus/conformance/interpreter/rules/testcasefactories/SimpleTestCaseFactory.scala
```diff
@@ -0,0 +1,162 @@
...
+    mockWhen(dao.getDataset("Example", 1)) thenReturn testCaseDataset
+    mockWhen(dao.getMappingTable("empty_mapping_table", 1)) thenReturn fixPathsInMappingTable(emptyMT)
+    mockWhen(dao.getMappingTable("non_existent_mapping_table", 1)) thenReturn fixPathsInMappingTable(nonExistentMT)
```
GeorgiChochov: These String literals can probably be converted to constants.

conformance/src/main/scala/za/co/absa/enceladus/conformance/datasource/DataSource.scala
GeorgiChochov: I see that `path does not exist` is another case for `AnalysisException`. Alright, let's stick with this. The tests will start failing if we move to a different spark version with a different message and we can adjust then.

PR #1008, opened 2019-11-04
Improve status code and error messages sent via HTTP on authentication failures.

yruslan:


menas/src/main/scala/za/co/absa/enceladus/menas/auth/MenasAuthenticationFailureHandler.scala
```diff
@@ -0,0 +1,46 @@
...
+
+  override def onAuthenticationFailure(request: HttpServletRequest,
+                                       response: HttpServletResponse,
+                                       ex: AuthenticationException): Unit = {
+    val stack = ex.getStackTrace.mkString("\n")
+    log.warn(s"Authentication exception: ${ex.getMessage}\n$stack")
+    ex match {
+      case _: AuthHostTimeoutException =>
+        response.sendError(HttpStatus.GATEWAY_TIMEOUT.value, ex.getMessage)
+      case _ : BadKrbHostException | _ : BadLdapHostException =>
```
Zejnilovic:
```suggestion
      case _: BadKrbHostException | _: BadLdapHostException =>
```

menas/src/main/scala/za/co/absa/enceladus/menas/auth/exceptions/AuthHostTimeoutException.scala
```diff
@@ -0,0 +1,28 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.enceladus.menas.auth.exceptions
+
+import org.springframework.security.core.AuthenticationException
+
+/**
+  * This exception is thrown then KDC or AD/LDAP server times out when authenticating a user.
```
Zejnilovic:
```suggestion
  * This exception is thrown when KDC or AD/LDAP server times out when authenticating a user.
```

menas/src/main/scala/za/co/absa/enceladus/menas/auth/exceptions/BadKrbHostException.scala
```diff
@@ -0,0 +1,28 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.enceladus.menas.auth.exceptions
+
+import org.springframework.security.core.AuthenticationException
+
+/**
+  * This exception is thrown then Menas failed to communicate to a KDC.
```
Zejnilovic:
```suggestion
  * This exception is thrown when Menas failed to communicate to a KDC.
```

menas/src/main/scala/za/co/absa/enceladus/menas/auth/MenasAuthenticationFailureHandler.scala
```diff
@@ -0,0 +1,46 @@
...
+import org.springframework.security.core.AuthenticationException
+import org.springframework.security.web.authentication.AuthenticationFailureHandler
+import za.co.absa.enceladus.menas.auth.exceptions.{AuthHostTimeoutException, BadKrbHostException, BadLdapHostException}
+
+/**
+  * This class is responsible for mapping authentication exceptions to status codes in HTTP responses.
+  */
+class MenasAuthenticationFailureHandler extends AuthenticationFailureHandler {
+
+  private val log: Logger = LoggerFactory.getLogger(this.getClass)
```
GeorgiChochov:
```suggestion
  private val log = LoggerFactory.getLogger(this.getClass)
```

menas/src/main/scala/za/co/absa/enceladus/menas/auth/MenasAuthenticationFailureHandler.scala
```diff
@@ -0,0 +1,46 @@
...
+
+  override def onAuthenticationFailure(request: HttpServletRequest,
+                                       response: HttpServletResponse,
+                                       ex: AuthenticationException): Unit = {
+    val stack = ex.getStackTrace.mkString("\n")
+    log.warn(s"Authentication exception: ${ex.getMessage}\n$stack")
```
GeorgiChochov:
```suggestion
    log.warn("Authentication exception", ex)
```
`Logger` has a smart API that allows you to log exceptions in an easy way. You don't need to construct the stack trace message yourself.

menas/src/main/scala/za/co/absa/enceladus/menas/auth/kerberos/MenasKerberosAuthentication.scala
```diff
@@ -140,7 +137,6 @@ class MenasKerberosAuthentication extends MenasAuthentication with InitializingB
     //something here changes the log level to WARN
     auth
       .authenticationProvider(new MenasKerberosAuthenticationProvider(adServer, ldapSearchFilter, ldapSearchBase))
-      .authenticationProvider(activeDirectoryLdapAuthenticationProvider())
```
lokm01: Why is this removed?

yruslan: Spring tries to authenticate using providers in the list. And if all of them fail it will provide the exception of the last provider tried.

This leads to a situation when `MenasKerberosAuthenticationProvider` fails, Menas tries to authenticate using local LDAP and returns an exception related to that attempt, instead of the exception from `MenasKerberosAuthenticationProvider`.

I tried to authenticate using login/password and using a keytab file without `activeDirectoryLdapAuthenticationProvider` and it worked, so I thought maybe we don't need it.

If we do need it, maybe we can just swap the order of providers.

PR #1009, opened 2019-11-05
Feature/950 improve spark app names

yruslan:


dao/src/main/scala/za/co/absa/enceladus/dao/menasplugin/MenasAuthUtils.scala
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  /**
+    * Checks if a keytab file is located in HDFS or in the local file system.
+    * If the file is in HDFS, it is copied to a temporary location.
+    *
+    * @param path A path fo a keytab file. Can be either local or HDFS location.
+    * @return A path to a keytab file in the local filesystem.
+    */
+  def getLocalKeytabPath(path: String): String = {
```
GeorgiChochov: This function itself is not `keytab` specific, and it kind of seems to duplicate a part of the `getPlainCredentials` function. 
I remember when I was refactoring this module thinking about extracting a common `getFile` function which checks local and HDFS, while `getKerberosCredentias` and `getPlainCredentials` can only deal with its result.

utils/src/main/scala/za/co/absa/enceladus/utils/fs/FileSystemVersionUtils.scala
```diff
@@ -127,6 +128,54 @@ class FileSystemVersionUtils(conf: Configuration) {
...
 
+  /**
+    * Checks if a file is located in HDFS or in the local file system.
+    * If the file is in HDFS, it is copied to a temporary location.
+    *
+    * @param path A path to a file.  Can be either local or HDFS location.
+    * @return A path to a file in the local filesystem.
+    */
+  @throws[FileNotFoundException]
+  def getLocalFile(path: String): String = {
```
Zejnilovic: I can't get over the fact that the method is called `getLocalFile` but is checking if the file is on HDFS and making the file local. Maybe `getLocalPathToFile` or something like that.

dao/src/main/scala/za/co/absa/enceladus/dao/menasplugin/MenasAuthUtils.scala
```diff
@@ -0,0 +1,53 @@
...
+import za.co.absa.enceladus.utils.fs.FileSystemVersionUtils
+
+/**
+  * This object contins Menas authentication tools.
+  */
+class MenasAuthUtils(conf: Configuration) {
+  private val fsUtils = new FileSystemVersionUtils(conf)
+
+  /**
+    * Creates Menas credentials from a credentials file located either on local file system or on HDFS.
```
Zejnilovic:
```suggestion
    * Instantiates MenasCredentials from a credentials file located either on local file system or on HDFS.
```

Create Menas credentials seems like it will create a new user for you.

dao/src/main/scala/za/co/absa/enceladus/dao/menasplugin/MenasAuthUtils.scala
```diff
@@ -0,0 +1,53 @@
...
+    MenasPlainCredentials(conf.getString("username"), conf.getString("password"))
+  }
+
+  /**
+    * Creates Menas credentials from file either on local file system or on HDFS.
+    *
+    * @param keytabPath A path to a Kerberos keytab file.
+    * @return An instance of Menas Credentials.
+    */
+  def getKerberosCredentias(keytabPath: String): MenasKerberosCredentials = {
```
Zejnilovic:
```suggestion
  def getKerberosCredentials(keytabPath: String): MenasKerberosCredentials = {
```

dao/src/main/scala/za/co/absa/enceladus/dao/menasplugin/MenasAuthUtils.scala
```diff
@@ -0,0 +1,53 @@
...
+
+package za.co.absa.enceladus.dao.menasplugin
+
+import com.typesafe.config.ConfigFactory
+import org.apache.hadoop.conf.Configuration
+import sun.security.krb5.internal.ktab.KeyTab
+import za.co.absa.enceladus.utils.fs.FileSystemVersionUtils
+
+/**
+  * This object contins Menas authentication tools.
```
Zejnilovic:
```suggestion
  * This object contains Menas authentication tools.
```

utils/src/main/scala/za/co/absa/enceladus/utils/fs/FileSystemVersionUtils.scala
```diff
@@ -127,6 +128,54 @@ class FileSystemVersionUtils(conf: Configuration) {
     }
   }
 
+  /**
+    * Checks if a file is located in HDFS or in the local file system.
```
Zejnilovic:
```suggestion
    * Checks if a file is located on the HDFS or the local file system.
```

GeorgiChochov: I'd say `on HDFS or the local file system` if you write `on the HDFS` you're giving `HDFS` which is a name (even though its an acronym it is never used as such) a definite article and that doesn't work grammatically.

utils/src/main/scala/za/co/absa/enceladus/utils/fs/FileSystemVersionUtils.scala
GeorgiChochov: I have the same problem.

conformance/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
```diff
@@ -162,9 +164,12 @@ object DynamicConformanceJob {
...
   }
 
-  private def obtainSparkSession(): SparkSession = {
+  private def obtainSparkSession()(implicit cmd: CmdConfig): SparkSession = {
+    val enceladusVersion = ProjectMetadataTools.getEnceladusVersion
+    log.info(s"Enceladus version $enceladusVersion")
+    val reportVersion = cmd.reportVersion.map(_.toString).getOrElse("")
     val spark: SparkSession = SparkSession.builder()
-      .appName("Dynamic Conformance")
+      .appName(s"Dynamic Conformance $enceladusVersion ${cmd.datasetName} ${cmd.datasetVersion} ${cmd.reportDate} $reportVersion")
```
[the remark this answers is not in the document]
yruslan: Yeah, looks better.

@DzMakatun, what do you think?

utils/src/main/scala/za/co/absa/enceladus/utils/fs/FileSystemVersionUtils.scala
yruslan: Struggling to come up with a good name. `getLocalPathToFile` seems decent. Do you agree on this, guys?

dao/src/main/scala/za/co/absa/enceladus/dao/menasplugin/MenasAuthUtils.scala
yruslan: Why `Menas credentials` -> `MenasCredentials`?

Zejnilovic: Because it is the name of the abstract class, it can be `MenasPlainCredentials`

utils/src/main/scala/za/co/absa/enceladus/utils/fs/FileSystemVersionUtils.scala
GeorgiChochov: Sure, good enough.

utils/src/main/scala/za/co/absa/enceladus/utils/fs/FileSystemVersionUtils.scala
yruslan: Good to know, thanks!

conformance/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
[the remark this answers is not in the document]
DzMakatun: 1. As one of the potential uses is to filter json from Spark History API, I would suggest to keep the parsing as simple as possible. So lets keep it with the spaces only.
2. I would also remove the word "Dynamic", to keep it consistent with Standardization naming, and, also a bit shorter.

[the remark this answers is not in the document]
yruslan: What are the challenges parsing job names in Spark History?
Do you use Scala to parse those strings?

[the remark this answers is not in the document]
yruslan: Maybe we can merge this now and continue the conversation is a separate issue?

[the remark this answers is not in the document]
yruslan: Raised #1032

Issue #1030, opened 2019-11-11
Helper scripts write to the same log file if run at the same time 


yruslan:
## Describe the bug
The helper scripts generate log file names by appending a timestamp suffix. The timestamp precision is up to a second.

* If 2 jobs run at the same time, even by different users, the will try to write to the same file
* If 2 jobs run by different users one of the jobs will fail on permission errors, since one of the users can't write to a file created by another user.

## To Reproduce
* Run 2 Standardization or Conformance jobs at the same time using helper scripts.

## Expected behaviour
Helper scripts should guarantee that many jobs can be started concurrently.

PR #1031, opened 2019-11-11
#1030 Fix helper scripts concurrent log files creation.

yruslan:


scripts/bash/run_enceladus.sh
```diff
@@ -272,6 +272,14 @@ echoerr() {
     echo "$@" 1>&2;
 }
 
+get_temp_log_file() {
+    DATE=`date +%Y_%m_%d-%H_%M_%S`
+    NAME=`sed -e 's#.*\.\(\)#\1#' <<< $CLASS`
```
Zejnilovic: Everything works, and seems fine. I am just curious. We want everything after the last dot, right?
can I ask why the `\(\)`?

yruslan: It is used to specify which subpattern to keep.
But yeah, after a careful look this can can be simplified:
```
NAME=`sed -e 's#.*\.##' <<< $CLASS`
```

Zejnilovic: If you are planning to change it, go for it. If not just merge you already have two approvalz

PR #1043, opened 2019-11-18
1041 Updated pom & Jenkinsfile to add SonarCloud related properties/step

hamc17:
Pom has additional SonarCloud properties required for analysis & Jenkinsfile has additional build step to run analysis.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -184,7 +184,7 @@
         <log.level.testcode>ERROR</log.level.testcode>
         <log.pattern>[%p] %d{yy/MM/dd HH:mm:ss} %c{4}: %m%n</log.pattern>
         <log.specialfilters.acceptonmatch>false</log.specialfilters.acceptonmatch>
-        <sonar.projectKey>AbsaOSS_enceladus</sonar.projectKey>
+        <sonar.projectKey>AbsaOSS_enceladus_project</sonar.projectKey>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -184,6 +184,11 @@
         <log.level.testcode>ERROR</log.level.testcode>
         <log.pattern>[%p] %d{yy/MM/dd HH:mm:ss} %c{4}: %m%n</log.pattern>
         <log.specialfilters.acceptonmatch>false</log.specialfilters.acceptonmatch>
+        <sonar.projectKey>AbsaOSS_enceladus</sonar.projectKey>
+        <sonar.moduleKey>${artifactId}</sonar.moduleKey>
+        <sonar.organization>absaoss</sonar.organization>
+        <sonar.host.url>https://sonarcloud.io</sonar.host.url>
+        <sonar.login>df0b1360db96da22d43d2d7d6a1d7b321137357a</sonar.login>
```
lokm01: Is this ok to be in the pom? Should it rather come from an environment setttings at the build time?

hamc17: This is just a token within SonarCloud; while it should be passed somehow into the pom or into the environment when using manual integration, now that the file .sonar.properties has been added we can remove those properties from the pom entirely.

I'm also going to revoke the tokens as we are not using them anymore.

hamc17:
Merging.

PR #1048, opened 2019-11-20
92 Use JWT for stateless authentication

GeorgiChochov:


menas/src/main/resources/application.properties.template
```diff
@@ -26,6 +26,10 @@ spring.resources.chain.strategy.fixed.enabled=true
 spring.resources.chain.strategy.fixed.paths=/components/**,/css/**,/generic/**,/service/**
 spring.resources.chain.strategy.fixed.version=@project.version@
 
+# Generate your own signing-key, must be be at least 256-bit
+za.co.absa.enceladus.menas.auth.signing-key=u7w!z%C*F-JaNdRgUkXp2s5v8y/A?D(G+KbPeShVmYq3t6w9z$C&E)H@McQfTjWn
```
yruslan: This is actually a *secret*, not a *signing key*.

menas/src/main/scala/za/co/absa/enceladus/menas/auth/jwt/JwtFactory.scala
```diff
@@ -0,0 +1,52 @@
...
+import io.jsonwebtoken.{JwtBuilder, JwtParser, Jwts}
+import org.springframework.beans.factory.annotation.{Autowired, Value}
+import org.springframework.context.annotation.Bean
+import org.springframework.stereotype.Component
+
+@Component
+class JwtFactory @Autowired()(@Value("${za.co.absa.enceladus.menas.auth.signing-key}")
+                             signingKey: String) {
+
+  private lazy val secretKey = Keys.hmacShaKeyFor(signingKey.getBytes)
```
yruslan:
```suggestion
  private lazy val signingKey = Keys.hmacShaKeyFor(secret.getBytes)
```

yruslan:
Probably not an issue, definitely not about this PR, just a generic thought.

When we have used sessions (this is only my assumption), the session expiration time was counted from a time of the most recent inactivity. If you keep UI active, the session will get prolonged automatically for as long as you want.
But after switching from sessions to JWT, both JWT and cookie expire at the same time. That means after a certain amount of time (like 8h) a user will get a login screen no matter the previous activity.

GeorgiChochov:
> I get 500 code response. Doesn't allow me to log in.
> 
> ```json
> {
> 	"timestamp": "2019-11-22T17:19:47.595+0000",
> 	"status": 500,
> 	"error": "Internal Server Error",
> 	"message": "The datetime zone id '\"UTC\"' is not recognised",
> 	"path": "/menas/api/login"
> }
> ```
> 
> Error found at `MenasAuthenticationSuccessHandler.scala:48`

Yeah, the `timezone="UTC"` quotes are taken as literals from the application.properties. Should be `timezone=UTC`.

menas/src/test/resources/application.properties
```diff
@@ -57,7 +61,7 @@ za.co.absa.enceladus.menas.spark.master=local[1]
 za.co.absa.enceladus.spline.urlTemplate=http://localhost:8080/spline/dataset/lineage/_search?path=%s&application_id=%s
 
 #system-wide time zone
-timezone="UTC"
+timezone=UTC
```
Zejnilovic: Please also change this in the template

GeorgiChochov:
Let's get back to refresh tokens in a separate task.

PR #1052, opened 2019-11-20
1049 Fix @lokm01 username in CODEOWNERS

GeorgiChochov:


GeorgiChochov:
Sorry, should have copy-pasted the usernames.

PR #1054, opened 2019-11-22
Feature/1017 A skeleton implementation of the broadcasting mapping rule

yruslan:
This PR adds a new `BroadcastingMappingRule` that uses the strategy of broadcasting the mapping table across executors instead of doing a join.
I decided to push several small PRs instead of creating a PR for the whole thing once it is done. There are things that might require discussion and can affect the design o f the feature.
* The rule currently works only on flat input fields.
* The rule is not used in the current production code, not even with a feature switch. The way I tested it is by using `MappingRuleBroadcastSuite`. The suite should pass on the current and th broadcast implementation of the mapping rule.  I change `DynamicInterpreter` to tun it:
```
        //if (ictx.featureSwitches.experimentalMappingRuleEnabled) {
        //  MappingRuleInterpreterGroupExplode(r, ictx.conformance, explodeContext).conform(df)
        //} else {
        //  MappingRuleInterpreter(r, ictx.conformance).conform(df)
        //}
        MappingRuleInterpreterBroadcast(r, ictx.conformance).conform(df)
```

The logic for dealing with nested structs and for deciding which strategy to use will be added in later PRs. Also optimizing the number of broadcasts is on ToDo list as well.


utils/src/test/scala/za/co/absa/enceladus/utils/broadcast/BroadcastUtilsSuite.scala
```diff
@@ -0,0 +1,353 @@
...
+        assertResults(dfOut1, expectedResultsMatchFound)
+        assertResults(dfOut2, expectedResultsMatchNotFoundDefautlt)
+      }
+    }
+
+    "throw an exception" when {
+
+      "a join without key fields is attempted" in {
+        intercept[IllegalArgumentException] {
+          val localMt = LocalMappingTable(dfMt, Nil, "val")
```
Zejnilovic:
```suggestion
          LocalMappingTable(dfMt, Nil, "val")
```

conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/rules/MappingRuleInterpreterBroadcast.scala
```diff
@@ -0,0 +1,103 @@
...
+
+    val mt = LocalMappingTable(mapTable, mappingTableFields, rule.targetAttribute)
+    val broadcastedMt = spark.sparkContext.broadcast(mt)
+    val mappingUDF = BroadcastUtils.getMappingUdf(broadcastedMt, defaultValueOpt)
+
+    val errorUDF = BroadcastUtils.getErrorUdf(broadcastedMt, rule.outputColumn, mappings)
+
+    val tempErrorColName = SchemaUtils.getUniqueName("err", Some(df.schema))
+
+    val res1 = df.withColumn(rule.outputColumn, mappingUDF(inputDfFields.map(a => col(a)): _ *))
```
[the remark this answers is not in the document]
yruslan: 1. Yes, this is me being lazy rewriting from one implementation to another.
2. Yes, `withColumn()` is slow, so if used in a loop for every column (like it was for Standardization), the performance degrades. For mapping rule this is not the slowest thing. 
Anyway, this bit will be replaced by `nestedWithColumnAndErrorMap()` to support mapping inside an array.

utils/src/main/scala/za/co/absa/enceladus/utils/broadcast/BroadcastUtils.scala
```diff
@@ -0,0 +1,253 @@
...
+      errorMessageSchema.dataType,
+      Some(mappingTable.value.keyTypes))
+
+    errorUdf
+  }
+
+  private def getMappingLambdaParam1(mappingTable: Broadcast[LocalMappingTable], defaultValueOpt: Option[Any]): AnyRef = {
+    defaultValueOpt match {
+      case None =>
+        param1: Any => {
```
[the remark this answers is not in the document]
yruslan: It is a short syntax for a lambda from `Any` to `Any` when this lambda is returned from a function.
```
def returnLambda() = {
  (a: Any) => {
    doSomething1()
    doSomething2()
  }: Any
}
```
can be shortened to
```
def returnLambda() = {
  a: Any => {
    doSomething1()
    doSomething2()
  }
}
```
I'm wondering if I should be more explicit in this code so it is easier to read?

utils/src/main/scala/za/co/absa/enceladus/utils/broadcast/BroadcastUtils.scala
```diff
@@ -0,0 +1,248 @@
...
+    *
+    * @param expression A Spark expression.
+    * @return A UDF that maps joins keys to a target attribute value.
+    */
+  def getValueOfSparkExpression(expression: String)(implicit spark: SparkSession): Any = {
+    import spark.implicits._
+
+    List(1).toDF()
+      .withColumn("a", expr(expression))
+      .collect()(0)(1)
```
GeorgiChochov: This might need a little more explanation about how it works. Not in the scaladoc

conformance/src/main/scala/za/co/absa/enceladus/conformance/interpreter/rules/MappingRuleInterpreterBroadcast.scala
```diff
@@ -0,0 +1,101 @@
...
+    val mt = LocalMappingTable(mapTable, mappingTableFields, rule.targetAttribute)
+    val broadcastedMt = spark.sparkContext.broadcast(mt)
+    val mappingUDF = BroadcastUtils.getMappingUdf(broadcastedMt, defaultValueOpt)
+
+    val errorUDF = BroadcastUtils.getErrorUdf(broadcastedMt, rule.outputColumn, mappings)
+
+    val tempErrorColName = SchemaUtils.getUniqueName("err", Some(df.schema))
+
+    val withMappedFieldsDf = df.withColumn(rule.outputColumn, mappingUDF(inputDfFields.map(a => col(a)): _ *))
+      .withColumn(rule.outputColumn, mappingUDF(inputDfFields.map(a => col(a)): _ *))
```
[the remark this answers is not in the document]
yruslan: 👍 it actually repeated - good catch!

utils/src/main/scala/za/co/absa/enceladus/utils/broadcast/BroadcastUtils.scala
```diff
@@ -42,6 +44,27 @@ object BroadcastUtils {
...
+    // that we will ignore.
+    //
+    // List(1).toDF()       // This creates a DataFrame with a single column 'val' and a single row with a value 1.
+    //
+    // Then we add another column, named 'a' (but the name is not important since we are going to ignore it),
+    // and the value of the Spark expression provided by the user.
+    //
+    // .withColumn("a", expr(expression))
+    //
+    // Then we collect the DataFrame '.collect()', take the first row '(0)' and the second column '(1)', which is
```
GeorgiChochov: This last bit is the only thing that I was thinking of and that the `List(1)` and ` column name "a"` are arbitrary and ignored, `expr` is clear enough.

yruslan: After thinking about it, `List(1)` is needed since we need an initial dataframe with 1 row in it. If we create an empty dataset, adding a column with a user-defined expression won't create any row that would contain the actual value of the expression. So will leave `List(1)` and extend the explanation.

The name "a" indeed seems arbitrary, will change it to "dummy".

GeorgiChochov: I would actually cut most of the explanation. I just wanted to have a bit explaining the choice of `List(1)`, column name `a` and `collect()(0)(1)`

PR #1058, opened 2019-11-22
1038 Add README for GH Pages

Zejnilovic:


README.md
```diff
@@ -1 +1,15 @@
...
+### Dependencies:
+- ruby >= 2.3.0
+
+### To build locally:
+```bash
+# In the root of the project
+$> git checkout gh-pages
+$> bundle install
+$> bundle exec jekyll serve
+# => Now browse to http://localhost:4000
```
benedeki: Make this a actual clickable link in the `readme.md`

README.md
```diff
@@ -1 +1,15 @@
...
+This branche's sole purpose is to serve as Documentation for the Enceladus project
+
+### Dependencies:
+- ruby >= 2.3.0
+
+### To build locally:
+```bash
+# In the root of the project
+$> git checkout gh-pages
+$> bundle install
```
benedeki: I guess this doesn't have to run always? Can you make split it to *run first time* and *run to build locally* or something like that.

Zejnilovic: I would rather not. I have thought about it and checking that the moving dependencies (GitHub's pre-packaged gems) are always up to date is better then rethinking what went wrong. And the bundle install takes 2 secs if you are up to date with everything.

README.md
Zejnilovic: Again, I would rather not. This would point to a "site" that is available only at a certain point after you do everything needed. It is not a site that is always available or something. It can even lead people to wrong sites that they have previously built and left running.

PS: These 3 lines were "stolen" from official `Jekyll` README. They also don't use link

README.md
```diff
@@ -1 +1,15 @@
 # Enceladus's Github Pages branch
+
+This branche's sole purpose is to serve as Documentation for the Enceladus project
```
GeorgiChochov:
```suggestion
This branch's sole purpose is to serve as Documentation for the Enceladus project.
```

PR #1066, opened 2019-11-27
1065 Create new JwtBuilder on each request

GeorgiChochov:


menas/src/test/scala/za/co/absa/enceladus/menas/integration/controllers/AuthenticationIntegrationSuite.scala
```diff
@@ -0,0 +1,66 @@
...
+
+  override def fixtures: List[FixtureService[_]] = List()
+
+  @Autowired
+  val jwtFactory: JwtFactory = null
+
+  private val jwtRegex = "JWT=([^;]+);?.*".r
+
+  "Username and password authentication" should {
+    "handle multiple users loggin in concurrently" in {
```
Zejnilovic:
```suggestion
    "handle multiple users login in concurrently" in {
```

menas/src/test/scala/za/co/absa/enceladus/menas/integration/TestInMemoryMenasAuthentication.scala
```diff
@@ -0,0 +1,43 @@
...
+@Component
+@Primary
+class TestInMemoryMenasAuthentication extends InMemoryMenasAuthentication {
+
+  override def configure(auth: AuthenticationManagerBuilder): Unit = {
+    super.configure(auth)
+
+    val inMemoryAuth = auth
+      .inMemoryAuthentication()
+      .passwordEncoder(NoOpPasswordEncoder.getInstance())
```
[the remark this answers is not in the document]
GeorgiChochov: Yeah, I know, but it doesn't seem to be going away. They've deprecated it so people don't use it in production, but since it is just for tests, I thought "no harm done". I'll replace it just to be rid of the deprecation warning.

menas/src/main/scala/za/co/absa/enceladus/menas/auth/InMemoryMenasAuthentication.scala
```diff
@@ -17,29 +17,40 @@ package za.co.absa.enceladus.menas.auth
...
+      .passwordEncoder(passwordEncoder)
+
+    addUsers(inMemoryAuth)
+  }
+
+  protected def addUsers(auth: InMemoryUserDetailsManagerConfigurer[_]): Unit = {
+    auth
       .withUser(username)
-      .password(password)
+      .password(passwordEncoder.encode(password))
```
benedeki: Question: Why is the encoder added?
Is it somehow connected to the change or is it an improvement  "thanks to being around"?

menas/src/test/scala/za/co/absa/enceladus/menas/auth/jwt/JwtFactoryTest.scala
```diff
@@ -0,0 +1,49 @@
...
+  private val jwtFactory = new JwtFactory(secret)
+
+  "JwtFactory" should {
+    "throw an exception" when {
+      "creating a JwtBuilder with a weak secret" in {
+        intercept[WeakKeyException] {
+          new JwtFactory("NotEnoughBits").jwtBuilder()
+        }
+      }
+      "creating a JwtParser with a weak secret" in {
```
benedeki: Isn't this the same code as above?

GeorgiChochov:
> Sadly I can do all the tests I want but this needs to be checked on pre-prod

I wrote an integration test for it, failed before, passes now.

menas/src/main/scala/za/co/absa/enceladus/menas/auth/InMemoryMenasAuthentication.scala
GeorgiChochov: The encoder is unnecessary because it's in-memory and used only for tests and running a local development environment. But the NoOpEncoder (which doesn't encode the password before storing it in memory is deprecated and @Zejnilovic suggested we work around the deprecation warning)

menas/src/test/scala/za/co/absa/enceladus/menas/auth/jwt/JwtFactoryTest.scala
GeorgiChochov: It checks that the `JWT Parser` the above one checks the `JWT Builder`

benedeki: Blind me, sorry.

Issue #1067, opened 2019-11-27
Remove Usage chapter from README.md


benedeki:
## Background
There's is now documentation on Gihub pages. Remove the excessive info from `README.md` covered in the documentation.

## Feature
Remove **Usage** chapter from `README.md` and add link to documentation.

benedeki:
### For Release Notes:
* README.md links to documentation

PR #1068, opened 2019-11-27
#1067: Remove Usage chapter from README.md

benedeki:
* Removed *How to use* chapter
* Fixed some typos
* Links to headers changed to standard and removed anchor names
* some table alignments for better readability in text-only display

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,34 +11,33 @@ ___
 
 <!-- toc -->
 - [What is Enceladus?](#what-is-enceladus)
-    - [Menas](#wie-menas)
-    - [Standardization](#wie-standardization)
-    - [Conformance](#wie-conformance)
-- [How to build](#build)
-- [How to run](#run)
-- [How to contribute](#contribute)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -11,34 +11,33 @@ ___
...
+    - [Standardization](#standardization)
+    - [Conformance](#conformance)
+- [How to build](#how-to-build)
+- [How to run](#how-to-run)
+- [How to contribute](#how-to-contribute)
+- [Documentation](#documentation)
 <!-- tocstop -->
 
-## <a name="what-is-enceladus"/>What is Enceladus?
+## What is Enceladus?
```
Zejnilovic: Why do this? I know some pre-processors will handle it, but some don't. I would rather be safe.

benedeki: I believe auto-anchoring of headers is the standard of markdown. And injected html is not encouraged.

GeorgiChochov: I think this is fine, it works and it looks good.

PR #1076, opened 2019-11-29
Bug/1073 unexpected schema validation error with null in default value

benedeki:
* Added explaining text into the documentation


Zejnilovic:
@benedeki I would ask you to fix the branch first and create a new PR. I don't get why there is so much and even something from github pages.

benedeki:
> 
> 
> @benedeki I would ask you to fix the branch first and create a new PR. I don't get why there is so much and even something from github pages.
> 
> EDIT: Now I see the intention, but still, you are trying to merge it into `develop` instead of `gh-pages`

Oh, silly me! 🤦‍♂ I made the PR into `develop` instead of 'gh-pages'. Will create a new PR

PR #1081, opened 2019-12-02
1079 Fix Entity Edit unpopulated form

GeorgiChochov:


Zejnilovic:
When creating entities the schema versions in the dropdown is kept cached. This is mostly a visual bug and breaks nothing in the end. If you select a schema, the version list is refreshed.
image

GeorgiChochov:
> When creating entities the schema versions in the dropdown is kept cached. This is mostly a visual bug and breaks nothing in the end. If you select a schema, the version list is refreshed.
> image

This bug is present from previous versions, I can't find it logged anywhere, but it wasn't introduced here and I think we should fix it separately.

Zejnilovic:
> > When creating entities the schema versions in the dropdown is kept cached. This is mostly a visual bug and breaks nothing in the end. If you select a schema, the version list is refreshed.
> > image
> 
> This bug is present from previous versions, I can't find it logged anywhere, but it wasn't introduced here and I think we should fix it separately.

#1084

Issue #1080, opened 2019-12-02
User being in too many groups creates too big cookie, browser then rejects it


Zejnilovic:
## Describe the bug
When the user has too many groups, the cookie created becomes too big and the browser rejects it.

## To Reproduce
Steps to reproduce the behavior OR commands run:
1. Have/be in too many groups
2. Try to log in

## Expected behaviour
I should be able to login

GeorgiChochov:
## For release notes
Fixed issue with JWT cookie inflating beyond the cookie size-limit that browsers accept.

PR #1082, opened 2019-12-02
#1080 Fix cookie size by not including the user's groups in the JWT

GeorgiChochov:


menas/src/main/scala/za/co/absa/enceladus/menas/auth/MenasAuthenticationSuccessHandler.scala
```diff
@@ -48,16 +48,11 @@ class MenasAuthenticationSuccessHandler @Autowired()(jwtFactory: JwtFactory,
...
-    val groups = authentication
-      .getAuthorities.asScala
-      .map(_.getAuthority).asJava
-
     val jwt = jwtFactory
       .jwtBuilder()
       .setSubject(user.getUsername)
       .setExpiration(jwtExpirationTime)
       .claim(CsrfTokenKey, csrfToken)
-      .claim(GroupsKey, groups)
```
benedeki: What was the purpose of including the groups?

GeorgiChochov: They were part of the principal and showed up on `/api/user/info`, so I preserved them when switching to JWT. But they were never used anywhere and there might not be a reason for them any time soon if ever.

PR #1089, opened 2019-12-04
Bugfix/1069 enceladus jars materialization fails oozie

lokm01:


menas/src/main/scala/za/co/absa/enceladus/menas/repositories/OozieRepository.scala
```diff
@@ -180,6 +183,11 @@ class OozieRepository @Autowired() (oozieClientRes: Either[OozieConfigurationExc
           hadoopFS.delete(hdfsConfPath, true)
       }
 
+      //For some reason these downloads were always cut half-way.. The blocking
+      //request is only made once per version (this step is skipped if jars are loaded already)
+      Await.ready(resFutureStd, Duration.Inf)
+      Await.ready(resFutureConf, Duration.Inf)
```
GeorgiChochov: Not a fan of infinite wait. I'd say this should be some configurable property like `download.duration.conformance.jars`.
Maybe also `logger.info("Downloading Standardization and Conformance jars, this may take some time.")`

yruslan:
The changes look good.

The odd thing is that these changes require so many commits. Maybe rebase or squash them when they are approved?

menas/src/main/scala/za/co/absa/enceladus/menas/repositories/OozieRepository.scala
benedeki: Agreed on the infinity waiting.
The `logger` is covered in the `downloadFile` function itself. But perhaps the wording could be changed. 🤔

menas/src/main/scala/za/co/absa/enceladus/menas/repositories/OozieRepository.scala
```diff
@@ -190,19 +198,21 @@ class OozieRepository @Autowired() (oozieClientRes: Either[OozieConfigurationExc
     Future {
       if (!hadoopFS.exists(hadoopPath) || hadoopFS.getStatus(hadoopPath).getCapacity == 0) {
         logger.info(s"Uploading jar from $url to $hadoopPath")
```
benedeki: `
```suggestion
        logger.info(s"Copying jar from $url to $hadoopPath, this may take some time.")
```

menas/src/main/scala/za/co/absa/enceladus/menas/repositories/OozieRepository.scala
```diff
@@ -190,19 +198,21 @@ class OozieRepository @Autowired() (oozieClientRes: Either[OozieConfigurationExc
...
-        in.read(targetArray)
         val os = hadoopFS.create(hadoopPath, true)
-        os.write(targetArray)
+
+        IOUtils.copy(in, os)
+
         os.flush()
         os.close()
+        in.close()
       }
```
benedeki: More info might be useful:
```
else {
  logger.info(s"Jar already present at $hadoopPath")
}
```

lokm01:
Let me open a new PR. Looks like that sync download wasn't necessary, we were testing it wrong :)

Issue #1069, opened 2019-11-28
Materialization of enceladus jars in oozie (quietly) fails when fetching from remote maven repo


lokm01:
It only saves first couple of hundred KBs.

GeorgiChochov:
There is a merged PR for this, but it's still open, can we close it? Could you add release notes for it?

lokm01:
Release notes
Fixed issue with initialization of enceladus jars for oozie integration when fetching binaries from remote maven repo.

PR #1094, opened 2019-12-05
#1069 Fixed incomplete download of enceladus jars for oozie

lokm01:


menas/src/main/scala/za/co/absa/enceladus/menas/repositories/OozieRepository.scala
```diff
@@ -197,12 +198,13 @@ class OozieRepository @Autowired() (oozieClientRes: Either[OozieConfigurationExc
...
-        val targetArray = Array.fill(in.available)(0.toByte)
-        in.read(targetArray)
         val os = hadoopFS.create(hadoopPath, true)
-        os.write(targetArray)
+
+        IOUtils.copy(in, os)
+
         os.flush()
         os.close()
+        in.close()
```
benedeki: Isn't it a good pattern to have the close after IO in `finally` block. To ensure closing even if something happens. 🤔

PR #1096, opened 2019-12-05
1039 Add convinience scripts for gh pages

### LINKED ISSUE #1039: Script to create new version of docs
## Background
When transitioning from one version to another, most of the previous one would remain unchanged, except some tiny alterations in each file. That would be nice to automatize.

## Feature
Create a script which copies and alters the files of previous version of documentation to the new version. And also registers the new version to be seen.

Zejnilovic:
1. I did a quick fix of the non-formating json in schema, it bugged me so I just used a gist to do it for now. We will solve later.
2. `create_docs.rb` creates a full folder and goes over every file and changes the versions. Needs an argument and that is the new version. Other things it gets by itself. In the future, I might do this automatically as well.
3. `get_release_notes.rb` connects to ZenHub and GitHub and automatically generates release notes. It takes the latest release, collects issues and then collects comments from these issues that start with `Release Notes`. So basically now you have to comment release note as:
```
Release Notes
This fixes this thing and the other. 
I can even use special formating characters here.
```

This should close #1039

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ Usage: ruby utils/get_release_notes.rb VERSION [options]
 Specific options:
         --github-token TOKEN         Github token. Can be specified using environment variable GITHUB_TOKEN
         --zenhub-token TOKEN         Zenhub token. This means we will use Release object for release notes. You don't have to use --use-zenhub in case you use this. Can be specified using environment variable ZENHUB_TOKEN
-    -z, --use-zenhub                 Run using zenhub. IT needs environment variable ZENHUB_TOKEN. If you use --zenhub-token option, you don't need to use this. This means we will use Release object for release notes.
+    -z, --use-zenhub                 Run using zenhub. It needs environment variable ZENHUB_TOKEN. If you use --zenhub-token option, you don't need to use this. This means we will use Release object for release notes.
         --organization ORGANIZATION  Github Organization
         --repository REPOSITORY      Github Repository name
         --repository-id REPOSITORYID Zenhub Repository ID
--- a/README.md
+++ b/README.md
@@ -22,18 +22,22 @@ ruby utils/create_docs.rb <version>
 ```
 
 #### Generate release notes
+A Ruby script to generate release notes. 
+
 ```bash
 Usage: ruby utils/get_release_notes.rb VERSION [options]
 
 Specific options:
-        --github-token TOKEN         Github token. Can be specified using environment variable GITHUB_TOKEN
-        --zenhub-token TOKEN         Zenhub token. This means we will use Release object for release notes. You don't have to use --use-zenhub in case you use this. Can be specified using environment variable ZENHUB_TOKEN
-    -z, --use-zenhub                 Run using zenhub. It needs environment variable ZENHUB_TOKEN. If you use --zenhub-token option, you don't need to use this. This means we will use Release object for release notes.
+        --github-token TOKEN         Github token. Can be specified using environment variable GITHUB_TOKEN or in get_release_notes.json in resources using github_token key
+        --zenhub-token TOKEN         Zenhub token. This means we will use Release object for release notes. You don't have to use --use-zenhub in case you do this. Can be specified using environment variable ZENHUB_TOKEN or in get_release_notes.json in resources using zenhub_token key
+    -z, --use-zenhub                 Run using zenhub. It needs zenhub token set. If you use --zenhub-token option, you don't need to use this. This means we will use Release object for release notes.
         --organization ORGANIZATION  Github Organization
         --repository REPOSITORY      Github Repository name
         --repository-id REPOSITORYID Zenhub Repository ID
         --zenhub-url ZENURL          Zenhub API URL
         --github-url GITURL          Github API URL
     -p, --[no-]print-empty           Should Issue with no release notes comment be included in the output file
+        --[no-]print-only-title      Should Issue with no release notes comment be preceeded with 'Couldn't find comment'
+    -s, --[no-]strict                Treats warnings as errors
     -h, --help                       Show this message
 ```
--- a/README.md
+++ b/README.md
@@ -29,7 +29,6 @@ Specific options:
         --github-token TOKEN         Github token. Can be specified using environment variable GITHUB_TOKEN
         --zenhub-token TOKEN         Zenhub token. This means we will use Release object for release notes. You don't have to use --use-zenhub in case you use this. Can be specified using environment variable ZENHUB_TOKEN
     -z, --use-zenhub                 Run using zenhub. IT needs environment variable ZENHUB_TOKEN. If you use --zenhub-token option, you don't need to use this. This means we will use Release object for release notes.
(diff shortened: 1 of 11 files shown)


utils/get_release_notes.rb
```diff
@@ -0,0 +1,129 @@
...
+      acc
+    end
+  else
+    whatever
+  end
+end
+
+
+all_releases = call(URI("#{ZENHUB_URL}/p1/repositories/#{REPO_ID}/reports/releases"))
+latest_release = Release.new(all_releases.last)
```
benedeki: This relies on the fact that the notes will be generated on the last release. What if the next release(s) is/are already defined. What about having an option input parameter for that?

GeorgiChochov: Let's make it so we are able to specify the version when we run the script

utils/get_release_notes.rb
```diff
@@ -0,0 +1,129 @@
...
+
+  def initialize(number)
+    @number = number
+  end
+
+  def get_release_notes
+    comments = call(URI("#{GITHUB_URL}/issues/#{@number}/comments"))
+    release_notes = comments.select { |comment| comment[:body] =~ /^Release Notes.*/ }
+    if release_notes.nil? || release_notes.empty?
+      @release_comment = "Couldn't find Release Notes for #{number}"
```
benedeki: Add the _Title_ of the item

benedeki: We talked  about outputting this to console instead of the `@release_comment`

utils/get_release_notes.rb
```diff
@@ -0,0 +1,129 @@
+require 'uri'
+require 'net/http'
+require 'openssl'
+require 'json'
+
+PATH_TO_ZENHUB_TOKEN = 'token'
```
benedeki: Can this be an environment variable - so I can set it up once and not fix it on every (fresh) checkout?

utils/get_release_notes.rb
```diff
@@ -0,0 +1,129 @@
...
+class Issue
+  attr_accessor :number, :release_comment
+
+  def initialize(number)
+    @number = number
+  end
+
+  def get_release_notes
+    comments = call(URI("#{GITHUB_URL}/issues/#{@number}/comments"))
+    release_notes = comments.select { |comment| comment[:body] =~ /^Release Notes.*/ }
```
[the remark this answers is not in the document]
Zejnilovic: If we want to be like that then I propose `/\A[[:blank:]]*(R|r)elease (N|n)otes([[:blank:]]|-|:)*\n/`

1. `\A` -> Ruby strict beginning of a string and not just on every line
2. `[[:blank:]]*` -> Any number of spaces or tabs
3. `(R|r)elease (N|n)otes` -> To again allow user to choose
4. `([[:blank:]]|-|:)*` -> Any number of spaces, tabs, etc. and dashes `-` or columns `:`
5. '\n' -> Forced newline

utils/get_release_notes.rb
Zejnilovic: Could be. @GeorgiChochov  what do you think? Make a token ENV variable?

Zejnilovic: Could actually even use a Github token

Zejnilovic:
Readme will come. I did not forget.

_docs/1.0.0/usage-schema.md
```diff
@@ -65,85 +63,8 @@ fields within fields.
...
-        "type": "timestamp",
-        "nullable": true,
-        "metadata": {
-            "pattern": "yyyyMMdd.HHmmss"
-        }
-    }
-    ]
-}
-```
+{% gist 394d2fe3a23366c081af4aa54441fe8b %}
```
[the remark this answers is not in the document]
Zejnilovic: reverted

utils/create_docs.rb
```diff
@@ -16,8 +30,10 @@
 
   Dir.glob("#{NEW_VERSION}/*").each do |file_path|
     puts "Updating version for #{file_path}"
-    file_content = File.read(file_path)
-    new_content = file_content.gsub(last_version, NEW_VERSION)
+    file_content = File.read(file_path).partition( /---.*?(---)/m )
```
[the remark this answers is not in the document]
Zejnilovic: `String#partition` returns always 3 parts. Preceding, searched, succeeding. That is the first rule. The second rule is Jekyll's liquid matter which guarantees this to be first.

utils/get_release_notes/opt_parser.rb
```diff
@@ -0,0 +1,106 @@
...
+                                    ' If you use --zenhub-token option, you don\'t need to use this.' +
+                                    ' This means we will use Release object for release notes.') do |z|
+        if options.zenhub_token.nil? || options.zenhub_token.empty?
+          raise ArgumentError, "Can't find ZENHUB_TOKEN environemnt variable", caller
+        end
+        options.use_zenhub = z
+        options.zenhub_token = options.zenhub_token
+      end
+
+      opts.on('-v', '--version VERSION', 'Version of release notes') do |v|
```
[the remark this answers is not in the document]
Zejnilovic: Since everything is in arguments. This should be as well. Makes no sense to pull something out unless it is a sub-command.

README.md
```diff
@@ -13,3 +13,28 @@ $> bundle install
...
+```
+
+#### Generate release notes
+```bash
+Usage: ruby utils/get_release_notes.rb [options]
+
+Specific options:
+        --github-token TOKEN         Github token.
+        --zenhub-token TOKEN         Zenhub token. This means we will use Release object for release notes.
+    -z, --use-zenhub                 Run using zenhub. IT needs environment variable ZENHUB_TOKEN. If you use --zenhub-token option, you don't need to use this. This means we will use Release object for release notes.
```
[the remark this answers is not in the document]
Zejnilovic: ZENHUB_TOKEN and GITHUB_TOKEN can be specified by Environment variables. I will try to put it there

README.md
```diff
@@ -13,3 +13,28 @@ $> bundle install
...
+    -v, --version VERSION            Version of release notes
+        --organization ORGANIZATION  Github Organization
+        --repository REPOSITORY      Github Repository name
+        --repository-id REPOSITORYID Zenhub Repository ID
+        --zenhub-url ZENURL          Zenhub API URL
+        --github-url GITURL          Github API URL
+    -p, --print_empty                Github API URL
```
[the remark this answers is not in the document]
Zejnilovic: You don't know. It might be.

Zejnilovic:
1. @wajda found a syntax error
2.`--help` of `-h` after version number was made into a command rather than an option.

GeorgiChochov:
Looks great

Zejnilovic:
We found out with @wajda that Github API default state is open, so we needed to add state all to rest calls for github

Zejnilovic:
Last addition for this PR. 
- add strict option so it handles warnings as errors (having open issues)
- add a possibility for the script to deduce Oganization and Repo name by itself

PR #1109, opened 2019-12-17
1090 Update Group ID

GeorgiChochov:


external/fixed-width/pom.xml
```diff
@@ -15,12 +15,12 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
-    <artifactId>enceladus-fixedWidth</artifactId>
+    <artifactId>fixed-width</artifactId>
```
yruslan: 👍  Nice that the camelCase notation is switched to dash-based one.

pom.xml
```diff
@@ -108,14 +108,14 @@
...
-        <module>dataModel</module>
+        <module>data-model</module>
         <module>examples</module>
         <module>dao</module>
         <module>menas</module>
         <module>migrations</module>
         <module>migrations-cli</module>
         <module>external/kafka</module>
-        <module>external/fixedWidth</module>
+        <module>external/fixed-width</module>
```
yruslan: Do you remember why these modules are 'external'?

standardization/pom.xml
```diff
@@ -104,8 +104,8 @@
             <version>${mongo.java.driver.version}</version>
         </dependency>
         <dependency>
-            <groupId>za.co.absa</groupId>
-            <artifactId>enceladus-fixedWidth</artifactId>
+            <groupId>za.co.absa.enceladus</groupId>
+            <artifactId>fixedWidth</artifactId>
```
yruslan:
```suggestion
            <artifactId>fixed-width</artifactId>
```

pom.xml
GeorgiChochov: Not really, I think because they were not originally meant to be part of the project, but got smuggled out during the process of open sourcing it. Maybe we can factor them out as well? Have to double check where they might be used and if there is any coupling. They might end up being top level modules as well.

yruslan: This investigation might worth a separate task on GutHub. I can log it.

PR #1121, opened 2020-01-02
273 Add grouped Runs navigation

GeorgiChochov:


menas/src/main/scala/za/co/absa/enceladus/menas/repositories/RunMongoRepository.scala
```diff
@@ -193,6 +193,94 @@ class RunMongoRepository @Autowired()(mongoDb: MongoDatabase)
...
+        Document("""{latestRunStartDateTime: {
+                   |  $dateToString: {
+                   |    date: "$latestStart",
+                   |    format: "%d-%m-%Y %H:%M:%S %z",
+                   |    timezone: "$timezone"
+                   |  }
+                   |}},""".stripMargin),
+        excludeId()
+      )),
+      sort(ascending("datasetVersion"))
```
yruslan: I'm wondering if descending order here is more convenient?

GeorgiChochov: Makes sense

menas/ui/components/run/runDatasetNameMaster.view.xml
```diff
@@ -0,0 +1,45 @@
...
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<core:View id="runMasterView" xmlns:core="sap.ui.core" xmlns="sap.m"
+           controllerName="components.run.runDatasetNameMaster"
+           xmlns:cust="http://schemas.sap.com/sapui5/extension/sap.ui.core.CustomData/1">
+    <Page id="masterPage" title="Dataset Runs" showNavButton="true" navButtonPress="onPressMasterBack">
```
HuvarVer: I think there should be the same arrows ( `>` ) as are at the Navigation menu. For better understanding that it rolled out after pressing the button.

HuvarVer: image
image

GeorgiChochov: That's an excellent point! Thanks

menas/ui/components/app.controller.js
```diff
@@ -92,7 +94,19 @@ sap.ui.define([
...
+      this._app.toMaster(this.createId("runsDatasetNamePage"));
+    },
+
+    onRunsDatasetNamePress: function (channel, event, data) {
+      this._eventBus.publish("runs", "list-grouped-version", data);
+      this._app.toMaster(this.createId("runsDatasetVersionPage"));
+    },
+
+    onRunsDatasetVersionPress: function (channel, event, dataset) {
+      console.log(dataset)
```
Zejnilovic: Is this `console.log` intentional? If so, it is missing a `;`

GeorgiChochov: Hah, nope, missed it

Issue #1115, opened 2019-12-18
Webjar doesn't work within an iframe


benedeki:
## Describe the bug
Menas, particuallry Spline webjar, doesn't work within an iframe

## To Reproduce
Put _Menas_ (_`http://localhost:8080/menas/`_) or its _Spline_ (_`http://localhost:8080/menas/lineage/app/`_) webjar link into `iframe`.
See it's not rendering

## Expected behaviour
It works.

Issue #1117, opened 2019-12-18
Proper error pages if lineage is not found/too many exists


benedeki:
## Background
Integrate Spline 0.4.x with Enceladus

## Feature
When no proper Lineage is found for a Run, there's a message but it's comes in a somewhat improper time. Better implementation would serve users better.

PR #1137, opened 2020-01-06
#1115: Webjar doesn't work within an iframe

benedeki:
* header to allow iframe display `<frame-options policy="SAMEORIGIN"/>`
* hard-coding the link to lineage ui (using webjar within the app)
* removed API not used anymore, due to hard-coding


menas/ui/service/RunService.js
```diff
@@ -126,27 +126,20 @@ var RunService = new function () {
...
-          errorMessage = "No lineage found";
-        } else {
-          errorMessage = "Multiple lineage records found";
-        }
-        sap.m.MessageBox.error(errorMessage);
-        return "";
+        errorMessage = "Multiple lineage records found";
       }
-    } else {
+      sap.m.MessageBox.error(errorMessage);
```
Zejnilovic: I have a code suggestion for line 133-144
```javascript
if (lineageIdInfo.totalCount !== 1) {
  let errorMessage = !!lineageIdInfo.totalCount ? "Multiple lineage records found" : "No lineage found";
  sap.m.MessageBox.error(errorMessage);
  return "";
}

return urlTemplate.replace("%s", lineageIdInfo.executionEventId);
```

basically, if something went wrong deal with it, otherwise let the method flow to its end. This, in my opinion, reads more like JS, than again, I am no JS guru.

benedeki: I plan to address this section of the code in #1117 . I might use your suggestion then.

PR #1138, opened 2020-01-09
Feature/1078 multilevel array transform

yruslan:
Extended array transformation functions to be able to access fields on all levels of nesting.


utils/src/main/scala/za/co/absa/enceladus/utils/transformations/ArrayContext.scala
```diff
@@ -0,0 +1,69 @@
...
+import za.co.absa.enceladus.utils.transformations.DeepArrayTransformations.splitByLongestParent
+
+import scala.collection.mutable.ArrayBuffer
+
+/**
+  * The class provides a storage for array transformation context for a transformation of a dataframe field.
+  * The context contains all arrays in the path of the field and their corresponding array element lambda variables
+  * provided by 'transform()' function of Spark SQL.
+  */
+class ArrayContext(val arrayPaths: ArrayBuffer[String] = new ArrayBuffer[String],
```
lokm01: 👍

utils/src/main/scala/za/co/absa/enceladus/utils/transformations/DeepArrayTransformations.scala
```diff
@@ -652,11 +789,79 @@ object DeepArrayTransformations {
...
+    *  split into parent and child parts and these parts are returned as a pair.</li>
+    *  <li>If the input field is not a child of any of the listed parents the method returns a pair of empty string and
+    *  the full qualified input field name.</li>
+    * </ul>
+    *
+    * @param field        A fully qualified input field name
+    * @param parentFields A parent field to split in case the input field is its child
+    * @return A pair consisting of the parent and the child field parts
+    */
+  private[transformations] def splitByLongestParent(field: String, parentFields: Seq[String]): (String, String) = {
```
Zejnilovic: What does it mean by the longest parent? Is `alfa.beta` (9 chars) longer than `a.b.c.d` (7 chars)? One is longer, the other is deeper. I see you use length, but the tests don't cover this, so I am still curious.

utils/src/test/scala/za/co/absa/enceladus/utils/transformations/DeepArrayErrorTransformationSuite.scala
```diff
@@ -857,6 +857,97 @@ class DeepArrayErrorTransformationSuite extends FunSuite with SparkTestBase with
...
+        |  }, {
+        |    "errType" : "confCastError",
+        |    "errCode" : "E00003",
+        |    "errMsg" : "Conformance Error - Null returned by casting conformance rule",
+        |    "errCol" : "legs.conditions.checks.optimizedNums",
+        |    "rawValues" : [ "5c" ],
+        |    "mappings" : [ ]
+        |  } ]
+        |} ]"""
+        .stripMargin.replace("\r\n", "\n")
```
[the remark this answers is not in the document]
yruslan: I wanted to be per-file consistent. Since all other tests in this file are written this way, a new test follows the pattern.

My plan is to move this to 'spark-hats' permanently (code and tests), and then make everything consistent inside 'spark-hats'.

utils/src/main/scala/za/co/absa/enceladus/utils/transformations/DeepArrayTransformations.scala
yruslan: Yeah, the terminology is too fuzzy. Going to rename to `splitByDeepestParent()`. But these methods are called only for fields in a path to one field, so the longest and the deepest here are the same.  I will clarify this in the ScalaDoc comments.

E.g. if a field being transformed is `a.b.c.d`, and 'a' and 'c' are arrays, the list of arrays will contain `a`, `a.b.c` so that 'a.x' and `a.b.x` will have `a` as the parent, while `a.b.c.d.e` will have `a.b.c` as the parent.

utils/src/main/scala/za/co/absa/enceladus/utils/broadcast/BroadcastUtils.scala
```diff
@@ -250,10 +250,18 @@ object BroadcastUtils {
...
-          param4.toString, param5.toString), mappings)
+        ErrorMessage.confMappingErr(outputColumn, Seq(safeToString(param1), safeToString(param2), safeToString(param3),
+          safeToString(param4), safeToString(param5)), mappings)
       }
     }
   }
 
+  private final def safeToString(a: Any): String = {
+    if (a == null) {
+      null
```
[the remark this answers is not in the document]
yruslan: This is to be consistent with the current behavior of other implementations of mapping rule. They write `null` in such cases.

utils/src/test/scala/za/co/absa/enceladus/utils/transformations/ExtendedTransformationsSuite.scala
```diff
@@ -0,0 +1,208 @@
...
+
+  private val nestedTestCaseFactory = new NestedTestCaseFactory()
+
+  test("Test extended array transformations work on root level fields") {
+    val expectedSchema = getResourceString("/test_data/nested/nested1Schema.txt")
+    val expectedResults = getResourceString("/test_data/nested/nested1Results.json")
+
+    val df = nestedTestCaseFactory.getTestCase
+
+    val dfOut = DeepArrayTransformations.nestedExtendedStructMap(df, "", "id_str", (_, gf) =>
```
GeorgiChochov: What does `gf` stand for?

utils/src/main/scala/za/co/absa/enceladus/utils/transformations/ArrayContext.scala
```diff
@@ -0,0 +1,69 @@
...
+    if (parentArray.isEmpty) {
+      col(childField)
+    } else {
+      val i = arrayPaths.indexOf(parentArray)
+      if (fieldName == arrayPaths(i)) {
+        // If the array itself is specified - return the array
+        lambdaVars(i)
+      } else {
+        // If a field inside an array is specified - return the field
+        // by using '.getField()' on each child (which could be a nested struct
```
GeorgiChochov:
```suggestion
        // by using '.getField()' on each child (which could be a nested struct)
```

utils/src/main/scala/za/co/absa/enceladus/utils/transformations/DeepArrayTransformations.scala
Zejnilovic: In the tests, there is a list `a, a.b, aa.bb` etc. So what if I keep the logic but modify it a bit to a list of `a, a.b.c, a.dddd`. Here `length` wouldn't work, right? Is this UseCase even plausible or is that just that tests are written that way?

Zejnilovic: counting the number of `.` (dosts) seems safer right now to me

yruslan: Cool idea! This solves a more generic case.

utils/src/test/scala/za/co/absa/enceladus/utils/transformations/DeepArrayTransformationSuite.scala
```diff
@@ -829,6 +829,41 @@ class DeepArrayTransformationSuite extends FunSuite with SparkTestBase with Logg
...
+
+    assert(splitParentField("a.b.c", "") == ("", "a.b.c"))
+    assert(splitParentField("a.b.c", "b") == ("", "a.b.c"))
+    assert(splitParentField("a.b.c", "c") == ("", "a.b.c"))
+    assert(splitParentField("aa.bb.cc", "a") == ("", "aa.bb.cc"))
+    assert(splitParentField("aa.bb.cc", "aa.b") == ("", "aa.bb.cc"))
+    assert(splitParentField("aa.bb.cc", "aa.bb") == ("aa.bb", "cc"))
+    assert(splitParentField("a.b.c", "a") == ("a", "b.c"))
+    assert(splitParentField("a.b.c", "a.b") == ("a.b", "c"))
+    assert(splitParentField("a.b.c", "a.b.c") == ("a.b.c", ""))
```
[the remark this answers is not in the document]
yruslan: Yeah, it won't hurt to handle whitespaces as well

utils/src/main/scala/za/co/absa/enceladus/utils/transformations/ArrayContext.scala
```diff
@@ -0,0 +1,69 @@
...
+
+import scala.collection.mutable.ArrayBuffer
+
+/**
+  * The class provides a storage for array transformation context for a transformation of a dataframe field.
+  * The context contains all arrays in the path of the field and their corresponding array element lambda variables
+  * provided by 'transform()' function of Spark SQL.
+  */
+class ArrayContext(val arrayPaths: ArrayBuffer[String] = new ArrayBuffer[String],
+                   val lambdaVars: ArrayBuffer[Column] = new ArrayBuffer[Column]) {
```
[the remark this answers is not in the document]
yruslan: I needed an indexed collection and mutability is just an implementation detail.
`immutable.Vector` might be a better fix from the clearness of the concept point of view. But `Vector` is so 'heavy'...

Now that you've mentioned it - I have a better idea. I'm going to rewrite this piece of code.

utils/src/main/scala/za/co/absa/enceladus/utils/transformations/DeepArrayTransformations.scala
yruslan: > In the tests, there is a list `a, a.b, aa.bb` etc. So what if I keep the logic but modify it a bit to a list of `a, a.b.c, a.dddd`. Here `length` wouldn't work, right? Is this UseCase even plausible or is that just that tests are written that way?

After thinking a bit I think counting dots and using `.length()` is equivalent since all parent path must be subpath of the input field. The test case includes a situation when the list of parent fields does not consist of a single subpath. This case is supported but should not happen in practice in array transformations.

Zejnilovic: You are right. Wanna change it back, or keep it?

yruslan: Will change back and add a test that will fail on dot counting

utils/src/main/scala/za/co/absa/enceladus/utils/transformations/ArrayContext.scala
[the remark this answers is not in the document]
yruslan: I experimented with usage of `s: Set[String,Column]`, but saw that for `splitByDeepestParent()` I need an array of strings so I would need to do `s.map(_._1)` (or similar matching) and decided that it is ugly and has a slow performance. But then I realized I could just use `Array`, lol 😄 .
So I've rewritten this using `Seq[]` as the static type and `Array[]` as the dynamic type. This should combine the best of conceptual cleanness and performance.

conformance/src/test/scala/za/co/absa/enceladus/conformance/interpreter/rules/MappingRuleBroadcastSuite.scala
```diff
@@ -253,7 +251,7 @@ class MappingRuleBroadcastSuite extends FunSuite with SparkTestBase with LoggerT
     }
   }
 
-  private def cleanupSchema(inputSchemaTree: String): String = {
+  private def cleanupContainsNullProperty(inputSchemaTree: String): String = {
```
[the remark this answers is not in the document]
yruslan: The mapping rule implementation that uses join creates arrays with `containsNull = false`, while the broadcasting strategy creates arrays that have `containsNull = true` (or it is the opposite - I don't remember). This method just removes this schema property from schema strings so that unit tests ignore that difference.

utils/src/main/scala/za/co/absa/enceladus/utils/broadcast/BroadcastUtils.scala
```diff
@@ -250,10 +250,18 @@ object BroadcastUtils {
...
       } else {
-        ErrorMessage.confMappingErr(outputColumn, Seq(param1.toString, param2.toString, param3.toString,
-          param4.toString, param5.toString), mappings)
+        ErrorMessage.confMappingErr(outputColumn, Seq(safeToString(param1), safeToString(param2), safeToString(param3),
+          safeToString(param4), safeToString(param5)), mappings)
       }
     }
   }
 
+  private final def safeToString(a: Any): String = {
```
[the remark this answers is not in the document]
yruslan: The 'safe' here means that if I do `a.toString` when `a` is null I'd get `NullPointerException`, but `safeToString()` handles this situation.

utils/src/main/scala/za/co/absa/enceladus/utils/transformations/DeepArrayTransformations.scala
```diff
@@ -652,11 +789,84 @@ object DeepArrayTransformations {
...
+    *   parts are returned as a pair.</li>
+    *   <li>If the input field is not a child of the parent field the method returns a pair of empty string and
+    *   the full qualified input field name.</li>
+    * </ul>
+    *
+    * @param field       A fully qualified input field name
+    * @param parentField A parent field to split in case the input field is its child
+    * @return A pair consisting of the parent and the child field parts
+    */
+  private[transformations] def splitParentField(field: String, parentField: String): (String, String) = {
```
benedeki: Question: Why `private`?

utils/src/main/scala/za/co/absa/enceladus/utils/transformations/DeepArrayTransformations.scala
```diff
@@ -652,11 +789,84 @@ object DeepArrayTransformations {
...
+    *  split into parent and child parts and these parts are returned as a pair.</li>
+    *  <li>If the input field is not a child of any of the listed parents the method returns a pair of empty string and
+    *  the full qualified input field name.</li>
+    * </ul>
+    *
+    * @param field        A fully qualified input field name
+    * @param parentFields A parent field to split in case the input field is its child
+    * @return A pair consisting of the parent and the child field parts
+    */
+  private[transformations] def splitByDeepestParent(field: String, parentFields: Seq[String]): (String, String) = {
```
benedeki: Same question:  Why `private`?

utils/src/main/scala/za/co/absa/enceladus/utils/transformations/DeepArrayTransformations.scala
```diff
@@ -652,11 +789,84 @@ object DeepArrayTransformations {
 
   }
 
-  /** Checks if a path is a leaf element
+  private def validateErrorColumnName(errorColumnName: String): Unit = {
+    if (errorColumnName.contains('.')) {
+      throw new IllegalArgumentException(s"Error columns should be at the root schema level. " +
```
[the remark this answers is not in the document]
benedeki: Fair enough. Thank you for the explanation 👍

utils/src/main/scala/za/co/absa/enceladus/utils/transformations/DeepArrayTransformations.scala
yruslan: I didn't want to bother with validations so make it for my particular use case. But yeah, it can be promoted to `SchemaUtils` later probably.

utils/src/main/scala/za/co/absa/enceladus/utils/transformations/DeepArrayTransformations.scala
yruslan: This function definitely does not cover all use cases that could be provided by a client.

PR #1139, opened 2020-01-09
1120 Allow entity recreation

GeorgiChochov:


Zejnilovic:
Works fine now. Just a question. There is no info about it being disabled previously or anything like that. Audit trails don't show anything and you can still look at previous versions. Is that ok?

GeorgiChochov:
> Works fine now. Just a question. There is no info about it being disabled previously or anything like that. Audit trails don't show anything and you can still look at previous versions. Is that ok?

There never was an audit thing for disabling (it's the mutable part of our data model). It used to be that you couldn't see disabled versions, now you can, but they're still not listed in the navigation menu.

PR #1146, opened 2020-01-13
1145 Remove external modules

GeorgiChochov:


Zejnilovic:
BTW this should fail on build until you create the fixed-width project

GeorgiChochov:
> BTW this should fail on build until you create the fixed-width project

No, it can just use the latest version of the released enceladus-fixedWidth library (which has been unchanged since 1.0.0), no need to fail anything.
Not sure if I want to create a project for fixed-width at all at the moment.
Definitely not creating a project for the `kafka` module as it was dead weight.

Zejnilovic:
Fair enough

GeorgiChochov:
> Just so the PR is not accidentally merged before the rebase

It can't be merged, there are merge conflicts, but thanks.

Issue #1136, opened 2020-01-06
Switch Enceladus to use spark-hats for array transfromations


yruslan:
## Background
`spark-hats` (https://github.com/AbsaOSS/spark-hats) is created from Enceladus by extracting deep array transformation functions.

## Feature
Now that array transformation helpers are in a separate project it makes sense to use it from Enceladus.

yruslan:
Time spent in progress: 4 hours.

yruslan:
Release notes
Nested array transformations API were extracted into a separate project: [spark-hats](https://github.com/AbsaOSS/spark-hats).

PR #1155, opened 2020-01-16
#1136 Use nested array transformation routines from 'spark-hats'.

yruslan:
* Deep array transformations are removed from the project
* 'spark-hats' is used instead.
* Use the newer version of 'spark-hofs'.


spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/interpreter/rules/ConcatenationRuleInterpreter.scala
```diff
@@ -22,9 +22,10 @@ import za.co.absa.enceladus.conformance.ConfCmdConfig
 import za.co.absa.enceladus.conformance.interpreter.{ExplosionState, RuleValidators}
 import za.co.absa.enceladus.dao.MenasDAO
 import za.co.absa.enceladus.model.conformanceRule.{ConcatenationConformanceRule, ConformanceRule}
-import za.co.absa.enceladus.utils.transformations.DeepArrayTransformations
 
 case class ConcatenationRuleInterpreter(rule: ConcatenationConformanceRule) extends RuleInterpreter {
+  import za.co.absa.spark.hats.Extensions._
```
benedeki: To discuss: Based on the repeated import occurrence, I wonder, if it would be possible and useful to extract the logic using `df.nestedWithColumnExtended' into one code block in the ancestor?

yruslan:
> I guess the hats files are equivalent to `DeepArrayTransformations`?

Yes, and it stands for "Helpers for Array TransformationS"

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/interpreter/rules/ConcatenationRuleInterpreter.scala
yruslan: I'm not sure. `df.nestedYYYY()` are meant to extend Dataframe DSL and to have a feel of being part of its API. I'd like them not to be grouped as something special.

benedeki: Fair enough.

Issue #1154, opened 2020-01-16
Merge of Standardization and Conformance is not reflected in Oozie


benedeki:
## Describe the bug
**Standardization** and **Conformance** were merged into one `SparkJob.jar` . This is not reflected in Oozie template (namely `Menas/src/main/resources/scheduling/oozie/workflow_template.xml`)

yruslan:
👍 Good catch!

PR #1156, opened 2020-01-16
#1154: Merge of Standardization and Conformance is not reflected in Oozie

benedeki:
* Oozie template updated
* New shared property for *Standardization* and *Conformance* jar location
* Few code improvement suggestions

### DIFF
--- a/menas/src/main/scala/za/co/absa/enceladus/menas/repositories/OozieRepository.scala
+++ b/menas/src/main/scala/za/co/absa/enceladus/menas/repositories/OozieRepository.scala
@@ -20,18 +20,16 @@ import java.io.InputStreamReader
 import java.net.HttpURLConnection
 import java.net.URL
 import java.text.SimpleDateFormat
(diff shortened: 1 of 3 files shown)


menas/src/main/scala/za/co/absa/enceladus/menas/repositories/OozieRepository.scala
```diff
@@ -381,10 +359,8 @@ class OozieRepository @Autowired() (oozieClientRes: Either[OozieConfigurationExc
    */
   private def getCoordinatorFromTemplate(ds: Dataset, wfPath: String): Array[Byte] = {
     val schedule = ds.schedule.get
-    val runtimeParams = schedule.runtimeParams
```
benedeki: I would like to ask particularly @lokm01 if `runtimeParams` and `timezoneOffset ` (couple of lines bellow) not being used is not some omission and it's OK to remove

Issue #1129, opened 2020-01-06
Use broadcasting mapping rule for batch jobs


yruslan:
## Background
Currently, the broadcasting strategy for the mapping conformance rule is never selected.

## Feature
Add a logic that checks the size of the mapping table to decide which strategy to use. 

## Proposed Solution [Optional]
Use the broadcasting strategy if the mapping table size is less than a configured limit.

yruslan:
Release notes
The broadcasting mapping rule strategy can now be used in experimental mode. To do this set conformance.mapping.rule.broadcast=auto as the option to run dynamic conformance. The conformance.mapping.rule.max.broadcast.size.mb option specifies the maximum mapping table size that is allowed to be broadcasted.

PR #1158, opened 2020-01-21
#1129 Add a configuration switch to control the broadcasting strategy  for mapping rules.

yruslan:
* auto - will always use the broadcasting strategy.
* never - never user the broadcasting strategy.
* auto - use the strategy depending on the size of the mapping table.

### DIFF
--- a/spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/interpreter/DynamicInterpreter.scala
+++ b/spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/interpreter/DynamicInterpreter.scala
@@ -15,6 +15,7 @@
 
 package za.co.absa.enceladus.conformance.interpreter
 
+import com.typesafe.config.{Config, ConfigFactory}
 import org.apache.spark.sql.execution.command.ExplainCommand
 import org.apache.spark.sql.functions._
 import org.apache.spark.sql.types.StructType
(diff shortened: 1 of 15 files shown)


spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/interpreter/DynamicInterpreter.scala
```diff
@@ -188,20 +190,21 @@ object DynamicInterpreter {
...
-  private def getMappingRuleInterpreter(rule: MappingConformanceRule,
-                                        schema: StructType)
+  private def getMappingRuleInterpreter(rule: MappingConformanceRule)
                                        (implicit ictx: InterpreterContext): RuleInterpreter = {
-    if (ictx.featureSwitches.experimentalMappingRuleEnabled) {
-      if (canMappingRuleBroadcast(rule, schema)) {
-        MappingRuleInterpreterBroadcast(rule, ictx.conformance)
-      } else {
+    if (canMappingRuleBroadcast(rule)) {
+      log.info("Broadcast strategy for mapping rules is used")
```
benedeki: I wonder if he log level `info` is appropriate here. Wouldn't `debug` be more suitable?

yruslan: I'm more inclined towards `info` since it is possible that the choice of the strategy affects the results. So this information is useful in prod logs.

benedeki: Fair enough

yruslan:
Release notes
The broadcasting mapping rule strategy can now be used in experimental mode. To do this set `conformance.mapping.rule.broadcast=auto` as the option to run dynamic conformance. The `conformance.mapping.rule.max.broadcast.size.mb` option specifies the maximum mapping table size that is allowed to be broadcasted.

Issue #1101, opened 2019-12-10
I am unable to set a standardization pattern for dates and timestamps in an array.


HuvarVer:
## Describe the bug
If I need to standardize data with dates or timestamps in an array, I have to set a pattern for them, but there is no possibility. 

## Expected behavior
I should be able to set a pattern for dates and timestamps in arrays. Because standardization needs it for these data types.

## Additional context
The pattern in metadata of array doesn't work, the error in a terminal is: `Dates & times represented as array values need specified 'pattern' metadata.`
If the pattern is in the metadata struct or alone in array type struct that `the file isn't a valid Spark StructType JSON schema`. This error is at the uploading of the schema in Menas.

`{
        "name": "timestamps_array",
        "type": {
            "type": "array",
            "elementType": "timestamp",
            "containsNull": true,
            "pattern": "dd/MM/yyyy HH:mm:ss zzz"
            },
        "nullable": true,
        "metadata": {}
    }`
`{
        "name": "timestamps_array",
        "type": {
            "type": "array",
            "elementType": "timestamp",
            "containsNull": true,
            "metadata": {
               "pattern": "dd/MM/yyyy HH:mm:ss zzz"
                 }          
         },
        "nullable": true,
        "metadata": {}
    }`

benedeki:
Release note:
Support of metadata in arrays of simple types. E.g. `pattern` can be specified for array of timestamps.

PR #1161, opened 2020-01-22
#1101: Metadata for array element types

benedeki:
* support to pass the array metadata to the underlying element type
* verification of metadata of array type based on the element type

### DIFF
--- a/spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
+++ b/spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
@@ -237,7 +237,7 @@ object TypeParser {
       SchemaUtils.getFieldType(inputFullPathName, origSchema).get
     }
 
-    protected def origTypeButArray: DataType = {
+    protected def underlyingType: DataType = {
       origType match {
         case arrayType: ArrayType => arrayType.elementType
         case anyOther => anyOther
@@ -260,7 +260,7 @@ object TypeParser {
     }
 
     override def assemblePrimitiveCastLogic: Column = {
-      if (origTypeButArray == StringType) {
+      if (underlyingType == StringType) {
         // in case of string as source some adjustments might be needed
         val decimalSymbols = field.pattern.toOption.flatten.map(
           _.decimalSymbols
@@ -304,7 +304,7 @@ object TypeParser {
     override protected def assemblePrimitiveCastErrorLogic(castedCol: Column): Column = {
       val basicLogic: Column = super.assemblePrimitiveCastErrorLogic(castedCol)
 
-      origTypeButArray match {
+      underlyingType match {
         case  dt: DecimalType =>
           // decimal can be too big, to catch overflow or imprecision  issues compare to original
           basicLogic or (column =!= castedCol.cast(dt))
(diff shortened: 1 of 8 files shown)


benedeki:
Code smells are code duplication warnings in UTs.

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
```diff
@@ -237,6 +237,12 @@ object TypeParser {
       SchemaUtils.getFieldType(inputFullPathName, origSchema).get
     }
 
+    protected def origTypeButArray: DataType = {
```
GeorgiChochov: I got very confused by this name. I thought it might mean "original type... but as an array".
How about `origElemType` for "original element type" or `underlyingType` maybe?
Not an easy thing to name...

benedeki: I was actually hoping for some good name suggestions. Wasn't happy with this one myself (but best I came up with).
It's the `origType` except for array's when it's the `origElementType`.
`underlyingType` is nicely vague - so not misleading. But I am not sure it actually says in any way, what is it. But perhaps better then what I have. 
Let's see if others comment.

spark-jobs/src/test/scala/za/co/absa/enceladus/standardization/interpreter/StandardizationInterpreter_ArraySuite.scala
```diff
@@ -0,0 +1,218 @@
...
+  test("Array of timestamps with pattern defined") {
+    val seq  = Seq(
+      Array("00:00:00 01.12.2018", "00:10:00 02.12.2018","00:20:00 03.12.2018"),
+      Array("00:00:00 01.12.2019", "00:10:00 02.12.2019","00:20:00 03.12.2019"),
+      Array("2020-01-12 00:00:00" ,"2020-12-02 00:10:00","2020-12-03 00:20:00")
+    )
+    val src = seq.toDF(fieldName)
+    val desiredSchema = generateDesiredSchema(TimestampType, s""""${MetadataKeys.Pattern}": "HH:mm:ss dd.MM.yyyy"""")
+
+    val expectedData =
```
yruslan: When switching from Spark 2.2.x to 2.4.4 we have noticed that the format of `.show()` is changed. So in order to make unit tests lee fragile we prefer exporting expected data as JSONs in resources.
Dataframe JSONs can be prettified for easier comparison between the expected and actual data.
This method can be used for that: `JsonUtils.prettyJSON(inputJson)`.

But it is just a suggestion - I'm fine with this implementation as well. The switch to the different way could be done later in bulk.

spark-jobs/src/test/scala/za/co/absa/enceladus/standardization/interpreter/StandardizationInterpreter_ArraySuite.scala
```diff
@@ -0,0 +1,218 @@
...
+import za.co.absa.enceladus.utils.validation.ValidationException
+
+class StandardizationInterpreter_ArraySuite extends FunSuite with SparkTestBase with LoggerTestBase {
+  import spark.implicits._
+
+  private implicit val udfLib: UDFLibrary = new UDFLibrary
+
+  private val fieldName = "arrayField"
+
+  private val errColSchema =  "\n |-- errCol: array (nullable = true)\n"+
```
yruslan: Just a suggestion. Moving test case generators to a separate class (test case factories) can help reuse the test cases in different suites.

Issue #1173, opened 2020-02-03
Add Run search by Spark app_id 


DzMakatun:
## Background
It is needed to find corresponding Run documents for given Spark History data

## Feature
Add API api/runs/bySparkAppId/{spark_app_id}
which will return Run document where std_application_id OR conform_application_id equals {spark_app_id}

## Proposed Solution [Optional]
Use mongo query
```
.find(
  {$or: [
      {'controlMeasure.metadata.additionalInfo.std_application_id': spark_app_id},
      {'controlMeasure.metadata.additionalInfo.conform_application_id': spark_app_id}
      ]
  }
)
```

GeorgiChochov:
@DzMakatun I think this is done, right? Could you mark it as part of release 2.1.0, add some release notes here and close it?

DzMakatun:
Release Notes
Add run search by Spark application id via API /runs/bySparkAppId/{spark_app_id}

PR #1174, opened 2020-02-03
Feature/1173 add run search by spark app

DzMakatun:
#1173 Add Run search by Spark app_id

It is needed to find corresponding Run documents for given Spark History data

Add API api/runs/bySparkAppId/{spark_app_id}
which will return Run document where std_application_id OR conform_application_id equals {spark_app_id}

### DIFF
--- a/menas/src/main/scala/za/co/absa/enceladus/menas/controllers/RunController.scala
+++ b/menas/src/main/scala/za/co/absa/enceladus/menas/controllers/RunController.scala
@@ -72,6 +72,12 @@ class RunController @Autowired()(runService: RunService) extends BaseController 
     runService.getRunSummariesPerDatasetVersion(datasetName)
   }
 
+  @GetMapping(Array("/bySparkAppId/{appId}"))
+  @ResponseStatus(HttpStatus.OK)
+  def getRunBySparkAppId(@PathVariable appId: String): CompletableFuture[String] = {
+    runService.getRunBySparkAppId(appId).map(ControlUtils.asJson)
+  }
+
   @GetMapping(Array("/{datasetName}/{datasetVersion}"))
   @ResponseStatus(HttpStatus.OK)
   def getSummariesByDatasetNameAndVersion(@PathVariable datasetName: String,
--- a/menas/src/main/scala/za/co/absa/enceladus/menas/repositories/RunMongoRepository.scala
+++ b/menas/src/main/scala/za/co/absa/enceladus/menas/repositories/RunMongoRepository.scala
@@ -283,7 +283,7 @@ class RunMongoRepository @Autowired()(mongoDb: MongoDatabase)
 
   def getRunBySparkAppId(appId: String): Future[Option[Run]] = {
     val stdAppIdFilter = equal("controlMeasure.metadata.additionalInfo.std_application_id", appId)
-    val conformAppIdFilter =  equal("controlMeasure.metadata.additionalInfo.conform_application_id", appId)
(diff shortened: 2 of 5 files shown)


Zejnilovic:
Just a quick question. App id is "usable" only using cluster mode as per [this code](https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/SparkContext.scala#L329) (really hope I am not reading this wrongly). Aren't old runs done using client mode? Might there be a problem or duplicity?

DzMakatun:
> Just a quick question. App id is "usable" only using cluster mode as per [this code of 2.4.4](https://github.com/apache/spark/blob/v2.4.4/core/src/main/scala/org/apache/spark/SparkContext.scala#L307) (really hope I am not reading this wrongly). Aren't old runs done using client mode? Might there be a problem or duplicity?

Good question, thnx. Here are my thoughts:
1. Spark app_id is assigned by scheduler in both cluster and client modes
2. The app_id format depends on the scheduler used (local, YARN or MESOS)
3. It is a responsibility of a scheduler to ensure the app_id is unique for the particular scheduler instance. 

To stay on the safe side, I change the implementation, so that it will return a list of all Runs with the given app_id.

menas/src/main/scala/za/co/absa/enceladus/menas/repositories/RunMongoRepository.scala
```diff
@@ -281,6 +281,16 @@ class RunMongoRepository @Autowired()(mongoDb: MongoDatabase)
       .toFuture()
   }
 
+  def getRunBySparkAppId(appId: String): Future[Seq[Run]] = {
```
GeorgiChochov: Could you write some tests for this in the `RunRepositoryIntegrationSuite`?

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/RunController.scala
```diff
@@ -72,6 +72,12 @@ class RunController @Autowired()(runService: RunService) extends BaseController
     runService.getRunSummariesPerDatasetVersion(datasetName)
   }
 
+  @GetMapping(Array("/bySparkAppId/{appId}"))
+  @ResponseStatus(HttpStatus.OK)
+  def getRunBySparkAppId(@PathVariable appId: String): CompletableFuture[String] = {
```
GeorgiChochov: Could you write some tests for this in the `RunApiIntegrationSuite`?

menas/src/main/scala/za/co/absa/enceladus/menas/repositories/RunMongoRepository.scala
DzMakatun: Added in 013c508 , thnx

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/RunController.scala
DzMakatun: Added in 1119d36, thnx

DzMakatun:
@GeorgiChochov @Zejnilovic 
Now the SonarCloud complains about duplication because of the code which existed in the integration tests before the commit. What can be done about it?

GeorgiChochov:
> @GeorgiChochov @Zejnilovic
> Now the SonarCloud complains about duplication because of the code which existed in the integration tests before the commit. What can be done about it?

It's fine, don't worry about sonarcloud. Maybe we'll revisit and refactor in the future.

Issue #1197, opened 2020-02-21
Push information from Enceladus to Kafka


benedeki:
## Background
WE need to enable pushing metadata, metrics etc. from Enceladus to Kafka

PR #1208, opened 2020-03-04
Push control info measurements to a Kafka topic.

### LINKED ISSUE #1202: Increase checks when invoking Enceladus plugins factories
## Background
Enceladus plugin factories can return `null` or throw exceptions.

## Feature
Enceladus Spark Jobs should handle these situations.

yruslan:
This fixes #1202 and #1197 .

The Kafka plugin is moved to Enceladus as a built-in plugin. And it also serves as an example of how to create plugins for Enceladus.


pom.xml
```diff
@@ -102,18 +102,26 @@
...
+                <updatePolicy>never</updatePolicy>
+            </releases>
+        </repository>
     </repositories>
 
     <modules>
-        <module>plugins-api</module>
-        <module>utils</module>
-        <module>spark-jobs</module>
+        <module>dao</module>
```
yruslan: In case you are wondering, I've sorted modules by name alphabetically.

plugins-builtin/src/main/scala/za/co/absa/enceladus/plugins/builtin/kafka/KafkaSecurityParams.scala
```diff
@@ -0,0 +1,62 @@
...
+   * @param conf A configuration.
+   * @return An instance of Kafka security parameters.
+   */
+  def fromConfig(conf: Config): Option[KafkaSecurityParams] = {
+    if (conf.hasPath(SecurityProtocolKey)) {
+      Option(KafkaSecurityParams(conf.getString(SecurityProtocolKey),
+        getOptionConf(conf, SaslMechanismKey)))
+    } else {
+      None
+    }
```
GeorgiChochov:
```suggestion
    getOptionConf(conf, SaslMechanismKey).map { securityProtocol =>
     KafkaSecurityParams(securityProtocol,
        getOptionConf(conf, SaslMechanismKey))
    }
```

plugins-builtin/src/main/scala/za/co/absa/enceladus/plugins/builtin/kafka/ControlInfoAvroSerializer.scala
```diff
@@ -0,0 +1,140 @@
...
+
+    try {
+      val avroMetadata = convertMetadata(controlInfo.controlMeasure.metadata, recordAvroSchema)
+      val avroCheckpoints = convertCheckpoints(controlInfo.controlMeasure.checkpoints, recordAvroSchema)
+
+      avroDceControlInfo.put("datasetName", controlInfo.datasetName)
+      avroDceControlInfo.put("datasetVersion", controlInfo.datasetVersion)
+      avroDceControlInfo.put("reportDate", controlInfo.reportDate)
+      avroDceControlInfo.put("reportVersion", controlInfo.reportVersion)
+      avroDceControlInfo.put("runStatus", controlInfo.runStatus)
```
GeorgiChochov: These are repeated in `KafkaInfoPlugin` and are prime candidates for constant extraction. (The rest of the String literals in this class might not be repeated elsewhere, but I'd still say they can be extracted as constants to a companion object)

plugins-builtin/src/main/resources/info_file_avro_schema.avsc
```diff
@@ -0,0 +1,144 @@
...
+                "fields": [
+                  {
+                    "name": "key",
+                    "type": "string"
+                  },
+                  {
+                    "name": "value",
+                    "type": "string"
+                  }
+                ]
```
Zejnilovic: Why not use `map` here instead of `array`?

yruslan: This is something to discuss with the downstream team.

plugins-builtin/src/main/scala/za/co/absa/enceladus/plugins/builtin/kafka/ControlInfoAvroSerializer.scala
yruslan: Let's discuss this

pom.xml
```diff
@@ -102,18 +102,26 @@
             <name>OpenUI5-Absa</name>
             <url>https://dl.bintray.com/absa/openui5</url>
         </repository>
+        <repository>
+            <id>confluent</id>
+            <url>https://packages.confluent.io/maven/</url>
+            <releases>
+                <updatePolicy>never</updatePolicy>
```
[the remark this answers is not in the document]
yruslan: I just took the definition from Abris as is. But this might speed up the build by not checking updates frequently. Once an artifact is downloaded we assume it is always recent. I guess...

plugins-builtin/src/main/scala/za/co/absa/enceladus/plugins/builtin/kafka/ControlInfoAvroSerializer.scala
```diff
@@ -0,0 +1,140 @@
+/*
```
benedeki: For consideration: I would add a `controlMetricsPlugin` package before `kafka`. And maybe not use `builtin`.

spark-jobs/src/main/scala/za/co/absa/enceladus/common/plugin/PluginLoader.scala
```diff
@@ -46,7 +46,13 @@ class PluginLoader[+A <: Plugin:ClassTag:universe.TypeTag] {
       val key = s"$configKeyPrefix.$i"
       val factoryName = config.getString(key)
       log.info(s"Going to load a plugin factory for configuration: '$key'. Factory name: $factoryName")
-      plugins += buildPlugin(factoryName, config)
+      val plugin = buildPlugin(factoryName, config)
+      if (plugin == null) {
+        log.error(s"A NULL is returned when building a plugin: '$key'. Factory name: $factoryName")
```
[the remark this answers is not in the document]
yruslan: I left it as an option for a plugin developer. If the plugin is critical to the job, the factory will throw an exception. If it is not critical, it will return `null` that will be converted to Option as per Georgi suggestion.

plugins-builtin/src/main/scala/za/co/absa/enceladus/plugins/builtin/kafka/ControlInfoProducer.scala
```diff
@@ -0,0 +1,29 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.enceladus.plugins.builtin.kafka
+
+/**
+ * Base interface for control info metrics (aka INFO file) producer for messaging queues.
+ */
+trait ControlInfoProducer {
```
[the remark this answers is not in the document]
yruslan: Yeah, I might put it somewhere above.

spark-jobs/src/main/scala/za/co/absa/enceladus/common/plugin/PluginLoader.scala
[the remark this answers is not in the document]
benedeki: Fair enough

plugins-builtin/src/main/scala/za/co/absa/enceladus/plugins/builtin/kafka/KafkaInfoPlugin.scala
```diff
@@ -0,0 +1,52 @@
...
+
+import com.typesafe.config.Config
+import za.co.absa.atum.model.ControlMeasure
+import za.co.absa.enceladus.plugins.api.control.{ControlMetricsPlugin, ControlMetricsPluginFactory}
+
+/**
+ * Implementation of a plugin that sends control information to a Kafka topic when
+ * a checkpoint is created or when run status is changed.
+ */
+class KafkaInfoPlugin(producer: ControlInfoProducer) extends ControlMetricsPlugin {
```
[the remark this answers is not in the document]
benedeki: Sorry if I was not clear in my description. The name `KafkaInfoPlugin` suggests that it will have something to do with Kafka. But there is nothing that would connect it with Kafka in it's logic. Even the producer is `ControlInfoProducer` trait. That "Kafkaness" comes in the form that producer added in the factory is `ControlInfoProducerKafka`.
The discrepancy should be imho addressed, and I see two possibilities:
1) Rename the class and move it higher in the package hierarchy. E.g. `SenderInfoPlugin`. (My suggestion might be poor, feel free to use better)
2) Limit the producer to `ControlInfoProducerKafka`, bringing in the Kafka on constructor level.
Personally I think solution 1 is better.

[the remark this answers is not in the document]
yruslan: Yeah, the plugin itself has no dependency on Kafka. But the factory is dependent on Kafka. And I wanted to have the factory and the plugin to have the same name.

plugins-builtin/src/main/scala/za/co/absa/enceladus/plugins/builtin/kafka/ControlInfoAvroSerializer.scala
yruslan: These names come directly from Avro schema. There is a serializer unit test. If either the schema or parameter changes so that they are no longer in sync the unit test will break.

plugins-builtin/src/main/scala/za/co/absa/enceladus/plugins/builtin/kafka/ControlInfoAvroSerializer.scala
yruslan: I agree on the order of elements of the package tree. But I want to have `builtin` as the part of the same so that when someone uses a plugin in the configuration it would be obvious if that plugin is an internal (built-in) or external one. For instance, if `ClassNotFound` exception is thrown for an external plugin a JAR might be missing. If the plugin is a built-in one the issue is differet, maybe a bug in fat JAR creation.

plugins-builtin/src/main/scala/za/co/absa/enceladus/plugins/builtin/kafka/ControlInfoAvroSerializer.scala
GeorgiChochov: That's all right.

plugins-builtin/src/main/scala/za/co/absa/enceladus/plugins/builtin/kafka/ControlInfoAvroSerializer.scala
benedeki: Yeah, good point about the `builtin` 👍

README.md
```diff
@@ -232,6 +232,55 @@ The list of additional options available for running Conformance:
 | --catalyst-workaround **true/false**       | Turns on (`true`) or off (`false`) workaround for Catalyst optimizer issue. It is `true` by default. Turn this off only is you encounter timing freeze issues when running Conformance. | 
 | --autoclean-std-folder **true/false**      | If `true`, the standardized folder will be cleaned automatically after successful execution of a Conformance job. |
 
+## Plugins
+
+Standardization and Conformance support plugins that allow executing additional actions any time a checkpoint is created
+or job status changes. In order to write a plugin to Enceladus you need to implement the `ControlMetricsPlugin` and
```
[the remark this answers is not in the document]
yruslan: I was hoping when we add new type of plugin or at least a plugin facility, then we update the documentation and the readme. I didn't want to include readme and documentation of things that are not implemented yet.

spark-jobs/src/main/scala/za/co/absa/enceladus/common/plugin/menas/EventListenerMenas.scala
```diff
@@ -139,6 +139,10 @@ class EventListenerMenas(config: Config,
     }
   }
 
+  def close(): Unit = {
+    controlMetricPlugins.foreach(plugin => plugin.close())
```
benedeki: Wouldn't it be safer to call the `close` within a `try` block?

plugins-api/src/main/scala/za/co/absa/enceladus/plugins/api/Plugin.scala
```diff
@@ -18,4 +18,9 @@ package za.co.absa.enceladus.plugins.api
 /**
  * Base [marker] interface for all Enceladus plugins.
  */
-trait Plugin
+trait Plugin {
+  /**
+   * This method will be called when a plugin is no longer needed.
+   */
+  def close(): Unit
```
benedeki: If having a `close` method and this is a `trait` to be interface for outer development wouldn't an `init` method be also appropriate?

spark-jobs/src/main/scala/za/co/absa/enceladus/common/plugin/menas/EventListenerMenas.scala
yruslan: Agreed. We should be protective when running external code.

plugins-api/src/main/scala/za/co/absa/enceladus/plugins/api/Plugin.scala
yruslan: Initialization is expected to be done in the constructor. But if later `init()` is required for some reason, we can always add it later. We can have a default implementation of doing nothing to not break existing plugins.

benedeki: Yeah, I realized after posting - the factory is good enough. Sorry.

spark-jobs/src/main/scala/za/co/absa/enceladus/common/plugin/menas/EventListenerMenas.scala
benedeki: Actually if you will do it, will you do it for close only, or all outer calls?

yruslan: All other plugins (`factory.apply()` and `plugin.onCheckpoint()`) calls are already protected.

yruslan:
Thanks guys!

PR #1213, opened 2020-03-05
1211 Roll back Spring Boot version to 2.0.0 because of SPNEGO issue

GeorgiChochov:


pom.xml
```diff
@@ -151,8 +151,8 @@
         <scopt.version>3.7.0</scopt.version>
         <kafka.spark.version>0-10</kafka.spark.version>
         <abris.version>2.1.0</abris.version>
-        <spring.version>2.2.2.RELEASE</spring.version>
-        <spring.ldap.version>5.2.1.RELEASE</spring.ldap.version>
+        <spring.version>2.0.0.RELEASE</spring.version>
+        <spring.ldap.version>5.0.14.RELEASE</spring.ldap.version>
```
Zejnilovic: Did we test this on some environment? The previous version was `5.0.3.RELEASE`. I know it seems insignificant but since we don't know what was the cause of the problem in the first place, it might be worth to do so.

yruslan: A week ago I tried different versions of `spring.ldap` leaving the version of `spring` the same. And it turned out that the version of `spring.ldap` does not affect the issue in any way. I think switching to `5.0.14` is safe.

Zejnilovic: In that case I am happy

GeorgiChochov: I tested it with several versions, the issue appears from spring.boot version 2.0.1 onwards.

Issue #1209, opened 2020-03-04
Set report date and version in streaming jobs


yruslan:
## Background
Currently, when Conformance is run as a Hyperdrive component, `Report Date` and `Report Version` is taken from `ingestion.config`. But the configuration file is not intended to be changed daily to set the report date. There should be a different way to populate these fields.

'Report date' and 'report version' concepts are related to snapshot type of data. Together they specify when the snapshot was created (at the end of which day) and amendments to the snapshot. But for data that has an event stream nature (such as transactions for the day), these notions are replaced by 'event time' and 'processing time'. We need to understand do we really need to specify 'report date' and 'report version' for event type of data at all.

## Feature
We need to have a way of specifying report date and version so that the data that comes through the fast lane matches the data that comes through slow lane.

## Proposed Solution [Optional]
_To be discussed_
- Make report date and report version optional fields so they can be omitted for event data.
- The report date can be set as the ingestion date.
  _Can it happen that somebody wants to ingest data for a different report date via Hyperdrive_
- The report version can be set as the micro match number.
_Can't this lead to duplicating report versions when several ingestions happen during a single day?_
- Report version can be always set to 1 since streaming jobs do not assume overwrites anyway.

PR #1221, opened 2020-03-10
#1209 Make report date and version non mandatory fields.

yruslan:
We'll figure out if we need to calculate report date and version differently in the future.

### DIFF
--- a/spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/HyperConformance.scala
+++ b/spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/HyperConformance.scala
@@ -18,7 +18,9 @@ package za.co.absa.enceladus.conformance
 import org.apache.commons.configuration2.Configuration
 import org.apache.spark.sql.functions.{lit, to_date}
 import org.apache.spark.sql.{DataFrame, SparkSession}
+import org.joda.time.DateTime
 import org.slf4j.{Logger, LoggerFactory}
+import za.co.absa.enceladus.common.Constants
 import za.co.absa.enceladus.common.Constants._
 import za.co.absa.enceladus.conformance.interpreter.{Always, DynamicInterpreter, FeatureSwitches}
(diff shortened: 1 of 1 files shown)


spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/HyperConformance.scala
```diff
@@ -109,8 +113,8 @@ object HyperConformance extends StreamTransformerFactory {
     implicit val cmd: ConfCmdConfig = ConfCmdConfig(
       datasetName = conf.getString(datasetNameKey),
       datasetVersion = conf.getInt(datasetVersionKey),
-      reportDate = conf.getString(reportDateKey),
-      reportVersion = Option(conf.getInt(reportVersionKey)),
+      reportDate = getReportDate(conf),
+      reportVersion = Option(1),
```
Zejnilovic: shouldn't `getReportVersion`  be used here?

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/HyperConformance.scala
```diff
@@ -157,5 +161,21 @@ object HyperConformance extends StreamTransformerFactory {
     }
   }
 
+  private def getReportDate(conf: Configuration): String = {
+    if (conf.containsKey(reportDateKey)) {
+      conf.getString(reportDateKey)
+    } else {
+      DateTime.now().toLocalDate.toString(Constants.ReportDateFormat)
```
Zejnilovic: If you use just `ReportDateFormat` you don't have to import constants twice. Or switch line 53-56 to this way of using constants.

benedeki: I believe `org.joda.time` is considered deprecated
```suggestion
import java.text.SimpleDateFormat
import java.util.Date
            new SimpleDateFormat(ReportDateFormat).format(new Date())
```

PR #1223, opened 2020-03-13
Feature/1164 avro schema

dk1844:
Avro Schema support added.

- Unit test added, integration test adjusted to guard against regressions.
- UI updated.

Manually tested, seems to work.
1164-ui-avsc
1164-avsc-upload-schema successful
( ⬆️  ... result of uploading [info_file_avro_schema.avsc.txt](https://github.com/AbsaOSS/enceladus/files/4330595/info_file_avro_schema.avsc.txt))


menas/src/main/scala/za/co/absa/enceladus/menas/controllers/SchemaController.scala
```diff
@@ -36,18 +38,21 @@ import za.co.absa.enceladus.menas.utils.converters.SparkMenasSchemaConvertor
 import za.co.absa.enceladus.model.Schema
 import za.co.absa.enceladus.model.menas._
 
+import scala.util.control.NonFatal
 
+// TODO: use enumeration instead to type-force the schema types?
```
yruslan: Good idea! Enumeration or sealed case objects can improve things here.
https://pedrorijo.com/blog/scala-enums/

menas/pom.xml
```diff
@@ -65,6 +65,23 @@
...
+            <scope>compile</scope>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.slf4j</groupId>
+                    <artifactId>slf4j-log4j12</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>javax.validation</groupId>
+                    <artifactId>validation-api</artifactId>
+                </exclusion>
```
yruslan: Just a question. These artifacts caused binary incompatibility conflicts, right?

menas/src/test/scala/za/co/absa/enceladus/menas/controllers/SchemaControllerTest.scala
```diff
@@ -0,0 +1,81 @@
...
+import za.co.absa.enceladus.menas.controllers.SchemaController._
+import za.co.absa.enceladus.menas.models.rest.exceptions.SchemaParsingException
+import za.co.absa.enceladus.menas.utils.converters.SparkMenasSchemaConvertor
+
+import scala.io.Source
+
+class SchemaControllerTest extends WordSpec with Matchers with MockitoSugar with Inside {
+
+  val mockSchemaConvertor: SparkMenasSchemaConvertor = mock[SparkMenasSchemaConvertor]
+  val schemaController = new SchemaController(null, null, mockSchemaConvertor) // SUT
```
yruslan: Just a question - what's `SUT`?

yruslan:
Btw, the Jenkins build failure is nothing to do with your changes. We will sort this on Monday.

menas/src/test/scala/za/co/absa/enceladus/menas/controllers/SchemaControllerTest.scala
dk1844: System under test.

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/SchemaController.scala
```diff
@@ -115,18 +121,18 @@ class SchemaController @Autowired() (
...
-    * @param structTypeJson A StructType JSON string.
-    * @return The parsed schema as an instance of [[StructType]].
-    */
-  private def parseStructType(structTypeJson: String): StructType = {
+   * Parses an StructType JSON file contents and converts it to Spark [[StructType]].
+   *
+   * @param structTypeJson A StructType JSON string.
+   * @return The parsed schema as an instance of [[StructType]].
+   */
+  private[controllers] def parseStructType(structTypeJson: String): StructType = {
```
benedeki: I am not fond of opening `private` methods just for the sake of testing.
I would prefer testing or higher level. Or even using reflection to get the private method. We can discuss it on CQC (in chat).

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/SchemaController.scala
```diff
@@ -115,18 +121,18 @@ class SchemaController @Autowired() (
...
 
   /**
-    * Parses an StructType JSON file contents and converts it to Spark [[StructType]].
-    *
-    * @param structTypeJson A StructType JSON string.
-    * @return The parsed schema as an instance of [[StructType]].
-    */
-  private def parseStructType(structTypeJson: String): StructType = {
+   * Parses an StructType JSON file contents and converts it to Spark [[StructType]].
+   *
```
Zejnilovic: Avoid changing styles of comments and others. Otherwise, we can find ourselves in PRs full of mainly style changes.

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/SchemaController.scala
yruslan: Yeah, let's discuss this on CQC (meeting or channel). And if we decide this is an issue, we can fix this in the next PR.
My personal take on this - I'm fine with the way it is.

yruslan:
Thanks @Zejnilovic for raising your concerns. It made me think again about our support for Avro. Indeed, there are types that do not map directly to Spark StructType. But,
* We are getting the input streaming data as a DataFrame from Hyperdrive. So we don't need to do Standardization and do not need to apply the schema (and associated metadata).
* This imported schema will be used only for creating conformance rules in Menas (at least for now). 
HyperConformance just applies conformance rules and does not validate schemas (maybe that will be needed in the future - we will see).
* The conversion Daniel is using comes directly from Spark, so your questions are more about Spark-Avro compatibility. Since Hyperdrive uses the same API, we can expect that DataFrames provided by Hyperdrive to HyperConformance will be compatible enough to apply conformance rules. 

Overall, this is a new area for us, still PoC. Let's keep these issues in mind and figure out as we go.

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/SchemaController.scala
dk1844: It is tested on a higher level as well, e.g. at `SchemaApiIntegrationSuite.scala:722`

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/SchemaController.scala
dk1844: sorry, reverted. (I have autoformatted the file in IDEA thinking that this would be ok for everyone.)

menas/pom.xml
dk1844: Honestly, no. I have included the same exclusions that spark-core had (based on some wrong assumptions). After being adviced the `mvn install` part, I was now able to run `mvn dependency:tree` properly to find out that the exclusions do not seem to be needed after all. I am thus removing them.

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/SchemaController.scala
benedeki: Well then if also the failure tests would be moved to higher level, the lower ones wouldn't be needed. (Too many tests testing the same test cost upkeep efforts and build time.)
Also IMHO the opening is usually a code smell, in this case that `SchemaController` breaks the S from SOLID.

menas/src/test/scala/za/co/absa/enceladus/menas/controllers/SchemaControllerTest.scala
```diff
@@ -0,0 +1,81 @@
...
+import org.mockito.Mockito
+import org.scalatest.mockito.MockitoSugar
+import org.scalatest.{Inside, Matchers, WordSpec}
+import za.co.absa.enceladus.menas.controllers.SchemaController._
+import za.co.absa.enceladus.menas.models.rest.exceptions.SchemaParsingException
+import za.co.absa.enceladus.menas.utils.converters.SparkMenasSchemaConvertor
+
+import scala.io.Source
+
+class SchemaControllerTest extends WordSpec with Matchers with MockitoSugar with Inside {
```
benedeki: Usually the test classes use the suffix `Suite` not `Test` (I know there are exceptions). Let's keep a unified naming, looks more tidied 😉

menas/src/test/scala/za/co/absa/enceladus/menas/controllers/SchemaControllerTest.scala
```diff
@@ -0,0 +1,81 @@
...
+          schemaController.parseStructType("bad struct type def")
+        }
+        caughtException shouldBe SchemaParsingException(SchemaTypeStruct, someException.getMessage, cause = someException)
+      }
+    }
+  }
+
+  "parse avro schema to StructType" when {
+    "correct avsc file content is given" in {
+      val expectedJsonFormatSchema = Source.fromFile("src/test/resources/test_data/schemas/avro/equivalent-to-avroschema.json").mkString
```
benedeki: Small: IntelliJ warns me that `Source` is not closed (also line 62)

yruslan: This might work better since it is independent of the current directory:
```scala
val expectedJsonFormatSchema = IOUtils.toString(getClass.getResourceAsStream("/test_data/schemas/avro/equivalent-to-avroschema.json"), "UTF-8")
```

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/SchemaController.scala
yruslan: Ok.
How about moving `parse*` method to a separate object, say, `SchemaParsers`, make all these methods `public` and leave tests as is?

The higher-level test here is an integration test. The lower level ones are normal unit tests. Let's leave both, they serve their purpose.

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/SchemaController.scala
GeorgiChochov: I agree, good idea! Should improve our adherence to the open-closed principle (if we abstract the pattern-matching and use polymorphism to make a call to `SchemaParser::parse` instead of keeping the pattern matching here and callng concrete methods like `parseStructType` and `parseCopybook`)

menas/src/test/scala/za/co/absa/enceladus/menas/controllers/SchemaControllerTest.scala
dk1844: Thanks, renamed.

menas/src/test/scala/za/co/absa/enceladus/menas/controllers/SchemaControllerTest.scala
dk1844: Thanks, I have gradually used the `utils.fs.FileReader` (that actually performs the desired closing) and then replaced it with what Ruslan suggested :)

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/SchemaController.scala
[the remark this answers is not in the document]
dk1844: Makes sense, issue created: https://github.com/AbsaOSS/enceladus/issues/1228

dk1844: Related issue created: https://github.com/AbsaOSS/enceladus/issues/1228

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/SchemaController.scala
GeorgiChochov: I think @yruslan's suggestion to move them outside the Controller makes perfect sense. No need to open up the access modifiers, and lightens the load of what the Controller is meant to do - provide piping for services which hold business logic without holding any itself.

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/SchemaController.scala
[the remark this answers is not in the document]
benedeki: Please add a reference to the TODO comment or remove the TODO. We have an agreement not no add TODO to development branch without an accompanied issue.

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/SchemaController.scala
dk1844: > Ok.
> How about moving `parse*` method to a separate object, say, `SchemaParsers` ...

I was actually contemplating something like this before, but did not want to cause too _ado  about nothing_ 😄

Cool, I will try to introduce a separate unit-testable object for it.

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/SchemaController.scala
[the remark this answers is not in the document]
dk1844: Done. That is actually a very smart rule 👍  - I have definitely witnessed a good deal of orphaned TODOs in the past.

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/SchemaController.scala
dk1844: Done as suggested (more or less), hope you guys like it.

Issue #1228, opened 2020-03-16
SchemaTypes should be enum or sealed trait


dk1844:
The Schema types in the[ `SchemaController`](https://github.com/AbsaOSS/enceladus/blob/develop/menas/src/main/scala/za/co/absa/enceladus/menas/controllers/SchemaController.scala#L40) mimic enumeration style values. I suggest making them a proper Enumeration or equivalent (sealed trait) to correctly type-force their usage. And adjust the rest of the project.

```
object SchemaController {
  val SchemaTypeStruct = "struct"
  val SchemaTypeCopybook = "copybook"
  val SchemaTypeAvro = "avro"
}
```
This came up as part of PR https://github.com/AbsaOSS/enceladus/pull/1223.

Some specific ideas on the topic:
![pr-1223-enum-suggestion-additional-ideas](https://user-images.githubusercontent.com/4457378/76765734-2c596480-6797-11ea-9aae-2f6877f3d117.png)

dk1844:
Release notes
SchemaType is now an Enumeration.
SchemaParser is now a trait with a factory providing individual (hidden) parsers.

PR #1233, opened 2020-03-17
#1228 SchemaType is now an Enumeration.

dk1844:
SchemaType is now an Enumeration.

 - Tests and other logic adjusted to allow for this change.
 - Note, that `SchemaFormatException` and `SchemaFormatError` hold string-based `schemaType`, because they may need to address unparsable values, too.


menas/src/main/scala/za/co/absa/enceladus/menas/controllers/SchemaController.scala
```diff
@@ -62,16 +78,18 @@ class SchemaController @Autowired()(
...
     val (sparkStruct, schemaType) = origFormat match {
-      case Some(SchemaTypeStruct) | Some("") | None => (schemaParsers.parseStructType(fileContent), SchemaTypeStruct)
-      case Some(SchemaTypeCopybook) => (schemaParsers.parseCopybook(fileContent), SchemaTypeCopybook)
-      case Some(SchemaTypeAvro) => (schemaParsers.parseAvro(fileContent), SchemaTypeAvro)
-      case Some(schemaType) =>
-        throw SchemaFormatException(schemaType, s"'$schemaType' is not a recognized schema format. " +
-          s"Menas currently supports: '$SchemaTypeStruct' and '$SchemaTypeCopybook'.")
+      case Some("") | None => (schemaParsers.parseStructType(fileContent), SchemaType.Struct) // legacy compatibility cases
+      case Some(explicitSchemaName) =>
+        SchemaType.fromSchemaName(explicitSchemaName) match {
```
dk1844: The "double match" structure looks a bit weird, but the most obvious way of writing it with
```
case Some(Struct) => ...
case Some(Copybook) => ...
case Some(Avro) => ...
```
would fail on being exhaustive, which would need to introduce another default case/error handling -- that looked even worse.

dk1844:
@GeorgiChochov, I tried to take your and @yruslan's to move the schema-type decision logic out of the `SchemaController` and use a `SchemaParser::parse`  polymorphic calling of the actual parsing implementation. However, it was also suggested to consider "sealed trait+case objects" instead of using a plain scala Enumeration. I have tried both, but the former seems unnecessarily complicated to me (lacking the `withName`, `values` method/field and some extra effort of having to deal with case object serialization in jackson). So my choice would be to stay the enum. Do you have an opinion on this? Is there and sealed trait+case objects advantage I am not seeing?

To find the different implentations more easily, this should help:
https://github.com/AbsaOSS/enceladus/pull/1233/files/d65fdddcecc47bdd6acbc8a4202b6445ae5ed2e2 (sealed-trait variant)

https://github.com/AbsaOSS/enceladus/pull/1233/files (enum types; `SchemaParserFactory` logic is kept)

yruslan:
Despite all Scala Enumeration disadvantages, I still prefer Enumeration here, even if case object serialization issues are removed. Simply because Enumeration better communicates the intent. The lack of exhaustive pattern matching is a shame, but not critical in this particular use case.

GeorgiChochov:
> Despite all Scala Enumeration disadvantages, I still prefer Enumeration here, even if case object serialization issues are removed. Simply because Enumeration better communicates the intent. The lack of exhaustive pattern matching is a shame, but not critical in this particular use case.

Alright, it's up to you.

Issue #1222, opened 2020-03-13
Check Spark version compatibility at startup


yruslan:
## Background
Currently, Standardization and Conformance will fail with unobvious error messages if an incompatible version of Spark is used.

## Feature
We can check Spark version at startup and fail immediately with a proper error message if the version is not compatible with the job.

yruslan:
Forgot to add. The code for Spark version checking might be reused in other projects, such as Hyperdrive (@jozefbakus , @kevinwallimann).

dk1844:
Release notes
 - Spark version checking (currently >=2.4.2, <3.0.0) for Conformance and Standardization

PR #1237, opened 2020-03-23
#1222 spark version guard for SparkJobs

dk1844:
Version of provided spark (via `org.apache.spark.SPARK_VERSION`) is checked to be:
 - higher or equal than `minVersionInclusive`
 - lower than `maxVersionExclusive` (the compared version is "finalized" so that we do not even let e.g. `3.0.0-whatever` to run when `3.0.0` is the `maxVersionExclusive`)

Sematic `Version` from [AbsaOSS commons](https://github.com/AbsaOSS/commons#version-utils) is used for version comparison logic.


spark-jobs/src/main/scala/za/co/absa/enceladus/common/version/VersionExt.scala
```diff
@@ -0,0 +1,34 @@
...
+  implicit class RichVersion(version: Version) {
+    /**
+     * This strips the version to what the final version should look like - only keeping the leading numeric components, e.g.
+     * `1.2.3-rc.2` -> `1.2.3`
+     * @return
+     */
+    def finalVersion: Version = {
+      val leadingNumericComponents = version.components.takeWhile(_.isInstanceOf[NumericComponent])
+      Version(leadingNumericComponents: _*)
+    }
```
yruslan: This looks like something the commons could benefit having. What do you think, @wajda?

dk1844: Initially, I thought so, too, but this is only specific for the semver implementation that I am specifically using in the code. If we added this to `Version` that was initialized a simple, this would not work correctly, as 

`1.2.3-dev.finalVersion` would yield `1.2`, because `3-dev` is not a `NumericComponent`.
```scala
import za.co.absa.commons.version.Version
import za.co.absa.enceladus.common.version.VersionExt._

Version.asSimple("1.2.3-dev").finalVersion // Version(WrappedArray(NumericComponent(1), NumericComponent(2)))
```

So, I think the idea itself is valid, but making this actually correct would require more effort than it might seem at first glance.

wajda: Agree with Deniel. Although nothing stops us from adding a new `SemVer` object that would hold this method and maybe a few others, that are specific to the semantic version spec.

wajda: Created a proposal in Commons - AbsaOSS/commons#13
Let's discuss it there

wajda: Meantime, to avoid exposing internal version component types I'd recommend to simply take the first 3 items, regardless of their types. It would be equally correct from the SemVer spec perspective. And it's also shorter.
```scala
Version(version.components.take(3): _*)
```

dk1844: Agreed, I will update it as you suggest. The proposal makes sense, too.

dk1844: Adjusted as outlined. Hope you find it reasonable, @wajda.

spark-jobs/src/test/scala/za/co/absa/enceladus/common/version/SparkVersionGuardSuite.scala
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    assertThrows[AssertionError] {
+      SparkVersionGuard(semver"2.4.4", semver"3.0.0").checkSparkVersionCompatibility(semver"2.4.4-rc.7") // do not allow pre-min version
+    }
+  }
+
+  it should "fail on invalid version strings" in {
+    val caughtBogus1 = intercept[IllegalArgumentException] {
+      SparkVersionGuard(semver"2.4.4", semver"3.0.0").checkSparkVersionCompatibility("bogus")
```
Zejnilovic: just for a sanity check, I would have a positive check as well.

spark-jobs/src/test/scala/za/co/absa/enceladus/common/version/SparkVersionGuardSuite.scala
```diff
@@ -0,0 +1,63 @@
...
+import za.co.absa.commons.version.Version._
+
+class SparkVersionGuardSuite extends FlatSpec with Matchers {
+
+  "SparkVersionGuard" should "check basic version compatibility" in {
+    SparkVersionGuard(semver"2.4.4", semver"3.0.0").checkSparkVersionCompatibility(semver"2.4.5")
+    SparkVersionGuard(semver"2.4.4", semver"3.0.0").checkSparkVersionCompatibility(semver"2.4.4") // min is inclusive
+    SparkVersionGuard(semver"1.6.0", semver"3.0.0").checkSparkVersionCompatibility(semver"2.4.4")
+
+    assertThrows[AssertionError] {
```
Zejnilovic: Could you add a check for the message thrown as well?

spark-jobs/src/test/scala/za/co/absa/enceladus/common/version/VersionExtSuite.scala
```diff
@@ -0,0 +1,50 @@
...
+class VersionExtSuite extends FlatSpec with Matchers {
+
+  import VersionExt._
+
+  val expectedFinalVersion = Version.asSemVer("1.2.3")
+
+  "VersionExt" should "correctly derive final version for SemVers" in {
+    val toBeFinalized = Seq(
+      "1.2.3-alpha", "1.2.3-alpha.1", "1.2.3-0.3.7", "1.2.3-x.7.z.92",
+      "1.2.3-dev", "1.2.3-rc.1", "1.2.3-alpha.beta", "1.2.3-alpha.beta+222", "1.2.3"
```
Zejnilovic: `1.0.0+20130313144700` this does not pass the test but is one of the examples of SemVer [spec](https://semver.org/#spec-item-10)

I have a feeling this is due to commons implementation, but I would like a second opinion.

wajda: Filed AbsaOSS/commons#14
Thanks Sasa!

spark-jobs/src/main/scala/za/co/absa/enceladus/common/SparkCompatibility.scala
```diff
@@ -0,0 +1,31 @@
...
+ * This is where one can set which versions of Spark are supported. Shared between
+ * [[za.co.absa.enceladus.standardization.StandardizationJob]] and [[za.co.absa.enceladus.conformance.DynamicConformanceJob]]
+ *
+ * Expected to be used by [[version.SparkVersionGuard]]
+ */
+
+import za.co.absa.commons.version.Version
+import za.co.absa.commons.version.Version._
+
+object SparkCompatibility {
```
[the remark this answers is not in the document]
yruslan: Maybe, I'm not sure. Maybe we need to look at that at some point. Here are just some thoughts.
- If the job and the env have a mismatch between Scala 2.11 and Scala 2.12, the job is likely to fail before it has a chance to validate the version number.
- Hadoop compatibility doesn't seem very strict. What are our requirements for Hadoop?

spark-jobs/src/test/scala/za/co/absa/enceladus/common/version/SparkVersionGuardSuite.scala
yruslan: Maybe a `-SNAPSHOT` version of Spark is worth checking as well.

spark-jobs/src/main/scala/za/co/absa/enceladus/common/version/SparkVersionGuard.scala
```diff
@@ -0,0 +1,69 @@
...
+   * would be allowed when 3.0.0 is disallowed)
+   *
+   * @param yourVersion provided spark version
+   */
+  def checkSparkVersionCompatibility(yourVersion: Version): Unit = {
+    import VersionExt._
+
+    // `someVersion.finalVersion < maxExclusive` will guard against e.g. 3.0.0-rc.1 being allowed when 3.0.0 should not be
+    assert(yourVersion >= minVersionInclusive && yourVersion.finalVersion < maxVersionExclusive,
+      // todo use Version.asString for better error message!
```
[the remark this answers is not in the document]
dk1844: Yes, I would very much like to.

spark-jobs/src/test/scala/za/co/absa/enceladus/common/version/VersionExtSuite.scala
dk1844: Nice, thanks for catching this!

spark-jobs/src/test/scala/za/co/absa/enceladus/common/version/SparkVersionGuardSuite.scala
dk1844: Will do, thanks for the ideas.

spark-jobs/src/test/scala/za/co/absa/enceladus/common/version/SparkVersionGuardSuite.scala
dk1844: Done, I am checking that the part of the `SparkVersionGuard` message appears to make sure that the source of the exception is what is expected.

spark-jobs/src/main/scala/za/co/absa/enceladus/common/version/SparkVersionGuard.scala
```diff
@@ -0,0 +1,69 @@
...
+      s"This SparkJob can only be ran on Spark version [$minVersionInclusive, $maxVersionExclusive) (min inclusive, max exclusive). " +
+        s"Your detected version was $yourVersion.")
+  }
+
+}
+
+
+
+
```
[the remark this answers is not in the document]
dk1844: Removed.

spark-jobs/src/main/scala/za/co/absa/enceladus/common/version/SparkVersionGuard.scala
```diff
@@ -0,0 +1,69 @@
...
+    checkSparkVersionCompatibility(Version.asSemVer(yourVersion))
+
+  /**
+   * Supplied version will be checked against the [[SparkVersionGuard]]'s. Note, `yourVersion` is
+   * finalized when comparing to max in order to disallow non-final versions againt a final guard (3.0.0-rc.1
+   * would be allowed when 3.0.0 is disallowed)
+   *
+   * @param yourVersion provided spark version
+   */
+  def checkSparkVersionCompatibility(yourVersion: Version): Unit = {
```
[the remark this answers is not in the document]
dk1844: Done, thanks.

spark-jobs/src/main/scala/za/co/absa/enceladus/common/version/SparkVersionGuard.scala
```diff
@@ -0,0 +1,69 @@
...
+    import VersionExt._
+
+    // `someVersion.finalVersion < maxExclusive` will guard against e.g. 3.0.0-rc.1 being allowed when 3.0.0 should not be
+    assert(yourVersion >= minVersionInclusive && yourVersion.finalVersion < maxVersionExclusive,
+      // todo use Version.asString for better error message!
+      s"This SparkJob can only be ran on Spark version [$minVersionInclusive, $maxVersionExclusive) (min inclusive, max exclusive). " +
```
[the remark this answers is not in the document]
dk1844: Thanks for the tip, done.

spark-jobs/src/test/scala/za/co/absa/enceladus/common/version/VersionExtSuite.scala
dk1844: If https://github.com/AbsaOSS/commons/pull/16 is merged and released soon, I would upgrade the dependency and include the test for the BigInt-based SemVer, but it is not essential for the core functionality here, I'd say.

dk1844: New Version used.

dk1844: The new version of utils (Version) used, version string of `1.0.0+20130313144700` is part of the test case now.

spark-jobs/src/main/scala/za/co/absa/enceladus/common/SparkCompatibility.scala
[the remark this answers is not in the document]
dk1844: Maybe this could be another issue to be solved, I do not think I should widen the scope of the problem here this much.

spark-jobs/src/main/scala/za/co/absa/enceladus/common/version/SparkVersionGuard.scala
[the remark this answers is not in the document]
dk1844: Done.

spark-jobs/pom.xml
```diff
@@ -71,6 +72,13 @@
             <scope>provided</scope>
         </dependency>
 
+        <!-- Commons - for Version utilities -->
```
benedeki: Remove this comment. I am going to use commons as well (`TempFIle`) this comment would become misleading.

dk1844: Sure, done.

yruslan:
Looks good! Just need to re-run the build maybe.

spark-jobs/src/main/scala/za/co/absa/enceladus/common/version/SparkVersionGuard.scala
```diff
@@ -0,0 +1,67 @@
...
+  /**
+   * Populates the version guard with the defaults from [[za.co.absa.enceladus.common.SparkCompatibility]]
+   */
+  def fromDefaultSparkCompatibilitySettings: SparkVersionGuard =
+    SparkVersionGuard(SparkCompatibility.minSparkVersionIncluded, SparkCompatibility.maxSparkVersionExcluded)
+}
+
+/**
+ * Setup Spark version guard with allowed min & max sem-ver version. Note that:
+ *  - min version is included to be allowed and may be non-final
```
benedeki: Why not to add these text to the parameters themsleves?

benedeki: Also I admit, I don't understand very well the comment 😊

spark-jobs/src/main/scala/za/co/absa/enceladus/common/version/SparkVersionGuard.scala
```diff
@@ -0,0 +1,67 @@
...
+ *  - max version is excluded from allowed (supremum) and may be non-final
```
benedeki: Sumpremum?

spark-jobs/src/main/scala/za/co/absa/enceladus/common/version/SparkVersionGuard.scala
```diff
@@ -0,0 +1,67 @@
...
+   * String wrapper for [[SparkVersionGuard#checkSparkVersionCompatibility(SemanticVersion)]]
+   *
+   * @param yourVersion provided spark version
+   */
+  def ensureSparkVersionCompatibility(yourVersion: String): Unit =
+    ensureSparkVersionCompatibility(Version.asSemVer(yourVersion))
+
+  /**
+   * Supplied version will be checked against the [[SparkVersionGuard]]'s. Note, `yourVersion` is
+   * finalized when comparing to max in order to disallow non-final versions againt a final guard (3.0.0-rc.1
```
benedeki:
```suggestion
   * finalized when comparing to max in order to disallow non-final versions against a final guard (3.0.0-rc.1
```

spark-jobs/src/main/scala/za/co/absa/enceladus/common/version/SparkVersionGuard.scala
```diff
@@ -0,0 +1,67 @@
...
+   * would be allowed when 3.0.0 is disallowed)
+   *
+   * @param yourVersion provided spark version
+   */
+  def ensureSparkVersionCompatibility(yourVersion: SemanticVersion): Unit = {
+
+    // `someVersion.core < maxExclusive` guards against e.g. 3.0.0-rc.1 being allowed when 3.0.0 should not be
+    assert(yourVersion >= minVersionInclusive && yourVersion.core < maxVersionExclusive,
```
benedeki: Perhaps add `// scalastyle:off line.size.limit string clearer in one line` for this line. I think it's legit to be longer. (and `on` after)

spark-jobs/src/main/scala/za/co/absa/enceladus/common/SparkCompatibility.scala
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * Expected to be used by [[version.SparkVersionGuard]]
+ */
+
+import za.co.absa.commons.version.Version._
+import za.co.absa.commons.version.impl.SemVer20Impl.SemanticVersion
+
+object SparkCompatibility {
+  val minSparkVersionIncluded: SemanticVersion = semver"2.4.2"
+  val maxSparkVersionExcluded: SemanticVersion = semver"3.0.0"
```
benedeki: Why to limit to max version? Do we already know about some incompatibility?
With the max version the SW can age faster then necessary.

spark-jobs/src/main/scala/za/co/absa/enceladus/common/version/SparkVersionGuard.scala
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * Setup Spark version guard with allowed min & max sem-ver version. Note that:
+ *  - min version is included to be allowed and may be non-final
+ *  - max version is excluded from allowed (supremum) and may be non-final
+ *
+ * @param minVersionInclusive
+ * @param maxVersionExclusive
+ */
+case class SparkVersionGuard(minVersionInclusive: SemanticVersion, maxVersionExclusive: SemanticVersion) {
```
benedeki: With accordance to my above comment, I would make the `maxVersionExclusive` optional.

spark-jobs/src/main/scala/za/co/absa/enceladus/common/version/SparkVersionGuard.scala
dk1844: Well, the issue/feature description (further explained by @yruslan) was that we want to set bounds .e.g. 2.4.2-3.0.0, where the lower is in fact minimum (allowed), while the "max" is what is already excluded. So everything up to a "maximum" = [supremum](https://cs.wikipedia.org/wiki/Supremum).

spark-jobs/src/main/scala/za/co/absa/enceladus/common/version/SparkVersionGuard.scala
dk1844: Thanks for catching the typo.

spark-jobs/src/main/scala/za/co/absa/enceladus/common/version/SparkVersionGuard.scala
dk1844: Good suggestion, will do, thanks.

spark-jobs/src/main/scala/za/co/absa/enceladus/common/SparkCompatibility.scala
dk1844: I see your point, @yruslan's request was to limit at <3.0.0, which I did.
image

 @yruslan, should we have the option to disable the upper bound then?
I can certainly do it (as @benedeki suggests - make the "max" optional in the `SparkVersionGuard`, but pass defined "max" value for our spark jobs...?

spark-jobs/src/main/scala/za/co/absa/enceladus/common/version/SparkVersionGuard.scala
dk1844: Let's decide in https://github.com/AbsaOSS/enceladus/pull/1237#discussion_r398560517 then

spark-jobs/src/main/scala/za/co/absa/enceladus/common/version/SparkVersionGuard.scala
dk1844: - The `min` is true minimum (value is included).
 - May be non-final: the guard will gladly accept e.g. `1.2.3-rc.5` as minium and it should work - `1.2.3-rc.8` will pass while `1.2.3-rc.2` would not.

The wording of which part of the comment do you think should be improved?

spark-jobs/src/main/scala/za/co/absa/enceladus/common/version/SparkVersionGuard.scala
benedeki: Ok, I didn't know the word - not in Czech nor in English 😊

spark-jobs/src/main/scala/za/co/absa/enceladus/common/version/SparkVersionGuard.scala
dk1844: I will reword it a bit.

dk1844: True, will do.

dk1844: Done.

dk1844: Done, hope it makes more sense now.

benedeki: I can't make grammatically heads&tails of sentence. Particularly _version is included to be allowed_
I kind of figured what you meant, but by reverse engineering as I know how the function works.  So the code explained the comment instead of visa-versa. 😄

spark-jobs/src/main/scala/za/co/absa/enceladus/common/SparkCompatibility.scala
dk1844: As suggested, I have adjusted the `SparkVersionGuard` so that the upper bound is now optional. The spark-jobs currently still enforce `3.0.0` for the upper bound, but it would now be very easy to remove, should we decide to do so. 😊

benedeki: Chatted with @yruslan and concluded, that a good compromise would be too issue a warning in case the version is over max. Not to prevent run altogether.

yruslan: Hey guys, sorry for replying late. Will try to be more responsive.

dk1844: Done - there is a `log.warn` now for the case where spark version exceeds the version upper bound.

spark-jobs/src/main/scala/za/co/absa/enceladus/common/version/SparkVersionGuard.scala
```diff
@@ -0,0 +1,73 @@
...
+}
+
+/**
+ * Setup Spark version guard with allowed min & max sem-ver version.
+ *
+ * @param minVersionInclusive lowest acceptable spark version (may be non-final)
+ * @param maxVersionExclusive version supremum - first discouraged spark version (this and higher version usage issues a warning) if not None
+ */
+case class SparkVersionGuard(minVersionInclusive: SemanticVersion, maxVersionExclusive: Option[SemanticVersion])
+                            (implicit log:Logger = LoggerFactory.getLogger(SparkVersionGuard.getClass)) {
```
benedeki: I am just confused/unsure where is this logger coming from?

spark-jobs/src/test/scala/za/co/absa/enceladus/common/version/SparkVersionGuardSuite.scala
```diff
@@ -0,0 +1,98 @@
...
+    val caught = the[T] thrownBy (fun)
+    caught.getMessage should include(messageSubstringToAppear)
+  }
+
+   // Specific checks for the [[SparkVersionGuard]]
+  private def ensureFailsVersionTooLow =
+    ensureThrowsWithMessageIncluding[AssertionError]("Your Spark version is too low.") _
+
+  private def ensureLogsVersionTooHigh(fun: Logger => scala.Any): Unit = {
+    val testLogger: Logger = mock[Logger]
```
benedeki: Oh, I really like this. 👍

dk1844: Thanks.

spark-jobs/src/main/scala/za/co/absa/enceladus/common/version/SparkVersionGuard.scala
dk1844: I am not completely sure if I follow, but let me try:
if a Logger is given (explicitly or implicitly), it is used, otherwise, this default implementation is used. In the actual sparkjobs, the default impl is used; In test, both the default or explicit logger are used (as you commented below).

If you are wondering if this "parameter group" needs to be implicit, it does necessarily does not; it could have also been just an optional parameter -- currently, the implicit param passing is not used. But it can be if we wanted so, so I included it.

benedeki: Sorry, I was blind. I failed to see the default value for the parameter. 🙄

Issue #1163, opened 2020-01-30
Schema registry integration


yruslan:
## Background
Streaming sources encode data in `Avro` format. Avro schema definition can be provided as an Avro JSON or as a URL to the schema registry. In order to be able to define datasets for streaming sources in Menas, we need to be able to import Avro schemes from schema registries.

## Feature
Add support for importing schemas from a schema registry in Menas.

dk1844:
Release notes:
Avro schema can be loaded from URL (without auth), too. (Expecting URL from Schema registry)

PR #1243, opened 2020-03-31
#1163 Loading Schema from URL

dk1844:
1163-menas-load-from-ui

 - Tested against local HTTP server, e.g. `http://localhost:8000/avro_users-v1-struct.json`
 - also works against a REST client with the `x-www-form-urlencoded` content type, provided there is an appropriate `X-CSRF-TOKEN` present.
menas-remote-url-postman

 - integtest updated to cover the `/remote` load variant. WireMock HTTP server mocking is used there, including mocking proper mime types of the served files.

`schemaDetail.controller`:
- the UI also works, but jQuery ajax call is used (`handleRemoteUrlSubmit`) whose response is quite similar, but not too similar to the uploading variant (`handleRemoteLoadComplete` vs `handleUploadComplete` - not sure it pays off to generalize. Or is there a way to convert one response to the type of the other?)

 - also tested to work with `Avro Schema (avsc)` and URL `http://zapalnrapp1079.corp.dsarena.com:8081/subjects/ruslan.avro.users-value/versions/1/schema`

## Update:
The UI now detects Schema registry URL missing the `/schema` suffix:
1163-menas-schema-url-autocorrect
 - The reason is that the schema is also available on the `/version/$x` endpoint, but is only available in the `schema` field there as a JSON-in-a-JSON.

### DIFF
--- a/menas/pom.xml
+++ b/menas/pom.xml
@@ -272,12 +272,11 @@
             <version>2.26.3</version>
             <scope>test</scope>
         </dependency>
-        <!-- guava 27 for wire mock -->
+        <!-- guava 27 for wire mock test, but not just test scope, because guava is needed for spring, too -->
         <dependency>
             <groupId>com.google.guava</groupId>
             <artifactId>guava</artifactId>
             <version>27.0.1-jre</version>
-            <scope>test</scope>
         </dependency>
     </dependencies>
     <build>
--- a/menas/pom.xml
+++ b/menas/pom.xml
@@ -265,6 +265,20 @@
             <version>${spring.version}</version>
             <scope>test</scope>
         </dependency>
+        <!-- WireMock - to act as SchemaRegistry for IntegTest-->
+        <dependency>
+            <groupId>com.github.tomakehurst</groupId>
+            <artifactId>wiremock-jre8</artifactId>
+            <version>2.26.3</version>
+            <scope>test</scope>
+        </dependency>
+        <!-- guava 27 for wire mock -->
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+            <version>27.0.1-jre</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
     <build>
         <plugins>
--- a/menas/pom.xml
+++ b/menas/pom.xml
@@ -29,6 +29,8 @@
         <scala.xml.version>1.0.4</scala.xml.version>
         <webui.basedir>${project.basedir}/ui</webui.basedir>
         <scalastyle.configLocation>${project.parent.basedir}/scalastyle-config.xml</scalastyle.configLocation>
+        <guava.version>27.0.1-jre</guava.version>
+        <wiremock.version>2.26.3</wiremock.version>
     </properties>
 
     <dependencies>
@@ -258,6 +260,12 @@
             <artifactId>spark-cobol</artifactId>
             <version>${cobrix.version}</version>
         </dependency>
+        <!-- guava 27 for spring & wire mock testing, too -->
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+            <version>${guava.version}</version>
+        </dependency>
         <!-- Test -->
         <dependency>
             <groupId>org.springframework.boot</groupId>
@@ -269,14 +277,8 @@
         <dependency>
             <groupId>com.github.tomakehurst</groupId>
             <artifactId>wiremock-jre8</artifactId>
-            <version>2.26.3</version>
+            <version>${wiremock.version}</version>
             <scope>test</scope>
-        </dependency>
-        <!-- guava 27 for wire mock test, but not just test scope, because guava is needed for spring, too -->
-        <dependency>
-            <groupId>com.google.guava</groupId>
-            <artifactId>guava</artifactId>
-            <version>27.0.1-jre</version>
         </dependency>
     </dependencies>
     <build>
--- a/menas/src/main/scala/za/co/absa/enceladus/menas/controllers/SchemaController.scala
+++ b/menas/src/main/scala/za/co/absa/enceladus/menas/controllers/SchemaController.scala
@@ -38,7 +38,6 @@ import za.co.absa.enceladus.model.menas._
 
 import scala.io.Source
 
-
 @RestController
 @RequestMapping(Array("/api/schema"))
 class SchemaController @Autowired()(
@@ -48,7 +47,6 @@ class SchemaController @Autowired()(
   extends VersionedModelController(schemaService) {
 
   import za.co.absa.enceladus.menas.utils.implicits._
-
   import scala.concurrent.ExecutionContext.Implicits.global
 
   @PostMapping(Array("/remote"))
@@ -69,8 +67,7 @@ class SchemaController @Autowired()(
     val schemaType = SchemaType.fromOptSchemaName(format)
     val sparkStruct = SchemaParser.getFactory(sparkMenasConvertor).getParser(schemaType).parse(fileContent)
 
-    // TODO generalize with [[handleFileUpload]]
-    val origFile = MenasAttachment(refCollection = RefCollection.SCHEMA.name().toLowerCase,
+    val menasFile = MenasAttachment(refCollection = RefCollection.SCHEMA.name().toLowerCase,
       refName = name,
       refVersion = version + 1,
       attachmentType = MenasAttachment.ORIGINAL_SCHEMA_ATTACHMENT,
@@ -78,16 +75,7 @@ class SchemaController @Autowired()(
       fileContent = fileContent.getBytes,
       fileMIMEType = mimeType)
 
-    try {
-
-      for {
-        // the parsing of sparkStruct can fail, therefore we try to save it first before saving the attachment
-        update <- schemaService.schemaUpload(principal.getUsername, name, version, sparkStruct)
-        _ <- attachmentService.uploadAttachment(origFile)
-      } yield update
-    } catch {
-      case e: SchemaParsingException => throw e.copy(schemaType = schemaType) //adding schema type
-    }
+    uploadSchemaToMenas(principal.getUsername, menasFile, sparkStruct, schemaType)
   }
 
   @PostMapping(Array("/upload"))
@@ -103,7 +91,7 @@ class SchemaController @Autowired()(
     val schemaType = SchemaType.fromOptSchemaName(format)
     val sparkStruct = SchemaParser.getFactory(sparkMenasConvertor).getParser(schemaType).parse(fileContent)
 
-    val origFile = MenasAttachment(refCollection = RefCollection.SCHEMA.name().toLowerCase,
+    val menasFile = MenasAttachment(refCollection = RefCollection.SCHEMA.name().toLowerCase,
       refName = name,
       refVersion = version + 1,
       attachmentType = MenasAttachment.ORIGINAL_SCHEMA_ATTACHMENT,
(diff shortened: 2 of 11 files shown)


menas/src/main/scala/za/co/absa/enceladus/menas/controllers/SchemaController.scala
```diff
@@ -48,6 +51,35 @@ class SchemaController @Autowired()(
...
 
+  @PostMapping(Array("/remote"))
+  @ResponseStatus(HttpStatus.CREATED)
+  def handleRemoteFile(@AuthenticationPrincipal principal: UserDetails,
+                       @RequestParam remoteUrl: String,
+                       @RequestParam version: Int,
+                       @RequestParam name: String,
+                       @RequestParam format: Optional[String]): CompletableFuture[Option[Schema]] = {
+
+    val url = new URL(remoteUrl)
```
dk1844: If there was a reason for it, the URL retrieval could also be wrapped in some kind of catch-try block in case we are not happy with the default stack trace. Unsure if we wanted a specific Exception for this case or an existing one to be used here, I did not introduce such exception handling. Could be done if we wanted to.

Zejnilovic: There should be something, line 62 throws an error in case the host doesn't exist and line 63 if the host is something like www.google.com. Both propagate to the user as 500 internal server errors. Could we tell user something better? Or give them a status code that it is their fault?

menas/ui/components/schema/schemaDetail.controller.js
```diff
@@ -221,6 +295,12 @@ sap.ui.define([
         this._schemaRestDAO.getLatestVersionByName(currentSchema.name)
           .then(version => this._model.setProperty("/editingEnabled", currentSchema.version === version));
       }
+    },
+
+    isHttpUrl: function (url) {
+      if (!url) return false;
+      const pattern = new RegExp('^https?://[^ ]+$', 'i');
```
Zejnilovic, line 302: I see what you did. But what about making it more general? 
```js
var pattern = new RegExp('^(https?:\\/\\/)?'+ // protocol
    '((([a-z\\d]([a-z\\d-]*[a-z\\d])*)\\.)+[a-z]{2,}|'+ // domain name
    '(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)|'+ // OR IPv4 address
    '((?:[a-f0-9]{1,4}:){7}[a-f0-9]{1,4}))' + // OR IPv6 address
    '(\\:\\d+)?' + // port
    '(\\/[-a-z\\d%_.~+]*)*'+ // path
    '(\\?[;&a-z\\d%_.~+=-]*)?'+ // query string
    '(\\#[-a-z\\d_]*)?$','i'); // fragment locator
```

menas/ui/components/schema/schemaDetail.view.xml
```diff
@@ -56,6 +56,22 @@
                         </u:FileUploader>
                         <Button text="Upload File" press="handleUploadPress"/>
                     </IconTabFilter>
+                    <IconTabFilter id="LoadRemote" icon="sap-icon://internet-browser" text="Load from URL">
```
Zejnilovic, line 59: I have been thinking about the usability part here. I am not sure I like a new tab. Can we put it in `Upload New` and have both options there? Something like:
image
(excuse the quality)

menas/ui/components/schema/schemaDetail.controller.js
dk1844: I did find this, too. I even considered using some well-known URL regexes from https://mathiasbynens.be/demo/url-regex. However, they all seem to have one important downside for us, they only seem to allow public domains; things like http://localhost/something could cause problems. Which is why did not want to include something this advanced and they butcher it with dev extensions 😆 

So, sure, I could, just let's be aware of the limitations...

menas/src/test/scala/za/co/absa/enceladus/menas/integration/controllers/BaseRestApiTest.scala
```diff
@@ -102,6 +102,15 @@ abstract class BaseRestApiTest extends BaseRepositoryTest {
     upload(urlPath, headers, fileParamName, fileName, parameters)
   }
 
+  def sendPostRemoteFile[T](urlPath: String,
+                            parameters: Map[String, Any],
+                            headers: HttpHeaders = new HttpHeaders())
+                           (implicit ct: ClassTag[T]): ResponseEntity[T] = {
+    require(parameters.keySet.contains("remoteUrl"), s"parameters map must contain the 'remoteUrl' entry, but only $parameters was found")
```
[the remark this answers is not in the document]
dk1844: Thanks.

menas/pom.xml
```diff
@@ -265,6 +265,19 @@
             <version>${spring.version}</version>
             <scope>test</scope>
         </dependency>
+        <!-- WireMock - to act as SchemaRegistry for IntegTest-->
+        <dependency>
+            <groupId>com.github.tomakehurst</groupId>
+            <artifactId>wiremock-jre8</artifactId>
+            <version>2.26.3</version>
```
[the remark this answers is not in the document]
dk1844: Thanks, done.

menas/ui/components/schema/schemaDetail.view.xml
dk1844: This can certainly be done. In this case, do we want to duplicate the schemaType select?

Should we, then?

Zejnilovic: I would duplicate it for now. If anyone has a better UX proposition I am happy to listen.

menas/ui/components/schema/schemaDetail.controller.js
Zejnilovic: `localhost` is `127.0.0.1` which is allowed by my solution as well as `::1`
You can implement a different one as well. The website you provided is quite nice!

menas/ui/components/schema/schemaDetail.view.xml
yruslan: Maybe in this case, we can remove the schema type from the bottom part and just add a label saying "Get from Schema Registry" or something? And it will always imply Avro for the URL for now.

We can keep the support of other formats in the backend part for the future. So far we don't support it yet anyway (e.g. we don't have StructType and Copybook schema registries).

dk1844: OK, will change it as suggested.

menas/ui/components/schema/schemaDetail.controller.js
```diff
@@ -301,6 +331,18 @@ sap.ui.define([
       if (!url) return false;
       const pattern = new RegExp('^https?://[^ ]+$', 'i');
       return pattern.test(url);
+    },
+
+    // this is what the schema registry url should look like to be "fixable", e.g. http://zapalnrapp1079.corp.dsarena.com:8081/subjects/somename/versions/1
```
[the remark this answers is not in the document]
dk1844: Sorry, removed.

menas/ui/components/schema/schemaDetail.controller.js
yruslan: I like the pattern provided by Saša. But I'd suggest to do it in a separate task/PR and implement it in a separate reusable component, covered by unit tests etc, to make sure everything we want to pass validation actually passes.

In this PR the simple RegExp is good for now. It really helps if a user mistypes an URL and clicks 'Submit URL'. And if the URL is wrong the user will get the message from the backend in a popup message box.

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/SchemaController.scala
yruslan: Agree. If by creating our own exception wrapper the user will get a better error message so they know how to fix it, it is worth doing it.

menas/ui/components/schema/schemaDetail.controller.js
dk1844: OK, related issue created: https://github.com/AbsaOSS/enceladus/issues/1250

menas/ui/components/schema/schemaDetail.view.xml
dk1844: upload-remote-load-in-one
The only downside is that I had to set the width of the `HBox`es to `35em` because I don't know how to make the FileUploader stretchable (it seems that there is no support for that).

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/SchemaController.scala
dk1844: Done, now it returns code `400` and we determine by the error type on the front-end which error message to show to the user.

menas/src/test/scala/za/co/absa/enceladus/menas/package.scala
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+package za.co.absa.enceladus
+
+package object menas {
+
+  /**
+   * Resource paths of files for schema testing
+   */
+  object TestResourcePath {
```
yruslan, line 23: Putting constants in nested objects inside a package object lools like a practice of some kind. If it is a common practice, is there something to read about it? We might use this more - it looks like a good way to group constants and make a single place to store them.

menas/src/test/scala/za/co/absa/enceladus/menas/integration/controllers/SchemaApiIntegrationSuite.scala
```diff
@@ -33,7 +43,20 @@ import scala.collection.immutable.HashMap
 
 @RunWith(classOf[SpringRunner])
 @SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
-class SchemaApiIntegrationSuite extends BaseRestApiTest {
+class SchemaApiIntegrationSuite extends BaseRestApiTest with BeforeAndAfterAll {
+
+  private val port = 8877
+  private val wireMockServer = new WireMockServer(WireMockConfiguration.wireMockConfig().port(port))
```
yruslan, line 49: 👍  This is a really nice way for integration testing something that relies on queries to an external web server.

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/SchemaController.scala
```diff
@@ -45,9 +48,47 @@ class SchemaController @Autowired()(
...
+        val mimeType = connection.getContentType
+        val fileStream = Source.fromInputStream(connection.getInputStream)
+        val fileContent = fileStream.mkString
+        fileStream.close()
+
+        (url, fileContent, mimeType)
+
+      } catch {
+        case NonFatal(e) =>
+          throw RemoteSchemaRetrievalException(schemaType, s"$remoteUrl is currently unreachable to download a schema file from.", e)
```
Zejnilovic, line 75:
```suggestion
          throw RemoteSchemaRetrievalException(schemaType, s"Could not retrieve a schema file from $remoteUrl. Please check the correctness of the URL and a presence of the schema at the mentioned endpoint", e)
```

"Is unreachable" invokes in me "network issue and nothing else". What do you think about this proposition?

dk1844: Sure, it sounds better.

menas/src/test/scala/za/co/absa/enceladus/menas/package.scala
dk1844: I might have seen this already in some codebases, generally speaking, this is just an interpretation of
> Scala provides package objects as a convenient container shared across an entire package.

(src: https://docs.scala-lang.org/tour/package-objects.html).

But mostly, having seen these literal string to be repeated over and over, I just felt to need to _do something_. 😆

PR #1249, opened 2020-04-01
Fix Request timeouts returning non-error status codes with empty body

yruslan:


menas/src/main/resources/application.properties.template
```diff
@@ -17,6 +17,9 @@
 # unlimited set the property to -1.
 spring.servlet.multipart.max-file-size=16MB
 
+# Set the timeout a single connection request can wait for a controller to return a value (in milliseconds)
+spring.mvc.async.request-timeout=600000
```
Zejnilovic, line 21: A default of 10 minutes sounds like a lot. 1 minute (60000)?

Zejnilovic:
I think that the build is failing because the `hotfix/1.7.1` branch is created on top of `release v2.2.0` commit. Anyway, this cannot be used to release 1.7.1 since we would be releasing everything from 2 onwards as 1.X

menas/src/main/resources/application.properties.template
yruslan: Yeah, I suppose the reasonable default value is 1 minute.

yruslan:
Also, I think I should add the index creation to the hotfix as well. This is for anybody who would use 1.7.x version with a clean DB. Marked the PR as WIP.

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/RestExceptionHandler.scala
```diff
@@ -40,6 +41,11 @@ class RestExceptionHandler {
 
   private val logger = LoggerFactory.getLogger(this.getClass)
 
+  @ExceptionHandler(value = Array(classOf[AsyncRequestTimeoutException]))
+  def handleAsyncRequestTimeoutException(exception: AsyncRequestTimeoutException): ResponseEntity[Any] = {
+    ResponseEntity.status(HttpStatus.BANDWIDTH_LIMIT_EXCEEDED).build[Any]()
```
yruslan, line 46: Guys, what's your opinion on this status code? This was something I wasn't sure about.

Zejnilovic: Do we want to keep it as 509, server issue, we should have scaled better, or do we want to blame the people with 429, too many requests?

I am fine with both.

yruslan: The server code it is. AsyncTimeoutExpired is more server related. It is something unexpected rather than something we deliberately restrict.

dk1844: I would not blame an individual user (4xx), unless absolutely certain that they are at fault. 

Since the nature of the exception does not guarantee that (both that such a timeout may have generally been caused by the user only indirectly or not at all (another user's action or another "process" being run)), I would rather stick 5xx error code.

I'd personally use 500/503 (rather standardized with an explanatory message), but 509 with a message may not be any worse.

yruslan: Agree, 503 is much better fits the purpose.

We try not to use 500 explicitly so we know if 500 occurs, that means something went wrong, but are not sure what.

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/RestExceptionHandler.scala
```diff
@@ -40,6 +41,11 @@ class RestExceptionHandler {
 
   private val logger = LoggerFactory.getLogger(this.getClass)
 
+  @ExceptionHandler(value = Array(classOf[AsyncRequestTimeoutException]))
+  def handleAsyncRequestTimeoutException(exception: AsyncRequestTimeoutException): ResponseEntity[Any] = {
+    ResponseEntity.status(HttpStatus.SERVICE_UNAVAILABLE).build[Any]()
```
GeorgiChochov, line 46: Might be worth logging something here.

menas/src/main/scala/za/co/absa/enceladus/menas/models/rest/errors/RequestTimeoutExpiredError.scala
```diff
@@ -0,0 +1,23 @@
+/*
+ * Copyright 2018-2019 ABSA Group Limited
```
Zejnilovic, line 2: How did this pass the checks? It is failing my local build on scala style issue. @yruslan

benedeki: Ha maybe this is the reason my release build fails. 🤔 
I ran in twice during CQC, but obviously wasn't able to investigate deeper.

PR #1256, opened 2020-04-07
Release hotfix 1.7.1

yruslan:


menas/src/main/scala/za/co/absa/enceladus/menas/models/rest/errors/RequestTimeoutExpiredError.scala
```diff
@@ -0,0 +1,23 @@
...
+ */
+
+package za.co.absa.enceladus.menas.models.rest.errors
+
+import za.co.absa.enceladus.menas.models.rest.ResponseError
+
+/**
+ * This error is produced when a request timeout expires.
+ */
+case class RequestTimeoutExpiredError(errorType: String = "request_timeout_expired") extends ResponseError
```
benedeki, line 23: Why the default message here is lower cases with underscores and in `RestExceptionHandler.scala` line 46 actually proper sentence is used?

yruslan: It's `errorType`, so that clients (e.g. frontend) can do switch/case on it. A message is constructed separately as part of the response.

The idea behind this is making errors as strongly typed as possible so that clients can reuse the context. This emerged from schema parsing errors so that frontend can point at the exact position in a schema file where it is failed to parse.

yruslan:
Yay, finally the build has succeeded. 😄

yruslan:
I'm merging it since it is released already. If any new issue/suggestion is found in the code, we'll create a new task and PR.

menas/src/main/scala/za/co/absa/enceladus/menas/models/rest/errors/RequestTimeoutExpiredError.scala
benedeki: Eh, sorry. It's `errorType` not `message`, misread.

PR #1259, opened 2020-04-07
Refactoring/730 schemapathvalidator uses return

AdrianOlosutean:
SchemaPathValidator implementation without returns

### DIFF
--- a/utils/src/main/scala/za/co/absa/enceladus/utils/validation/SchemaPathValidator.scala
+++ b/utils/src/main/scala/za/co/absa/enceladus/utils/validation/SchemaPathValidator.scala
@@ -145,7 +145,6 @@ object SchemaPathValidator {
     if (path.isEmpty) {
       Nil
     } else {
-
       val currentField = path(0)
       val fullPath = s"$parentPath${path.mkString(".")}"
       if (parentOnly && path.length == 1) {
@@ -155,7 +154,6 @@ object SchemaPathValidator {
           Nil
         }
       } else {
-
         val matched = checkMatchType(schema, currentField)
         matched match {
           case ExactMatch(field) =>
@@ -176,13 +174,13 @@ object SchemaPathValidator {
     }
   }
 
-  private def handleParentMatch(schema: StructType, parentPath: String, currentField: String, fullPath: String) = {
+  private def handleParentMatch(schema: StructType, parentPath: String, currentField: String, fullPath: String): Seq[ValidationError] = {
     val matched = checkMatchType(schema, currentField)
     matched match {
       case ExactMatch(_) =>
         Seq(ValidationError(s"Column '$parentPath$currentField' already exists so it cannot be used as an output column '$fullPath'."))
       case CaseInsensitiveMatch(_) =>
-        Seq(ValidationError(s"Case insensitive variant of a cloumn '$parentPath$currentField' already exists so it cannot be used as an output column '$fullPath'."))
+        Seq(ValidationError(s"Case insensitive variant of a cloumn '$parentPath$currentField' already exists so it cannot be used as an output column '$fullPath'."))//scalastyle:ignore line.size.limit
       case _ => Nil
     }
   }
--- a/utils/src/main/scala/za/co/absa/enceladus/utils/validation/SchemaPathValidator.scala
+++ b/utils/src/main/scala/za/co/absa/enceladus/utils/validation/SchemaPathValidator.scala
@@ -149,18 +149,19 @@ object SchemaPathValidator {
       val currentField = path(0)
       val fullPath = s"$parentPath${path.mkString(".")}"
       if (parentOnly && path.length == 1) {
-        schema.fields match {
-          case f if fullPathNew && f.exists(field => field.name == currentField) =>
+        val matched = checkMatchType(schema, currentField)
+        matched match {
+          case ExactMatch(_) if fullPathNew =>
             Seq(ValidationError(s"Column '$parentPath$currentField' already exists so it cannot be used as an output column '$fullPath'."))
-          case f if fullPathNew && f.exists(field => field.name.compareToIgnoreCase(currentField) == 0) =>
+          case CaseInsensitiveMatch(_) if fullPathNew =>
             Seq(ValidationError(s"Case insensitive variant of a cloumn '$parentPath$currentField' already exists so it cannot be used as an output column '$fullPath'."))
           case _ => Nil
         }
       } else {
 
-        val exactMatch = schema.fields.find(field => field.name == currentField)
-        val failures = exactMatch match {
-          case Some(field) =>
+        val matched = checkMatchType(schema, currentField)
+        val failures = matched match {
+          case ExactMatch(field) =>
             val dataType = getUnderlyingType(field.dataType)
             dataType match {
               case st: StructType =>
@@ -169,22 +170,31 @@ object SchemaPathValidator {
                 Seq(ValidationError(s"Column '$parentPath$currentField' is a primitive type and can't contain child fields '$fullPath'."))
               case _ => Nil
             }
-
-          case None => caseInsensitiveErrors(schema, parentPath, currentField)
+          case CaseInsensitiveMatch(field) =>
+            Seq(ValidationError(s"Column name '$parentPath$currentField' does not case-sensitively match '$parentPath${field.name}'."))
+          case _ =>
+            Seq(ValidationError(s"Column name '$parentPath$currentField' does not exist."))
         }
         failures
       }
     }
-
   }
 
-  def caseInsensitiveErrors(schema: StructType, parentPath: String, currentField: String): Seq[ValidationError] = {
-    val caseInsensitiveMatch = schema.fields.find(field => field.name.compareToIgnoreCase(currentField) == 0)
-    caseInsensitiveMatch match {
-      case Some(field) =>
-        Seq(ValidationError(s"Column name '$parentPath$currentField' does not case-sensitively match '$parentPath${field.name}'."))
-      case None =>
-        Seq(ValidationError(s"Column name '$parentPath$currentField' does not exist."))
+  trait Match
+
+  case class ExactMatch(field: StructField) extends Match
+
+  case class CaseInsensitiveMatch(field: StructField) extends Match
+
+  case object NoMatch extends Match
+
+  private def checkMatchType(schema: StructType, currentField: String): Match = {
+
+    val maybeField = schema.fields.find(field => field.name.compareToIgnoreCase(currentField) == 0)
+    maybeField match {
+      case Some(field) if field.name == currentField => ExactMatch(field)
+      case Some(field) => CaseInsensitiveMatch(field)
+      case _ => NoMatch
     }
   }
 
--- a/utils/src/main/scala/za/co/absa/enceladus/utils/validation/SchemaPathValidator.scala
+++ b/utils/src/main/scala/za/co/absa/enceladus/utils/validation/SchemaPathValidator.scala
@@ -20,55 +20,55 @@ import org.apache.spark.sql.types._
 import scala.annotation.tailrec
 
 /**
- * Object responsible for validating paths to fields, it's existence and case sensitivity
- */
+  * Object responsible for validating paths to fields, it's existence and case sensitivity
+  */
 object SchemaPathValidator {
 
   /**
-   * Validate path existence
-   *
-   * @param schema    A Spark schema
-   * @param fieldPath A path to a field (e.g. "data.employees.employee.name")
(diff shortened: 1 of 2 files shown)


utils/src/main/scala/za/co/absa/enceladus/utils/validation/SchemaPathValidator.scala
```diff
@@ -141,54 +141,51 @@ object SchemaPathValidator {
...
-      return Nil
+      Nil
     }
+    else {
+
+      val currentField = path(0)
+      val fullPath = s"$parentPath${path.mkString(".")}"
+      if (parentOnly && path.length == 1) {
+        schema.fields match {
+          case f if fullPathNew && f.exists(field => field.name == currentField) =>
```
benedeki, line 153: I think it would be more effective to go through the fields just once and only in case of insensitive hit check witch message to send based on sensitive match or not.
(We have datasets with thousands of columns)

utils/src/main/scala/za/co/absa/enceladus/utils/validation/SchemaPathValidator.scala
```diff
@@ -141,54 +141,51 @@ object SchemaPathValidator {
...
         val exactMatch = schema.fields.find(field => field.name == currentField)
-        if (exactMatch.nonEmpty) {
-          return Seq(ValidationError(s"Column '$parentPath$currentField' already exists so it cannot be used as an output column '$fullPath'."))
-        }
-        val caseInsensitiveMatch = schema.fields.find(field => field.name.compareToIgnoreCase(currentField) == 0)
-        if (caseInsensitiveMatch.nonEmpty) {
-          return Seq(ValidationError(s"Case insensitive variant of a cloumn '$parentPath$currentField' already exists so it cannot be used as an output column '$fullPath'."))
+        val failures = exactMatch match {
+          case Some(field) =>
+            val dataType = getUnderlyingType(field.dataType)
```
benedeki, line 164: Why this case you left inline while the other you made a function?

utils/src/main/scala/za/co/absa/enceladus/utils/validation/SchemaPathValidator.scala
```diff
@@ -141,54 +141,51 @@ object SchemaPathValidator {
...
+        failures
       }
-      return Nil
     }
 
-    val exactMatch = schema.fields.find(field => field.name == currentField)
-    val failures = exactMatch match {
+  }
+
+  def caseInsensitiveErrors(schema: StructType, parentPath: String, currentField: String): Seq[ValidationError] = {
```
benedeki, line 181: Make `private`

utils/src/main/scala/za/co/absa/enceladus/utils/validation/SchemaPathValidator.scala
```diff
@@ -141,54 +141,61 @@ object SchemaPathValidator {
...
-          case None =>
+          case _ =>
             Seq(ValidationError(s"Column name '$parentPath$currentField' does not exist."))
         }
+        failures
+      }
+    }
+  }
+
+  trait Match
```
dk1844, line 183: sealed trait?

utils/src/main/scala/za/co/absa/enceladus/utils/validation/SchemaPathValidator.scala
```diff
@@ -141,54 +141,61 @@ object SchemaPathValidator {
...
+
+  case class ExactMatch(field: StructField) extends Match
+
+  case class CaseInsensitiveMatch(field: StructField) extends Match
+
+  case object NoMatch extends Match
+
+  private def checkMatchType(schema: StructType, currentField: String): Match = {
```
dk1844, line 191: I am wondering if the code got simpler if this method's return type was actually `Option[Match]` and there would be no `NoMatch` case object, just the two actual match types.

I have a hunch that the processing of such a result could then be done in 
```
matched.map{actualMatch => ... }.toSeq
```

utils/src/main/scala/za/co/absa/enceladus/utils/validation/SchemaPathValidator.scala
AdrianOlosutean: right, forgot about that one

utils/src/main/scala/za/co/absa/enceladus/utils/validation/SchemaPathValidator.scala
AdrianOlosutean: I actually thought of doing the same the first time, but I thought that nesting of the return type when it's possible not to would be better

utils/src/main/scala/za/co/absa/enceladus/utils/validation/SchemaPathValidator.scala
```diff
@@ -141,54 +141,51 @@ object SchemaPathValidator {
...
-    // scalastyle:off  return Covered in Issue #730
     if (path.isEmpty) {
-      return Nil
+      Nil
     }
+    else {
+
+      val currentField = path(0)
+      val fullPath = s"$parentPath${path.mkString(".")}"
+      if (parentOnly && path.length == 1) {
```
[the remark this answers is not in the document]
AdrianOlosutean: How would you suggest to extract?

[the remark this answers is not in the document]
Zejnilovic: Something like this. You don't have to stick to the structure or names.

```scala
private def validateSchemaPathArray(schema: StructType,
                                      path: Array[String],
                                      parentOnly: Boolean = false,
                                      fullPathNew: Boolean = false,
                                      parentPath: String = ""): Seq[ValidationIssue] = {
    if (path.isEmpty) {
      Nil
    } else {
      val currentField = path(0)
      val fullPath = s"$parentPath${path.mkString(".")}"
      if (parentOnly && path.length == 1) {
        validateParentOnlySchemaPath(schema, fullPathNew, parentPath, currentField, fullPath)
      } else {
        validateDeepSchemaPath(schema, path, parentOnly, fullPathNew, parentPath, currentField, fullPath)
      }
    }
  }

  private def validateDeepSchemaPath(schema: StructType,
                                     path: Array[String],
                                     parentOnly: Boolean,
                                     fullPathNew: Boolean,
                                     parentPath: String,
                                     currentField: String,
                                     fullPath: String): Seq[ValidationIssue] = {
    val exactMatch = schema.fields.find(field => field.name == currentField)
    val failures = exactMatch match {
      case Some(field) =>
        val dataType = getUnderlyingType(field.dataType)
        dataType match {
          case st: StructType =>
            validateSchemaPathArray(st, path.drop(1), parentOnly, fullPathNew, s"$parentPath${field.name}.")
          case _ if path.length > 1 =>
            Seq(ValidationError(s"Column '$parentPath$currentField' is a primitive type and can't contain child fields '$fullPath'."))
          case _ => Nil
        }

      case None => caseInsensitiveErrors(schema, parentPath, currentField)
    }
    failures
  }

  private def validateParentOnlySchemaPath(schema: StructType,
                                           fullPathNew: Boolean,
                                           parentPath: String,
                                           currentField: String,
                                           fullPath: String): Seq[ValidationError] = {
    schema.fields match {
      case f if fullPathNew && f.exists(field => field.name == currentField) =>
        Seq(ValidationError(s"Column '$parentPath$currentField' already exists so it cannot be used as an output column '$fullPath'."))
      case f if fullPathNew && f.exists(field => field.name.compareToIgnoreCase(currentField) == 0) =>
        Seq(ValidationError(s"Case insensitive variant of a cloumn '$parentPath$currentField' already exists so it cannot be used as an output column '$fullPath'."))
      case _ => Nil
    }
  }
```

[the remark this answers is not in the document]
AdrianOlosutean: Ok, sure I will do. I just thought that methods with so many parameters would not be good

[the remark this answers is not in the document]
Zejnilovic: You are absolutely right.

We also shouldn't be sending booleans as parameters. You can try and "renove" them.

utils/src/main/scala/za/co/absa/enceladus/utils/validation/SchemaPathValidator.scala
dk1844: (Futurama reference)

utils/src/main/scala/za/co/absa/enceladus/utils/validation/SchemaPathValidator.scala
```diff
@@ -136,59 +136,73 @@ object SchemaPathValidator {
...
+    }
+  }
+
+  private def handleParentMatch(schema: StructType, parentPath: String, currentField: String, fullPath: String) = {
+    val matched = checkMatchType(schema, currentField)
+    matched match {
+      case ExactMatch(_) =>
+        Seq(ValidationError(s"Column '$parentPath$currentField' already exists so it cannot be used as an output column '$fullPath'."))
+      case CaseInsensitiveMatch(_) =>
+        Seq(ValidationError(s"Case insensitive variant of a cloumn '$parentPath$currentField' already exists so it cannot be used as an output column '$fullPath'."))
```
benedeki, line 185: Add ` //scalastyle:ignore line.size.limit`. I think it's justified here.

utils/src/main/scala/za/co/absa/enceladus/utils/validation/SchemaPathValidator.scala
```diff
@@ -136,59 +136,73 @@ object SchemaPathValidator {
...
-    if (parentOnly && path.length == 1) {
-      if (fullPathNew) {
-        val exactMatch = schema.fields.find(field => field.name == currentField)
-        if (exactMatch.nonEmpty) {
-          return Seq(ValidationError(s"Column '$parentPath$currentField' already exists so it cannot be used as an output column '$fullPath'."))
-        }
-        val caseInsensitiveMatch = schema.fields.find(field => field.name.compareToIgnoreCase(currentField) == 0)
-        if (caseInsensitiveMatch.nonEmpty) {
-          return Seq(ValidationError(s"Case insensitive variant of a cloumn '$parentPath$currentField' already exists so it cannot be used as an output column '$fullPath'."))
+      val currentField = path(0)
```
benedeki, line 149: Small: I think the empty lines after else just prolong the function. But if you think it's improve readability keep them (also line 158),

utils/src/main/scala/za/co/absa/enceladus/utils/validation/SchemaPathValidator.scala
```diff
@@ -136,59 +136,71 @@ object SchemaPathValidator {
...
+        Seq(ValidationError(s"Column '$parentPath$currentField' already exists so it cannot be used as an output column '$fullPath'."))
+      case CaseInsensitiveMatch(_) =>
+        Seq(ValidationError(s"Case insensitive variant of a cloumn '$parentPath$currentField' already exists so it cannot be used as an output column '$fullPath'."))//scalastyle:ignore line.size.limit
+      case _ => Nil
+    }
+  }
+
+  private sealed trait Match
+
+  private case class ExactMatch(field: StructField) extends Match
```
dk1844, line 190: I find often convenient to put the trait implementation classes/object inside a companion object of that trait. It is not strictly necessary, of course, but it both [seems to be a common pattern](https://users.scala-lang.org/t/sealed-traits-and-their-implementations/5559) and making the companion object private would suffice here. Just a thought. 😉

utils/src/main/scala/za/co/absa/enceladus/utils/validation/SchemaPathValidator.scala
```diff
@@ -136,59 +136,71 @@ object SchemaPathValidator {
...
-        dataType match {
-          case st: StructType =>
-            validateSchemaPathArray(st, path.drop(1), parentOnly, fullPathNew, s"$parentPath${field.name}.")
-          case _ =>
-            if (path.length > 1) {
-              Seq(ValidationError(s"Column '$parentPath$currentField' is a primitive type and can't contain child fields '$fullPath'."))
-            } else {
-              Nil
+      } else {
+        val matched = checkMatchType(schema, currentField)
```
Zejnilovic, line 157: I would think about even extracting this huge else. If you decide to keep it, keep it. I think there are arguments for both.

AdrianOlosutean: I tried to think of a way to simplify and extract, but I could not find one without overcomplicating or extracting to a method with a lot of parameters

Issue #1157, opened 2020-01-16
Make the group join mapping rule non-experimental


yruslan:
## Background
The group join mapping rule implementation became mature enough to make it non-experimental.

## Feature
Make the group join implementation non-experimental. Reserve the notion of experimental to the broadcasting strategy.

yruslan:
After discussing with @GeorgiChochov decided to do this later. In `2.0` the configs will be changed to turn on experimental mapping rule by default. In `3.0` the old implementation will be turned off completely and the broadcasting strategy will be used as the experimental mapping rule.

yruslan:
The semantics of this issue has changed slightly.
Here are the steps of switching to the new implementation of the mapping rule:
1. Make the experimental rule the default option (this issue)
2. Turn on broadcasting strategy by default
3. Remove the old implementation of the mapping rule (in version `3.0`).

benedeki:
Release notes
The group explosion strategy for mapping conformance rules is turned on by default.

PR #1260, opened 2020-04-08
#1157 Use the new optimized implementation of mapping rule by default

yruslan:
In addition, make the size of mapping table to broadcast to 10 MB so that
if many mapping tables are used there won't be out of memory issues.


yruslan:
Release notes
The group explosion strategy for mapping conformance rules is turned on by default.

Issue #1236, opened 2020-03-19
Menas IntegTest needs a local mongodb running


dk1844:
## Background
The current implementation of the Menas integration tests requires that there is a local mongodb instance running and some of the test cases are run against this instance. If not found, the tests fail.

## Feature
If the test suite could provide a mongo instance or a mock, the tests could be completed without additional setup on the machine. Which seems highly desirable. Ideally, by using a testing framework that will spawn a mongodb and cleanup after tests are done

## Proposed Solution
Solution Ideas 
1. Embedded MongoDB - https://github.com/flapdoodle-oss/de.flapdoodle.embed.mongo

dk1844:
Release notes
Integration tests for Menas are now run against an in-place-spawn embedded Mongo instead of relying on a locally running instance.

PR #1263, opened 2020-04-08
#1236 embedded mongo for menas integtests

dk1844:
Menas integTests now runs against an embedded Mongo instance instead of having to rely on an actual MongoDB server present running (by default on port 27017).
 - Config value `menas.mongo.connection.string` is no longer used for the integTest; `menas.mongo.connection.database` is kept.
 - the integ tests that want to use embedded mongo instead of an actual one can do so using a spring profile `withEmbeddedMongo`. That way, conceptually, both types of tests can coexits (with embedded mongo or with actual one if we really wanted to).
 - A single instance of the (embedded mongo db) is spawned at a random free port - 
evidence that only one mongoDb is launched (previous implementations may have spawn multiple):
```
mvn clean && mvn -Pintegration test > output1.txt
$ cat ./output1.txt | grep "MongoDB starting"
[mongod output] 2020-04-08T10:12:17.052+0200 I CONTROL  [initandlisten] MongoDB starting : pid=85184 port=61629 dbpath=...(stripped)
$ cat ./output1.txt | grep "now exiting"
[mongod output] 2020-04-08T10:12:35.355+0200 I CONTROL  [conn4] now exiting
```

(Big thanks @wajda for all the tips and hints you gave me, I really appreciate it)


menas/src/test/resources/application.properties
```diff
@@ -56,7 +56,8 @@ menas.hadoop.auth.krb5.keytab=hdfs.keytab
 menas.auth.inmemory.user=user
 menas.auth.inmemory.password=chang< )eme
 
-menas.mongo.connection.string=mongodb://localhost:27017
+# embedded mongo for integTest will be launched instead on a random free port
+menas.mongo.connection.string=mongodb://not-used-for-testing:12345
```
dk1844, line 60: This cannot be currently removed, because the `defaultMongoDb` still requires it (e.g. for the theoretical case where would want to run a test without the embedded Mongo).

menas/src/test/scala/za/co/absa/enceladus/menas/integration/controllers/RunApiIntegrationSuite.scala
```diff
@@ -28,6 +29,7 @@ import za.co.absa.enceladus.model.{Run, SplineReference}
 
 @RunWith(classOf[SpringRunner])
 @SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
+@ActiveProfiles(Array("withEmbeddedMongo"))
```
dk1844, line 32: The profile decoration could have probably been used in the base class instead - with the same effect. This makes it more verbose.

menas/src/test/scala/za/co/absa/enceladus/menas/integration/repositories/BaseRepositoryTest.scala
```diff
@@ -29,7 +29,7 @@ import scala.concurrent.{Await, Future}
 
 abstract class BaseRepositoryTest extends WordSpec with TestContextManagement with BeforeAndAfter {
 
-  val awaitDuration: Duration = Duration(500, TimeUnit.MILLISECONDS)
+  val awaitDuration: Duration = Duration(800, TimeUnit.MILLISECONDS)
```
dk1844, line 32: The embedded mongo take a little extra time, so sometimes 500ms was not enough.

menas/src/test/resources/log4j.properties
```diff
@@ -22,3 +22,6 @@ log4j.appender.general.layout=org.apache.log4j.PatternLayout
 log4j.appender.general.layout.ConversionPattern=@log.pattern@
 
 log4j.logger.menas.services.RunService=@log.level.testcode@
+
+# (useful) uncomment to see spring beans wiring and related information:
+# log4j.logger.org.springframework.beans.factory.support.DefaultListableBeanFactory=DEBUG
```
Zejnilovic, line 27: Please add a new line here and in IntelliJ check that you have an editorconfig plugin and that it is enabled.

menas/src/test/scala/za/co/absa/enceladus/menas/integration/repositories/BaseRepositoryTest.scala
Zejnilovic: `[2020-04-08T10:03:56.432Z]   java.util.concurrent.TimeoutException: Futures timed out after [800 milliseconds]`

Might this be why the build is failing?

dk1844: Yes, apparently it seems to be machine-dependent.

Zejnilovic: I am talking with @hamc17 . It also took 23 minutes which is too long. Let's see if he finds anything.

menas/src/test/resources/log4j.properties
dk1844: Aha! I had both, but that was not all that was needed. The `.editorconfig` file was missing a specific setting for `.properties` file type. Added:
```
[*.properties]
insert_final_newline = true
```

menas/src/test/scala/za/co/absa/enceladus/menas/integration/repositories/BaseRepositoryTest.scala
dk1844: The next build on Jenkins took about 17 minutes, seems about normal to me. But sure, @hamc17's consultation couldn't hurt 👍

.editorconfig
```diff
@@ -11,6 +11,9 @@ indent_size = 4
 indent_style = space
 insert_final_newline = true
 
+[*.properties]
+insert_final_newline = true
```
yruslan, line 15: 👍

menas/src/test/resources/application.properties
yruslan: It's okay, since it is inside *test* resources.

menas/src/test/scala/za/co/absa/enceladus/menas/integration/repositories/BaseRepositoryTest.scala
dk1844: @Zejnilovic was there any thoughts? Should I contact @hamc17 directly on the topic?

Zejnilovic: No, it's fine.

Issue #1277, opened 2020-04-16
An exception is thrown at the end of the job


yruslan:
## Describe the bug

This exception is thrown at the end of a job if Kafka control metrics plugin is enabled. The job successfully finishes though.

```
20/04/16 09:47:46 INFO ControlInfoProducerKafka: Sending control measurements to dq.control.info...
20/04/16 09:47:46 ERROR ControlInfoProducerKafka: Error sending control info metrics to Kafka.
java.lang.IllegalStateException: Cannot perform operation after producer has been closed
	at org.apache.kafka.clients.producer.KafkaProducer.throwIfProducerClosed(KafkaProducer.java:853)
	at org.apache.kafka.clients.producer.KafkaProducer.doSend(KafkaProducer.java:862)
	at org.apache.kafka.clients.producer.KafkaProducer.send(KafkaProducer.java:846)
	at org.apache.kafka.clients.producer.KafkaProducer.send(KafkaProducer.java:733)
	at za.co.absa.enceladus.plugins.builtin.common.mq.kafka.ControlInfoProducerKafka.send(ControlInfoProducerKafka.scala:70)
	at za.co.absa.enceladus.plugins.builtin.controlinfo.mq.ControlInfoSenderPlugin.onCheckpoint(ControlInfoSenderPlugin.scala:41)
	at za.co.absa.enceladus.common.plugin.menas.EventListenerMenas$$anonfun$za$co$absa$enceladus$common$plugin$menas$EventListenerMenas$$notifyPlugins$1$$anonfun$apply$1.apply(EventListenerMenas.scala:177)
	at za.co.absa.enceladus.common.plugin.menas.EventListenerMenas$$anonfun$za$co$absa$enceladus$common$plugin$menas$EventListenerMenas$$notifyPlugins$1$$anonfun$apply$1.apply(EventListenerMenas.scala:175)
	at scala.collection.immutable.List.foreach(List.scala:392)
	at scala.collection.generic.TraversableForwarder$class.foreach(TraversableForwarder.scala:35)
	at scala.collection.mutable.ListBuffer.foreach(ListBuffer.scala:45)
	at za.co.absa.enceladus.common.plugin.menas.EventListenerMenas$$anonfun$za$co$absa$enceladus$common$plugin$menas$EventListenerMenas$$notifyPlugins$1.apply(EventListenerMenas.scala:175)
[...]

PR #1278, opened 2020-04-17
#1277 Use onApplicationEnd event to close Enceladus plugins

yruslan:
This fixes the attempt to send control info when the Kafka producer is already closed.


yruslan:
Thanks, guys!

Issue #1285, opened 2020-04-20
Make sure 'Source' and 'Raw' checkpoints are present when Standardization starts


yruslan:
## Background
We need to start more strict validations of incoming _INFO files.

## Feature
Make sure 'Source' and 'Raw' checkpoints are present when Standardization starts.

## Additional context
It might require changes to Atum to allow clients access to checkpoints (instance of `ControlMeasure`)

This is related to #1186

AdrianOlosutean:
Release notes:
Validate source and raw checkpoints are present in Standardization

PR #1288, opened 2020-04-20
#1285 Make sure 'Source' and 'Raw' checkpoints are present when Standardization starts

AdrianOlosutean:
Ensure raw and source checkpoints are present in Standardization


spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
```diff
@@ -424,53 +425,58 @@ object StandardizationJob {
...
+  /**
+   * Gets the number of records in source data by traversing Atum's checkpoints.
+   *
+   * @return The number of records in a checkpoint corresponding to source data (if available)
+   */
+  private def getSourceRecordCountFromCheckpoints: Either[String, Long] = {
+    getFieldRawRecordFromCheckpoints("source")
+  }
+
+  private def getFieldRawRecordFromCheckpoints(field: String) = {
```
yruslan, line 465: This method can encounter 3 types of error conditions. But It seems it distinguishes only 2 of them.
The 3 error conditions are:
- No checkpoint that has the name `name`
- There is a checkpoint `name`, but it doesn't have a `count` measurement
- There is a checkpoint `name`, it has a `count` measurement, but `count<0`

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
```diff
@@ -424,53 +425,58 @@ object StandardizationJob {
...
-    )
+    val optionResult = for {
+      checkpoint <- Atum.getControlMeasure
+        .checkpoints
+        .find(c => c.name.equalsIgnoreCase(field) || c.workflowName.equalsIgnoreCase(field))
+
+      measurement <- checkpoint.controls.find(m => m.controlType.equalsIgnoreCase(controlTypeRecordCount))
+      result <- Try({
+        val rowCount = measurement.controlValue.toString.toLong
+        if (rowCount >= 0) rowCount else throw new Exception("negative rowcount")
```
yruslan, line 475: Why not return `Left` here as well? `Option` might become unnecessary.

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
```diff
@@ -424,53 +425,58 @@ object StandardizationJob {
...
 
-    val rawRecordCount = checkpointRawRecordCount match {
-      case Some(num) => num
-      case None      => df.count
+    (checkpointRawRecordCount, checkpointSourceRecordCount) match {
+      case (Right(x), Right(y)) => {
+        Atum.setAdditionalInfo(s"raw_record_count" -> x.toString)
+        Atum.setAdditionalInfo(s"source_record_count" -> y.toString)
+      }
+      case (Right(_), Left(er)) => throw new ValidationException("Field errors", Seq(er))
```
yruslan, line 441:
```suggestion
      case (Right(_), Left(er)) => throw new ValidationException("Checkpoint validation failed:", Seq(er))
```

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
```diff
@@ -424,53 +425,58 @@ object StandardizationJob {
...
   }
 
   /**
     * Gets the number of records in raw data by traversing Atum's checkpoints.
     *
     * @return The number of records in a checkpoint corresponding to raw data (if available)
     */
-  private def getRawRecordCountFromCheckpoints: Option[Long] = {
+  private def getRawRecordCountFromCheckpoints: Either[String, Long] = {
+    getFieldRawRecordFromCheckpoints("row")
```
yruslan, line 453:
```suggestion
    getFieldRawRecordFromCheckpoints("raw")
```

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
```diff
@@ -424,53 +425,58 @@ object StandardizationJob {
...
+  }
+
+  /**
+   * Gets the number of records in source data by traversing Atum's checkpoints.
+   *
+   * @return The number of records in a checkpoint corresponding to source data (if available)
+   */
+  private def getSourceRecordCountFromCheckpoints: Either[String, Long] = {
+    getFieldRawRecordFromCheckpoints("source")
```
yruslan, line 462: It could be useful to extract "Source" and "Raw" as constants.

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
```diff
@@ -424,53 +425,58 @@ object StandardizationJob {
...
-        } else {
-          None
-        }
-      }
-      catch {
-        case NonFatal(_) => None
-      }
-    )
+    val optionResult = for {
+      checkpoint <- Atum.getControlMeasure
```
yruslan, line 468: I know this is not part of your code, but it can be made better.
If `getSourceRecordCountFromCheckpoints` gets a `ControlMeasure` instance, it can be unit tested for various cases. `Atum.getControlMeasure` can be called outsode the method.

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
AdrianOlosutean: I think I did that because I could not use Either in the for comprehension, but I may be wrong

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
AdrianOlosutean: Makes sense

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
```diff
@@ -424,53 +425,58 @@ object StandardizationJob {
...
+        .find(c => c.name.equalsIgnoreCase(field) || c.workflowName.equalsIgnoreCase(field))
+
+      measurement <- checkpoint.controls.find(m => m.controlType.equalsIgnoreCase(controlTypeRecordCount))
+      result <- Try({
+        val rowCount = measurement.controlValue.toString.toLong
+        if (rowCount >= 0) rowCount else throw new Exception("negative rowcount")
+      }).toOption
+    } yield result
+
+    if (optionResult.isDefined) Right(optionResult.get) else Left(s"Invalid field ${field}")
```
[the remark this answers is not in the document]
AdrianOlosutean: Curiously this was the shortest and accessible to use

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
AdrianOlosutean: For this kind of use case, the Validated class would be quite a good fit

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
yruslan: Yep, you a correct. I've missed that, sorry.

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
```diff
@@ -424,53 +429,75 @@ object StandardizationJob {
...
-    * If raw record count is not available in checkpoints the method will calculate that count
-    * based on the provided raw dataframe.
+    * Adds metadata about the number of records in raw and source data by checking Atum's checkpoints first.
     *
     * @return The number of records in a checkpoint corresponding to raw data (if available)
     */
-  private def addRawRecordCountToMetadata(df: DataFrame): Unit = {
-    val checkpointRawRecordCount = getRawRecordCountFromCheckpoints
+  def addRawAndSourceRecordCountsToMetadata: Unit = {
+    val checkpoints: immutable.Seq[Checkpoint] = Atum.getControlMeasure.checkpoints
```
benedeki, line 437: Why `immutable.`? (I believe it's the default)

spark-jobs/src/test/scala/za/co/absa/enceladus/standardization/StandardizationValidationSuite.scala
```diff
@@ -0,0 +1,63 @@
+package za.co.absa.enceladus.standardization
+
+import org.scalatest.FunSuite
+import za.co.absa.atum.model.{Checkpoint, Measurement}
+
+class StandardizationValidationSuite extends FunSuite {
```
benedeki, line 6: Not sure I like this unit test is not clear what it bounds to. 
For discussion: what about extracting the atum related logic to a dedicated `object` (or `class`), make it easier to test. Also `StandardizationJob` is becoming little fatty of a mixture of unrelated code.

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
```diff
@@ -102,7 +107,7 @@ object StandardizationJob {
     Atum.setAdditionalInfo(Constants.InfoVersionColumn -> reportVersion.toString)
 
     // Add the raw format of the input file(s) to Atum's metadta as well
-    Atum.setAdditionalInfo("raw_format" -> cmd.rawFormat)
+    Atum.setAdditionalInfo(rawFieldName + "_format" -> cmd.rawFormat)
```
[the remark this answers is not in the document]
yruslan: It's odd, but for a different reason. We don't have a notion of 'source_format'. The row format is the format the data landed to Hadoop (e.g. CSV, XML, JSON). I'd leave the field name hardcoded.

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
```diff
@@ -424,53 +429,75 @@ object StandardizationJob {
...
+  /**
+   * Gets the number of records in source data by traversing Atum's checkpoints.
+   *
+   * @return The number of records in a checkpoint corresponding to source data (if available)
+   */
+  def getSourceRecordCountFromCheckpoints(checkpoints: immutable.Seq[Checkpoint]): Either[String, Long] = {
+    getFieldRawRecordFromCheckpoints(sourceFieldName, checkpoints)
+  }
+
+  private def getFieldRawRecordFromCheckpoints(checkpointName: String, checkpoints: immutable.Seq[Checkpoint]) = {
```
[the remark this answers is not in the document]
yruslan: I think it is to collect all errors in `Left` and then throw an exception specifying the list of all errors. @AdrianOlosutean to confirm.

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
```diff
@@ -424,53 +429,75 @@ object StandardizationJob {
...
+
+    measurement match {
+      case Right(m) =>
+        val trial = Try({
+          val rowCount = m.controlValue.toString.toLong
+          if (rowCount >= 0) rowCount else throw new Exception(s"Negative value")
+        })
+        trial match {
+        case Success(value) => Right(value)
+        case Failure(ex) => Left(s"Wrong $checkpointName $controlTypeRecordCount value: ${ex.getMessage}")
```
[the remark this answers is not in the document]
yruslan: We need `Long` to sanity check for negative values. Maybe we can workaround bogus values leading to the throw of `.toLong` by extracting a `safeToLong()` method that throws a more specific error? What do you think, guys?

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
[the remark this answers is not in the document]
AdrianOlosutean: It was extracted by the IDE when I extracted the constant, I did not have a particular reason for that

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
```diff
@@ -424,53 +429,75 @@ object StandardizationJob {
...
-    Atum.setAdditionalInfo(s"raw_record_count" -> rawRecordCount.toString)
   }
 
   /**
     * Gets the number of records in raw data by traversing Atum's checkpoints.
     *
     * @return The number of records in a checkpoint corresponding to raw data (if available)
     */
-  private def getRawRecordCountFromCheckpoints: Option[Long] = {
+  def getRawRecordCountFromCheckpoints(checkpoints: immutable.Seq[Checkpoint]): Either[String, Long] = {
```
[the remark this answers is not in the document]
AdrianOlosutean: True, I'll remove them

spark-jobs/src/test/scala/za/co/absa/enceladus/standardization/StandardizationValidationSuite.scala
yruslan: I like this test suite and I like the idea of putting Standardization prerequisites validation into a separate object. Since
* It might be used in Conformance as well, for instance for jobs that do not require Standardization.
* It is expected that we will add more validations in the near future

When validations are extracted into a separate class or object, this test suite will fit perfectly.

@AdrianOlosutean , sorry that these ideas were not described in the GitHub issue. My bad, I didn't think it through properly.

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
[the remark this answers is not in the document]
AdrianOlosutean: Yes, you want to put those errors in the ValidationException further on, but the method will have just one error at a time. @benedeki , do you mean to have a for comprehension with optionals?

spark-jobs/src/test/scala/za/co/absa/enceladus/standardization/StandardizationValidationSuite.scala
AdrianOlosutean: Ok, I will extract it into a separate object

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
```diff
@@ -424,53 +429,75 @@ object StandardizationJob {
...
+    val checkpointRawRecordCount = getRawRecordCountFromCheckpoints(checkpoints)
+    val checkpointSourceRecordCount = getSourceRecordCountFromCheckpoints(checkpoints)
+
+    val errorMessage = "Checkpoint validation failed:"
 
-    val rawRecordCount = checkpointRawRecordCount match {
-      case Some(num) => num
-      case None      => df.count
+    (checkpointRawRecordCount, checkpointSourceRecordCount) match {
+      case (Right(x), Right(y)) => {
```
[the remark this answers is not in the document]
AdrianOlosutean: Actually here it's needed cause the code inside is imperative. I should restructure it

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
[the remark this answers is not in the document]
AdrianOlosutean: Where would this method be located?

spark-jobs/src/test/scala/za/co/absa/enceladus/standardization/StandardizationValidationSuite.scala
yruslan: `ControlInfoValidation` maybe?

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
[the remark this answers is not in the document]
yruslan: It's not implemented, but we can put it into `Utils`.

spark-jobs/src/test/scala/za/co/absa/enceladus/standardization/StandardizationValidationSuite.scala
benedeki: For the record, I don't have any issue with the suite itself. 😄 
It's just it's not clear what it unit tests. (e.g. the `Shift`+`Ctrl`+`T` key combination doesn't tap here

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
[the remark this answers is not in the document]
benedeki: Not sure why this make them needed and what does it change? 🤔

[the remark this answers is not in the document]
AdrianOlosutean: The Atum.setAdditionalInfo lines are side effects

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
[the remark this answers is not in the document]
AdrianOlosutean: Indeed, something like that I was thinking to do, but I thought I could not use Try properly in the for comprehension. Actually I can

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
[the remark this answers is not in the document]
benedeki: Yup. But the Atum functions are of same return type as this one here (`Unit`). And if not I don't think the curly braces would fix it. 😉 
So I don't think they are needed.

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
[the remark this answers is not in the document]
AdrianOlosutean: I wouldn't say it's criptic. As you can see in the testing suite, for an empty string value there it would something like : Wrong raw $controlTypeRecordCount value: For input string ""

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/ControlInfoValidation.scala
```diff
@@ -0,0 +1,78 @@
...
+      case (Success(x), Success(y)) => FieldCounts(x, y)
+      case (Success(_), Failure(er)) => throw new ValidationException(errorMessage, Seq(er.getMessage))
+      case (Failure(er), Success(_)) => throw new ValidationException(errorMessage, Seq(er.getMessage))
+      case (Failure(er1), Failure(er2)) =>
+        throw new ValidationException(errorMessage, Seq(er1.getMessage, er2.getMessage))
+    }
+  }
+
+
+  def getFieldRecordFromCheckpoints(checkpointName: String, checkpoints: Seq[Checkpoint]): Try[Long] = {
```
benedeki, line 54:
```suggestion
  def getFieldCountFromCheckpoint(checkpointName: String, checkpoints: Seq[Checkpoint]): Try[Long] = {
```
The function is public and gets a particular field from the checkpoint, but the name is very general. This is just a suggestion tho. There might easily be better names. 😉

Issue #1293, opened 2020-04-21
Add Spark version check to HyperConformance


yruslan:
## Background
If the version of Spark HyperConformanc is running is less than `2.4.3`, the following exception can be thrown if a conformance rule operates on an array:
```
za.co.absa.hyperdrive.shared.exceptions.IngestionStartException: NOT STARTED ingestion dbcff102-2e4c-49d4-8ad6-14b0165b5f2e. This exception was thrown during the starting of the ingestion job. Check the logs for details.
	at za.co.absa.hyperdrive.driver.SparkIngestor.ingest(SparkIngestor.scala:83)
	at za.co.absa.hyperdrive.driver.IngestionDriver.ingest(IngestionDriver.scala:49)
	at za.co.absa.hyperdrive.driver.drivers.CommandLineIngestionDriver$.main(CommandLineIngestionDriver.scala:38)
	at za.co.absa.hyperdrive.driver.drivers.CommandLineIngestionDriver.main(CommandLineIngestionDriver.scala)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.apache.spark.deploy.yarn.ApplicationMaster$$anon$2.run(ApplicationMaster.scala:678)
Caused by: org.apache.spark.sql.catalyst.analysis.UnresolvedException: Invalid call to dataType on unresolved object, tree: lambda 'v1[InnerTestField]
	at org.apache.spark.sql.catalyst.analysis.UnresolvedExtractValue.dataType(unresolved.scala:415)
	at org.apache.spark.sql.catalyst.expressions.Cast.checkInputDataTypes(Cast.scala:210)
	at org.apache.spark.sql.Dataset.org$apache$spark$sql$Dataset$$withPlan(Dataset.scala:3407)
	at org.apache.spark.sql.Dataset.select(Dataset.scala:1335)
	at za.co.absa.spark.hats.transformations.NestedArrayTransformations$.nestedWithColumnMapHelper(NestedArrayTransformations.scala:686)
	at za.co.absa.spark.hats.transformations.NestedArrayTransformations$.nestedWithColumnMap(NestedArrayTransformations.scala:55)
[...]

PR #1297, opened 2020-04-22
#1293 Add Spark version check to HyperConformance

AdrianOlosutean:
Enforce minimum 2.4.3 Spark version for HyperConformance

### DIFF
--- a/spark-jobs/src/main/scala/za/co/absa/enceladus/common/SparkCompatibility.scala
+++ b/spark-jobs/src/main/scala/za/co/absa/enceladus/common/SparkCompatibility.scala
@@ -27,5 +27,6 @@ import za.co.absa.commons.version.impl.SemVer20Impl.SemanticVersion
 
 object SparkCompatibility {
   val minSparkVersionIncluded: SemanticVersion = semver"2.4.2"
+  val minSparkHyperConformanceIncluded: SemanticVersion = semver"2.4.3"
   val maxSparkVersionExcluded: Option[SemanticVersion] = Some(semver"3.0.0")
 }
--- a/spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/HyperConformance.scala
+++ b/spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/HyperConformance.scala
@@ -24,15 +24,12 @@ import org.apache.spark.sql.functions.{lit, to_date}
 import org.apache.spark.sql.{DataFrame, SparkSession}
 import org.slf4j.{Logger, LoggerFactory}
 import za.co.absa.enceladus.common.Constants._
-import za.co.absa.enceladus.common.SparkCompatibility
 import za.co.absa.enceladus.common.version.SparkVersionGuard
 import za.co.absa.enceladus.conformance.interpreter.{Always, DynamicInterpreter, FeatureSwitches}
 import za.co.absa.enceladus.dao.MenasDAO
 import za.co.absa.enceladus.dao.auth.{MenasCredentialsFactory, MenasKerberosCredentialsFactory, MenasPlainCredentialsFactory}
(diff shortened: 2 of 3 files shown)


spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/HyperConformance.scala
```diff
@@ -93,11 +97,15 @@ object HyperConformance extends StreamTransformerFactory with HyperConformanceAt
   val log: Logger = LoggerFactory.getLogger(this.getClass)
 
   private val defaultReportVersion = 1
+  private val minimumSparkVersion: SemanticVersion = semver"2.4.3"
 
   @throws[IllegalArgumentException]
   override def apply(conf: Configuration): StreamTransformer = {
     log.info("Building HyperConformance")
 
+    SparkVersionGuard(minimumSparkVersion, SparkCompatibility.maxSparkVersionExcluded)
```
benedeki, line 106: Do we expect the Spark compatibility to differ between core DCE and HyperConformance?
If not, why not to use `fromDefaultSparkCompatibilitySettings`?

Zejnilovic: Well the versions already differ. But we could have all versions requirements centralized somewhere. Thoughts?

benedeki: Ah, indeed. 2.4.2 vs. 2.4.3. Then let's have it centralized indeed. (`object SparkCompatibility`)

AdrianOlosutean: So the minimum should be changed to 2.4.3?

Zejnilovic: Just to make myself clearer. Centralized, as in one place, yes but we could keep multiple versions there. Like `minSparkVersionGeneral`, `minSparkVersionHyperConformance`, etc.

Question is, do we want to and how hard would it be to maintain?

AdrianOlosutean: This would also mean adding a new entry to SparkVersionGuard

dk1844: I think more specifically, it would involve adding new field(s) to `SparkCompatibility`.

`SparkVersionGuard` case class and logic can stay the same, only calling it with different params should suffice. Either directly or via a different wrapper, e.g.
```scala
SparkVersionGuard.fromHyperConformaceSparkCompatibilitySetting
```

spark-jobs/src/main/scala/za/co/absa/enceladus/common/SparkCompatibility.scala
```diff
@@ -27,5 +27,6 @@ import za.co.absa.commons.version.impl.SemVer20Impl.SemanticVersion
 
 object SparkCompatibility {
   val minSparkVersionIncluded: SemanticVersion = semver"2.4.2"
+  val minSparkHyperConformanceIncluded: SemanticVersion = semver"2.4.3"
```
yruslan, line 30: Sorry guys, a post-merge comment.
The requirement for HyperConformance is the same as normal Conformance.
I'd say, 2.4.2 for both. Adrian was asking me in the chat, I haven't guessed the context.
We can fix this some other time.
Sorry.

Zejnilovic: Should be fine, at least we have precedence for the next time.

Issue #1280, opened 2020-04-17
DAO module is a fat jar


yruslan:
## Describe the bug
DAO is a library module and shouldn't be a fat jar. Presumably, this is the case since there is a shaded artifact.

This might be the reason HyperConformance fails to deserialize a JSON on some builds (not confirmed).

## Expected behaviour
`dao` should be a thin jar that doesn't include transitive dependencies. If an artifact needs to be shaded, implement the shading in the far jar module (e.g. `spark-jobs`)

## Solution
Remove `shade` plugin from `dao`. Do the shading in `spark-jobs`.

yruslan:
Release notes
The DAO module was made a thin JAR so that spark-jobs don't depend on a fat JAR. The DAO artifact is now 90KB instead of 90MB.

PR #1298, opened 2020-04-23
#1280 Make DAO a thin jar and move shading logic to 'spark-jobs'.

yruslan:


spark-jobs/pom.xml
```diff
@@ -191,6 +190,20 @@
...
+                             If not shaded the following exception will be thrown:
+                               Caused by: java.lang.ClassNotFoundException: za.co.absa.atum.model.RunState
+	                           at java.net.URLClassLoader.findClass(URLClassLoader.java:382)
+	                           ...
+	                           at com.fasterxml.jackson.module.scala.deser.EnumerationDeserializer.deserialize(EnumerationDeserializerModule.scala:40)
+                               java.lang.NoClassDefFoundError: com/fasterxml/jackson/datatype/jsr310/JavaTimeModule
+                        -->
+                        <relocation>
+                            <pattern>com.fasterxml.jackson</pattern>
+                            <shadedPattern>com.shaded.fasterxml.jackson</shadedPattern>
```
dk1844, line 204: Maybe it does not matter here (moreover, if the purpose for relocation is only not to interfere), but usually, I find a good practice to prefix the newyl relocated package name by our own namespace (e.g. `za.co.absa.shaded.com.fasterxml.jackson`). This also has the benefit of minimizing the chance that another jar chose the same shaded location.

yruslan: Good point! Will do

spark-jobs/pom.xml
```diff
@@ -191,6 +190,20 @@
                             <exclude>net.jpountz.lz4:*</exclude>
                         </excludes>
                     </artifactSet>
+                    <relocations>
+                        <!-- This was introduced in #86 Set up Standardization and Conformance Menas auth.
```
benedeki, line 194: Question: I like comments. Just wonder, if this is not too detailed explanation?

yruslan: I don't mind removing it. It just took some time to find out why we are doing this relocation. I'm slightly inclined to keep it for the future maintainers but will remove it if you think this is too much 😄

benedeki: I definitely like that it's explained why the relocation was added. Just found it untypical, compared to what we usually have, that it got down to the level of mentioning the exception class. Fine if it stays, we can always shorten the comment if future. in case it's found to be long. 😉

Issue #1294, opened 2020-04-22
Add 'enceladus_record_id' during processing


yruslan:
## Background
We need a unique record identifier so that errors sent to a DQ topic can be traced back to the original record.

Various data sources might have very complicated primary keys, some might not have primary keys. 

## Feature
We should generate our own 'enceladus_record_id' based on GUUID, for instance.

This should be done in Standardization and Conformance similarly to how we generate 'enceladus_info_date' and 'enceladus_info_version'. Conformance should generate these filelds only if they are absent.

**Note** Usage of GUUIDs is a straightforward solution. If you have better ideas - please let us know, let's discuss.

Zejnilovic:
Thinking about testability. Could we generate GUUID so it is "always same"? My idea is if we run standardization twice, can we compare the datasets? It could also help us with "merging" the data and figuring out what changed between report versions.

dk1844:
I am thinking if we based this ID generation on a random generator, one could provide a seed - so, if the tests always provided the same seed, the ID generation could be always the same.

We could use the default seed for production then. What do you think, @Zejnilovic?

Zejnilovic:
A seed sounds like an interesting idea. If I then run standardization twice, would it generate the same IDs?

dk1844:
I think it should, but have not coded it yet, just thinking. If we agree this is the way to go, I can jump on it.

Based on some [ideas presented here](https://stackoverflow.com/a/49736088/1773349), I could use the `org.apache.spark.sql.functions.rand()` with or without seed. I could expose the ability to provide the seed via a method or a parameter and then use the result of the rand to `java.util.UUID.nameUUIDFromBytes`.

This way, we could, I hope, get a reproducible random set of record ids.

tagging @benedeki as he wanted to be part of the discussion.

yruslan:
UUID generation routines guarantee that generated values are globally unique. We can use this property in prod.

For testing, we can replace the generator with the one that uses a fixed seed.
I'd limit the usage of `rand()` with `java.util.UUID.nameUUIDFromBytes` for unit test only since we need to make sure generated ids are unique in prod. And `rand()` doesn't provide such guarantees.

Alternatively, tests can just drop the id column before doing comparisons.

Anyway, there is another issue potentially. If the same data is conformed using batch and streaming jobs, we can't match unique ids since they are generated independently. But this is just a hypothetical issue, so let's not to focus on it for now.

Zejnilovic:
[...]

Issue #1220, opened 2020-03-10
Implement Enceladus plugin to send error columns to DQ topic


yruslan:
## Background
Data Operations require to get notifications of any standardization and conformance errors as soon as possible. On their side, they subscribe to a Kafka topic, the data quality (DQ) topic.

## Feature
From Enceladus's side we need to implement a built-in plugin that sends contents of all non-empty error columns to the DQ topic.

## Proposed solution
* The sending should happen *after* the output file is already saved. That way we ensure that we send errors in the error column only after the job has succeeded.
* The plugin should use `PostProcessor` interface (and the corresponding factory). Currently, a post-processor gets a dataframe and returns a dataframe. Maybe we should change it to return `Unit` since the output data won't be changed as the result. Also, we need to rethink the name `PostProcessor` maybe.
* The schema for the topic is hard-coded and can be placed in the `resources` folder similar to how it works for the control measurements info plugin.
* Kafka configs can be reused between info file and error plugins:
```
kafka.schema.registry.url:"http://127.0.0.1:8081"
kafka.bootstrap.servers="127.0.0.1:9092"
kafka.security.protocol="SASL_SSL"
kafka.sasl.mechanism="GSSAPI"
```
* The `topic name` and the `client id` should be specified specifically for this particular plugin.

dk1844:
Release notes
Sending erros to a kafka topic can now be used enabling built-in plugins:
```
standardization.plugin.postprocessor.1=za.co.absa.enceladus.plugins.builtin.errorsender.mq.kafka.KafkaErrorSenderPlugin
conformance.plugin.postprocessor.1=za.co.absa.enceladus.plugins.builtin.errorsender.mq.kafka.KafkaErrorSenderPlugin
```

Note, that the existence of supporting keys is necessary: `kafka.schema.registry.url`, `kafka.bootstrap.servers`, `kafka.error.client.id`, `kafka.error.topic.name`.

Issue #1029, opened 2019-11-11
Dataset common settings


benedeki:
## Background
There are some settings that would be useful to be defined as properties of a dataset.

## Feature
Allow to have properties to be defined for datasets, similarly as fields have _metadata_

## Example
Some metadata of the fields would be nice to be able to define only once per dataset, because the chance all the fields share them is high.
Good contenders:
* Time zone
* Locale
* Decimal symbols
* ...

PR #1315, opened 2020-04-29
#1294 enceladus_record_id added to Standardization (& Conformance)

dk1844:
`enceladus_record_id` can now be added to the result of standardization and conformance
 
There are 3 modes (`uuid`, `stableHashId`, and `none`), usage of which can be driven by supplying the `-Denceladus.recordid.generation.strategy={uuid|stableHashId|none}`:
 - `uuid` (default mode) - the `enceladus_record_id` column will be added with values resulting from `java.util.UUID.randomUUID()`
 - `stableHashId` - the `enceladus_record_id` column will be added with values resulting from (Murmur3) `hash` of all values in the row. This way, each run of standardization or conformance will yield the same ID.
 - `none` - the `enceladus_record_id` column will not be added at all.

## StableHashId considerations
 - `stableHashId` mode always yields the same `Int` value for the same data. In this approach, we value _stability over uniqueness_, because if there were 2 identical rows, their hash/id would end up being the same. This, however, should not happen in practice and should suffice for this mode's specific purposes.

## ID generation in Conformance
 - Both Standardization and Conformance checksthe presence of the `enceladus_record_id` in its input and the `enceladus.recordid.generation.strategy` value. If the field exists, nothing is done, no matter the strategy. If the field does not exist, the column might be added based on the strategy set.
Note, that the `stableHashId` strategy may assign different values in Conformance compared to what standardization would because there might be different data in the row after conformance (especially if there were some conformance error). In general, the basic condition holds - over the same input dataset, the strategy will assign the same values each time.

## Testing
- Individual classes responsible for ID generation are accompanied by unit tests, so is the standardization usage.
- I have did a few test-runs, the output indeed confirmed the expected - `none` turns off the id generation, `stableHashId` always generates the same IDs (even true for standardization & conformance combinations), `uuid`/omitting the override results in each-run-different UUIDs.

### DIFF
--- a/spark-jobs/src/main/resources/application.conf.template
+++ b/spark-jobs/src/main/resources/application.conf.template
@@ -17,6 +17,10 @@ enceladus.version=${project.version}
 # each can have multiple comma-separated hosts, these are used for fault-tolerance
 menas.rest.uri="http://localhost:8080,host2:9000/menas;https://localhost:8080,host2:9000/menas"
 
+# 'enceladus_record_id' with an id can be added containing either true UUID, always the same UUIDs (pseudo) or the column
+# will not be added at all. Allowed values: "no", "pseudo", "true"
+enceladus.recordid.generation.strategy="true"
+
 standardized.hdfs.path="/tmp/conformance-output/standardized-{0}-{1}-{2}-{3}"
 
 # Pattern to look for mapping table for the specified date
--- a/spark-jobs/src/main/resources/application.conf.template
+++ b/spark-jobs/src/main/resources/application.conf.template
@@ -19,7 +19,7 @@ menas.rest.uri="http://localhost:8080,host2:9000/menas;https://localhost:8080,ho
 
 # 'enceladus_record_id' with an id can be added containing either true UUID, always the same UUIDs (pseudo) or the column
 # will not be added at all. Allowed values: "true", "pseudo", "no"
-enceladus.recordid.generation.strategy="true"
+enceladus.recordId.generation.strategy="true"
 
 standardized.hdfs.path="/tmp/conformance-output/standardized-{0}-{1}-{2}-{3}"
 
--- a/spark-jobs/src/main/resources/application.conf.template
+++ b/spark-jobs/src/main/resources/application.conf.template
@@ -17,9 +17,9 @@ enceladus.version=${project.version}
 # each can have multiple comma-separated hosts, these are used for fault-tolerance
 menas.rest.uri="http://localhost:8080,host2:9000/menas;https://localhost:8080,host2:9000/menas"
 
-# 'enceladus_record_id' with an id can be added containing either true UUID, always the same UUIDs (pseudo) or the column
-# will not be added at all. Allowed values: "true", "pseudo", "no"
-enceladus.recordId.generation.strategy="true"
+# 'enceladus_record_id' with an id can be added containing either true UUID, always the same IDs (row-hash-based) or the
+# column will not be added at all. Allowed values: "uuid", "stableHashId", "none"
+enceladus.recordId.generation.strategy="uuid"
 
 standardized.hdfs.path="/tmp/conformance-output/standardized-{0}-{1}-{2}-{3}"
 
--- a/spark-jobs/src/main/resources/application.conf.template
+++ b/spark-jobs/src/main/resources/application.conf.template
@@ -18,7 +18,7 @@ enceladus.version=${project.version}
 menas.rest.uri="http://localhost:8080,host2:9000/menas;https://localhost:8080,host2:9000/menas"
 
 # 'enceladus_record_id' with an id can be added containing either true UUID, always the same UUIDs (pseudo) or the column
-# will not be added at all. Allowed values: "no", "pseudo", "true"
+# will not be added at all. Allowed values: "true", "pseudo", "no"
 enceladus.recordid.generation.strategy="true"
 
 standardized.hdfs.path="/tmp/conformance-output/standardized-{0}-{1}-{2}-{3}"
--- a/spark-jobs/src/main/scala/za/co/absa/enceladus/common/RecordIdGeneration.scala
+++ b/spark-jobs/src/main/scala/za/co/absa/enceladus/common/RecordIdGeneration.scala
@@ -17,9 +17,8 @@ package za.co.absa.enceladus.common
 
 import com.typesafe.config.{Config, ConfigException}
 import org.apache.spark.sql.DataFrame
-import org.apache.spark.sql.functions.{callUDF, col, hash}
+import org.apache.spark.sql.functions.{col, hash, expr}
 import org.slf4j.{Logger, LoggerFactory}
-import za.co.absa.enceladus.utils.udf.{UDFLibrary, UDFNames}
 
 object RecordIdGeneration {
 
@@ -53,10 +52,9 @@ object RecordIdGeneration {
    * @param origDf       dataframe to be possibly extended
    * @param idColumnName name of the id column to be used (usually [[Constants.EnceladusRecordId]])
    * @param strategy     decides if and what ids will be appended to the origDf
-   * @param udfLib       library that registred UDFs [[UDFNames.uuid]]
    * @return possibly updated `origDf`
    */
-  def addRecordIdColumnByStrategy(origDf: DataFrame, idColumnName: String, strategy: IdType)(implicit udfLib: UDFLibrary): DataFrame = {
+  def addRecordIdColumnByStrategy(origDf: DataFrame, idColumnName: String, strategy: IdType): DataFrame = {
     strategy match {
       case IdType.NoId =>
         log.info("Record id generation is off.")
@@ -68,7 +66,7 @@ object RecordIdGeneration {
 
       case IdType.TrueUuids =>
         log.info("Record id generation is on and true UUIDs will be added to output.")
-        origDf.withColumn(Constants.EnceladusRecordId, callUDF(UDFNames.uuid))
+        origDf.withColumn(Constants.EnceladusRecordId, expr("uuid()"))
     }
   }
 
--- a/spark-jobs/src/main/scala/za/co/absa/enceladus/common/RecordIdGeneration.scala
+++ b/spark-jobs/src/main/scala/za/co/absa/enceladus/common/RecordIdGeneration.scala
@@ -23,13 +23,17 @@ import za.co.absa.enceladus.utils.udf.{UDFLibrary, UDFNames}
 
 object RecordIdGeneration {
 
-  private val log: Logger = LoggerFactory.getLogger(this.getClass)
+  sealed trait UuidType
 
-  object UuidType extends Enumeration {
-    val TrueUuids, PseudoUuids, NoUuids = Value
+  object UuidType {
+    case object TrueUuids extends UuidType
+    case object PseudoUuids extends UuidType
+    case object NoUuids  extends UuidType
   }
 
-  def getRecordIdGenerationStrategyFromConfig(conf: Config): UuidType.Value = {
+  private val log: Logger = LoggerFactory.getLogger(this.getClass)
+
+  def getRecordIdGenerationStrategyFromConfig(conf: Config): UuidType = {
     conf.getString("enceladus.recordid.generation.strategy") match {
       case "true" => UuidType.TrueUuids
       case "pseudo" => UuidType.PseudoUuids
@@ -49,7 +53,7 @@ object RecordIdGeneration {
    * @param udfLib
    * @return possibly updated `origDf`
    */
-  def addRecordIdColumnByStrategy(origDf: DataFrame, strategy: UuidType.Value)
+  def addRecordIdColumnByStrategy(origDf: DataFrame, strategy: UuidType)
                                  (implicit udfLib: UDFLibrary) : DataFrame = {
     strategy match {
       case UuidType.NoUuids => origDf
--- a/spark-jobs/src/main/scala/za/co/absa/enceladus/common/RecordIdGeneration.scala
+++ b/spark-jobs/src/main/scala/za/co/absa/enceladus/common/RecordIdGeneration.scala
@@ -23,60 +23,55 @@ import za.co.absa.enceladus.utils.udf.{UDFLibrary, UDFNames}
 
 object RecordIdGeneration {
 
-  sealed trait UuidType
+  sealed trait IdType
 
-  object UuidType {
-    case object TrueUuids extends UuidType
-    case object PseudoUuids extends UuidType
-    case object NoUuids extends UuidType
+  object IdType {
+    case object TrueUuids extends IdType
+    case object StableHashId extends IdType
+    case object NoId extends IdType
   }
 
   private val log: Logger = LoggerFactory.getLogger(this.getClass)
 
-  def getRecordIdGenerationStrategyFromConfig(conf: Config): UuidType = {
+  def getRecordIdGenerationStrategyFromConfig(conf: Config): IdType = {
     val strategyValue = conf.getString("enceladus.recordId.generation.strategy")
 
     strategyValue.toLowerCase match {
-      case "true" => UuidType.TrueUuids
-      case "pseudo" => UuidType.PseudoUuids
-      case "no" => UuidType.NoUuids
+      case "uuid" => IdType.TrueUuids
+      case "stablehashid" => IdType.StableHashId
+      case "none" => IdType.NoId
       case _ => throw new ConfigException.BadValue("enceladus.recordId.generation.strategy",
-        s"Invalid value $strategyValue was encountered for id generation strategy, use one of: true, pseudo, no.")
+        s"Invalid value '$strategyValue' was encountered for id generation strategy, use one of: uuid, stableHashId, none.")
     }
   }
 
   /**
-   * The supplied dataframe `origDf` is either kept as-is (`strategy` = [[UuidType.NoUuids]]) or appended the a column named
-   * [[Constants.EnceladusRecordId]] with an ID for each record. These ID true UUID (`strategy` = [[UuidType.TrueUuids]])
-   * or always the same ones for testing purposes (`strategy` = [[UuidType.PseudoUuids]]
+   * The supplied dataframe `origDf` is either kept as-is (`strategy` = [[IdType.NoId]]) or appended the a column named
+   * [[Constants.EnceladusRecordId]] with an ID for each record. These ID true UUID (`strategy` = [[IdType.TrueUuids]])
+   * or always the same ones for testing purposes (`strategy` = [[IdType.StableHashId]]
    *
-   * @param origDf dataframe to be possibly extended
+   * @param origDf   dataframe to be possibly extended
    * @param strategy decides if and what ids will be appended to the origDf
-   * @param udfLib library that registred UDFs [[UDFNames.pseudoUuidFromHash]] and [[UDFNames.uuid]]
+   * @param udfLib   library that registred UDFs [[UDFNames.uuid]]
    * @return possibly updated `origDf`
    */
-  def addRecordIdColumnByStrategy(origDf: DataFrame, strategy: UuidType)
-                                 (implicit udfLib: UDFLibrary): DataFrame = {
-
+  def addRecordIdColumnByStrategy(origDf: DataFrame, strategy: IdType)(implicit udfLib: UDFLibrary): DataFrame = {
     strategy match {
-      case UuidType.NoUuids =>
+      case IdType.NoId =>
         log.info("Record id generation is off.")
         origDf
-      case UuidType.PseudoUuids =>
-        log.info("Record id generation is set to 'pseudo' - all runs will yield the same IDs.")
 
-        val hashColName = "enceladusTempHashForPseudoUuid"
-        def hashFromAllColumns(df: DataFrame) = df.withColumn(hashColName, hash(df.columns.map(col): _*))
+      case IdType.StableHashId =>
+        log.info(s"Record id generation is set to 'stableHashId' - all runs will yield the same IDs.")
+        origDf.transform(hashFromAllColumns(Constants.EnceladusRecordId, _)) // adds hashId
 
-        import org.apache.spark.sql.functions.col
-        origDf.transform(hashFromAllColumns) // adds hash
-          .withColumn(Constants.EnceladusRecordId, callUDF(UDFNames.pseudoUuidFromHash, col(hashColName)))
-          .drop(hashColName) // hash is no longer needed (pseudo uuids were generated from it)
-
-      case UuidType.TrueUuids =>
+      case IdType.TrueUuids =>
         log.info("Record id generation is on and true UUIDs will be added to output.")
         origDf.withColumn(Constants.EnceladusRecordId, callUDF(UDFNames.uuid))
     }
   }
+
+  private def hashFromAllColumns(hashColName: String, df: DataFrame): DataFrame =
+    df.withColumn(hashColName, hash(df.columns.map(col): _*))
 
 }
--- a/spark-jobs/src/main/scala/za/co/absa/enceladus/common/RecordIdGeneration.scala
+++ b/spark-jobs/src/main/scala/za/co/absa/enceladus/common/RecordIdGeneration.scala
@@ -46,9 +46,9 @@ object RecordIdGeneration {
   }
 
   /**
-   * The supplied dataframe `origDf` is either kept as-is (`strategy` = [[IdType.NoId]]) or appended an ID column
-   * with an ID for each record. These ID true UUID (`strategy` = [[IdType.TrueUuids]]) or always the same ones for testing
-   * purposes (`strategy` = [[IdType.StableHashId]]
+   * The supplied dataframe `origDf` is either kept as-is (`strategy` = [[IdType.NoId]]) or has a column appended
+   * with an (presumably) unique value for each record. These are true UUIDs (`strategy` = [[IdType.TrueUuids]]) or
+   * values always the same for the same row, mainly for testing purposes (`strategy` = [[IdType.StableHashId]]
    *
    * @param origDf       dataframe to be possibly extended
    * @param idColumnName name of the id column to be used (usually [[Constants.EnceladusRecordId]])
--- a/spark-jobs/src/main/scala/za/co/absa/enceladus/common/RecordIdGeneration.scala
+++ b/spark-jobs/src/main/scala/za/co/absa/enceladus/common/RecordIdGeneration.scala
@@ -1,0 +1,64 @@
+/*
+ * Copyright 2018 ABSA Group Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.enceladus.common
+
+import com.typesafe.config.{Config, ConfigException}
+import org.apache.spark.sql.DataFrame
+import org.apache.spark.sql.functions.callUDF
+import org.slf4j.{Logger, LoggerFactory}
+import za.co.absa.enceladus.utils.udf.{UDFLibrary, UDFNames}
+
+object RecordIdGeneration {
+
+  private val log: Logger = LoggerFactory.getLogger(this.getClass)
+
+  object UuidType extends Enumeration {
+    val TrueUuids, PseudoUuids, NoUuids = Value
+  }
+
+  def getRecordIdGenerationStrategyFromConfig(conf: Config): UuidType.Value = {
+    conf.getString("enceladus.recordid.generation.strategy") match {
+      case "true" => UuidType.TrueUuids
+      case "pseudo" => UuidType.PseudoUuids
+      case "no" => UuidType.NoUuids
+      case other => throw new ConfigException.BadValue("enceladus.recordid.generation.strategy",
+        s"Invalid value $other was encountered for id generation strategy, use one of: true, pseudo, no.")
+    }
+  }
+
+  /**
+   * The supplied dataframe `origDf` is either kept as-is (`stragegy` = [[UuidType.NoUuids]]) or appended the a column named
+   * [[Constants.EnceladusRecordId]] with an ID for each record. These ID true UUID (`stragegy` = [[UuidType.TrueUuids]])
+   * or always the same ones for testing purposes (`stragegy` = [[UuidType.PseudoUuids]]
+   *
+   * @param origDf dataframe to be possibly extended
+   * @param strategy decides if and what ids will be appended to the origDf
+   * @param udfLib
+   * @return possibly updated `origDf`
+   */
+  def addRecordIdColumnByStrategy(origDf: DataFrame, strategy: UuidType.Value)
+                                 (implicit udfLib: UDFLibrary) : DataFrame = {
+    strategy match {
+      case UuidType.NoUuids => origDf
+      case other => {
+        val udfName = if (other == UuidType.TrueUuids) UDFNames.uuid else UDFNames.pseudoUuid
+        log.info(s"Adding UUIDs ($other) to the records.")
+        origDf.withColumn(Constants.EnceladusRecordId, callUDF(udfName))
+      }
+    }
+  }
+
+}
--- a/spark-jobs/src/main/scala/za/co/absa/enceladus/common/RecordIdGeneration.scala
+++ b/spark-jobs/src/main/scala/za/co/absa/enceladus/common/RecordIdGeneration.scala
@@ -46,16 +46,17 @@ object RecordIdGeneration {
   }
 
   /**
-   * The supplied dataframe `origDf` is either kept as-is (`strategy` = [[IdType.NoId]]) or appended the a column named
-   * [[Constants.EnceladusRecordId]] with an ID for each record. These ID true UUID (`strategy` = [[IdType.TrueUuids]])
-   * or always the same ones for testing purposes (`strategy` = [[IdType.StableHashId]]
+   * The supplied dataframe `origDf` is either kept as-is (`strategy` = [[IdType.NoId]]) or appended an ID column
+   * with an ID for each record. These ID true UUID (`strategy` = [[IdType.TrueUuids]]) or always the same ones for testing
+   * purposes (`strategy` = [[IdType.StableHashId]]
    *
-   * @param origDf   dataframe to be possibly extended
-   * @param strategy decides if and what ids will be appended to the origDf
-   * @param udfLib   library that registred UDFs [[UDFNames.uuid]]
+   * @param origDf       dataframe to be possibly extended
+   * @param idColumnName name of the id column to be used (usually [[Constants.EnceladusRecordId]])
+   * @param strategy     decides if and what ids will be appended to the origDf
+   * @param udfLib       library that registred UDFs [[UDFNames.uuid]]
    * @return possibly updated `origDf`
    */
-  def addRecordIdColumnByStrategy(origDf: DataFrame, strategy: IdType)(implicit udfLib: UDFLibrary): DataFrame = {
+  def addRecordIdColumnByStrategy(origDf: DataFrame, idColumnName: String, strategy: IdType)(implicit udfLib: UDFLibrary): DataFrame = {
     strategy match {
       case IdType.NoId =>
         log.info("Record id generation is off.")
--- a/spark-jobs/src/main/scala/za/co/absa/enceladus/common/RecordIdGeneration.scala
+++ b/spark-jobs/src/main/scala/za/co/absa/enceladus/common/RecordIdGeneration.scala
@@ -34,12 +34,14 @@ object RecordIdGeneration {
   private val log: Logger = LoggerFactory.getLogger(this.getClass)
 
   def getRecordIdGenerationStrategyFromConfig(conf: Config): UuidType = {
-    conf.getString("enceladus.recordid.generation.strategy") match {
+    val strategyValue = conf.getString("enceladus.recordid.generation.strategy")
+
+    strategyValue.toLowerCase match {
       case "true" => UuidType.TrueUuids
       case "pseudo" => UuidType.PseudoUuids
       case "no" => UuidType.NoUuids
       case other => throw new ConfigException.BadValue("enceladus.recordid.generation.strategy",
-        s"Invalid value $other was encountered for id generation strategy, use one of: true, pseudo, no.")
+        s"Invalid value $strategyValue was encountered for id generation strategy, use one of: true, pseudo, no.")
     }
   }
 
--- a/spark-jobs/src/main/scala/za/co/absa/enceladus/common/RecordIdGeneration.scala
+++ b/spark-jobs/src/main/scala/za/co/absa/enceladus/common/RecordIdGeneration.scala
@@ -17,7 +17,7 @@ package za.co.absa.enceladus.common
 
 import com.typesafe.config.{Config, ConfigException}
 import org.apache.spark.sql.DataFrame
-import org.apache.spark.sql.functions.callUDF
+import org.apache.spark.sql.functions.{callUDF, col, hash}
 import org.slf4j.{Logger, LoggerFactory}
 import za.co.absa.enceladus.utils.udf.{UDFLibrary, UDFNames}
 
@@ -28,42 +28,54 @@ object RecordIdGeneration {
   object UuidType {
     case object TrueUuids extends UuidType
     case object PseudoUuids extends UuidType
-    case object NoUuids  extends UuidType
+    case object NoUuids extends UuidType
   }
 
   private val log: Logger = LoggerFactory.getLogger(this.getClass)
 
   def getRecordIdGenerationStrategyFromConfig(conf: Config): UuidType = {
-    val strategyValue = conf.getString("enceladus.recordid.generation.strategy")
+    val strategyValue = conf.getString("enceladus.recordId.generation.strategy")
 
     strategyValue.toLowerCase match {
       case "true" => UuidType.TrueUuids
       case "pseudo" => UuidType.PseudoUuids
       case "no" => UuidType.NoUuids
-      case other => throw new ConfigException.BadValue("enceladus.recordid.generation.strategy",
+      case _ => throw new ConfigException.BadValue("enceladus.recordId.generation.strategy",
         s"Invalid value $strategyValue was encountered for id generation strategy, use one of: true, pseudo, no.")
     }
   }
 
   /**
-   * The supplied dataframe `origDf` is either kept as-is (`stragegy` = [[UuidType.NoUuids]]) or appended the a column named
-   * [[Constants.EnceladusRecordId]] with an ID for each record. These ID true UUID (`stragegy` = [[UuidType.TrueUuids]])
-   * or always the same ones for testing purposes (`stragegy` = [[UuidType.PseudoUuids]]
+   * The supplied dataframe `origDf` is either kept as-is (`strategy` = [[UuidType.NoUuids]]) or appended the a column named
+   * [[Constants.EnceladusRecordId]] with an ID for each record. These ID true UUID (`strategy` = [[UuidType.TrueUuids]])
+   * or always the same ones for testing purposes (`strategy` = [[UuidType.PseudoUuids]]
    *
    * @param origDf dataframe to be possibly extended
    * @param strategy decides if and what ids will be appended to the origDf
-   * @param udfLib
+   * @param udfLib library that registred UDFs [[UDFNames.pseudoUuidFromHash]] and [[UDFNames.uuid]]
    * @return possibly updated `origDf`
    */
   def addRecordIdColumnByStrategy(origDf: DataFrame, strategy: UuidType)
-                                 (implicit udfLib: UDFLibrary) : DataFrame = {
+                                 (implicit udfLib: UDFLibrary): DataFrame = {
+
     strategy match {
-      case UuidType.NoUuids => origDf
-      case other => {
-        val udfName = if (other == UuidType.TrueUuids) UDFNames.uuid else UDFNames.pseudoUuid
-        log.info(s"Adding UUIDs ($other) to the records.")
-        origDf.withColumn(Constants.EnceladusRecordId, callUDF(udfName))
-      }
+      case UuidType.NoUuids =>
+        log.info("Record id generation is off.")
+        origDf
+      case UuidType.PseudoUuids =>
+        log.info("Record id generation is set to 'pseudo' - all runs will yield the same IDs.")
+
+        val hashColName = "enceladusTempHashForPseudoUuid"
+        def hashFromAllColumns(df: DataFrame) = df.withColumn(hashColName, hash(df.columns.map(col): _*))
+
+        import org.apache.spark.sql.functions.col
+        origDf.transform(hashFromAllColumns) // adds hash
+          .withColumn(Constants.EnceladusRecordId, callUDF(UDFNames.pseudoUuidFromHash, col(hashColName)))
+          .drop(hashColName) // hash is no longer needed (pseudo uuids were generated from it)
+
+      case UuidType.TrueUuids =>
+        log.info("Record id generation is on and true UUIDs will be added to output.")
+        origDf.withColumn(Constants.EnceladusRecordId, callUDF(UDFNames.uuid))
     }
   }
 
--- a/spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
+++ b/spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
@@ -45,7 +45,7 @@ import za.co.absa.enceladus.utils.performance.{PerformanceMeasurer, PerformanceM
 import za.co.absa.enceladus.utils.time.TimeZoneNormalizer
 import za.co.absa.enceladus.utils.udf.UDFLibrary
 
-import scala.util.Try
+import scala.util.{Failure, Success, Try}
 import scala.util.control.NonFatal
 
 object DynamicConformanceJob {
@@ -228,7 +228,7 @@ object DynamicConformanceJob {
     performance
   }
 
-  private def conform(conformance: Dataset, inputData: sql.Dataset[Row], enableCF: Boolean, recordIdGenerationStrategy: UuidType.Value)
+  private def conform(conformance: Dataset, inputData: sql.Dataset[Row], enableCF: Boolean, recordIdGenerationStrategy: UuidType)
                      (implicit spark: SparkSession, cmd: ConfCmdConfig, fsUtils: FileSystemVersionUtils, dao: MenasDAO): DataFrame = {
     implicit val featureSwitcher: FeatureSwitches = FeatureSwitches()
       .setExperimentalMappingRuleEnabled(isExperimentalRuleEnabled())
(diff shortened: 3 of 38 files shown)


spark-jobs/src/main/resources/application.conf.template
```diff
@@ -17,6 +17,10 @@ enceladus.version=${project.version}
 # each can have multiple comma-separated hosts, these are used for fault-tolerance
 menas.rest.uri="http://localhost:8080,host2:9000/menas;https://localhost:8080,host2:9000/menas"
 
+# 'enceladus_record_id' with an id can be added containing either true UUID, always the same UUIDs (pseudo) or the column
+# will not be added at all. Allowed values: "no", "pseudo", "true"
+enceladus.recordid.generation.strategy="true"
```
dk1844, line 22: @Zejnilovic, I hope you are ok with this form of overridability. 😉

spark-jobs/src/test/scala/za/co/absa/enceladus/standardization/interpreter/StdInterpreterSuite.scala
```diff
@@ -22,6 +22,7 @@ import org.apache.spark.sql.types._
 import org.scalatest.FunSuite
 import za.co.absa.enceladus.utils.error.{ErrorMessage, UDFLibrary}
 import za.co.absa.enceladus.utils.testUtils.{LoggerTestBase, SparkTestBase}
+import za.co.absa.enceladus.utils.udf
```
AdrianOlosutean, line 25: import udf.UDFLibrary

spark-jobs/src/main/resources/application.conf.template
Zejnilovic: this is perfect 👍

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
```diff
@@ -221,7 +228,7 @@ object DynamicConformanceJob {
     performance
   }
 
-  private def conform(conformance: Dataset, inputData: sql.Dataset[Row], enableCF: Boolean)
+  private def conform(conformance: Dataset, inputData: sql.Dataset[Row], enableCF: Boolean, recordIdGenerationStrategy: UuidType.Value)
```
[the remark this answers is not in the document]
dk1844: Sure, done.

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
```diff
@@ -230,7 +237,7 @@ object DynamicConformanceJob {
       .setBroadcastStrategyMode(broadcastingStrategyMode)
       .setBroadcastMaxSizeMb(broadcastingMaxSizeMb)
 
-    try {
+    val conformedDF = try {
```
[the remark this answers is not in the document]
dk1844: Done.

spark-jobs/src/test/scala/za/co/absa/enceladus/standardization/interpreter/StdInterpreterSuite.scala
dk1844: good idea, thanks, done.

utils/src/main/scala/za/co/absa/enceladus/utils/udf/UDFLibrary.scala
```diff
@@ -64,6 +69,20 @@ case class UDFLibrary()(implicit val spark: SparkSession) {
   spark.udf.register(errorColumnAppend,
                      UDFLibrary.errorColumnAppend,
                      ArrayType.apply(ErrorMessage.errorColSchema, containsNull = false))
+
+
+  spark.udf.register(uuid, { () => UUID.randomUUID().toString })
+
+  private val pseudoRandom = new Random(seed = 22L) // scalastyle:ignore magic.number
```
yruslan, line 76: The instance of `Random` might be created on each executor separately. Meaning that if more than 1 executor is used, it will repeat keys for each executor.

This is theory. But I'm wondering if pseudo UUID is going to create duplicate keys when more than 1 executor is used?

utils/src/main/scala/za/co/absa/enceladus/utils/schema/SparkUtils.scala
```diff
@@ -66,6 +67,8 @@ object SparkUtils {
     }
   }
 
+  def containsColumn(df: DataFrame, colName: String): Boolean = df.schema.exists(field => field.name.equalsIgnoreCase(colName))
```
yruslan, line 70: This might better belong to `SchemaUtils` and the method might accept `Schema` as the input parameter.

spark-jobs/src/main/scala/za/co/absa/enceladus/common/RecordIdGeneration.scala
```diff
@@ -0,0 +1,68 @@
...
+  object UuidType {
+    case object TrueUuids extends UuidType
+    case object PseudoUuids extends UuidType
+    case object NoUuids  extends UuidType
+  }
+
+  private val log: Logger = LoggerFactory.getLogger(this.getClass)
+
+  def getRecordIdGenerationStrategyFromConfig(conf: Config): UuidType = {
+    conf.getString("enceladus.recordid.generation.strategy") match {
```
yruslan, line 37: Just a tiny suggestion. Consider lowercasing the config value before matching so that 'True' and 'PsEuDo' can be valid.

dk1844: Couldn't hurt, true. Done.

utils/src/main/scala/za/co/absa/enceladus/utils/schema/SparkUtils.scala
dk1844: SchemaUtils already contains such a functionality - `SchemaUtils.fieldExists`. But this method has the advantage of being available as implicit directly on the dataframe.

So as a middle ground, I have changed this method to be calling the above-mentioned `SchemaUtils.fieldExists`
```scala
def containsColumn(colName: String): Boolean = SchemaUtils.fieldExists(colName, df.schema)
```
Would you consider this acceptable or you object primarily to the fact that the column existence is queried on the dataframe instead of being queried on the schema instead?

dk1844: Hmm, on the other hand, its true, that then the implicit is too much. I am changing it to using the `SchemaUtils` directly.
```scala
SchemaUtils.fieldExists(Constants.EnceladusRecordId, conformedDF.schema))```

Comment (dk1844) [reply utils/src/main/scala/za/co/absa/enceladus/utils/schema/SparkUtils.scala]: Done.

File: utils/src/main/scala/za/co/absa/enceladus/utils/udf/UDFLibrary.scala (hunk 5 above)
Comment (dk1844) [reply utils/src/main/scala/za/co/absa/enceladus/utils/udf/UDFLibrary.scala:76]: Sadly, spot on. The current implementation indeed only works as expected with only one executor, otherwise, the ids get duplicate :(

File: spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala [hunk 8]
```diff
@@ -230,17 +238,29 @@ object DynamicConformanceJob {
...
+        if (SchemaUtils.fieldExists(Constants.EnceladusRecordId, conformedDF.schema)) {
+          conformedDF // no new id regenereration
+        } else {
+          recordIdGenerationStrategy match {
+            case UuidType.NoUuids => log.info("Record id generation is off.")
+            case UuidType.PseudoUuids => log.info("Record id generation is set to 'pseudo' - all runs will yield the same IDs.")
+            case UuidType.TrueUuids => log.info("Record id generation is on and true UUIDs will be added to output.")
+          }
+          RecordIdGeneration.addRecordIdColumnByStrategy(conformedDF, recordIdGenerationStrategy)(UDFLibrary())
+        }
```
Zejnilovic, line 262: I need to ask. Could we extract this. Non-objective problem is that too much indentation is hard to read.

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
```diff
@@ -230,17 +238,29 @@ object DynamicConformanceJob {
...
         AtumImplicits.SparkSessionWrapper(spark).setControlMeasurementError("Conformance", e.getMessage, sw.toString)
         throw e
+      case Success(conformedDF) =>
+        if (SchemaUtils.fieldExists(Constants.EnceladusRecordId, conformedDF.schema)) {
+          conformedDF // no new id regenereration
+        } else {
+          recordIdGenerationStrategy match {
+            case UuidType.NoUuids => log.info("Record id generation is off.")
+            case UuidType.PseudoUuids => log.info("Record id generation is set to 'pseudo' - all runs will yield the same IDs.")
+            case UuidType.TrueUuids => log.info("Record id generation is on and true UUIDs will be added to output.")
```
Zejnilovic, line 259: This could be part of `RecordIdGeneration.addRecordIdColumnByStrategy`

spark-jobs/src/main/scala/za/co/absa/enceladus/common/RecordIdGeneration.scala
```diff
@@ -0,0 +1,70 @@
...
+   * @return possibly updated `origDf`
+   */
+  def addRecordIdColumnByStrategy(origDf: DataFrame, strategy: UuidType)
+                                 (implicit udfLib: UDFLibrary) : DataFrame = {
+    strategy match {
+      case UuidType.NoUuids => origDf
+      case other => {
+        val udfName = if (other == UuidType.TrueUuids) UDFNames.uuid else UDFNames.pseudoUuid
+        log.info(s"Adding UUIDs ($other) to the records.")
+        origDf.withColumn(Constants.EnceladusRecordId, callUDF(udfName))
```
Zejnilovic, line 65: what is the column is already taken?

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
```diff
@@ -22,23 +22,22 @@ import java.util.regex.Pattern
 
 import org.apache.spark.sql.Column
 import org.apache.spark.sql.expressions.UserDefinedFunction
-import org.apache.spark.sql.functions._
+import org.apache.spark.sql.functions.{lit, _}
```
Zejnilovic, line 25: Does the `lit` need to be specified likethis?

utils/src/main/scala/za/co/absa/enceladus/utils/schema/SparkUtils.scala
yruslan: Sorry, I just read these comments. Whatever your decision about this, I'm happy. 😄

utils/src/main/scala/za/co/absa/enceladus/utils/udf/UDFLibrary.scala
Zejnilovic: Great catch

yruslan: Spark supports a monotonically increasing numbers using this function: `monotonicallyIncreasingId`. The pseudo-UUID generation UDF can take a `long` s a parameter and add it to the generated numbers.

Still, the results will depend on the number of executors. But there shouldn't be duplicates.

Let's not rush with this to do it properly.
@HuvarVer , @Zejnilovic , what are your ideal requirements for testability?

spark-jobs/src/main/scala/za/co/absa/enceladus/common/RecordIdGeneration.scala
dk1844: It would be overridden. There should be no such column, though.
What do you think is the correct behavior in such a case? Throw an exception and fail? Keep the original `enceladus_record_id`?

utils/src/main/scala/za/co/absa/enceladus/utils/udf/UDFLibrary.scala
Zejnilovic: Well if we believe in such a "simple function" as adding UUIDs to the data, from testing perspective we would probably be ok with just `uuid` and `no` options. I would like to hear @HuvarVer ideas

spark-jobs/src/main/scala/za/co/absa/enceladus/common/RecordIdGeneration.scala
Zejnilovic: That's a question. If it is already there. The data has some keys. Users might expect those keys to be used. It would also give an option to generate one's own keys. I would **print a warning and leave the previous keys**.

@yruslan ?

utils/src/main/scala/za/co/absa/enceladus/utils/udf/UDFLibrary.scala
HuvarVer: I think, from the testing perspective, that we should be ok with checking that there is the column with id in the `UUID` case and that there is no column like that in the `no` case. I think there is no reason for checking the value of nonstatic value.

yruslan: Then maybe leave it as is. If we need more sophisticated pseudo UUID generation, we can always improve it later.

spark-jobs/src/main/scala/za/co/absa/enceladus/common/RecordIdGeneration.scala
yruslan: It could be a situation where Standardization or Conformance runs more than once. It could be that Conformance runs without Standardization. And in all such cases, the existing columns added by Enceladus previously are always preserved.

yruslan: Take a look at how `enceladus_info_date`,  `enceladus_info_version` and `enceladus_info_date_string` are created. We can follow the pattern for `enceladus_record_id` as well.

yruslan: Sorry for spamming. But the current implementation for `enceladus_record_id` preserves the existing column, isn't it? Let me put a comment there.

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
```diff
@@ -230,17 +238,29 @@ object DynamicConformanceJob {
...
         throw e
-      case NonFatal(e)            =>
+      case Failure(NonFatal(e)) =>
         val sw = new StringWriter
         e.printStackTrace(new PrintWriter(sw))
         AtumImplicits.SparkSessionWrapper(spark).setControlMeasurementError("Conformance", e.getMessage, sw.toString)
         throw e
+      case Success(conformedDF) =>
+        if (SchemaUtils.fieldExists(Constants.EnceladusRecordId, conformedDF.schema)) {
+          conformedDF // no new id regenereration
```
yruslan, line 254: So when the column already exists no id column will be generated. This is the expected behavior.

Zejnilovic: I have totally missed this. Thank you. Closing my previous comment.;

spark-jobs/src/main/scala/za/co/absa/enceladus/common/RecordIdGeneration.scala
```diff
@@ -0,0 +1,70 @@
...
+  object UuidType {
+    case object TrueUuids extends UuidType
+    case object PseudoUuids extends UuidType
+    case object NoUuids  extends UuidType
+  }
+
+  private val log: Logger = LoggerFactory.getLogger(this.getClass)
+
+  def getRecordIdGenerationStrategyFromConfig(conf: Config): UuidType = {
+    val strategyValue = conf.getString("enceladus.recordid.generation.strategy")
```
benedeki, line 37: **Question:** This makes the setting mandatory. Is it what we want?

spark-jobs/src/main/scala/za/co/absa/enceladus/common/RecordIdGeneration.scala
```diff
@@ -0,0 +1,70 @@
...
+
+    strategyValue.toLowerCase match {
+      case "true" => UuidType.TrueUuids
+      case "pseudo" => UuidType.PseudoUuids
+      case "no" => UuidType.NoUuids
```
benedeki, line 42: **Question:** What about '' (empty string) meaning the same?

benedeki:
Also please consider describing the new setting in the documentation. It's function and values are non-trivial and of the kind to be easily forgotten.

spark-jobs/src/main/scala/za/co/absa/enceladus/common/RecordIdGeneration.scala
dk1844: Yes, it does make it mandatory. This is in line with having the default value in the config.
I chose the combination deliberately. What would you prefer otherwise?

spark-jobs/src/test/scala/za/co/absa/enceladus/common/RecordIdGenerationSuite.scala
```diff
@@ -0,0 +1,100 @@
...
+
+    val df1 = spark.createDataFrame(data1)
+    val updatedDf1 = addRecordIdColumnByStrategy(df1, PseudoUuids)(UDFLibrary())
+    val updatedDf2 = addRecordIdColumnByStrategy(df1, PseudoUuids)(UDFLibrary())
+
+    updatedDf1.as[SomeDataWithId].collect() should contain theSameElementsInOrderAs updatedDf2.as[SomeDataWithId].collect()
+
+    Seq(updatedDf1, updatedDf2).foreach { updatedDf =>
+      val updatedData = updatedDf.as[SomeDataWithId].collect()
+      updatedData.size shouldBe 3
```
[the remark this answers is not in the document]
dk1844: They have the same function, why are you suggesting the change?

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
```diff
@@ -230,17 +238,29 @@ object DynamicConformanceJob {
       .setBroadcastStrategyMode(broadcastingStrategyMode)
       .setBroadcastMaxSizeMb(broadcastingMaxSizeMb)
 
-    try {
+    Try {
```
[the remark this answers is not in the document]
dk1844: https://github.com/AbsaOSS/enceladus/pull/1315#discussion_r417292857

It was @AdrianOlosutean's suggestion. I do not mind either version, both have merits.

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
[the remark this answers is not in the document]
dk1844: This is not a typo, it is intentional -- as in _Some ID column exists already, so we are not going to regenerate it (= generate it again)_

spark-jobs/src/main/scala/za/co/absa/enceladus/common/RecordIdGeneration.scala
dk1844: My initial response is: no, why to introduce another value when the 3 are quite explanatory, while the empty string would not.

What is your motivation?

utils/src/main/scala/za/co/absa/enceladus/utils/udf/UDFNames.scala
```diff
@@ -28,4 +28,7 @@ object UDFNames {
   final val arrayDistinctErrors = "arrayDistinctErrors"
   final val cleanErrCol = "cleanErrCol"
   final val errorColumnAppend = "errorColumnAppend"
+
+  final val uuid = "uuid"
```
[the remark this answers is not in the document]
dk1844: I have adjusted the UDF name to avoid possible clashes by including `enceladus` in it.

utils/src/main/scala/za/co/absa/enceladus/utils/udf/UDFLibrary.scala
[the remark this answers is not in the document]
dk1844: No longer relevant in this case (changed imp). Generally: 👍

utils/src/main/scala/za/co/absa/enceladus/utils/udf/UDFLibrary.scala
```diff
@@ -64,6 +69,20 @@ case class UDFLibrary()(implicit val spark: SparkSession) {
...
+
+  def pseudoUuidFn: UUID = {
+    val arr = new Array[Byte](10) // scalastyle:ignore magic.number
```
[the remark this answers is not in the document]
dk1844: I have changed the implementation so this is no longer relevant

utils/src/main/scala/za/co/absa/enceladus/utils/udf/UDFLibrary.scala
```diff
@@ -64,6 +69,20 @@ case class UDFLibrary()(implicit val spark: SparkSession) {
   spark.udf.register(errorColumnAppend,
                      UDFLibrary.errorColumnAppend,
                      ArrayType.apply(ErrorMessage.errorColSchema, containsNull = false))
+
+
+  spark.udf.register(uuid, { () => UUID.randomUUID().toString })
+
+  private val pseudoRandom = new Random(seed = 22L) // scalastyle:ignore magic.number
+
+  def pseudoUuidFn: UUID = {
```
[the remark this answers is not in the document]
dk1844: Done, thanks.

spark-jobs/src/test/scala/za/co/absa/enceladus/common/RecordIdGenerationSuite.scala
```diff
@@ -0,0 +1,100 @@
...
+  import spark.implicits._
+
+  val data1 = Seq(
+    SomeData("abc", 12),
+    SomeData("def", 34),
+    SomeData("xyz", 56)
+  )
+
+  "RecordIdColumnByStrategy" should s"do noop with $NoUuids" in {
+    implicit val udfLib = UDFLibrary()
```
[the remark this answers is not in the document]
dk1844: Thanks, fixed.

spark-jobs/src/test/scala/za/co/absa/enceladus/common/RecordIdGenerationSuite.scala
```diff
@@ -0,0 +1,100 @@
...
+
+import com.typesafe.config.{Config, ConfigException, ConfigFactory, ConfigValueFactory}
+import org.scalatest.{FlatSpec, Matchers}
+import za.co.absa.enceladus.common.RecordIdGenerationSuite.{SomeData, SomeDataWithId}
+import za.co.absa.enceladus.utils.testUtils.SparkTestBase
+import za.co.absa.enceladus.utils.udf.UDFLibrary
+
+class RecordIdGenerationSuite extends FlatSpec with Matchers with SparkTestBase {
+
+  import RecordIdGeneration._
```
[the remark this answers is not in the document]
dk1844: My reasoning was limiting the scope as much as possible, but in the end, it can be top-level. Done.

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/interpreter/StandardizationInterpreter.scala
```diff
@@ -19,29 +19,36 @@ import org.apache.spark.sql._
...
+   *
+   * @param df                      Dataframe to be standardized
+   * @param expSchema               The schema for the df to be standardized into
+   * @param failOnInputNotPerSchema if true a discrepancy between expSchema and input data throws an exception
+   *                                if false the error is marked in the error column
+   * @param addUuids                Decides if true uuid, pseudo (always the same) is used for the
+   *                                [[Constants.EnceladusRecordId]] or if the column is not added at all [[UuidType.NoUuids]] (default).
+   */
+  def standardize(df: Dataset[Row], expSchema: StructType, inputType: String, failOnInputNotPerSchema: Boolean = false,
+                  addUuids: UuidType = UuidType.NoUuids)
```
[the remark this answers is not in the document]
dk1844: Sure, sounds better. Thanks & done.

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/interpreter/StandardizationInterpreter.scala
```diff
@@ -61,8 +68,16 @@ object StandardizationInterpreter{
 
     logger.info(s"Step 3: Clean the final error column")
     val cleanedStd = cleanTheFinalErrorColumn(std)
+
+    addUuids match {
```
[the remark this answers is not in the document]
dk1844: Logging generalized together as described above.

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
```diff
@@ -19,28 +19,34 @@ import java.io.{PrintWriter, StringWriter}
...
 import za.co.absa.enceladus.common.plugin.menas.MenasPlugin
+import za.co.absa.enceladus.common.version.SparkVersionGuard
+import za.co.absa.enceladus.common.{Constants, RecordIdGeneration}
 import za.co.absa.enceladus.conformance.interpreter.rules.ValidationException
 import za.co.absa.enceladus.conformance.interpreter.{DynamicInterpreter, FeatureSwitches, ThreeStateSwitch}
 import za.co.absa.enceladus.dao.MenasDAO
 import za.co.absa.enceladus.dao.auth.MenasCredentials
 import za.co.absa.enceladus.dao.rest.{MenasConnectionStringParser, RestDaoFactory}
 import za.co.absa.enceladus.model.Dataset
+import za.co.absa.enceladus.standardization.interpreter.StandardizationInterpreter.logger
```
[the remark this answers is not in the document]
dk1844: Thanks, fixed.

spark-jobs/src/main/scala/za/co/absa/enceladus/common/RecordIdGeneration.scala
```diff
@@ -0,0 +1,70 @@
...
+      case other => throw new ConfigException.BadValue("enceladus.recordid.generation.strategy",
+        s"Invalid value $strategyValue was encountered for id generation strategy, use one of: true, pseudo, no.")
+    }
+  }
+
+  /**
+   * The supplied dataframe `origDf` is either kept as-is (`stragegy` = [[UuidType.NoUuids]]) or appended the a column named
```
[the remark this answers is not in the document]
dk1844: Thanks, fixed.

spark-jobs/src/main/scala/za/co/absa/enceladus/common/RecordIdGeneration.scala
```diff
@@ -0,0 +1,70 @@
...
+   * @param strategy decides if and what ids will be appended to the origDf
+   * @param udfLib
+   * @return possibly updated `origDf`
+   */
+  def addRecordIdColumnByStrategy(origDf: DataFrame, strategy: UuidType)
+                                 (implicit udfLib: UDFLibrary) : DataFrame = {
+    strategy match {
+      case UuidType.NoUuids => origDf
+      case other => {
+        val udfName = if (other == UuidType.TrueUuids) UDFNames.uuid else UDFNames.pseudoUuid
```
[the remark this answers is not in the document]
dk1844: I have changed as suggested - now there are all 3 cases.

spark-jobs/src/main/scala/za/co/absa/enceladus/common/RecordIdGeneration.scala
```diff
@@ -0,0 +1,70 @@
...
+   * @param origDf dataframe to be possibly extended
+   * @param strategy decides if and what ids will be appended to the origDf
+   * @param udfLib
+   * @return possibly updated `origDf`
+   */
+  def addRecordIdColumnByStrategy(origDf: DataFrame, strategy: UuidType)
+                                 (implicit udfLib: UDFLibrary) : DataFrame = {
+    strategy match {
+      case UuidType.NoUuids => origDf
+      case other => {
```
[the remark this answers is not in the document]
dk1844: Removed.

spark-jobs/src/main/scala/za/co/absa/enceladus/common/RecordIdGeneration.scala
```diff
@@ -0,0 +1,70 @@
...
+  private val log: Logger = LoggerFactory.getLogger(this.getClass)
+
+  def getRecordIdGenerationStrategyFromConfig(conf: Config): UuidType = {
+    val strategyValue = conf.getString("enceladus.recordid.generation.strategy")
+
+    strategyValue.toLowerCase match {
+      case "true" => UuidType.TrueUuids
+      case "pseudo" => UuidType.PseudoUuids
+      case "no" => UuidType.NoUuids
+      case other => throw new ConfigException.BadValue("enceladus.recordid.generation.strategy",
```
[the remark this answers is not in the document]
dk1844: Thanks, fixed. (I have had previously used the value `other` in the message but later I removed it and forgot to adjust the case)

spark-jobs/src/main/resources/application.conf.template
```diff
@@ -17,6 +17,10 @@ enceladus.version=${project.version}
 # each can have multiple comma-separated hosts, these are used for fault-tolerance
 menas.rest.uri="http://localhost:8080,host2:9000/menas;https://localhost:8080,host2:9000/menas"
 
+# 'enceladus_record_id' with an id can be added containing either true UUID, always the same UUIDs (pseudo) or the column
+# will not be added at all. Allowed values: "true", "pseudo", "no"
+enceladus.recordid.generation.strategy="true"
```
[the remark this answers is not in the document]
dk1844: Thanks, done.

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
dk1844: Fixed, thanks

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
dk1844: Moved in the method, thanks.

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
dk1844: Moved in the method as stated elsewhere.

utils/src/main/scala/za/co/absa/enceladus/utils/udf/UDFLibrary.scala
```diff
@@ -79,8 +98,10 @@ object UDFLibrary {
   }
 
   private val errorColumnAppend = new UDF2[Seq[Row], Row, Seq[Row]] {
-    override def call(t1: Seq[Row], t2: Row) : Seq[Row] = {
+    override def call(t1: Seq[Row], t2: Row): Seq[Row] = {
       t1 :+ t2
     }
   }
+
```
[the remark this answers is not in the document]
dk1844: Removed, thanks.

spark-jobs/src/main/scala/za/co/absa/enceladus/common/RecordIdGeneration.scala
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  /**
+   * The supplied dataframe `origDf` is either kept as-is (`stragegy` = [[UuidType.NoUuids]]) or appended the a column named
+   * [[Constants.EnceladusRecordId]] with an ID for each record. These ID true UUID (`stragegy` = [[UuidType.TrueUuids]])
+   * or always the same ones for testing purposes (`stragegy` = [[UuidType.PseudoUuids]]
+   *
+   * @param origDf dataframe to be possibly extended
+   * @param strategy decides if and what ids will be appended to the origDf
+   * @param udfLib
```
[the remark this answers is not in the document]
dk1844: Description added, but..

spark-jobs/src/main/scala/za/co/absa/enceladus/common/RecordIdGeneration.scala
```diff
@@ -0,0 +1,70 @@
...
+   * @return possibly updated `origDf`
+   */
+  def addRecordIdColumnByStrategy(origDf: DataFrame, strategy: UuidType)
+                                 (implicit udfLib: UDFLibrary) : DataFrame = {
```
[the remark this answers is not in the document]
dk1844: ... I think it is the same case as with `TypeParser.standardize`. It does not use `udfLib` directly, but it requires it being implicitly supplied. My understanding of the reason was that further processing there relies on those necessary UDF being registered (which is what the UDFLibrary does in its constructor). I require it implicitly for the very same reason.

utils/src/main/scala/za/co/absa/enceladus/utils/udf/UDFLibrary.scala
dk1844: I have actually found another way to create the pseudoID values -- for the very least, it is oblivious to the division of work by executors: I calculate a hash all the columns and use this as a byte-array basis to feed the UUID creation. That way, it is completely stable no matter the way you run it.

Of course, it does directly guarantee uniqueness (the seeded-random version did not either), but I consider it to be an improvement, so it seems to me to be worth replacing what I had before. In a practical sense, if the whole rows are unique, these pseudoIDs will not collide much. From testing PoV, it could work. @Zejnilovic, @HuvarVer?

Leveraging the `monotonicallyIncreasingId` could be beneficial if we wanted to prioritize uniqueness over stability. I can implement this as well (as another mode) if it is desired but I believe quite strongly that the former hash-based solution complies better to the originally intended pseudo mode (i.e. prioritize stability over uniqueness)

Your thoughts?

spark-jobs/src/main/scala/za/co/absa/enceladus/common/RecordIdGeneration.scala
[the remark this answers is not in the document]
dk1844: No longer relevant after updates.

spark-jobs/src/main/scala/za/co/absa/enceladus/common/RecordIdGeneration.scala
```diff
@@ -0,0 +1,70 @@
...
+    case object NoUuids  extends UuidType
+  }
+
+  private val log: Logger = LoggerFactory.getLogger(this.getClass)
+
+  def getRecordIdGenerationStrategyFromConfig(conf: Config): UuidType = {
+    val strategyValue = conf.getString("enceladus.recordid.generation.strategy")
+
+    strategyValue.toLowerCase match {
+      case "true" => UuidType.TrueUuids
```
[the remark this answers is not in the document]
dk1844: Since I have changed the pseudo variant, the `seededuiid` does not make sense anymore.

However, I was trying to avoid naming the pseudo variant anything with `uuid`, because the result does not guarantee uniqueness as the 'uUid' would suggest.

So maybe `uuid`, `hashid`, (`monotonousid` if used), and `none`?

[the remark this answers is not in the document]
benedeki: Of course my values were just suggestions. 😉 I see the issues with `seededuuid`
Does `hashId` imply that the value is stable tho?

spark-jobs/src/main/scala/za/co/absa/enceladus/common/RecordIdGeneration.scala
benedeki: Just considering user expectations. Nothing specified means no key generation required - to some users.
But if it feels weird, let's not do it. In case we would get errors due to that, it's easy to extend.

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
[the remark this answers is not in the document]
benedeki: There's still a typo in _regenereration_. 😉 
And I see what you mean. Can I suggest _(re-)generation_ but fine with _regeneration_ too

utils/src/main/scala/za/co/absa/enceladus/utils/udf/UDFLibrary.scala
benedeki: That sounds really good. The hash solution 👍 
The `monotonicallyIncreasingId` would be good too. Decide if you add it to this PR or as a separate feature. (I vote for inclusion)

spark-jobs/src/main/scala/za/co/absa/enceladus/common/RecordIdGeneration.scala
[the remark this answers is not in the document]
benedeki: Fair enough.

spark-jobs/src/main/scala/za/co/absa/enceladus/common/RecordIdGeneration.scala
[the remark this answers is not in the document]
benedeki: ... no buts anymore 😄

spark-jobs/src/test/scala/za/co/absa/enceladus/common/RecordIdGenerationSuite.scala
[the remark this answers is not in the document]
benedeki: `length` makes in clear it refers to array items count. `size` is more ambiguous possibly refering memory footprint. 
And Idea marks it for me as a warning 😋

spark-jobs/src/main/scala/za/co/absa/enceladus/common/RecordIdGeneration.scala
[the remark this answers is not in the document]
dk1844: If one realizes what the hash is computed from, then it might. Maybe `stableHashId` then 😄

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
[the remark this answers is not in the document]
benedeki: I love the `Try` class. Here I find it less practical, but it's a personal opinion.
Which solution you choose I leave up to you.

spark-jobs/src/main/scala/za/co/absa/enceladus/common/RecordIdGeneration.scala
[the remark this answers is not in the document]
benedeki: I think it's good enough. Especially if described in documentation 😉

utils/src/main/scala/za/co/absa/enceladus/utils/udf/UDFLibrary.scala
benedeki: From discussion with @dk1844 a small drawback of the hash was realized - for the same row repeated in the dataframe it creates the same value, so it stops to be key per-se. But we believe for the purposes it's still good enough.
Please first of all testers confirm (@Zejnilovic @HuvarVer ), but anybody is welcomed to comment (@yruslan , @AdrianOlosutean , ... )

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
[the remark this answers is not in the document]
dk1844: 🤦‍♂️ I couldn't spot it. Fixing it.

utils/src/main/scala/za/co/absa/enceladus/utils/udf/UDFLibrary.scala
```diff
@@ -13,13 +13,18 @@
...
 import org.apache.spark.sql.{Row, SparkSession}
+import za.co.absa.enceladus.utils.error.{ErrorMessage, Mapping}
+import za.co.absa.enceladus.utils.udf.UDFNames._
+
 import scala.collection.mutable
-import UDFNames._
+import scala.util.Random
 
 
 case class UDFLibrary()(implicit val spark: SparkSession) {
```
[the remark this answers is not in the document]
dk1844: True, I have uncased the class.

spark-jobs/src/test/scala/za/co/absa/enceladus/standardization/StandardizationCsvSuite.scala
```diff
@@ -242,7 +242,7 @@ class StandardizationCsvSuite extends fixture.FunSuite with SparkTestBase with T
     assert(actual == expected)
   }
 
-  test("Test standardizing a CSV file with enabled check of maxColumns limit, permissive read mode explicit, no corrupt record") { tmpFileName =>
+  test("Test standardizing a CSV file with enabled check of maxColumns limit, permissive read mode explicit, no corrupt record") { tmpFileName => //scalastyle:ignore line.size.limit readability
```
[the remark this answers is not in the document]
dk1844: 👍 thanks for the explanation, removed.

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
dk1844: I am glad there is no confusion 😃

spark-jobs/src/main/resources/application.conf.template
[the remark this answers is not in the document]
dk1844: Sure, there could be a future change... What is the correct issue/epic to link it to?

spark-jobs/src/test/scala/za/co/absa/enceladus/common/RecordIdGenerationSuite.scala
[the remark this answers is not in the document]
dk1844: Very succinct! `length` it is. Done.

spark-jobs/src/main/scala/za/co/absa/enceladus/common/RecordIdGeneration.scala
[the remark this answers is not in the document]
dk1844: I have updated the `Readme.md` with the options and a short description `foreach` 😄 .

spark-jobs/src/test/scala/za/co/absa/enceladus/standardization/StandardizationParquetSuite.scala
```diff
@@ -387,8 +387,8 @@ class StandardizationParquetSuite extends fixture.FunSuite with SparkTestBase wi
       StructField("struct", StructType(Seq(StructField("bar", BooleanType))), nullable = false)
     )
     val schema = StructType(seq)
-    // PseudoUuids will always yield the same ids
-    val destDF = StandardizationInterpreter.standardize(sourceDF, schema, cmd.rawFormat, recordIdGenerationStrategy = UuidType.TrueUuids)
+    // True UUid will always yield the same ids
```
benedeki, line 390: Is this correct?

spark-jobs/src/main/scala/za/co/absa/enceladus/common/RecordIdGeneration.scala
dk1844: The original keys being preserved is definitely true for Conformance.

For standardization, it would get overridden. Is is undesirable in that case? @yruslan

spark-jobs/src/main/scala/za/co/absa/enceladus/common/RecordIdGeneration.scala
benedeki: Well, it's a pretty specific feature. And we require it to be set. I would make it optional and default to `None` obviously.

spark-jobs/src/main/scala/za/co/absa/enceladus/common/RecordIdGeneration.scala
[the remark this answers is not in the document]
benedeki: Sorry if being a nuisance. 😊  I don't think this is a description belongs to `README.md`.
Could you start a chapter in documentation Usage with `application.conf` parameters? Of course with this one there only for now, not all the others.

spark-jobs/src/main/resources/application.conf.template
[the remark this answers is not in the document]
benedeki: #1029

spark-jobs/src/main/scala/za/co/absa/enceladus/common/RecordIdGeneration.scala
```diff
@@ -0,0 +1,77 @@
...
+   * The supplied dataframe `origDf` is either kept as-is (`strategy` = [[IdType.NoId]]) or appended the a column named
+   * [[Constants.EnceladusRecordId]] with an ID for each record. These ID true UUID (`strategy` = [[IdType.TrueUuids]])
+   * or always the same ones for testing purposes (`strategy` = [[IdType.StableHashId]]
+   *
+   * @param origDf   dataframe to be possibly extended
+   * @param strategy decides if and what ids will be appended to the origDf
+   * @param udfLib   library that registred UDFs [[UDFNames.uuid]]
+   * @return possibly updated `origDf`
+   */
+  def addRecordIdColumnByStrategy(origDf: DataFrame, strategy: IdType)(implicit udfLib: UDFLibrary): DataFrame = {
```
benedeki, line 58: I would still have it id column name as an input parameter of this function.

spark-jobs/src/main/scala/za/co/absa/enceladus/common/RecordIdGeneration.scala
dk1844: Initially, I was under the impression since this is necessary for #1220, this feature would now be used by default. You are correct in the sense that it does not need to be, there just needs to be a dependency then (when attempting to send errors to Kafka, the ID Generation feature must be enabled. In other cases, it can be off).

Sure, I can change the default. Are we, though, sure what should be the default behavior? @benedeki @yruslan @Zejnilovic

spark-jobs/src/main/scala/za/co/absa/enceladus/common/RecordIdGeneration.scala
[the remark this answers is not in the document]
dk1844: OK, I have removed the added documentation from the `README.md` and I will create another PR to `gh-pages` instead.

spark-jobs/src/test/scala/za/co/absa/enceladus/standardization/StandardizationParquetSuite.scala
dk1844: Of course not, thanks, fixed.

utils/src/main/scala/za/co/absa/enceladus/utils/udf/UDFLibrary.scala
Zejnilovic: So making your own key using hashed rows is actually something that Hermes does when people don't supply their keys. So I like it.

The duplicities are something to look out for. I have an idea for Hermes that might be applied here, but it is not that performant. Generate your hashes, then check for duplicate rows and append something to them if there are duplicitous.

I know this idea is broad but might spark something.

`monotonicallyIncreasingId` are not needed then.

dk1844: > Generate your hashes, then check for duplicate rows and append something to them if there are duplicitous.

I briefly considered this as well, but I did not seem worthwhile. In my limited experience, the rows seem to be unique in practice.

spark-jobs/src/main/scala/za/co/absa/enceladus/common/RecordIdGeneration.scala
dk1844: Done.

spark-jobs/src/main/scala/za/co/absa/enceladus/common/RecordIdGeneration.scala
yruslan: I prefer your initial implementation slightly more. Most of our config options are mandatory with defaults specified in `application.config.template`. And this case looks like a good candidate. And I'd set to `uuid` by default.

I see `enceladus_record_id` the same way as `enceladus_info_date`, etc. So they should be generated by default.

benedeki: Considering the rest of options, yeah mandatory is probably OK.
But then I would vote for `None` as the default in template.

utils/src/main/scala/za/co/absa/enceladus/utils/udf/UDFLibrary.scala
benedeki: The beauty of the current implementation is, that we can add strategies if their need emerges. What @Zejnilovic  wrote is one possible solution for a more robust unique id. And someones till might want the `monotonicallyIncreasingId` as well.
😀

spark-jobs/src/main/scala/za/co/absa/enceladus/common/RecordIdGeneration.scala
yruslan: Yes, let's preserve the column in Standardization as well

utils/src/main/scala/za/co/absa/enceladus/utils/udf/UDFLibrary.scala
yruslan: If all columns of 2 rows are the same, we can consider them the same, and they are ok to have the same `enceladus_row_id`, especially for `stable uuid`.

yruslan: If we have hash solution, monotonic id is not needed. It has its caveats, let's not do this for now.

spark-jobs/src/main/scala/za/co/absa/enceladus/common/RecordIdGeneration.scala
```diff
@@ -46,16 +46,17 @@ object RecordIdGeneration {
   }
 
   /**
-   * The supplied dataframe `origDf` is either kept as-is (`strategy` = [[IdType.NoId]]) or appended the a column named
-   * [[Constants.EnceladusRecordId]] with an ID for each record. These ID true UUID (`strategy` = [[IdType.TrueUuids]])
-   * or always the same ones for testing purposes (`strategy` = [[IdType.StableHashId]]
+   * The supplied dataframe `origDf` is either kept as-is (`strategy` = [[IdType.NoId]]) or appended an ID column
+   * with an ID for each record. These ID true UUID (`strategy` = [[IdType.TrueUuids]]) or always the same ones for testing
+   * purposes (`strategy` = [[IdType.StableHashId]]
```
benedeki, line 51:
```suggestion
   * The supplied dataframe `origDf` is either kept as-is (`strategy` = [[IdType.NoId]]) or has a column appended
   * with an (presumably) unique value for each record. These are true UUIDs (`strategy` = [[IdType.TrueUuids]]) or
   * values always the same for the same row, mainly for testing purposes (`strategy` = [[IdType.StableHashId]]
```

Zejnilovic:
If this is a final implementation, please give me today to run a test or two. I will jump on it right now.

spark-jobs/src/main/scala/za/co/absa/enceladus/common/RecordIdGeneration.scala
dk1844: OK, adjusted to keep existing `enceladus_record_id` in Standardization, too.

Zejnilovic: That would be cool. people could come then with their own Kyes

spark-jobs/src/test/scala/za/co/absa/enceladus/standardization/StandardizationParquetSuite.scala
```diff
@@ -18,7 +18,7 @@ package za.co.absa.enceladus.standardization
 import java.util.UUID
 
 import org.apache.spark.sql.DataFrame
-import org.apache.spark.sql.types._
+import org.apache.spark.sql.types.{StructField, _}
```
Zejnilovic, line 21: Is there a reason for StructField to be out?

dk1844: No, not really. My import optimization somehow didn't catch it. Fixed.

dk1844:
Hopefully "one of the last updates": I have change the true UUID generation to leverage Spark SQL's built-in `uuid()` - so, no UDF is necessary after all. Thanks @yruslan for the tip!

PR #1321, opened 2020-05-04
1320 Add breadcrumbs to Docs

Zejnilovic:
- add new structure for docs sub-pages
  - add category
  - move sub-pages to a folder named same as the main page
- add breadcrumbs for docs
- move version selection to float left
- update create_docs.rb script to respect the multi-level docs.


Zejnilovic:
I almost forgot. You might google that breadcrumb plugins for Jekyll exist, but none are allowed by Github's "safe Jekyll mode".

PR #1327, opened 2020-05-06
1294 config docs

### LINKED ISSUE #1294: Add 'enceladus_record_id' during processing

dk1844:
_Usage > Configuration_ page was started, brief info about general config and overriding it has been added (mainly for the sake of documenting the `enceladus.recordId.generation.strategy`)

closes #1294

### DIFF
--- a/_docs/1.0.0/usage-schema.md
+++ b/_docs/1.0.0/usage-schema.md
@@ -139,7 +139,7 @@ You provide *Schema* to **Standardization** in a JSON file:
 }
 ```
 
-Example of data adhering to the above schema can be found [here](https://github.com/AbsaOSS/enceladus/blob/develop/standardization/src/test/scala/za/co/absa/enceladus/standardization/samples/TestSamples.scala).
+Example of data adhering to the above schema can be found [here](https://github.com/AbsaOSS/enceladus/blob/master/spark-jobs/src/test/scala/za/co/absa/enceladus/standardization/samples/TestSamples.scala).
 
 Automatically added columns
 ---------------------------
--- a/_docs/1.0.0/usage-schema.md
+++ b/_docs/1.0.0/usage-schema.md
@@ -139,7 +139,7 @@ You provide *Schema* to **Standardization** in a JSON file:
 }
 ```
 
-Example of data adhering to the above schema can be found [here](https://github.com/AbsaOSS/enceladus/blob/master/spark-jobs/src/test/scala/za/co/absa/enceladus/standardization/samples/TestSamples.scala).
+Example of data adhering to the above schema can be found [here](https://github.com/AbsaOSS/enceladus/blob/v1.7.1/standardization/src/test/scala/za/co/absa/enceladus/standardization/samples/TestSamples.scala).
 
 Automatically added columns
 ---------------------------
--- a/_docs/2.0.0/usage/config.md
+++ b/_docs/2.0.0/usage/config.md
@@ -1,0 +1,42 @@
+---
+layout: docs
+title: Usage - Configuration
+version: '2.0.0'
+categories:
+    - '2.0.0'
+    - usage
+redirect_from: /docs/usage/config
+---
+
+Configuration
+============
+
+<!-- toc -->
+- [Configuration](#configuration)
+  - [Intro](#intro) 
+  - [General options](#general-options)
+<!-- tocstop -->
+
+Intro
+-----------
+
+This page describes the usage of configuration of _Standardization_ and _Conformance_.
+There are number of default options that 
+[Project's readme](https://github.com/AbsaOSS/enceladus/blob/develop/README.md) documents
+this page describes the configuration values stored in `spark-jobs`'s `application.conf` (or its
+[template](https://github.com/AbsaOSS/enceladus/blob/develop/spark-jobs/src/main/resources/application.conf.template)).
+These values can be overridden using the `-D` property values as in:
+```shell
+spark-submit --conf "spark.driver.extraJavaOptions= -Dkey1=value1 -Dkey2=value2" ...
+```
+
+General options
+-----------
+
+|            Config path                   |                 Description                                                                                                              |
+|------------------------------------------|----------------------------------------------|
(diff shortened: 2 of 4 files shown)


_docs/2.0.0/usage/config.md
```diff
@@ -0,0 +1,42 @@
...
+  - [Intro](#intro) 
+  - [General options](#general-options)
+<!-- tocstop -->
+
+Intro
+-----------
+
+This page describes the usage of configuration of _Standardization_ and _Conformance_.
+There are number of default options that 
+[Project's readme](https://github.com/AbsaOSS/enceladus/blob/develop/README.md) documents
```
benedeki, line 25: I think it would be better to refer the `master`

_docs/2.0.0/usage/config.md
```diff
@@ -0,0 +1,42 @@
...
+These values can be overridden using the `-D` property values as in:
+```shell
+spark-submit --conf "spark.driver.extraJavaOptions= -Dkey1=value1 -Dkey2=value2" ...
+```
+
+General options
+-----------
+
+|            Config path                   |                 Description                                                                                                              |
+|------------------------------------------|----------------------------------------------|
```
benedeki, line 37: **Suggestion:** I would add a column describing the possible values.
list of enumerables, "any string", "integer between 1 and 100"

benedeki, line 37: **Suggestion 2**: Actually for enumeration, I would add its own line for every option (with no value in `Config path` with the meaning described in `Description`.

_docs/2.0.0/usage/config.md
```diff
@@ -0,0 +1,42 @@
...
+  - [General options](#general-options)
+<!-- tocstop -->
+
+Intro
+-----------
+
+This page describes the usage of configuration of _Standardization_ and _Conformance_.
+There are number of default options that 
+[Project's readme](https://github.com/AbsaOSS/enceladus/blob/develop/README.md) documents
+this page describes the configuration values stored in `spark-jobs`'s `application.conf` (or its
```
benedeki, line 26: The sentence looks like a (unwanted) merge of two sentences. Split or reformulate please.

_docs/2.0.0/usage/config.md
dk1844: I have thought about it first, but then I saw https://github.com/AbsaOSS/enceladus/blame/gh-pages/_docs/2.0.0/usage/schema.md#L150 where `develop` is also referred to.

So, I'd better change that, too, I guess.

_docs/2.0.0/usage/config.md
dk1844: Split, thanks.

_docs/2.0.0/usage/config.md
dk1844: Good idea, done.

dk1844: I have done something like that.

_docs/2.0.0/usage/config.md
dk1844: Updated.

benedeki: Yeah, it was an oversight. Thanks for fixing

_docs/2.0.0/usage/config.md
```diff
@@ -0,0 +1,44 @@
...
+
+General options
+-----------
+
+|            Config path                   | Possible value(s) |        Description        |
+|------------------------------------------|-------------------|---------------------------|
+| `enceladus.recordId.generation.strategy` |  `uuid` (default) | `enceladus_record_id` column will be added and will contain a UUID `String` for each row. |
+|                                          |  `stableHashId`   | `enceladus_record_id` column will be added and populated with an always-the-same `Int` hash (Murmur3-based, for testing). |
+|                                          |  `none`           | no column will be added to the output. |
+| `menas.rest.uri`                         |  string with URLs | Comma-separated list of URLs where Menas will be looked for. E.g.: `http://example.com/menas1,http://domain.com:8080/menas2` |
```
benedeki, line 41: **Comment: ** It really can be multiple URLs? Didn't know that. Cool.

Zejnilovic: This is our road to High Availability

_docs/1.0.0/usage-schema.md
```diff
@@ -139,7 +139,7 @@ You provide *Schema* to **Standardization** in a JSON file:
 }
 ```
 
-Example of data adhering to the above schema can be found [here](https://github.com/AbsaOSS/enceladus/blob/develop/standardization/src/test/scala/za/co/absa/enceladus/standardization/samples/TestSamples.scala).
+Example of data adhering to the above schema can be found [here](https://github.com/AbsaOSS/enceladus/blob/master/spark-jobs/src/test/scala/za/co/absa/enceladus/standardization/samples/TestSamples.scala).
```
Zejnilovic, line 142: Could I be a *** and ask you to update the URL to `https://github.com/AbsaOSS/enceladus/blob/v1.7.1/standardization/src/test/scala/za/co/absa/enceladus/standardization/samples/TestSamples.scala`

I checked. There is no more master or develop URL issues in the whole gh-pages branch

dk1844: Sure, done.

No need to be calling yourself any colorful names, though. Neither ***, nor ****** ***** ****, or even !@#$%^& 🤣

benedeki: @Zejnilovic  Yeah, we know them all anyway 😈 
@dk1844 Thanks for being the [good boy scout](https://deviq.com/boy-scout-rule/) 🏅

Zejnilovic: @benedeki more like [this](https://youtu.be/dQw4w9WgXcQ)

PR #1328, opened 2020-05-06
Feature/1316 indexes for mongodb performance

### LINKED ISSUE #1261: Introduce more MongbDB indexes
## Background
Currently, many queries to MongoDB fo full collection scans. For big collections such as `run` it creates a lot of pressure for the MongoDB cluster.

## Feature
Investigate production logs and find bottlenecks. Check if corresponding queries can be optimized or new indexes can be added to ease the pressure on MongoDB.

AdrianOlosutean:
The proposed created indexes
Closes #1261

### DIFF
--- a/migrations/src/main/scala/za/co/absa/enceladus/migrations/migrations/MigrationToV1.scala
+++ b/migrations/src/main/scala/za/co/absa/enceladus/migrations/migrations/MigrationToV1.scala
@@ -39,6 +39,7 @@ object MigrationToV1 extends MigrationBase with CollectionMigration with JsonMig
   createIndex("dataset", Seq(IndexField("name", ASC), IndexField("version", ASC)), unique = true)
   createIndex("schema", Seq(IndexField("name", ASC), IndexField("version", ASC)), unique = true)
   createIndex("mapping_table", Seq(IndexField("name", ASC), IndexField("version", ASC)), unique = true)
+
   createIndex("run", Seq(IndexField("dataset", ASC), IndexField("datasetVersion", ASC)))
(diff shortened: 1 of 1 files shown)


AdrianOlosutean:
@DzMakatun Do you think there are other indexes which can be added?

migrations/src/main/scala/za/co/absa/enceladus/migrations/migrations/MigrationToV1.scala
```diff
@@ -39,6 +39,8 @@ object MigrationToV1 extends MigrationBase with CollectionMigration with JsonMig
   createIndex("dataset", Seq(IndexField("name", ASC), IndexField("version", ASC)), unique = true)
   createIndex("schema", Seq(IndexField("name", ASC), IndexField("version", ASC)), unique = true)
   createIndex("mapping_table", Seq(IndexField("name", ASC), IndexField("version", ASC)), unique = true)
+
+  createIndex("run", Seq(IndexField("dataset", ASC), IndexField("datasetVersion", ASC)))
```
benedeki, line 43: I am not sure how MongoDB work, but isn't this index a subset/sub-index of the one right bellow? (with just little smaller selectivity). 🤔

benedeki: Yup, according to this [article](https://engineering.mixmax.com/blog/improving-mongo-performance-by-managing-indexes), the new one is redundant and MongoDB engine should be able to use the other index whenever the new indexed might seem useful.

dk1844: Aaaah, I did not realize it, but it actually makes sense 😊 .

PR #1330, opened 2020-05-07
Add support for secure Kafka in cluster mode

### LINKED ISSUE #1329: Remove Kafka security parameters from command line
## Background
Currently in order to connect to secure Kafka we need to provide 
```ini
javax.net.ssl.trustStore="..redacted..."
javax.net.ssl.trustStorePassword="..redacted..."
javax.net.ssl.keyStore="..redacted..."
javax.net.ssl.keyStorePassword="..redacted..."
java.security.auth.login.config="..redacted..."
```
as system properties (e.g. passing it as `-D...`.

## Feature
- [ ] Allow these properties to be specified in config `application.conf`.
- [ ] Use `System.setProperty()` to copy these configuration options to the system properties.
- [ ] While on it, as much config options as possible from `-D...` in helper scripts.

yruslan:
Closes #1329


plugins-builtin/src/main/scala/za/co/absa/enceladus/plugins/builtin/utils/ConfigUtils.scala
```diff
@@ -0,0 +1,72 @@
...
+   * @param key   Configuration key.
+   */
+  def setSystemPropertyStringFallback(conf: Config, key: String): Unit = {
+    if (System.getProperty(key) == null && conf.hasPath(key)) {
+      System.setProperty(key, conf.getString(key))
+    }
+  }
+
+  /**
+   * Puts a file location from a configuration to the system properties ensusing the file exists.
```
Zejnilovic, line 40:
```suggestion
   * Puts a file location from a configuration to the system properties ensuring the file exists.
```

plugins-builtin/src/main/scala/za/co/absa/enceladus/plugins/builtin/utils/ConfigUtils.scala
```diff
@@ -0,0 +1,72 @@
...
+   *
+   * A file provided can be at an absolute path (for client mode), e.g. /home/aabb/file.conf,
+   * but other locations are investigated as well in this order.
+   *
+   * - The exact path pased (/home/aabb/file.conf).
```
Zejnilovic, line 45:
```suggestion
   * - The exact path passed (/home/aabb/file.conf).
```

plugins-builtin/src/main/scala/za/co/absa/enceladus/plugins/builtin/utils/ConfigUtils.scala
```diff
@@ -0,0 +1,72 @@
...
+        System.setProperty(key, conf.getString(key))
+      } else {
+        log.warn(s"File does not exist: ${pathFileName.toString}")
+        val fileNameInCurDir = Paths.get(pathFileName).getFileName
+        if (Files.exists(fileNameInCurDir)) {
+          log.info(s"File exists: ${fileNameInCurDir.toString}")
+          System.setProperty(key, fileNameInCurDir.toString)
+        } else {
+          log.error(s"File does not exist: ${fileNameInCurDir.toString}")
+          val sparkRoot = System.getProperty("spark.dist.files.location")
```
Zejnilovic, line 65: How is this `sparkRoot` used? Is it for the future?

scripts/bash/run_enceladus.sh
```diff
@@ -289,7 +294,7 @@ fi
 
 SPARK_CONF="--conf spark.logConf=true"
 
-JVM_CONF="spark.driver.extraJavaOptions=-Dmenas.rest.uri=$MENAS_URI -Dstandardized.hdfs.path=$STD_HDFS_PATH \
+JVM_CONF="spark.driver.extraJavaOptions=-Dstandardized.hdfs.path=$STD_HDFS_PATH \
```
Zejnilovic, line 297: Why was the rest URI removed?

plugins-builtin/src/main/scala/za/co/absa/enceladus/plugins/builtin/utils/ConfigUtils.scala
```diff
@@ -0,0 +1,72 @@
...
+        log.info(s"File exists: ${pathFileName.toString}")
+        System.setProperty(key, conf.getString(key))
+      } else {
+        log.warn(s"File does not exist: ${pathFileName.toString}")
+        val fileNameInCurDir = Paths.get(pathFileName).getFileName
+        if (Files.exists(fileNameInCurDir)) {
+          log.info(s"File exists: ${fileNameInCurDir.toString}")
+          System.setProperty(key, fileNameInCurDir.toString)
+        } else {
+          log.error(s"File does not exist: ${fileNameInCurDir.toString}")
```
dk1844, line 64: I can see the logic that the file is looked up in the current directory, so it is queried just by name. The logging messages, however, could include the path, so the log-reader knows (especially when the file was not found), that the _current directory_ is what they expect it is.

(also with the log.info 2 rows above)

yruslan: Hmm... `Paths.get(pathFileName).getFileName` strips the path so all other usage should include only the file name.
Sorry, I might not completely following. How would you write it?

plugins-builtin/src/main/scala/za/co/absa/enceladus/plugins/builtin/utils/ConfigUtils.scala
yruslan: 👍 Haha, good catch. Will fix. This is an artifact from the previous implementations.
The thing is files sent with `--files <list>` to `spark-submit` are seems to be sent to different locations on executors than to the driver. So I removed that approach in later commits.

scripts/bash/run_enceladus.sh
yruslan: It is now placed in `application.conf` in the same directory as scripts. I'm gonna add the mentioning of in it to the `.template`.

scripts/bash/enceladus_env.template.sh
```diff
@@ -56,11 +52,26 @@ DEFAULT_DEPLOY_MODE="client"
...
+# Kafka security
+# Path to jaas.config
+#JAAS_CLIENT="-Djava.security.auth.login.config=/path/jaas.config"
+#JAAS_CLUSTER="-Djava.security.auth.login.config=jaas_cluster.config"
+
+APPLICATION_PROPERTIES_CLIENT="-Dconfig.file=/absolute/path/application.conf"
+APPLICATION_PROPERTIES_CLUSTER="-Dconfig.file=application.conf"
+
+# Files to send when running in cluster mode (comma separated)
+ENCELADUS_FILES="/absolute/path/application.conf#application.conf"
```
yruslan, line 64: Here is the file from where Menas URIs will be taken.

scripts/bash/run_enceladus.sh
yruslan: Just noticed, it is already there. Added a comment to that line.

Zejnilovic: Are we now expecting users to provide their own app.conf? Or is it ok to use APPLICATION_PROPERTIES_CLIENT as well?

plugins-builtin/src/main/scala/za/co/absa/enceladus/plugins/builtin/utils/ConfigUtils.scala
dk1844:
```suggestion
          log.error(s"File does not exist: ${pathFileName.toString}")
```

yruslan: Makes sense, Thanks!

scripts/bash/run_enceladus.sh
yruslan: Yes, we are asking users to provide their own `application.conf` and put all options there. 

Later we can remove our `application.template` logic and use `reference.conf` as you suggested.

yruslan: But it is not mandatory. Just easier. The older helper scripts will still work.

Instead of passing many options as `-D...`  to command line we just specify the location of the environment-specific `application.properties`. No credentials will now be passed to command line.

Zejnilovic: Just a quick question. Did we try playing with `-Dconfig.resource=path/to/application.conf` option?

yruslan: Exactly. With the difference that we will be using `-Dconfig.file=path/to/application.conf`.

We need to distinguish between client and cluster mode. In client mode, the absolute path is used. In cluster mode, the config is sent via `--files ...` and read from the current directory.

Zejnilovic: ok, I might be slow on the uptake here, so just to sum up. Using `config.file` in cluster mode does not propagate the config to the cluster. It need HDFS location.

yruslan: Not exactly. Let me explain.
1. `-Dconfig.file=/path/file` is a TypeSafe Config option saying where to look for the config.
2. When running in cluster mode the driver is started on a random cluster node, not on the edge node where `spark-submit` is called.
3. In order to put the configs to the node where the driver will be running `--files` option is used in `spark-submit`.
4. Files sent that way will appear on the driver's Yarn container in the current path. This includes `application.conf` sent this way.

yruslan: > Are we now expecting users to provide their own app.conf? Or is it ok to use APPLICATION_PROPERTIES_CLIENT as well?

I think I've misunderstood the question. Users will use the command line as always. But we will have per-environment `appllication.config` placed in the directory with the helper scripts.

Zejnilovic: Great explanation. Thank you. Will not click `resolve conversation` as somebody might be interested as I was, but please treat it as resolved.

scripts/bash/enceladus_env.template.sh
yruslan: It's a file 'alias': https://stackoverflow.com/a/49866757/1038282

plugins-builtin/src/main/scala/za/co/absa/enceladus/plugins/builtin/utils/ConfigUtils.scala
```diff
@@ -0,0 +1,71 @@
...
+   * @param key   Configuration key.
+   */
+  def setSystemPropertyFileFallback(conf: Config, key: String): Unit = {
+    if (System.getProperty(key) == null && conf.hasPath(key)) {
+      val pathFileName = conf.getString(key)
+      if (Files.exists(Paths.get(pathFileName))) {
+        log.info(s"File exists: $pathFileName")
+        System.setProperty(key, conf.getString(key))
+      } else {
+        log.warn(s"File does not exist: $pathFileName")
```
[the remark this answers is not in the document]
yruslan: I want to be able to trace where the file was looked by looking at logs. This helps analyze Kafka's security issues. But I can change to `info` here. Will keep `error` on the final file location.

plugins-builtin/src/main/scala/za/co/absa/enceladus/plugins/builtin/utils/ConfigUtils.scala
```diff
@@ -0,0 +1,71 @@
...
+   * - The file in the current directory, when spark-submit uses --files with aliases (file.conf).
+   *
+   * @param conf  A configuration.
+   * @param key   Configuration key.
+   */
+  def setSystemPropertyFileFallback(conf: Config, key: String): Unit = {
+    if (System.getProperty(key) == null && conf.hasPath(key)) {
+      val pathFileName = conf.getString(key)
+      if (Files.exists(Paths.get(pathFileName))) {
+        log.info(s"File exists: $pathFileName")
```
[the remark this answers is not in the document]
yruslan: Since Kakfa security files can be taken from different locations I want to be able to see what was used even if it succeded.

scripts/bash/run_enceladus.sh
yruslan: I removed this since I removed it from our environments. And I want these scripts here to reflect as much as possible to what was actually deployed.

plugins-builtin/src/main/scala/za/co/absa/enceladus/plugins/builtin/utils/ConfigUtils.scala
```diff
@@ -0,0 +1,71 @@
...
+  private val log = LoggerFactory.getLogger(this.getClass)
+
+  /**
+   * Puts a config key and value to the system properties if it is not defined there.
+   *
+   * @param conf  A configuration.
+   * @param key   Configuration key.
+   */
+  def setSystemPropertyStringFallback(conf: Config, key: String): Unit = {
+    if (System.getProperty(key) == null && conf.hasPath(key)) {
```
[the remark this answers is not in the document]
yruslan: Yeah, it is funny. But `null` is justified here - old Java API :)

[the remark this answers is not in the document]
benedeki: Of course I haven't  doubted the presence of `null`, I saw it's valid here. :-)

scripts/bash/enceladus_env.template.sh
benedeki: Need trick 👍 Maybe a comment explaining?

plugins-builtin/src/main/scala/za/co/absa/enceladus/plugins/builtin/utils/ConfigUtils.scala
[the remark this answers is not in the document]
benedeki: Fair enough

scripts/bash/run_enceladus.sh
Zejnilovic: > No sure why to remove the `application.conf.template` (?). I think it is a good example how to set the configuration values (and what are the configurations anyway).

reference.conf is a "template" when using lightbends typesafe config. If we move forward to using that we should follow their standards.

plugins-builtin/src/main/scala/za/co/absa/enceladus/plugins/builtin/common/mq/kafka/KafkaConnectionParams.scala
```diff
@@ -48,6 +49,8 @@ object KafkaConnectionParams {
   def fromConfig(conf: Config, clientIdKey: String, topicNameKey: String): KafkaConnectionParams = {
     validate(conf, clientIdKey, topicNameKey)
 
+    SecureKafka.setSecureKafkaProperties(conf)
```
benedeki, line 52: Is this still needed here if used in the `main` functions?

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
```diff
@@ -55,6 +56,10 @@ object DynamicConformanceJob {
   private val menasBaseUrls = MenasConnectionStringParser.parse(conf.getString("menas.rest.uri"))
 
   def main(args: Array[String]) {
+    // This should be the first thing the app does to make secure Kafka work with our CA.
+    // After Spring activates JavaX, it will be too late.
+    SecureKafka.setSecureKafkaProperties(conf)
```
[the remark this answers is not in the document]
yruslan: I just want to be explicit of what we are doing.

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
```diff
@@ -26,16 +26,16 @@ import org.apache.spark.sql.{Column, DataFrame, DataFrameReader, SparkSession}
 import org.slf4j.LoggerFactory
 import za.co.absa.atum.AtumImplicits
 import za.co.absa.atum.core.Atum
+import za.co.absa.enceladus.common.RecordIdGeneration.{IdType, _}
```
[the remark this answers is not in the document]
yruslan: Interesting...

plugins-builtin/src/main/scala/za/co/absa/enceladus/plugins/builtin/common/mq/kafka/KafkaConnectionParams.scala
yruslan: Yeah, I guess not. I will remove it.

More to it. This initialization affects the whole application and all frameworks that use SSL, so it is not the place to do it in plugins.

yruslan: But also there is a rationale to keep it. The method is idempotent. If the system properties are already set, it won't overwrite them.
But I guess these plugins can be used only in `spark-jobs`, so removing it is still the better option.

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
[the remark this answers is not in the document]
benedeki: It could be actually in a class that both `StandardizationJob`  and `DynamicConformanceJob` can extend. This would move us toward shared code base of the two applications.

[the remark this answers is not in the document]
yruslan: Yeah, we can do this as part of the task of combining Standardization and Conformance codebase.

scripts/bash/run_enceladus.sh
dk1844: I'd like to point out that there is a noteworthy downside to using `reference.conf` - that is, it makes it harder to make sure that you have overridden a default property from `reference.conf` in `application.conf` (e.g. mistyping a key will result in the default being used without warning). By using "only" `application.conf`, the config reading would fail fast instead.

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/StandardizationJob.scala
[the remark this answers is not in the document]
dk1844: Happened to me, too a few times. 😢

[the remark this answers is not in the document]
Zejnilovic: I think it is the new update of IntelliJ

PR #1335, opened 2020-05-12
#1333 Change the way 'enceladus_info_date' is generated for streaming jobs

### LINKED ISSUE #1331: Add support for long running jobs
## Background
Currently, Streaming Conformance writes the date of ingestion start to `enceladus_info_date`. For long-running jobs (e.g. Kafka to Kafka), this date should change when crossing midnight.

## Feature
Set `enceladus_info_date` to the processing date for long-running jobs.

### LINKED ISSUE #1333: Information date column support for streaming jobs
## Background
Streaming data may contain an event date or timestamp. We should use it if it is available to populate 'enceladus_info_date'.

## Feature
Add a configuration option to specify a column name to be used to derive `enceladus_info_date` from. It can be `date` or `timestamp` or `string`. If it is a string, users need to be able to specify a pattern for that field.

yruslan:
* If transformer.hyperconformance.report.date is specified, it will be used
 * If transformer.hyperconformance.event.timestamp.column is specified, the event date derived from it will be used
 * Otherwise the processing date will be used.

Closes #1331 and closes #1333

### DIFF
--- a/spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/HyperConformance.scala
+++ b/spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/HyperConformance.scala
@@ -15,6 +15,9 @@
 
 package za.co.absa.enceladus.conformance
 
+import java.text.SimpleDateFormat
+import java.util.Date
+
 import org.apache.commons.configuration2.Configuration
 import org.apache.spark.SPARK_VERSION
 import org.apache.spark.sql.functions.lit
@@ -112,6 +115,7 @@ object HyperConformance extends StreamTransformerFactory with HyperConformanceAt
     implicit val cmd: ConfCmdConfig = ConfCmdConfig(
       datasetName = conf.getString(datasetNameKey),
       datasetVersion = conf.getInt(datasetVersionKey),
+      reportDate = new SimpleDateFormat(ReportDateFormat).format(new Date()), // Still need a report date for mapping table patterns
       reportVersion = Option(getReportVersion(conf)),
       menasCredentialsFactory = menasCredentialsFactory,
       performanceMetricsFile = None,
--- a/spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/HyperConformance.scala
+++ b/spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/HyperConformance.scala
@@ -20,7 +20,7 @@ import java.util.Date
 
 import org.apache.commons.configuration2.Configuration
 import org.apache.spark.SPARK_VERSION
-import org.apache.spark.sql.functions.lit
+import org.apache.spark.sql.functions._
 import org.apache.spark.sql.{DataFrame, SparkSession}
 import org.slf4j.{Logger, LoggerFactory}
 import za.co.absa.enceladus.common.Constants._
@@ -58,7 +58,7 @@ class HyperConformance (implicit cmd: ConfCmdConfig,
 
     DynamicInterpreter.interpret(conformance, streamData)
       .withColumnIfDoesNotExist(InfoDateColumn, infoDateColumn)
-      .withColumnIfDoesNotExist(InfoDateColumnString, lit(cmd.reportDate))
+      .withColumnIfDoesNotExist(InfoDateColumnString, date_format(infoDateColumn,"yyyy-MM-dd"))
       .withColumnIfDoesNotExist(InfoVersionColumn, lit(reportVersion))
   }
 
--- a/spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/HyperConformance.scala
+++ b/spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/HyperConformance.scala
@@ -20,8 +20,9 @@ import java.util.Date
 
 import org.apache.commons.configuration2.Configuration
 import org.apache.spark.SPARK_VERSION
-import org.apache.spark.sql.functions.{lit, to_date}
-import org.apache.spark.sql.{DataFrame, SparkSession}
+import org.apache.spark.sql.functions.{current_timestamp, lit, to_date, col}
+import org.apache.spark.sql.types.DateType
+import org.apache.spark.sql.{Column, DataFrame, SparkSession}
 import org.slf4j.{Logger, LoggerFactory}
 import za.co.absa.enceladus.common.Constants._
 import za.co.absa.enceladus.common.version.SparkVersionGuard
@@ -33,7 +34,8 @@ import za.co.absa.hyperdrive.ingestor.api.transformer.{StreamTransformer, Stream
 
 class HyperConformance (implicit cmd: ConfCmdConfig,
                         featureSwitches: FeatureSwitches,
-                        menasBaseUrls: List[String]) extends StreamTransformer {
+                        menasBaseUrls: List[String],
+                        reportDateCol: Column) extends StreamTransformer {
   val log: Logger = LoggerFactory.getLogger(this.getClass)
 
   @throws[IllegalArgumentException]
@@ -53,7 +55,7 @@ class HyperConformance (implicit cmd: ConfCmdConfig,
     val conformance = dao.getDataset(cmd.datasetName, cmd.datasetVersion)
 
     DynamicInterpreter.interpret(conformance, streamData)
-      .withColumnIfDoesNotExist(InfoDateColumn, to_date(lit(cmd.reportDate), ReportDateFormat))
+      .withColumnIfDoesNotExist(InfoDateColumn, reportDateCol)
       .withColumnIfDoesNotExist(InfoDateColumnString, lit(cmd.reportDate))
       .withColumnIfDoesNotExist(InfoVersionColumn, lit(reportVersion))
   }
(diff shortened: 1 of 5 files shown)


spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/HyperConformance.scala
```diff
@@ -155,11 +159,18 @@ object HyperConformance extends StreamTransformerFactory with HyperConformanceAt
...
+  private def getReportDate(conf: Configuration): Column = {
     if (conf.containsKey(reportDateKey)) {
-      conf.getString(reportDateKey)
+      val reportDate = conf.getString(reportDateKey)
+      log.info(s"Information date: Explicit from the job configuration = $reportDate")
+      to_date(lit(reportDate), ReportDateFormat)
+    } else if (conf.containsKey(eventTimestampColumnKey)) {
+      val eventTimestampColumnName = conf.getString(eventTimestampColumnKey)
+      log.info(s"Information date: Derived from the event time column = $eventTimestampColumnName")
+      col(eventTimestampColumnName).cast(DateType)
```
Zejnilovic, line 170: Can we be sure this would pass? Is it going to fail fast with a proper message?

yruslan: If the column does not exist or cannot be converted to date, it will fail on a Spark error.

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/HyperConformance.scala
```diff
@@ -127,6 +129,8 @@ object HyperConformance extends StreamTransformerFactory with HyperConformanceAt
       .setBroadcastStrategyMode(Always)
       .setBroadcastMaxSizeMb(0)
 
+    implicit val reportDateCol: Column = getReportDate(conf)
```
AdrianOlosutean, line 132: Why has it become implicit?

yruslan: It is a new parameter. The old one was taken from `cmdConfig`. But now we need the info date value to be a `Column`, so I pass it as a separate parameter.

For batch jobs info date is a constant across the job. For streaming, it became an expression so that it can change depending on event or processing time.

I've noticed that all params to `HyperConformance` constructor are implicit and made this one as well.  Probably not a perfect reason. I can make it normal if you think it is better this way.

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/HyperConformance.scala
yruslan: You probably want me to add validations of the existence of the column and its type?

Yeah, I should do it. Also, we will need a pattern if it is a string type.

Zejnilovic: Yes. The thing is normal spark error would be lost (i think) and would look like normal std error. At least catch it and throw error with better message.

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/HyperConformance.scala
AdrianOlosutean: Then I think it may make sense to be implicit

yruslan: I had to make a specific factory type to provide info date strategy. So now that the type is very specific, it justifies implicit even more.

Since it is a bad practice to make a common type parameter implicit. Like `Boolean`, `Int`, and similarly `Column`.

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/datasource/PartitioningUtils.scala
```diff
@@ -78,7 +78,12 @@ object PartitioningUtils {
...
-  private def validateReportDate(reportDate: String): Unit = {
+  /**
+   * Validates the port date to be in tyhe expected format: 'yyyy-MM-dd'.
+   *
+   * @param reportDate          A report date string
+   */
+  def validateReportDate(reportDate: String): Unit = {
     val reportDateRegEx = """\d\d\d\d-\d\d-\d\d""".r
 
     if (!reportDateRegEx.pattern.matcher(reportDate).matches()) {
```
dk1844, line 89:
```suggestion
    reportDate match {
      case reportDateRegEx() =>
      case _ => throw new IllegalArgumentException("...")
    }
```

Also, maybe it would be useful if the regex was outside of the method, so it doesn't need to be recreated every time. it is used. 

Either private, if planned to be used only here and in this way, or maybe as `"""(\d{4})-(\d{2})-(\d{2})"""` with capturing groups if reused from somewhere else (in that case, the matching would have to change to ```reportDateRegEx(y, m, d)```).

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/streaming/InfoDateFactory.scala
```diff
@@ -0,0 +1,88 @@
...
+  val log: Logger = LoggerFactory.getLogger(this.getClass)
+
+  def getFactoryFromConfig(conf: Configuration): InfoDateFactory = {
+    if (conf.containsKey(reportDateKey)) {
+      val reportDate = conf.getString(reportDateKey)
+      log.info(s"Information date: Explicit from the job configuration = $reportDate")
+      new InfoDateLiteralFactory(reportDate)
+    } else if (conf.containsKey(eventTimestampColumnKey)) {
+      val eventTimestampColumnName = conf.getString(eventTimestampColumnKey)
+      val eventTimestampPattern = conf.getString(eventTimestampPatternKey, defaultEventTimestampPattern)
```
dk1844, line 80: I wonder if it is worth to somehow validate the supplied pattern beforehand in order to detect possibly invalid TS pattern before spark processing is started (aim: to fail fast).

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/datasource/PartitioningUtils.scala
yruslan: Sure, thanks. Will fix it since I'm there. But the piece of code you added the comment to is not part of this PR.

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/streaming/InfoDateFactory.scala
yruslan: I'm open to suggestions

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/datasource/PartitioningUtils.scala
dk1844: I know, sorry about that 😈

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/streaming/InfoDateFactory.scala
Zejnilovic: Is it even possible to validate patterns? I think it's like email validation. Too much time, too much effort, low returns.

dk1844: Well, since `to_timestamp` uses the pattern rules of `java.text.SimpleDateFormat` which does not provide a public method to check pattern validity, the easiest way would be to construct a `new SimpleDateFormat(pattern)` to see if it fails with an exception.

The more I think about it though, the less I am convinced that it is worth it. There is not a lot that can fail while the pattern is read (non-date chars are treated as normal chars anyway) and the key part where it would most probably fail (the `to_timestamp` application) cannot be known beforehand without the data.

All in all, failing fast is hard to do here. :(

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/datasource/PartitioningUtils.scala
yruslan: Nw, it is a good suggestion 😄

yruslan: Actually, now that I'm looking at it, I prefer the current implementation. The matcher implementation has an empty case. But this opinion is not strong. If people prefer it as a matcher, I can change.

AdrianOlosutean: Unless it will be used to extract those parts too, I think it would not make much of a difference

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/HyperConformance.scala
```diff
@@ -20,7 +20,7 @@ import java.util.Date
 
 import org.apache.commons.configuration2.Configuration
 import org.apache.spark.SPARK_VERSION
-import org.apache.spark.sql.functions.lit
+import org.apache.spark.sql.functions._
```
Zejnilovic, line 23: There are these imports at the top that don't seem to be used. Can we remove them?
```
import java.util.Date
```

yruslan: Now they are used again 😄

yruslan:
Confirmed from @kevinwallimann yesterday that this works as expected.

PR #1343, opened 2020-05-16
1341 Proposition for table updates

Zejnilovic:
- new config table dynamically generated
- alignment of heading styles
- remove of top-level heading from md files that have title


_docs/2.0.0/usage/errcol.md
```diff
@@ -1,25 +1,22 @@
...
-  - [Notes](#notes)
+- [Table Of Content](#table-of-content)
+- [Description](#description)
+- [Error Types](#error-types)
+- [Notes](#notes)
 <!-- tocstop -->
 
-Description
------------
+## Description
```
benedeki, line 18: Why changing the style of headers?

_includes/config_options.html
```diff
@@ -0,0 +1,27 @@
...
+  <thead>
+    <tr>
+      <th scope="col">Config Path</th>
+      <th scope="col">Possible Value(s)</th>
+      <th scope="col">Description</th>
+    </tr>
+  </thead>
+  <tbody>
+  {% for entry in tableFile %}
+    {% assign current_name = entry.name %}
```
benedeki, line 15: Nice. Can the _Code Path_ lines be in style `<code></code>` automatically?

_docs/2.0.0/usage/errcol.md
Zejnilovic: https://cirosantilli.com/markdown-style-guide#headers

multiple markdown style guides say not to mix header types. Some even prohibit it completely as it can break some markdown engines (usually found in chats)

Zejnilovic: Subjectively to keep it all in one style

_includes/config_options.html
Zejnilovic: I don't understand the question.

Zejnilovic: Are you asking about the usage of backticks? As it is HTML it does not translate a backtick into `<code>` so code needs to be used explicitly.

_docs/2.0.0/usage/errcol.md
Zejnilovic: Also, H1 heading is not needed as it is the title of the page.

benedeki: Fair enough.

PR #1346, opened 2020-05-18
Feature/1319 support specifying csv characters as codes

### LINKED ISSUE #1319: Add support for specifying CSV characters as codes
## Background
When Standardization is ran from a scripting environment sometimes recodes characters depending on the current encoding. This causes failures of jobs that use higher ASCII characters as delimiters, for instance.

## Feature
Add the ability to specify delimiter, quote and escape character as a UNICODE expression instead of the character itself.

From UNICODE table: https://en.wikipedia.org/wiki/List_of_Unicode_characters

## Example [Optional]
So that you can specify the character like this:
`--delimiter U+00A1` or `--delimiter u00a1` or `--delimiter 0161`
instead of this
`--delimiter ¡`
(`--delimiter ¡` should still be supported)

Same for `--escape` and `--quote`.

Check format of the option and decode it into the UNICODE `String` (changed from `Char` since an empty string is allowed).

We can have a character interpreter utility function that converts a command-line character argument to the actual character or an empty string.

AdrianOlosutean:
Closes #1319


utils/src/main/scala/za/co/absa/enceladus/utils/unicode/ParameterConversion.scala
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+package za.co.absa.enceladus.utils.unicode
+
+object ParameterConversion {
+
+  implicit class UnicodeValueString(s: String) {
+    val unicodeElement = """[uU]?\+?([0-9a-fA-F]{4})""".r
+
+    def includingUnicode: String = s match {
```
Zejnilovic, line 23: Not sure where we stand now. Do we implement "no" in this match as well for empty string? @yruslan

utils/src/main/scala/za/co/absa/enceladus/utils/unicode/ParameterConversion.scala
```diff
@@ -0,0 +1,28 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.enceladus.utils.unicode
+
+object ParameterConversion {
+
+  implicit class UnicodeValueString(s: String) {
+    val unicodeElement = """[uU]?\+?([0-9a-fA-F]{4})""".r
```
Zejnilovic, line 21: there is a POSIX class `\p{XDigit}`. What are our thoughts on using that? It is pre-defined and has rules that can be easily referenced.?

spark-jobs/src/test/scala/za/co/absa/enceladus/standardization/csv/EnhancedStandardizationCsvSuite.scala
```diff
@@ -0,0 +1,81 @@
...
+class EnhancedStandardizationCsvSuite extends fixture.FunSuite with CsvFileFixture {
+
+  import za.co.absa.enceladus.utils.implicits.DataFrameImplicits.DataFrameEnhancements
+
+  // A field containing the delimiter with the escape has to be enclosed in specified quotes
+  private val content: String =
+    """1¡2¡3¡4¡5
+      |Text1¡Text2¡Text3¡10¡11
+      |Text10"Add¡Text11¡$Text/¡12$¡100¡200
+      |$Text/¡15$¡Text16¡$Text/¡17$¡1000¡2000"""
```
yruslan, line 30: Maybe also include a line with `$Text/$18$` to see if escaping a quote works.

spark-jobs/src/test/scala/za/co/absa/enceladus/standardization/csv/NoneValueStandardizationCsvSuite.scala
```diff
@@ -0,0 +1,96 @@
...
+      |+----------+--------+---------+----+----+----------------------------------+
+      |
+      |"""
+  test("Test none for quote") { tmpFileName =>
+    val args = ("--dataset-name SpecialChars --dataset-version 1 --report-date 2019-07-23 --report-version 1 " +
+      "--menas-auth-keytab src/test/resources/user.keytab.example " +
+      "--raw-format csv --header false " +
+      "--charset ISO-8859-1 --delimiter ¡ --csv-quote none").split(" ")
+
+    val expected = noneQuoteResult.stripMargin.replace("\r\n", "\n")
```
yruslan, line 42: You can move `.stripMargin.replace("\r\n", "\n")` to the definition of `noneQuoteResult` saving an aditional operation.

utils/src/main/scala/za/co/absa/enceladus/utils/unicode/ParameterConversion.scala
yruslan: Looks good. Or I may miss something.
After a discussion yesterday we decided to go for "none" as the keyword for the empty string.
This case is handled in `NoneValueString`. If a string parameter does not support empty strings, but support Unicode, `includingUnicode` matcher will be used with `NoneValueString`. But we decided to support empty string for eny char parameter for now, even for the delimiter.

yruslan:
While answering Saša's comment I realized that this feature needs documentation.

Guys, we add documentation as a separate PR to gh-pages, right?

utils/src/main/scala/za/co/absa/enceladus/utils/unicode/ParameterConversion.scala
Zejnilovic: My comment was before the discussion happened.

Zejnilovic:
> While answering Saša's comment I realized that this feature needs documentation.
> 
> Guys, we add documentation as a separate PR to gh-pages, right?

Next PR

README.md
```diff
@@ -217,15 +217,17 @@ The list of additional options available for running Standardization:
...
 | --row-tag **tag**                    | A row tag if the input format is `xml`.                                                                                                            |
 | --header **true/false**              | Indicates if in the input CSV data has headers as the first row of each file.                                                                      |
-| --delimiter **character**            | Specifies a delimiter character to use for CSV format. By default `,` is used.                                                                     |
-| --csv-quote **character**            | Specifies a character to be used as a quote for creating fields that might contain delimiter character. By default `"` is used.                    |
-| --csv-escape **character**           | Specifies a character to be used for escaping other characters. By default '&#92;' (backslash) is used.                                            |
+| --delimiter **character**            | Specifies a delimiter character to use for CSV format. By default `,` is used. <sup>*                                                              |
```
dk1844, line 220:
```suggestion
| --delimiter **character**            | Specifies a delimiter character to use for CSV format. By default `,` is used. <sup>*</sup>                                                              |
```

dk1844: and both occurrences below

utils/src/main/scala/za/co/absa/enceladus/utils/unicode/ParameterConversion.scala
yruslan: Hey, how widely is this supported? Can I try this on https://regex101.com/ ?

yruslan: I'd also add an explicit end of string so that '0AA01' won't match as '0AA0'.
```suggestion
    val unicodeElement = """[uU]?\+?([0-9a-fA-F]{4})$""".r
```

Zejnilovic: > Hey, how widely is this supported? Can I try this on https://regex101.com/ ?

I think you cannot. It is a POSIX class, meaning that bash, scala, and other languages will support it. Regex101 I think has no such classes pre-defined.

utils/src/main/scala/za/co/absa/enceladus/utils/unicode/ParameterConversion.scala
```diff
@@ -0,0 +1,35 @@
...
+    def includingUnicode: String = s match {
+      case unicodeElement(hex) => Integer.parseInt(hex, 16).toChar.toString
+      case other => other
+    }
+  }
+
+  implicit class NoneValueString(s: String) {
+    def includingNone: String = s.toLowerCase() match {
+      case "none" => ""
+      case _ => s
```
dk1844, line 32: I think the case should also check that the string is exactly 1 character long (or at most 1 char long - do we allow empty string or the only possibility to express empty char is by `none`?) and fail otherwise. And the class/method name could also reflect this, IMHO.

... but thinking about it, using
```someString.includingUnicode``` would then go unchecked, if the length check was only in `includedNone`

It seems that these two are always used together. If that is always going to be the case, maybe a better suggestion is to allow API like:

```scala
someString.toUnicodeOrChar(includingNone = true)
```
That could work for:
`a`, `u1234`, `U+1234`, `none` and fail for `abc`, `U123`, ...

Just a suggestion, though.

AdrianOlosutean: The idea is that it is still currently possible to specify empty string. The none keyword is used in case the usage is from a scripting environment where ''(double single quote) would be ignored. 
Furthermore, Spark checks if there is at most one character, so values like `abc` would fail from Spark side. 

About using them together, I am open to suggestions. My idea was to keep them separate in case  some further usage would use just the None part without coupling it to the unicode part.

dk1844: What you say makes sense. On the other hand, to make the `includingNone` truly versatile, its signature should in my opinion probably be
```
implicit class NoneValueString(s: String) {
    def includingNone: Option[String] = s.toLowerCase match {
      case "none" => None
      case _ => Some(s)
}
```
Nevertheless, all these are just suggestions. The current code is acceptable and it looks like works as expected. 👍

AdrianOlosutean: The thing is that if None is passed as a csv option, then it means that the default values will be used, which will be `,`(comma) for delimiter,  `\`(backslash) for escape and `"` for quote. The value that we want to pass in the None case is `""`(empty string) or `u0000`

PR #1350, opened 2020-05-19
#1234 export file fix, menas attachment version fix

### LINKED ISSUE #1234: Export file feature should provide .avsc format for Avro schemas
## Background
We have added support of avro format import to Menas. Now it is time to add avro file format for the export file feature. Currently, people would get `.cob` as default in case of avro schema.

## Feature
1. Add avro format to export file feature
2. Add a general format. Like `.txt` if no format is matched

## Example
When I export spark struck file I get `.json`
When I export cobol copybook I get `.cob`
When I export avro schema original file it should be `avsc`

## Additional Info
Check Menas's `ui/components/schema/schemaDetail.controller.js` method  `_getFileExtension`

### LINKED ISSUE #1347: Menas attachment versioning is off by one
## Describe the bug
A schema version in Menas and the internal Menas attachment version is off by one (e.g. for schema version 2, the attachment version is 3)

## To Reproduce

1. Create a new schema. Its version will be "1".
2. Upload/Load schema file for it, version of the schema will be "2".
3. Attempting to export the schema file using "Export File" will fail, because Menas will look for attachment v2, but only attachment v3 would exist. 
4. Moreover, if you upload another version of the file and attempt to "Export File", you will get the older file instead of the latest uploaded.

## Expected behaviour
 - everytime there is a schema file attached (v2+), the UI should allow to "Export File" and get the correct result (e.g. the latest file that was just uploaded)

## Additional context
This could most probably be fixed by changing:
```scala
refVersion = version + 1
```
to 
```scala
refVersion = version
```
in `SchemaController.scala`.

However, this can also lead to unmatched expectations with our users (they may have exported schemas they thought were the lastest, but weren't and so on).

dk1844:
- AVSC is always saved and exported with the same mime-type: `application/vnd.schemaregistry.v1+json`
 - means attachment version is now in line with the schema version (shown in Menas)

### A possible naive remapping script:
// edit: will need to create something more convoluted, see comments.
The following MongoDB script can be used to remap means attachments version, provided there are no holes (i.e. means without and with the bug was not used over the same persistence and the ids are not e.g. `2, 3, (missing 4 here) 5, 6`):

- `2, 3, 4` -> `2, 3, 4` (no change, all should be already correct)
- `3, 4, 5, 6` -> `2, 3, 4, 5` (all -1, all should be correct)
- `2, 3, 5, 6` -> `2, 3, 5, 6` (no change, does not work for version sequences with holes, so the second half of the version would stay incorrect)

```javascript
db.getCollection('attachment_v1').aggregate(
    [
        {
            $match: {}
        },
        {
            $group: {
                _id: "$refName",
                minVersion: { $min: "$refVersion" }
            }
        },
        { "$out": "tmp_indicators" }
    ]
);
     
var bulkOps = db.tmp_indicators.find().map(function (doc) { 
    // for every minimal version larger than 3, lower each menas attachment version by 1
    // if no records are missing, the correct lowerst should be v2, the incorrect lowest version should be v3
    if (doc.minVersion >= NumberInt(3)) {
        return { 
            "updateMany": { 
                "filter": { "refName": doc._id } ,              
                "update": { "$inc": { "refVersion": NumberInt(-1) } } 
            }         
        };
    }
});

// bulkOps contain undefined for ops that did not satisfy the minVersion condition
var bulkOpsCleaned = bulkOps.filter(function(d) { return d != undefined})

db.getCollection('attachment_v1').bulkWrite(bulkOpsCleaned);
db.tmp_indicators.drop()
```

It would make sense to discuss what the state of the persistence with users might be and if we need a smarter script - e.g. checking for the holes, too. Or will the above do?

fixes #1234 fixed #1347

### DIFF
--- a/menas/src/main/scala/za/co/absa/enceladus/menas/controllers/SchemaController.scala
+++ b/menas/src/main/scala/za/co/absa/enceladus/menas/controllers/SchemaController.scala
@@ -63,7 +63,7 @@ class SchemaController @Autowired()(
       try {
         val url = new URL(remoteUrl)
         val connection = url.openConnection()
-        val mimeType = SchemaController.avscConentType // only AVSC is expected to come from the schema registry
+        val mimeType = SchemaController.avscContentType // only AVSC is expected to come from the schema registry
         val fileStream = Source.fromInputStream(connection.getInputStream)
         val fileContent = fileStream.mkString
         fileStream.close()
@@ -104,7 +104,7 @@ class SchemaController @Autowired()(
 
     // for avro schema type, always force the same mime-type to be persisted
     val mime = if (schemaType == SchemaType.Avro) {
-     SchemaController.avscConentType
+     SchemaController.avscContentType
     } else {
       file.getContentType
     }
@@ -166,5 +166,5 @@ class SchemaController @Autowired()(
 }
 
 object SchemaController {
-  val avscConentType = "application/vnd.schemaregistry.v1+json"
+  val avscContentType = "application/vnd.schemaregistry.v1+json"
 }
--- a/menas/src/main/scala/za/co/absa/enceladus/menas/controllers/SchemaController.scala
+++ b/menas/src/main/scala/za/co/absa/enceladus/menas/controllers/SchemaController.scala
@@ -58,12 +58,12 @@ class SchemaController @Autowired()(
                        @RequestParam name: String,
                        @RequestParam format: Optional[String]): CompletableFuture[Option[Schema]] = {
 
-    val schemaType = SchemaType.fromOptSchemaName(format)
+    val schemaType: SchemaType.Value = SchemaType.fromOptSchemaName(format)
     val (url, fileContent, mimeType) = {
       try {
         val url = new URL(remoteUrl)
         val connection = url.openConnection()
-        val mimeType = connection.getContentType
+        val mimeType = SchemaController.avscConentType // only AVSC is expected to come from the schema registry
         val fileStream = Source.fromInputStream(connection.getInputStream)
         val fileContent = fileStream.mkString
         fileStream.close()
@@ -80,7 +80,7 @@ class SchemaController @Autowired()(
 
     val menasFile = MenasAttachment(refCollection = RefCollection.SCHEMA.name().toLowerCase,
       refName = name,
-      refVersion = version + 1,
+      refVersion = version + 1,  // version is the current one, refVersion is the to-be-created one
       attachmentType = MenasAttachment.ORIGINAL_SCHEMA_ATTACHMENT,
       filename = url.getFile,
       fileContent = fileContent.getBytes,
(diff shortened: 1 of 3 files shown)


Zejnilovic:
A question about logic: You have attachments for schema, and if you are missing version, you subtract 1 from each following? 

What if the version rise was dues to description change?

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/SchemaController.scala
```diff
@@ -58,12 +58,12 @@ class SchemaController @Autowired()(
...
                        @RequestParam format: Optional[String]): CompletableFuture[Option[Schema]] = {
 
-    val schemaType = SchemaType.fromOptSchemaName(format)
+    val schemaType: SchemaType.Value = SchemaType.fromOptSchemaName(format)
     val (url, fileContent, mimeType) = {
       try {
         val url = new URL(remoteUrl)
         val connection = url.openConnection()
-        val mimeType = connection.getContentType
+        val mimeType = SchemaController.avscConentType // only AVSC is expected to come from the schema registry
```
Zejnilovic, line 66:
```suggestion
        val mimeType = SchemaController.avscContentType // only AVSC is expected to come from the schema registry
```

Or did you mean conent as in agreement?

menas/src/main/scala/za/co/absa/enceladus/menas/services/AttachmentService.scala
```diff
@@ -36,8 +36,8 @@ class AttachmentService @Autowired()(attachmentMongoRepository: AttachmentMongoR
   def uploadAttachment(attachment: MenasAttachment): Future[Completed] = {
     chooseRepository(attachment.refCollection).getLatestVersionValue(attachment.refName).flatMap {
       case Some(version) =>
-        val updated = attachment.copy(refVersion = version + 1)
-        attachmentMongoRepository.create(updated)
+        // the attachment version should already increased by one from the currently latest by the caller
```
Zejnilovic, line 39: `version` on the line above not used anymore, also

```suggestion
        // the attachment version should already be increased by one from the currently latest by the caller
```

dk1844: Thanks, fixed.

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/SchemaController.scala
dk1844: Of course not, this is a typo, thanks. Fixed.

dk1844:
> A question about logic: You have attachments for schema, and if you are missing version, you subtract 1 from each following?
> 
> What if the version rise was dues to description change?
Thanks for pointing that out, this complicates things even a bit more :(

Hi, the logic is quite rudimentary:
In case the lowest attachment version is 3 for a given means attachment by `refName`, all attachment version are lowered by 1 for that `refName`.

The reason to look for the lowest version 3 is that such an attachment versioning is likely wrong (should start at 2, if the first steps were to create a schema and upload the schema file). There are cases where this would **not have the desired effect**:
 - the schema was started on a non-buggy means version (first schema attachment is v2) and somewhere along the lines, the buggy version was deployed. At that moment, the latter part of attachment versioning is off by one, but it is hard to detect (because the description changes also create holes in the attachment version sequence -- so the version hole is not conclusive evidence of the boundary)
 - the schema was started on a non-buggy means version and immediately changed in its description and the first schema attachment uploaded after that, the first Menas attachment version would be v3, but correctly. This time, the script would mess things up.

So I am now torn **if there is a truly a foolproof way to determine which Menas attachments are wrong and which are correct**. Any ideas, @Zejnilovic, or anyone?

In theory, for each deployment, one could manually identify when the buggy version was deployed and write a script that would only take after-the-fact schema versions (schemas have timestamps) and their attachments into consideration -- lower there attachments by one. That, however, seems quite labor-intensive and error-prone to my taste.

dk1844:
Well, there might be way... It's a longshot, but:
[The Mongo ObjectId contains a timestamp!](https://docs.mongodb.com/manual/reference/method/ObjectId.getTimestamp/#ObjectId.getTimestamp)

Such a fact could help find the actual corresposponding `schema version <-> attachment version` relationships and identify the problems. The timestamps might not be the same, but close enough. I will think about if this is doable. How much do we need such a script, btw?

Zejnilovic:
> So I am now torn if there is a truly a foolproof way to determine which Menas attachments are wrong and which are correct.

Could we ["calculate"](https://docs.mongodb.com/manual/reference/method/ObjectId.getTimestamp/) creation time from `_id` and cross-check that with when 1.2.2 was deployed and everything after that can be lowered by one?

Zejnilovic:
> How much do we need such a script, btw?

Let's try and count the number of occurrences

PR #1355, opened 2020-05-22
#1347 remapping script added

dk1844:
Adding the remapping script from https://github.com/AbsaOSS/enceladus/issues/1347#issuecomment-631988013 discussion to codebase.

Relates to #1347 

@yruslan @Zejnilovic


scripts/mongo/1347-attachment-version-off-by-one-mapping-script.js
```diff
@@ -0,0 +1,65 @@
+// Directly related to Enceladus issue #1347
```
Zejnilovic, line 1: Could we add that this is 1.2.X to 2.5 problem?

dk1844: Done.

Issue #1353, opened 2020-05-20
Add support for ASCII encoded mainframe files


yruslan:
## Background
Currently, we don't support Cobrix's
`.option("encoding", "ascii")` 
and
`.option("ascii_charset", "ISO-8859-1") `
and
`.option("string_trimming_policy", "both")`
to specify the encoding of mainframe files.

## Feature
Add support for `--cobol-encoding`, `--charset`, `--cobol-trim` for `cobol` format.

yruslan:
Release notes
Added support for loading ASCII fixed-width files using `cobol` as raw data format.

PR #1356, opened 2020-05-22
#1353 Add support for ASCII for 'cobol' raw format.

yruslan:
This adds the support of fixed-width text files when using 'cobol' as raw format.
The only thing that needs to be provided is a copybook. But it is easy for such formats:
```
       01  RECORD.
           05  A        PIC X(1).
           05  B        PIC X(5).
           05  C        PIC X(10).
           05  FILLER   PIC X(1).
```
Each field defines just the size.

In order to load ASCII files you need to use `--cobol-encoding ascii` (`ebcdic` is Cobrix default). You can specify the charset the normal way using `--charset`.

Since fixed-with text files contain a lot of leading or trailing spaces, users want to control how these spaces will be handled. Use `--cobol-trimming-policy` for that. It can be `none`, `left`, `right` or `both` (Cobrix default).

### DIFF
--- a/spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/StdCmdConfig.scala
+++ b/spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/StdCmdConfig.scala
@@ -255,6 +255,16 @@ object StdCmdConfig {
           } else {
             failure("The --cobol-encoding option is supported only for COBOL data format")
           })
+
+      opt[String]("cobol-trimming-policy").optional().action((value, config) => {
+        config.copy(cobolOptions = cobolSetTrimmingPolicy(config.cobolOptions, value))
+      }).text("Specify string trimming policy for mainframe files (none, left, right, both)")
+        .validate(value =>
+          if (rawFormat.isDefined && rawFormat.get.equalsIgnoreCase("cobol")) {
+            success
+          } else {
+            failure("The --cobol-trimming-policy option is supported only for COBOL data format")
+          })
     }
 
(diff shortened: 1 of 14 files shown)


spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/StdCmdConfig.scala
```diff
@@ -245,6 +245,26 @@ object StdCmdConfig {
...
+            success
+          } else {
+            failure("The --cobol-encoding option is supported only for COBOL data format")
+          })
+
+      opt[String]("cobol-trimming-policy").optional().action((value, config) => {
+        config.copy(cobolOptions = cobolSetTrimmingPolicy(config.cobolOptions, value))
+      }).text("Specify string trimming policy for mainframe files (none, left, right, both)")
+        .validate(value =>
+          if (rawFormat.isDefined && rawFormat.get.equalsIgnoreCase("cobol")) {
```
Zejnilovic, line 263: won't you check if the value is one of `Seq("none", "left", "right", "both")`?

yruslan: Cobrix will validate this anyway. And if the set of trimming options is changed (new option is added for instance) in Cobrix we won't need to make changes in Enceladus.

Having said that, I can add a test for a bogus trimming policy.

Zejnilovic: I am sorry for treating Cobrix as a random 3rd party library at the time of review 🙇‍♂️

PR #1358, opened 2020-05-25
#1309 Add attempt_id to runs

### LINKED ISSUE #1309: Add attempt id to runs
## Background
When a Spark job is submitted to YARN in cluster mode it may have multiple attempts. Attempts have the same app_id and different attemptId. (see [Spark History API docs] https://spark.apache.org/docs/latest/monitoring.html#rest-api)
As a result, in Menas, the later attempts will create new runs for std OR rewrite run data for conformance.
Currently Menas logs std_application_id and conform_application_id.
For monitoring, it is important to know the attempt id as well.

## Feature
Store Spark attempt id to Run documents in Menas.

## Proposed Solution 
Solution Ideas 

1. Check if attempt id exists and append it to metadata using Atum

AdrianOlosutean:
Closes #1309 
Attempt id to Atum metadata in Standardization and Conformance

### DIFF
--- a/spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
+++ b/spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
@@ -119,8 +119,6 @@ object DynamicConformanceJob {
       postProcessingService.onSaveOutput(conformedDf) // all enabled postProcessors will be run with the std df
     } finally {
 
-      spark.sparkContext.applicationAttemptId
-        .foreach(attemptId => Atum.setAdditionalInfo("conf_spark_attempt_id" -> attemptId))
       MenasPlugin.runNumber.foreach { runNumber =>
         val name = cmd.datasetName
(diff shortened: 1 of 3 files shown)


spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/DynamicConformanceJob.scala
```diff
@@ -119,6 +119,8 @@ object DynamicConformanceJob {
       postProcessingService.onSaveOutput(conformedDf) // all enabled postProcessors will be run with the std df
     } finally {
 
+      spark.sparkContext.applicationAttemptId
+        .foreach(attemptId => Atum.setAdditionalInfo("conf_spark_attempt_id" -> attemptId))
```
yruslan, line 123: You can put this inside `PerformanceMetricTools.addJobInfoToAtumMetadata(...)` and use the metadata prefix (std/conf).
If Spark Session is not passed to `addJobInfoToAtumMetadata()`, you can extend this method with `attemptId: String`.

Issue #1250, opened 2020-04-02
URL validation in Menas UI


dk1844:
## Background
In the UI of Menas, the user might want to input a URL, that we want to be checked. The initial use case is for providing URL of a schema to a schema registry endpoint, but there could potentially be other uses.

This issue is directly inspired by the discussion at https://github.com/AbsaOSS/enceladus/pull/1243#discussion_r401604104

## Feature
Implement a validation mechanism of a URL to be usable in the Menas UI.
The validation should allow public domains as well as general testing locations as `http://localhost/somepath`.

The solution should be accompanied by a series of unit tests that will check that it works for the expected usecases in Menas as outlined above.

## Proposed Solution [Optional]
Solution Ideas 
1. use the Regex posted by @Zejnilovic in the linked discussion
2. use any of the https://mathiasbynens.be/demo/url-regex
3.

dk1844:
Release notes
"Schema load from URL" - URL validation added - supports localhost, domain names, IP v4 and v6.

PR #1359, opened 2020-05-25
#1250 URL validation

dk1844:
Essentially validated by Rx pattern:
```javascript
'((([a-z\\d]([a-z\\d-]*[a-z\\d])*)\\.)+[a-z]{2,}|'+ // domain name
'localhost|'+ // OR localhost
'(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)|'+ // OR IPv4 address
'(([0-9a-f]{0,4}:){1,7}[0-9a-f]{0,4}))' + // OR IPv6 address (simplified version, allows ::1 and short other short forms)
'(\\:\\d+)?' + // port
'(\\/[-a-z\\d%_.~+]*)*'+ // path
'(\\?[;&a-z\\d%_.~+=-]*)?'+ // query string
'(\\#[-a-z\\d_]*)?$' // fragment locator
```

url-validation-example

After discussion with @Zejnilovic, I have not added unit tests, because the UnitTest flow setup would be out of the scope of this issue.

### DIFF
--- a/menas/ui/components/schema/schemaDetail.controller.js
+++ b/menas/ui/components/schema/schemaDetail.controller.js
@@ -245,9 +245,6 @@ sap.ui.define([
         this._eventBus.publish("schemas", "list");
         // nav back to info
         this.byId("schemaIconTabBar").setSelectedKey("info");
-        // clear error if present
-        this.byId("remoteUrl").setValueState(sap.ui.core.ValueState.None);
-        this.byId("remoteUrl").setValueStateText("");
       } else if (status === 400) {
         const errorMessage = ajaxResponse.responseJSON.message;
         const errorMessageDetails = errorMessage ? `\n\nDetails:\n${errorMessage}` : "";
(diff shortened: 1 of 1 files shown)


menas/ui/components/schema/schemaDetail.controller.js
```diff
@@ -334,10 +335,19 @@ sap.ui.define([
       }
     },
 
+    httpUrlRx: new RegExp('^(https?:\\/\\/)?'+ // protocol
```
Zejnilovic, line 338: Tried a few usual URLs. The only that didn't work is `::1`

dk1844: That seems bearable.

dk1844: OK, I have updated the IPv6 part a bit to allow shorted forms and thus also ::1.

Issue #1098, opened 2019-12-06
Add dynamic resource allocation options to run scripts


yruslan:
## Background
Currently, run scripts allow specifying only fixed (static) resources for a job.

## Feature
We need to extend run scripts to allow specifying dynamic resource allocation options.

The exact details to be discussed and put here:
```
placeholder
```

DzMakatun:
To enable DRA the following is needed:
 - Enable external spark_shuffle service in YARN (note that it is version specific)
- Set spark.dynamicAllocation.enabled to true
- Add 'spark.shuffle.service.enabled true' to Spark defaults

Parameters for spark-submit:
--conf spark.shuffle.service.enabled=true \
--conf spark.dynamicAllocation.enabled=true \
--conf spark.dynamicAllocation.minExecutors=0 \
--conf spark.dynamicAllocation.maxExecutors=8 \

Note: should be careful with maxExecutors. Spark adds a new executors whenever there are tasks in the queue. Typically one task is created per HDFS partition (with a minimum of 2). E.g. for 90 GB input file 800 tasks will appear in a queue, and if  maxExecutors is not set appropriately (Integer.MAX_VALUE by default), it will take over entire cluster in a matter of few seconds.

additionally DRA uses the following params as a unit of allocation:
--executor-cores 1 \
--executor-memory 1G \
----num-executors 1 \ (initial number of executors in DRA)

Also (investigation ongoing):
--conf spark.files.openCostInBytes=134217728 \ (could help to avoid spanning too many tiny tasks)

DzMakatun:
resolved in #1362

DzMakatun:
Release notes:
[Allows to set Dynamic Resource Allocation as a default for Enceladus jobs. DRA scales resource allocation to the actual size of the job (input size). Compared to fixed allocation, this makes a) small jobs more resource efficient b) large jobs faster. It reduces the effort for resource configuration on per case basis. 

The updated scripts also ensure adaptive execution is enabled together with DRA. Upon shuffle operations, it adjusts the data partitioning to target sizes i.g. HDFS blocks sizes. This improves the efficiency of resource usage and solves the small files issue (when output was split into too many tiny files).

DRA gets auto-disabled, and spark submit falls back to Spark defaults (preserving other custom parameters) when:
- --num-executors is set. We assume, that this parameter is set when user knows how many resources are needed exactly. Remove this parameter, to enable DRA back. 
[...]

Issue #670, opened 2019-07-16
Small files problem


lokm01:
Figure out how to do automatic repartitioning of output data in order to match the HDFS block size as close as possible => tf. avoid the small files problem

DzMakatun:
This issue is addressed in # 1098.
The main reasons of excessive partitioning are:
- Mismatch between data size and number of total allocated cores (Addressed with DRA)
- Default partitioning in shuffles, resulting in 200 partitions. (Addressed with Adaptive execution , which is now included in default DRA config)

DzMakatun:
resolved in #1362

DzMakatun:
Release notes:
[Overpartitioning (too many tiny files) of output data is resolved by configuring DRA and adaptive execution.]

PR #1362, opened 2020-05-29
#1098 Add DRA to scripts

DzMakatun:
Adds DRA to scrips with safety checks (AbsaOSS/enceladus#1098 ):
- DRA 
- external shuffle service  -
- adaptive execution 
are set together (based on recent study). This allows to avoid overallocation when shuffles with default partitioning (200 partitions) is used. Also adaptive execution addresses the Small files problem #670

Sanity checks:
DRA gets auto-disabled, and spark submit falls back to Spark defaults (preserving other custom parameters) when:
- Max number of executors is not set. By default max executors is set to infinity. In production it may result in small job taking over entire cluster. Therefore, it is checked for save configuration.
- --num-executors is set. We assume, that this parameter is set when user knows how many resources are needed exactly. In order not to everride their setup with DRA, we disable it. Note: Using --num-executors with DRA is possible in Spark. It sets the starting number of executors. However, it is more practical to start with 0 and allocate upon need, especially when the job has some initialization overhead at start.


scripts/bash/run_enceladus.sh
```diff
@@ -28,6 +28,14 @@ EXECUTOR_CORES="$DEFAULT_EXECUTOR_CORES"
 NUM_EXECUTORS="$DEFAULT_NUM_EXECUTORS"
 FILES="$ENCELADUS_FILES"
 
+# DRA related defaults
+DRA_ENABLED="$STD_DEFAULT_DRA_ENABLED"
```
yruslan, line 32:
```suggestion
DRA_ENABLED="$DEFAULT_DRA_ENABLED"
```

DzMakatun: thanks!

scripts/bash/enceladus_env.template.sh
```diff
@@ -46,8 +57,20 @@ CONF_DEFAULT_DRIVER_MEMORY=""
...
 
+# Dynamic Resource Allocation
+# also enables external shuffle service and adaptive execution for consistent setup
+CONF_DEFAULT_DRA_ENABLED=false
+# max executors limit is a required parameter
+CONF_DEFAULT_DRA_MAX_EXECUTORS=4
+
+CONF_DEFAULT_DRA_MIN_EXECUTORS=0
+CONF_DEFAULT_DRA_ALLOCATION_RATIO=0.5
+CONF_DEFAULT_ADAPTIVE_TARGET_POSTSHUFFLE_INPUT_SIZE=134217728
```
benedeki, line 72: This number seems little random. Can you add a comment please, why this number? (128MB?)

DzMakatun: It is 1 HDF block by default. The idea is to match Spark partitions to HDFS blocks. It would allow less load to name nodes and potentially reduce amount of repartitioning happening in Spark.

benedeki: Oh, I see. Would be a nice educative comment in the script 😉

Issue #1306, opened 2020-04-27
Catalyst optimizer hangs streaming jobs when many conformance rules are used


yruslan:
## Describe the bug
When too many conformance rules are used (around 12), Streaming Conformance jobs hung.
It is related to the execution plan traversal logic when optimizing consecutive projections.

Related issues in batch: #190, #413 
This also logged in Spark Jira: [SPARK-28090](https://issues.apache.org/jira/browse/SPARK-28090).

Here is how the driver's thread dump looks like:
```
org.apache.spark.sql.catalyst.trees.TreeNode$$anonfun$2.apply(TreeNode.scala:259)
org.apache.spark.sql.catalyst.trees.TreeNode$$anonfun$2.apply(TreeNode.scala:259)
org.apache.spark.sql.catalyst.trees.CurrentOrigin$.withOrigin(TreeNode.scala:70)
org.apache.spark.sql.catalyst.trees.TreeNode.transformDown(TreeNode.scala:258)
org.apache.spark.sql.catalyst.trees.TreeNode$$anonfun$transformDown$1.apply(TreeNode.scala:264)
org.apache.spark.sql.catalyst.trees.TreeNode$$anonfun$transformDown$1.apply(TreeNode.scala:264)
org.apache.spark.sql.catalyst.trees.TreeNode.org$apache$spark$sql$catalyst$trees$TreeNode$$mapChild$2(TreeNode.scala:298)
org.apache.spark.sql.catalyst.trees.TreeNode$$anonfun$4$$anonfun$apply$13.apply(TreeNode.scala:357)
scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:234)
scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:234)
scala.collection.immutable.List.foreach(List.scala:392)
scala.collection.TraversableLike$class.map(TraversableLike.scala:234)
scala.collection.immutable.List.map(List.scala:296)
org.apache.spark.sql.catalyst.trees.TreeNode$$anonfun$4.apply(TreeNode.scala:357)
org.apache.spark.sql.catalyst.trees.TreeNode.mapProductIterator(TreeNode.scala:187)
org.apache.spark.sql.catalyst.trees.TreeNode.mapChildren(TreeNode.scala:327)
org.apache.spark.sql.catalyst.trees.TreeNode.transformDown(TreeNode.scala:264)
[...]

PR #1366, opened 2020-05-29
#1307 Integration test for Streaming Conformance

### LINKED ISSUE #1307: Add an end-to-end Kafka source&sink streaming conformance integration test
## Background
An end-to-end streaming conformance test will help to catch streaming-only issues.
Such as the Catalyst optimizer issue #1306  

## Feature
Add an end-to-end Kafka source&sink streaming conformance integration test.

* Data for the test can be usual JSONs. We can even reuse the existing end-to-end conformance test data for that.
* An instance of Kafka should be brought up.
* Data from the examples should be sent to raw Kafka topic converting the input data to Avro.
* A streaming conformance test should be run as Kafka -> Parquet and Kadka -> Kafka

Use https://www.testcontainers.org/modules/kafka/
It is used by Hyperdrive as well. Thanks Felipe for the suggestion.

AdrianOlosutean:
Closes #1307


AdrianOlosutean:
So far I've addded a test case with timeout which can catch #1306 . What other use cases could be useful for testing HyperConformance?

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/HyperConformance.scala
```diff
@@ -40,29 +41,30 @@ class HyperConformance (implicit cmd: ConfCmdConfig,
...
 
-    val reportVersion = getReportVersion
-
     logPreConformanceInfo(rawDf)
 
-    val infoDateColumn = infoDateFactory.getInfoDateColumn(rawDf)
-
     val conformance = dao.getDataset(cmd.datasetName, cmd.datasetVersion)
+    val conformedDf = applyConformanceTransformations(rawDf, conformance)
+    conformedDf
```
yruslan, line 54: Why the logging was removed. Maybe let's put it here.

spark-jobs/src/test/scala/za/co/absa/enceladus/conformance/streaming/HyperConformanceIntegrationSuite.scala
```diff
@@ -0,0 +1,40 @@
...
+import za.co.absa.enceladus.conformance.interpreter.fixtures.{NestedStructsFixture, StreamingFixture}
+
+class HyperConformanceIntegrationSuite extends FunSuite with StreamingFixture with NestedStructsFixture {
+
+  test("Test with catalyst workaround") {
+    val frame: DataFrame = testHyperConformance(standardizedDf,
+      "result",
+      nestedStructsDS)
+    frame.show()
+    assertResult(frame.count())(20)
```
yruslan, line 29: In addition to testing count, maybe we can test the output as well?

spark-jobs/src/test/scala/za/co/absa/enceladus/conformance/interpreter/fixtures/StreamingFixture.scala
```diff
@@ -0,0 +1,65 @@
...
+import za.co.absa.enceladus.conformance.streaming.{InfoDateFactory, InfoDateLiteralFactory}
+import za.co.absa.enceladus.dao.MenasDAO
+import za.co.absa.enceladus.model.Dataset
+import za.co.absa.enceladus.utils.testUtils.SparkTestBase
+
+trait StreamingFixture extends FunSuite with SparkTestBase with MockitoSugar {
+  implicit val menasBaseUrls: List[String] = List()
+  implicit val cmd: ConfCmdConfig = ConfCmdConfig.apply(reportVersion = Some(1))
+  implicit val infoDateFactory: InfoDateFactory = new InfoDateLiteralFactory("2020-05-23")
+  private val interval = 10000
```
yruslan, line 35: How this interval was chosen?

AdrianOlosutean: Mostly I lowered the interval as long as it was possible to reproduce the bug. Since on the test where the optimizer workaround is enabled, it works to write the results during this interval, but on the other one it doesn't. From testing with several values I noticed not enabling this interval would keep the test stuck, so adding the timeout made sense.

AdrianOlosutean:
1. Yes, I was able to reproduce it by enabling or disabling the workaround and running it on the test. As @kevinwallimann said before, this bug is happening even if you provide just a MemoryStream.
2. The class seems to take around 30 seconds

spark-jobs/src/test/scala/za/co/absa/enceladus/conformance/interpreter/fixtures/StreamingFixture.scala
```diff
@@ -0,0 +1,65 @@
...
+    val sink = conformed
+      .writeStream
+      .trigger(Trigger.ProcessingTime(interval))
+      .queryName(memoryTableName)
+      .outputMode("append")
+      .format("memory")
+      .start()
+
+    input.collect().foreach(e => memoryStream.addData(e))
+    sink.awaitTermination(interval)
```
kevinwallimann, line 60: `awaitTermination(interval)` is probably a little bit dangerous. If the query is terminated just before the microbatch has been committed, then the whole microbatch will be aborted and there will be zero rows.
Also, you have to assume that your data will be processed within `interval` ms. 10s is probably enough, but it's kind of a magic value.
In this case `processAllAvailable` would be a good fit, because then you can be sure that all the rows will have been read, no matter to what value `interval` has been set to. `processAllAvailable` terminates as long as the data source doesn't emit values indefinitely which is the case here

AdrianOlosutean: I agreee, but how can you use `processAllAvailable` on a test with the catalyst optimizer workaround disabled without having an unreasonable running time for testing?

kevinwallimann: Ok, now I see your need for a timeout. Either you give up on testing with multiple micro-batches and use `Trigger.Once` or you could use a margin of safety, say `3 * interval`.
Also, if you really want to cover multiple micro-batches, you need to make sure that the `foreach` loop runs during multiple micro-batches.

spark-jobs/src/test/scala/za/co/absa/enceladus/conformance/streaming/HyperConformanceIntegrationSuite.scala
```diff
@@ -0,0 +1,58 @@
...
+import org.scalatest.FunSuite
+import za.co.absa.enceladus.conformance.interpreter.fixtures.{NestedStructsFixture, StreamingFixture}
+
+class HyperConformanceIntegrationSuite extends FunSuite with StreamingFixture with NestedStructsFixture {
+
+  test("Test with catalyst workaround") {
+    val frame: DataFrame = testHyperConformance(standardizedDf,
+      "result",
+      nestedStructsDS)
+    frame.show()
```
Zejnilovic, line 29: Don't use show in tests

spark-jobs/src/test/scala/za/co/absa/enceladus/conformance/streaming/HyperConformanceIntegrationSuite.scala
```diff
@@ -0,0 +1,58 @@
...
+    val strings = data("strings").asInstanceOf[GenericRowWithSchema]
+    assertResult(strings.getAs("whitespaces_upper"))("K  C H E   RS Z     ")
+
+    val numerics = data("numerics").asInstanceOf[GenericRowWithSchema]
+    assertResult(numerics.getAs("big_negative_negated"))(783143645497786L)
+
+    assertResult(frame.count())(20)
+  }
+
+  test("Test without catalyst workaround") {
```
Zejnilovic, line 50: I am not sure what this tests?

spark-jobs/src/test/scala/za/co/absa/enceladus/conformance/interpreter/fixtures/StreamingFixture.scala
```diff
@@ -0,0 +1,69 @@
...
+      .trigger(Trigger.ProcessingTime(processingTime))
+      .queryName(sinkTableName)
+      .outputMode("append")
+      .format("memory")
+      .start()
+
+    input.collect().foreach(e => {
+      memoryStream.addData(e)
+      Thread.sleep(100)
+      spark.sql(s"select * from $sinkTableName").show()
```
yruslan, line 62: `.show` is redundant here since we don't want our build logs to be very big.

spark-jobs/src/test/scala/za/co/absa/enceladus/conformance/interpreter/fixtures/StreamingFixture.scala
```diff
@@ -0,0 +1,69 @@
...
+import org.scalatest.mockito.MockitoSugar
+import za.co.absa.enceladus.conformance.{ConfCmdConfig, HyperConformance}
+import za.co.absa.enceladus.conformance.interpreter.FeatureSwitches
+import za.co.absa.enceladus.conformance.streaming.{InfoDateFactory, InfoDateLiteralFactory}
+import za.co.absa.enceladus.dao.MenasDAO
+import za.co.absa.enceladus.model.Dataset
+import za.co.absa.enceladus.utils.testUtils.SparkTestBase
+
+trait StreamingFixture extends FunSuite with SparkTestBase with MockitoSugar {
+  implicit val menasBaseUrls: List[String] = List()
```
Zejnilovic, line 32: I know it is a matter of taste, but readability wise I like `List.empty[String]` the most.

spark-jobs/src/test/scala/za/co/absa/enceladus/conformance/interpreter/fixtures/StreamingFixture.scala
```diff
@@ -0,0 +1,69 @@
...
+      .writeStream
+      .trigger(Trigger.ProcessingTime(processingTime))
+      .queryName(sinkTableName)
+      .outputMode("append")
+      .format("memory")
+      .start()
+
+    input.collect().foreach(e => {
+      memoryStream.addData(e)
+      Thread.sleep(100)
```
yruslan, line 61: Is there a way it can be done so it won't need to rely on timing. Since if the timing is involved we might see the test fail non-deterministically since it is going to be run on machines that have very different specs.

Maybe you can wait for a micro-batch to be processed before sending another one by doing `select` and `count` on the memory table?

spark-jobs/src/test/scala/za/co/absa/enceladus/conformance/streaming/HyperConformanceIntegrationSuite.scala
```diff
@@ -0,0 +1,58 @@
...
+
+import org.apache.spark.sql.DataFrame
+import org.apache.spark.sql.catalyst.expressions.GenericRowWithSchema
+import org.scalatest.FunSuite
+import za.co.absa.enceladus.conformance.interpreter.fixtures.{NestedStructsFixture, StreamingFixture}
+
+class HyperConformanceIntegrationSuite extends FunSuite with StreamingFixture with NestedStructsFixture {
+
+  test("Test with catalyst workaround") {
+    val frame: DataFrame = testHyperConformance(standardizedDf,
```
yruslan, line 26: We name data frames `df` everywhere. I'm not arguing which one is better, just let's be consistent as much as possible.

spark-jobs/src/test/scala/za/co/absa/enceladus/conformance/interpreter/fixtures/NestedStructsFixture.scala
```diff
@@ -141,7 +138,6 @@ trait NestedStructsFixture extends BeforeAndAfterAll with SparkTestBase {
 
   protected val nestedStructsCastingDS = Dataset(
     name = "Nested Structs Conformance",
-    version = 1,
```
Zejnilovic, line 144: Why was this removed?

spark-jobs/src/test/scala/za/co/absa/enceladus/conformance/streaming/HyperConformanceIntegrationSuite.scala
```diff
@@ -0,0 +1,58 @@
...
+      "result",
+      nestedStructsDS)
+    frame.show()
+    val fields = frame.schema.fieldNames
+    val data = frame.first().getValuesMap(fields)
```
yruslan, line 31: Rather than comparing specific values, we can compare full output by doing:
```
df.toJSON.collect().mkString("\n")
```
and comparing results with the expected JSON somewhere in resources. If there. are non-deterministic columns, we can remove them from comparison by using `dropColumn()` or `select()`.

I'd prefer doing `select` on all fields expected in the output data frame. Since changes of column order are allowed, doing `select` will make th order specific for easier comparison. 

Ideally, the expected output JSON should be the same for batch and streaming jobs. If the order of records could be changes, I'd recommend doing `orderBy` before comparison.

spark-jobs/src/test/scala/za/co/absa/enceladus/conformance/streaming/HyperConformanceIntegrationSuite.scala
```diff
@@ -0,0 +1,58 @@
...
+
+    val infoDateStringColumnName = "enceladus_info_date_string"
+    assert(fields.contains(infoDateStringColumnName))
+    assertResult(data("enceladus_info_date_string"))("2020-05-23")
```
yruslan, line 35: We need to test different `enceladus_info_date` strategies, possibly as part of smaller tests.

spark-jobs/src/test/scala/za/co/absa/enceladus/conformance/streaming/HyperConformanceIntegrationSuite.scala
```diff
@@ -0,0 +1,58 @@
...
+    assert(fields.contains(infoVersionColumnName))
+    assertResult(data(infoVersionColumnName))(1)
+
+    val strings = data("strings").asInstanceOf[GenericRowWithSchema]
+    assertResult(strings.getAs("whitespaces_upper"))("K  C H E   RS Z     ")
+
+    val numerics = data("numerics").asInstanceOf[GenericRowWithSchema]
+    assertResult(numerics.getAs("big_negative_negated"))(783143645497786L)
+
+    assertResult(frame.count())(20)
```
yruslan, line 47: Let's make count checking the first test. If it fails it doesn't make sense to compare individual values or the output as a whole.

spark-jobs/src/test/scala/za/co/absa/enceladus/conformance/streaming/HyperConformanceIntegrationSuite.scala
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  test("Test without catalyst workaround") {
+    val frame: DataFrame = testHyperConformance(standardizedDf,
```
yruslan, line 51: What does this test?

spark-jobs/src/test/scala/za/co/absa/enceladus/conformance/streaming/HyperConformanceIntegrationSuite.scala
```diff
@@ -0,0 +1,58 @@
...
+      "result2",
+      nestedStructsDS,
+      catalystWorkaround = false)
+    frame.show()
```
yruslan, line 55: Redundant show

spark-jobs/src/test/scala/za/co/absa/enceladus/conformance/interpreter/fixtures/NestedStructsFixture.scala
AdrianOlosutean: It's redundant, the default is 1

spark-jobs/src/test/scala/za/co/absa/enceladus/conformance/streaming/HyperConformanceIntegrationSuite.scala
AdrianOlosutean: I mostly thought that show would make following what happens at the runtime easier

spark-jobs/src/test/scala/za/co/absa/enceladus/conformance/streaming/HyperConformanceIntegrationSuite.scala
AdrianOlosutean: The idea is that since the workaround is disabled, the test will timeout and no data will be added to the sink in a reasonable time. Probably makes sense to add more description

spark-jobs/src/test/scala/za/co/absa/enceladus/conformance/interpreter/fixtures/StreamingFixture.scala
AdrianOlosutean: Would printing the length of the table make sense instead for debugging reasons?

spark-jobs/src/test/scala/za/co/absa/enceladus/conformance/interpreter/fixtures/StreamingFixture.scala
AdrianOlosutean: About timing, I don't say the tests couldn't fail, but it's not likely. As @kevinwallimann suggested, the timeout between the insertion better simulates a more realistic use case and since inserting everything would take 2 seconds, the total waited time is 6 seconds and I can incrase it. 

The other discussion we had about not using `awaitTermination` and using `processAllAvailable` is that using the second one makes testing with the workaround disabled run an unreasonable amount of time.

spark-jobs/src/test/scala/za/co/absa/enceladus/conformance/streaming/HyperConformanceIntegrationSuite.scala
AdrianOlosutean: Makes sense

spark-jobs/src/test/scala/za/co/absa/enceladus/conformance/streaming/HyperConformanceIntegrationSuite.scala
Zejnilovic: This is testing of the presence of a bug is Spark. Seems quite weird to me and I would remove the test.

spark-jobs/src/test/scala/za/co/absa/enceladus/conformance/streaming/HyperConformanceIntegrationSuite.scala
Zejnilovic: I understand, but in an ideal world, you just look at the bottom where it says that everything passed. In case of an issue, every one of us can write this line by themselves and re-run

spark-jobs/src/test/scala/za/co/absa/enceladus/conformance/interpreter/fixtures/StreamingFixture.scala
yruslan: You can use `log.debug` if you really want to commit lines that are used only for debugging. But generally, we prefer removing lines added for debugging but instead, provide a detailed error message if the test fails.

spark-jobs/src/test/scala/za/co/absa/enceladus/conformance/streaming/HyperConformanceIntegrationSuite.scala
yruslan: I see Saša has the same question :)

spark-jobs/src/test/scala/za/co/absa/enceladus/conformance/interpreter/fixtures/StreamingFixture.scala
yruslan: Sorry, I wasn't very constructive since I don't have a solution to this. Structured Streaming jobs is a new topic for me as well.
Not sure if it can be done without relying on the timer. But if it could be done, such a solution would be preferable.

spark-jobs/src/test/scala/za/co/absa/enceladus/conformance/streaming/HyperConformanceIntegrationSuite.scala
yruslan: It might make sense to keep a commented test that, once uncommented, will demonstrate that the Catalyst issue still exists and the test hangs.
It might be useful in case new version of Spark has this issue fixed. We could try it then to see if the fix covers our use case.

spark-jobs/src/test/scala/za/co/absa/enceladus/conformance/interpreter/fixtures/StreamingFixture.scala
yruslan: I'm slightly confused. We need to test 2 things:
1. The correctness of the results of streaming conformance. All workarounds should be turned on so such tests can finish in a reasonable time.
2. We need to test that workaround works by adding a test that hangs if the workaround is turned off. There is no particular time limit on the test, but should be some reasonable time.

Tests that test (1) should not rely on timing if possible. If `processAllAvailable` helps with it - great..
Tests that test (2) might rely on timing, but not necessarily. Currently in batch, if the workaround stops working the corresponding test will just hang forever. Not ideal, but we expect workaround to work. If it stops having an effect, it will be noticed by the developer when the test hangs.
Alternatively, a big timeout can be used to interrupt the test if it takes too long. But the value of the timeout should be set so that test passes on all envs the test could run, including Jenkins VMs.

The reason I care about this so much is that timings introduce non-determinism. Such tests could work perfectly locally on Mac, but start failing on Windows or on Linux build box.

spark-jobs/src/test/scala/za/co/absa/enceladus/conformance/interpreter/fixtures/StreamingFixture.scala
yruslan: Thanks a lot, @kevinwallimann for the insight and for the proposed solution!

@AdrianOlosutean, I'm still not 100% sure if we should have a [uncommented] test that runs forever if the workaround is turned off and is interrupted by a timer. Let me know what you think. My current concern is that having such a test would bring more risks than benefits. We need to ensure that the processing interval does interrupt the query even if it is stuck inside the Catalyst layer.
Having a commented test that hangs without the workaround can be beneficial to test new versions of Spark. And since it is commented, it doesn't take time to run on normal builds :)

AdrianOlosutean: For me it seems to that the commented test would be the more sensible approach. Very interesting, obviously the solution proposed by @kevinwallimann is the rigurous one and I guess it should be used, thanks.

AdrianOlosutean:
Release notes: [here](https://github.com/absaoss/enceladus/issues/1307#issuecomment-644697488)

Issue #1198, opened 2020-02-21
Add Support for Missing Spark Types


benedeki:
## Background
Some native Spakr types are not supported by Enceladus. Their inclusion might be worthy. Namely:
* CalendarInterval
* Map

Zejnilovic:
BinaryType?

benedeki:
> 
> 
> BinaryType?

Yep, could be useful. Please create an Issue for this Epic.

dk1844:
@Zejnilovic, what is the problem with `BinaryType`? I have just tried it in Menas and it uploaded, displayed and exported fine. Is there a specific case that `BinaryType` has problems with?

Zejnilovic:
Binary type is missing CastingConformance Rule. Also, would Binary have any metadata in schema? Would they work?

PR #1369, opened 2020-06-01
Feature/1198 menas schema: map<string, V> support

### LINKED ISSUE #902: Menas supports MapType while Standardization does not. Establish parity
## Background
_Menas_ supports `MapType` entry in the schema while _Standardization_ does not reflect this type existence.

## Feature
Establish parity between the component feature, preferably with _Standardization_ extending its support to `MapType` rather then _Menas_ dropping it.

Additionaly - check _Conformance_ compatibility with the type

dk1844:
(Draft: Since this only covers the UI changes, for the full feature to make sense, the backend support should be addressed, too. ref: #902 )

Should the upload contain incompatible type to `Map<String, V>` (e.g. `Map<Integer, V>`), the UI let's the user know:
maptype-with-integerType-keytype

#1198, closes #902


Zejnilovic:
Do you plan to add casting rule with map option? For now, the key could be a literal specified.

dk1844:
Closed for now for no activity on the prereq issue #902 .

Issue #1235, opened 2020-03-18
Add support for Binary Type 


Zejnilovic:
## Background
`BinaryType` is a basic Spark type.

## Feature
Add support for this type both to **Standardization** and **Conformance**.

dk1844:
Release notes
BinaryType is now supported both in Standardization/Conformance and in menas UI. Mind the `encoding` metadata setting when converting from StringType.

PR #1370, opened 2020-06-01
#1235 BinaryType support in enceladus

dk1844:
`BinaryType` support added for Enceladus
 - encoding in metadata added, supported values are `base64` and `none`. No encoding behaves like `none` (taking strings as-is and just casting them to binary)
 - default value follows the encoding set (and is validated to be valid base64 string if base64 encoding is used)
 - test cases extended.
 - Standardization run ok with a custom binary field in the input (as parquet)
 - Conformance UI enriched by allowing `binary` for the _CastingConformanceRule_.
castingConformanceBinary

### Binary <-> String
 - Tested Conformance with such a rule to cast "binary <-> string": successfully
binary-string-vice-versa

### default value from metadata is used
 - default can be provided in the metadata (always encoded as base64):
default-meta-for-binary1
 - having input for Standardziation with a `null`-containing binary field and a schema that disallows null for this column in standardization:
```json
{
      "name": "binary1",
      "type": "binary",
      "nullable": false,
      "metadata": {"default" : "dGVzdA==", "encoding":"base64"}
    }
```
(`unbase64("dGVzdA==")` yields `test`)

std-input-binary1-null
 - correctly resulted in the default being applied
binary-default-applied-in-std

### DIFF
--- a/menas/ui/components/dataset/conformanceRule/ConformanceRuleForm.js
+++ b/menas/ui/components/dataset/conformanceRule/ConformanceRuleForm.js
@@ -165,7 +165,8 @@ class CastingConformanceRuleForm extends ConformanceRuleForm {
       {type: "decimal(38,18)"},
       {type: "string"},
       {type: "date"},
-      {type: "timestamp"}
+      {type: "timestamp"},
+      {type: "binary"},
     ]
   }
 
--- a/menas/ui/components/dataset/conformanceRule/ConformanceRuleForm.js
+++ b/menas/ui/components/dataset/conformanceRule/ConformanceRuleForm.js
@@ -166,7 +166,7 @@ class CastingConformanceRuleForm extends ConformanceRuleForm {
       {type: "string"},
       {type: "date"},
       {type: "timestamp"},
-      {type: "binary"},
+      {type: "binary"}
     ]
   }
 
--- a/spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
+++ b/spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
@@ -27,8 +27,9 @@ import org.apache.spark.sql.types._
 import org.slf4j.{Logger, LoggerFactory}
 import za.co.absa.enceladus.standardization.interpreter.dataTypes.ParseOutput
 import za.co.absa.enceladus.utils.error.ErrorMessage
-import za.co.absa.enceladus.utils.schema.{MetadataKeys, SchemaUtils}
+import za.co.absa.enceladus.utils.schema.{MetadataKeys, MetadataValues, SchemaUtils}
 import za.co.absa.enceladus.utils.schema.SchemaUtils.FieldWithSource
+import za.co.absa.enceladus.utils.implicits.StructFieldImplicits._
 import za.co.absa.enceladus.utils.time.DateTimePattern
 import za.co.absa.enceladus.utils.typeClasses.{DoubleLike, LongLike}
 import za.co.absa.enceladus.utils.types.TypedStructField._
@@ -417,18 +418,18 @@ object TypeParser {
                                        (implicit defaults: Defaults) extends PrimitiveParser[Array[Byte]] {
     override protected def assemblePrimitiveCastLogic: Column = {
       origType match {
-        case BinaryType => column.cast(BinaryType) // binary: just cast
+        case BinaryType => column
         case StringType =>
-          if (field.structField.metadata.contains(MetadataKeys.Encoding)) {
-            val encoding = field.structField.metadata.getString(MetadataKeys.Encoding)
-            encoding.toLowerCase match {
-              case "base64" => org.apache.spark.sql.functions.unbase64(column).cast(BinaryType)
-              case _ => throw new IllegalStateException(s"Unsupported encoding for Binary field ${field.structField.name}: '$encoding'")
-            }
+          val encoding = field.structField.getMetadataString(MetadataKeys.Encoding).map(_.toLowerCase)
 
-          } else {
-            logger.info(s"Binary field ${field.structField.name} does not have encoding setup in metadata. Reading as-is using 'getBytes'")
-            column.cast(field.dataType) // no metadata, just cast
+          encoding match {
+            case Some(MetadataValues.Encoding.Base64) => unbase64(column)
+            case Some(MetadataValues.Encoding.None) | None =>
+              if (encoding.isEmpty) {
+                logger.info(s"Binary field ${field.structField.name} does not have encoding setup in metadata. Reading as-is.")
+              }
+              column.cast(field.dataType) // use as-is
+            case _ => throw new IllegalStateException(s"Unsupported encoding for Binary field ${field.structField.name}: '${encoding.get}'")
           }
         case _ => throw new IllegalStateException(s"Unsupported conversion from BinaryType to ${field.dataType}")
       }
--- a/spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
+++ b/spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
@@ -168,6 +168,7 @@ object TypeParser {
       case _: DoubleType    => FractionalParser(TypedStructField.asNumericTypeStructField[Double](field), _, _, _, _, _)
       case _: DecimalType   => DecimalParser(TypedStructField.asNumericTypeStructField[BigDecimal](field), _, _, _, _, _)
       case _: StringType    => StringParser(TypedStructField(field), _, _, _, _, _)
+      case _: BinaryType    => BinaryParser(TypedStructField(field), _, _, _, _, _)
       case _: BooleanType   => BooleanParser(TypedStructField(field), _, _, _, _, _)
       case _: DateType      => DateParser(TypedStructField.asDateTimeTypeStructField(field), _, _, _, _, _)
       case _: TimestampType => TimestampParser(TypedStructField.asDateTimeTypeStructField(field), _, _, _, _, _)
@@ -406,6 +407,14 @@ object TypeParser {
                                         failOnInputNotPerSchema: Boolean,
                                         isArrayElement: Boolean)
                                        (implicit defaults: Defaults) extends ScalarParser[String]
+
+  private final case class BinaryParser(field: TypedStructField,
+                                        path: String,
+                                        column: Column,
+                                        origType: DataType,
+                                        failOnInputNotPerSchema: Boolean,
+                                        isArrayElement: Boolean)
+                                       (implicit defaults: Defaults) extends ScalarParser[Array[Byte]]
 
   private final case class BooleanParser(field: TypedStructField,
                                          path: String,
--- a/spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
+++ b/spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
@@ -53,7 +53,7 @@ import scala.util.{Random, Try}
   *         NumericParser !
   *         StringParser !
   *         BooleanParser !
- *        BinaryParser !
+  *       BinaryParser !
   *       DateTimeParser
   *         TimestampParser !
   *         DateParser !
--- a/spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
+++ b/spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
@@ -35,6 +35,7 @@ import za.co.absa.enceladus.utils.typeClasses.{DoubleLike, LongLike}
 import za.co.absa.enceladus.utils.types.TypedStructField._
 import za.co.absa.enceladus.utils.types.{Defaults, TypedStructField}
 import za.co.absa.enceladus.utils.udf.{UDFBuilder, UDFLibrary, UDFNames}
+import za.co.absa.enceladus.utils.validation.ValidationIssue
 import za.co.absa.spark.hofs.transform
 
 import scala.reflect.runtime.universe._
@@ -52,6 +53,7 @@ import scala.util.{Random, Try}
   *         NumericParser !
   *         StringParser !
   *         BooleanParser !
+ *        BinaryParser !
   *       DateTimeParser
   *         TimestampParser !
   *         DateParser !
@@ -169,7 +171,7 @@ object TypeParser {
       case _: DoubleType    => FractionalParser(TypedStructField.asNumericTypeStructField[Double](field), _, _, _, _, _)
       case _: DecimalType   => DecimalParser(TypedStructField.asNumericTypeStructField[BigDecimal](field), _, _, _, _, _)
       case _: StringType    => StringParser(TypedStructField(field), _, _, _, _, _)
-      case _: BinaryType    => BinaryParser(TypedStructField(field), _, _, _, _, _)
+      case _: BinaryType    => BinaryParser(TypedStructField.asBinaryTypeStructField(field), _, _, _, _, _)
       case _: BooleanType   => BooleanParser(TypedStructField(field), _, _, _, _, _)
       case _: DateType      => DateParser(TypedStructField.asDateTimeTypeStructField(field), _, _, _, _, _)
       case _: TimestampType => TimestampParser(TypedStructField.asDateTimeTypeStructField(field), _, _, _, _, _)
@@ -409,7 +411,7 @@ object TypeParser {
                                         isArrayElement: Boolean)
                                        (implicit defaults: Defaults) extends ScalarParser[String]
 
-  private final case class BinaryParser(field: TypedStructField,
+  private final case class BinaryParser(field: BinaryTypeStructField,
                                         path: String,
                                         column: Column,
                                         origType: DataType,
@@ -420,20 +422,24 @@ object TypeParser {
       origType match {
         case BinaryType => column
         case StringType =>
-          val encoding = field.structField.getMetadataString(MetadataKeys.Encoding).map(_.toLowerCase)
-
-          encoding match {
-            case Some(MetadataValues.Encoding.Base64) => unbase64(column)
-            case Some(MetadataValues.Encoding.None) | None =>
-              if (encoding.isEmpty) {
-                logger.info(s"Binary field ${field.structField.name} does not have encoding setup in metadata. Reading as-is.")
-              }
-              column.cast(field.dataType) // use as-is
-            case _ => throw new IllegalStateException(s"Unsupported encoding for Binary field ${field.structField.name}: '${encoding.get}'")
+          val validationIssues: Seq[ValidationIssue] = field.validate()
+          if (validationIssues.nonEmpty) {
+            throw new IllegalStateException(s"There are validation issues, cannot continue parsing binary field $field: $validationIssues")
+          } else {
+            field.normalizedEncoding match {
+              case Some(MetadataValues.Encoding.Base64) => unbase64(column)
+              case Some(MetadataValues.Encoding.None) | None =>
+                if (field.normalizedEncoding.isEmpty) {
+                  logger.info(s"Binary field ${field.structField.name} does not have encoding setup in metadata. Reading as-is.")
+                }
+                column.cast(field.dataType) // use as-is
+              case _ => throw new IllegalStateException(s"Unsupported encoding for Binary field ${field.structField.name}:" +
+                s" '${field.normalizedEncoding.get}'")
+            }
           }
+
         case _ => throw new IllegalStateException(s"Unsupported conversion from BinaryType to ${field.dataType}")
       }
-
     }
   }
 
--- a/spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
+++ b/spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
@@ -35,7 +35,7 @@ import za.co.absa.enceladus.utils.typeClasses.{DoubleLike, LongLike}
 import za.co.absa.enceladus.utils.types.TypedStructField._
 import za.co.absa.enceladus.utils.types.{Defaults, TypedStructField}
 import za.co.absa.enceladus.utils.udf.{UDFBuilder, UDFLibrary, UDFNames}
-import za.co.absa.enceladus.utils.validation.ValidationIssue
+import za.co.absa.enceladus.utils.validation.{ValidationError, ValidationIssue}
 import za.co.absa.spark.hofs.transform
 
 import scala.reflect.runtime.universe._
@@ -422,9 +422,12 @@ object TypeParser {
       origType match {
         case BinaryType => column
         case StringType =>
-          val validationIssues: Seq[ValidationIssue] = field.validate()
-          if (validationIssues.nonEmpty) {
-            throw new IllegalStateException(s"There are validation issues, cannot continue parsing binary field $field: $validationIssues")
+          val (validationErrors, validationWarnings) = field.validate().span(_.isInstanceOf[ValidationError])
+          if (validationWarnings.nonEmpty){
+            logger.warn(s"There are validation warnings for binary field parsing $field: $validationWarnings")
+          }
+          if (validationErrors.nonEmpty) {
+            throw new IllegalStateException(s"There are validation issues, cannot continue parsing binary field $field: $validationErrors")
           } else {
             field.normalizedEncoding match {
               case Some(MetadataValues.Encoding.Base64) => unbase64(column)
--- a/spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
+++ b/spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
@@ -27,15 +27,13 @@ import org.apache.spark.sql.types._
 import org.slf4j.{Logger, LoggerFactory}
 import za.co.absa.enceladus.standardization.interpreter.dataTypes.ParseOutput
 import za.co.absa.enceladus.utils.error.ErrorMessage
-import za.co.absa.enceladus.utils.schema.{MetadataKeys, MetadataValues, SchemaUtils}
 import za.co.absa.enceladus.utils.schema.SchemaUtils.FieldWithSource
-import za.co.absa.enceladus.utils.implicits.StructFieldImplicits._
+import za.co.absa.enceladus.utils.schema.{MetadataValues, SchemaUtils}
 import za.co.absa.enceladus.utils.time.DateTimePattern
 import za.co.absa.enceladus.utils.typeClasses.{DoubleLike, LongLike}
 import za.co.absa.enceladus.utils.types.TypedStructField._
 import za.co.absa.enceladus.utils.types.{Defaults, TypedStructField}
 import za.co.absa.enceladus.utils.udf.{UDFBuilder, UDFLibrary, UDFNames}
-import za.co.absa.enceladus.utils.validation.{ValidationError, ValidationIssue}
 import za.co.absa.spark.hofs.transform
 
 import scala.reflect.runtime.universe._
@@ -422,23 +420,16 @@ object TypeParser {
       origType match {
         case BinaryType => column
         case StringType =>
-          val (validationErrors, validationWarnings) = field.validate().span(_.isInstanceOf[ValidationError])
-          if (validationWarnings.nonEmpty){
-            logger.warn(s"There are validation warnings for binary field parsing $field: $validationWarnings")
-          }
-          if (validationErrors.nonEmpty) {
-            throw new IllegalStateException(s"There are validation issues, cannot continue parsing binary field $field: $validationErrors")
-          } else {
-            field.normalizedEncoding match {
-              case Some(MetadataValues.Encoding.Base64) => unbase64(column)
-              case Some(MetadataValues.Encoding.None) | None =>
-                if (field.normalizedEncoding.isEmpty) {
-                  logger.info(s"Binary field ${field.structField.name} does not have encoding setup in metadata. Reading as-is.")
-                }
-                column.cast(field.dataType) // use as-is
-              case _ => throw new IllegalStateException(s"Unsupported encoding for Binary field ${field.structField.name}:" +
-                s" '${field.normalizedEncoding.get}'")
-            }
+          // already validated in Standardization
+          field.normalizedEncoding match {
+            case Some(MetadataValues.Encoding.Base64) => callUDF(UDFNames.binaryUnbase64, column)
+            case Some(MetadataValues.Encoding.None) | None =>
+              if (field.normalizedEncoding.isEmpty) {
+                logger.info(s"Binary field ${field.structField.name} does not have encoding setup in metadata. Reading as-is.")
+              }
+              column.cast(field.dataType) // use as-is
+            case _ => throw new IllegalStateException(s"Unsupported encoding for Binary field ${field.structField.name}:" +
+              s" '${field.normalizedEncoding.get}'")
           }
 
         case _ => throw new IllegalStateException(s"Unsupported conversion from BinaryType to ${field.dataType}")
--- a/spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
+++ b/spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
@@ -27,7 +27,7 @@ import org.apache.spark.sql.types._
 import org.slf4j.{Logger, LoggerFactory}
 import za.co.absa.enceladus.standardization.interpreter.dataTypes.ParseOutput
 import za.co.absa.enceladus.utils.error.ErrorMessage
-import za.co.absa.enceladus.utils.schema.SchemaUtils
+import za.co.absa.enceladus.utils.schema.{MetadataKeys, SchemaUtils}
 import za.co.absa.enceladus.utils.schema.SchemaUtils.FieldWithSource
 import za.co.absa.enceladus.utils.time.DateTimePattern
 import za.co.absa.enceladus.utils.typeClasses.{DoubleLike, LongLike}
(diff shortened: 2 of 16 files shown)


spark-jobs/src/test/scala/za/co/absa/enceladus/standardization/interpreter/StandardizationInterpreter_BinarySuite.scala
```diff
@@ -0,0 +1,54 @@
...
+ */
+
+package za.co.absa.enceladus.standardization.interpreter
+
+import org.apache.spark.sql.types.{BinaryType, StructField, StructType}
+import org.scalatest.{FunSuite, Matchers}
+import za.co.absa.enceladus.utils.testUtils.{LoggerTestBase, SparkTestBase}
+import za.co.absa.enceladus.utils.udf.UDFLibrary
+
+class StandardizationInterpreter_BinarySuite extends FunSuite with SparkTestBase with LoggerTestBase with Matchers {
```
Zejnilovic, line 23: Are there any negative tests? Like `Seq(123)`. To show the casting error.

utils/src/main/scala/za/co/absa/enceladus/utils/types/Defaults.scala
```diff
@@ -52,6 +52,7 @@ object GlobalDefaults extends Defaults {
       case _: DoubleType    => 0.0d
       case _: LongType      => 0L
       case _: StringType    => ""
+      case _: BinaryType    => Array[Byte]()
```
Zejnilovic, line 55: Just my personal taste
```suggestion
      case _: BinaryType    => Array.empty[Byte]
```

spark-jobs/src/test/scala/za/co/absa/enceladus/standardization/interpreter/StandardizationInterpreter_BinarySuite.scala
benedeki: Not sure there can be a negative test (fail) for the type other then `null` on a  _not null_ column. 🤔

menas/ui/components/dataset/conformanceRule/ConformanceRuleForm.js
```diff
@@ -165,7 +165,8 @@ class CastingConformanceRuleForm extends ConformanceRuleForm {
       {type: "decimal(38,18)"},
       {type: "string"},
       {type: "date"},
-      {type: "timestamp"}
+      {type: "timestamp"},
+      {type: "binary"},
```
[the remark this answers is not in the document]
dk1844: No, thanks, fixed.

spark-jobs/src/test/scala/za/co/absa/enceladus/standardization/interpreter/StandardizationInterpreter_BinarySuite.scala
```diff
@@ -0,0 +1,54 @@
...
+  import spark.implicits._
+
+  private implicit val udfLib: UDFLibrary = new UDFLibrary
+
+  private val fieldName = "binaryField"
+
+  test("byteArray") {
+    val seq = Seq(
+      Array(1, 2, 3).map(_.toByte),
+      Array('a', 'b', 'c').map(_.toByte)
```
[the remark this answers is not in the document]
dk1844: I have added some more tests, so the case you are looking for should be addressed in the test _Binary from string with no encoding_.

spark-jobs/src/test/scala/za/co/absa/enceladus/standardization/interpreter/StandardizationInterpreter_BinarySuite.scala
dk1844: A test with null in a non-null column is also added (where an explicit `default` from metadata is used.)

utils/src/main/scala/za/co/absa/enceladus/utils/types/Defaults.scala
dk1844: Sure, why not, changed.

utils/src/main/scala/za/co/absa/enceladus/utils/types/TypedStructField.scala
```diff
@@ -178,6 +184,22 @@ object TypedStructField {
...
 
+  // BinaryTypeStructField
+  final class BinaryTypeStructField private[TypedStructField](structField: StructField)
+                                                             (implicit defaults: Defaults)
+    extends TypedStructFieldTagged[Array[Byte]](structField) {
+
+    // used to convert the default value from metadata's [[MetadataKeys.DefaultValue]]
+    override protected def convertString(string: String): Try[Array[Byte]] = {
+      //binary's default value is always expected to be in base64
+      Try(Base64.getDecoder.decode(string))
```
benedeki, line 195: This will always use the base64 decoder, which is not desired, IMO.
It should be consittent with the encoding, like in other cases. But glad to discuss.

utils/src/main/scala/za/co/absa/enceladus/utils/types/TypedStructField.scala
```diff
@@ -178,6 +184,22 @@ object TypedStructField {
...
+    }
+
+    override def validate(): Seq[ValidationIssue] = {
+      ScalarFieldValidator.validate(this)
```
benedeki, line 199: Considering there is the `encoding` metadata, a dedicated validator is required.

utils/src/main/scala/za/co/absa/enceladus/utils/types/TypedStructField.scala
```diff
@@ -123,6 +126,8 @@ sealed abstract class TypedStructField(structField: StructField)(implicit defaul
 }
 
 object TypedStructField {
+  private val logger = LoggerFactory.getLogger(this.getClass)
```
benedeki, line 129: Logger not needed, not used

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
```diff
@@ -407,6 +408,34 @@ object TypeParser {
...
+                                        origType: DataType,
+                                        failOnInputNotPerSchema: Boolean,
+                                        isArrayElement: Boolean)
+                                       (implicit defaults: Defaults) extends PrimitiveParser[Array[Byte]] {
+    override protected def assemblePrimitiveCastLogic: Column = {
+      origType match {
+        case BinaryType => column.cast(BinaryType) // binary: just cast
+        case StringType =>
+          if (field.structField.metadata.contains(MetadataKeys.Encoding)) {
+            val encoding = field.structField.metadata.getString(MetadataKeys.Encoding)
```
benedeki, line 423: Would make the `encoding` a property of `BinaryTypeStructField`.

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
```diff
@@ -407,6 +408,34 @@ object TypeParser {
...
+            encoding.toLowerCase match {
+              case "base64" => org.apache.spark.sql.functions.unbase64(column).cast(BinaryType)
+              case _ => throw new IllegalStateException(s"Unsupported encoding for Binary field ${field.structField.name}: '$encoding'")
```
benedeki, line 426: No option for "leave as it is'.

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
```diff
@@ -407,6 +408,34 @@ object TypeParser {
...
+  private final case class BinaryParser(field: TypedStructField,
+                                        path: String,
+                                        column: Column,
+                                        origType: DataType,
+                                        failOnInputNotPerSchema: Boolean,
+                                        isArrayElement: Boolean)
+                                       (implicit defaults: Defaults) extends PrimitiveParser[Array[Byte]] {
+    override protected def assemblePrimitiveCastLogic: Column = {
+      origType match {
+        case BinaryType => column.cast(BinaryType) // binary: just cast
```
benedeki, line 420: 'cast` is not needed here 😉

benedeki, line 420: **For consideration**: What if the `encoding` is 'base64`? 🤔

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
benedeki, line 426: Should never got here, if validation is done properly. But keep it with a comment, perhaps.

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
```diff
@@ -407,6 +408,34 @@ object TypeParser {
...
+        case StringType =>
+          if (field.structField.metadata.contains(MetadataKeys.Encoding)) {
+            val encoding = field.structField.metadata.getString(MetadataKeys.Encoding)
+            encoding.toLowerCase match {
+              case "base64" => org.apache.spark.sql.functions.unbase64(column).cast(BinaryType)
```
benedeki, line 425:
```suggestion
              case "base64" => unbase64(column)
``

File: spark-jobs/src/test/scala/za/co/absa/enceladus/standardization/interpreter/StandardizationInterpreter_BinarySuite.scala (hunk 1 above)
Comment (Zejnilovic) [reply spark-jobs/src/test/scala/za/co/absa/enceladus/standardization/interpreter/StandardizationInterpreter_BinarySuite.scala:24]: I like the test added

File: spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala (hunk 9 above)
Comment (dk1844) [reply spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala:426]: The option to leave 'as-is' is a couple of lines below - omitting the `encoding` metadata key altogether. Do you think there should be another option within `encoding`, too? e.g. `{"encoding":"none"}`?

File: utils/src/main/scala/za/co/absa/enceladus/utils/types/TypedStructField.scala (hunk 7 above)
Comment (dk1844) [reply utils/src/main/scala/za/co/absa/enceladus/utils/types/TypedStructField.scala:129]: True, I have been logging something and refrained from doing so afterward, will fix, thanks.

File: spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala (hunk 9 above)
Comment (benedeki) [reply spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala:426]: IMHO it would be better, yes.

File: utils/src/main/scala/za/co/absa/enceladus/utils/types/TypedStructField.scala (hunk 6 above)
Comment (dk1844) [reply utils/src/main/scala/za/co/absa/enceladus/utils/types/TypedStructField.scala]: Great observation, I totally forgot about that...

So there is now a specific validator in place 😀

File: utils/src/main/scala/za/co/absa/enceladus/utils/types/TypedStructField.scala (hunk 5 above)
Comment (dk1844) [reply utils/src/main/scala/za/co/absa/enceladus/utils/types/TypedStructField.scala]: I thought we wanted it that way. OK, redone to respect the encoding, too.

File: spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala (hunk 8 above)
Comment (dk1844) [reply spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala]: True, this helps a bit here (when dealing with the default). Sadly, in the case of the BinaryParser, I still have to pick the encoding from the metadata so the field cannot easily be reused, because I don't decode binary->binary with it. 

Done.

File: spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala (hunk 9 above)
Comment (dk1844) [reply spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala]: Done.

File: spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala (hunk 10 above)
Comment (dk1844) [reply spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala]: true, removed.

Comment (dk1844) [reply spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala]: My take on this is: do nothing, we do not act on this for binary->binary. Any other thoughts anyone?

File: spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala (hunk 11 above)
Comment (dk1844) [reply spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala]: Done, thanks.

**Comment by dk1844** [comment]
Another thing I have come across:
I can see that a ValidationException correctly block Standardization from running, but does not really state what the problem was, e.g.:
```
za.co.absa.enceladus.utils.validation.ValidationException: A fatal schema validation error occurred.
	at za.co.absa.enceladus.standardization.interpreter.StandardizationInterpreter$.validateSchemaAgainstSelfInconsistencies(StandardizationInterpreter.scala:86)
	at za.co.absa.enceladus.standardization.interpreter.StandardizationInterpreter$.standardize(StandardizationInterpreter.scala:55)
```

What about if we included the validation issues messages in the exception message, e.g. via:
```scala
class ValidationException(val msg: String, val errors: Seq[String])
  extends Exception(s"$msg due to errors: ${errors.mkString(",")}")
```

This could be quite helpful to the users. I am just pondering if this would not break things that the user relies on anywhere...?

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
```diff
@@ -407,6 +409,34 @@ object TypeParser {
...
+                                        column: Column,
+                                        origType: DataType,
+                                        failOnInputNotPerSchema: Boolean,
+                                        isArrayElement: Boolean)
+                                       (implicit defaults: Defaults) extends PrimitiveParser[Array[Byte]] {
+    override protected def assemblePrimitiveCastLogic: Column = {
+      origType match {
+        case BinaryType => column
+        case StringType =>
+          val encoding = field.structField.getMetadataString(MetadataKeys.Encoding).map(_.toLowerCase)
```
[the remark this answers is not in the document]
dk1844: So your suggestion is to force lowercasing in validation and then rely on this?
I was trying to be more flexible and letting users use e.g. "Base64", too.

Or am I misunderstanding?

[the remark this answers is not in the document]
benedeki: My suggestion is to read and normalize the encoding metadata value in `BinaryTypeStructField`. Then verify in validator.
Here just use the `BinaryTypeStructField` normalized value.
The advanced option is to use enumeration in `BinaryTypeStructField`, but that can prove clumsy with validation involved.

utils/src/main/scala/za/co/absa/enceladus/utils/validation/field/BinaryFieldValidator.scala
```diff
@@ -0,0 +1,69 @@
...
+    }).getOrElse(Seq.empty)
+  }
+
+  private def validateEncoding(field: TypedStructField): Seq[ValidationIssue] = {
+    val encoding: Option[String] = encodingForField(field)
+
+    encoding.map(_.toLowerCase) match {
+      case Some(MetadataValues.Encoding.Base64) | Some(MetadataValues.Encoding.None) =>
+        Seq.empty
+      case None => Seq.empty
```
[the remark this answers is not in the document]
benedeki: I wouldn't mind the _warning_ if no encoding is specified. But leave it up to you and what others think (maybe ask in chat for better exposure)

utils/src/main/scala/za/co/absa/enceladus/utils/validation/field/BinaryFieldValidator.scala
```diff
@@ -0,0 +1,69 @@
...
+import za.co.absa.enceladus.utils.implicits.StructFieldImplicits._
+import za.co.absa.enceladus.utils.schema.{MetadataKeys, MetadataValues}
+import za.co.absa.enceladus.utils.types.TypedStructField
+import za.co.absa.enceladus.utils.validation.{ValidationError, ValidationIssue}
+
+import scala.util.{Failure, Success, Try}
+
+object BinaryFieldValidator extends BinaryFieldValidator
+
+class BinaryFieldValidator extends FieldValidator {
```
benedeki, line 29: Please refactor this class.
It's strangely inconsistently written. Functions for simple one liners used ones, `vals` for trivial expressions, on the other hand complicated expressions are directly operated on...

utils/src/main/scala/za/co/absa/enceladus/utils/validation/field/BinaryFieldValidator.scala
```diff
@@ -0,0 +1,69 @@
...
+
+  private def validateDefaultValue(field: TypedStructField): Seq[ValidationIssue] = {
+    tryToValidationIssues(field.defaultValueWithGlobal)
+  }
+
+  private def validateExplicitDefaultValue(field: TypedStructField): Seq[ValidationIssue] = {
+    val defaultValue = field.structField.getMetadataString(MetadataKeys.DefaultValue)
+    val encoding: Option[String] = encodingForField(field)
+
+    (for {
```
benedeki, line 41: This for comprehension doesn't really enlighten the goal of the code. Also doesn't make it all that easy to add a check if new encoding will be added. Wouldn't simple `match` server the purpose better?

utils/src/main/scala/za/co/absa/enceladus/utils/validation/field/BinaryFieldValidator.scala
```diff
@@ -0,0 +1,69 @@
...
+class BinaryFieldValidator extends FieldValidator {
+  private def encodingForField(field: TypedStructField): Option[String] =
+    field.structField.getMetadataString(MetadataKeys.Encoding)
+
+  private def validateDefaultValue(field: TypedStructField): Seq[ValidationIssue] = {
+    tryToValidationIssues(field.defaultValueWithGlobal)
+  }
+
+  private def validateExplicitDefaultValue(field: TypedStructField): Seq[ValidationIssue] = {
+    val defaultValue = field.structField.getMetadataString(MetadataKeys.DefaultValue)
```
benedeki, line 38: Why these `val`s?

utils/src/main/scala/za/co/absa/enceladus/utils/validation/field/BinaryFieldValidator.scala
```diff
@@ -0,0 +1,69 @@
...
+    val encoding: Option[String] = encodingForField(field)
+
+    (for {
+      enc <- encoding if enc == MetadataValues.Encoding.Base64
+      value <- defaultValue
+    } yield {
+      // only base64 and explicit default value could result in validation issues here:
+      Try {
+        Base64.getDecoder.decode(value)
+      } match {
```
benedeki, line 48: I understand that you want a custom validation message. But perhaps enhancing the `tryToValidationIssues` or overloading it would make sense.

utils/src/main/scala/za/co/absa/enceladus/utils/validation/field/BinaryFieldValidator.scala
```diff
@@ -0,0 +1,69 @@
...
+    encoding.map(_.toLowerCase) match {
+      case Some(MetadataValues.Encoding.Base64) | Some(MetadataValues.Encoding.None) =>
+        Seq.empty
+      case None => Seq.empty
+      case _ => Seq(ValidationError(s"Unsupported encoding for Binary field ${field.structField.name}: '${encoding.get}'"))
+    }
+  }
+
+  override def validate(field: TypedStructField): Seq[ValidationIssue] = {
+    super.validate(field) ++ validateDefaultValue(field) ++ validateExplicitDefaultValue(field) ++ validateEncoding(field)
```
benedeki, line 67: Why the separation into `validateDefaultValue` and `validateExplicitDefaultValue`?

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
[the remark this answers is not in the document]
dk1844: Changed as suggested, I hope that is what you had in mind (basically I am reusing the encoding reading and the validation as well, on the other hand, the actual application of the encoding is not reused, because it operates on different types (fromString conversion vs Colum->Column conversion)

utils/src/main/scala/za/co/absa/enceladus/utils/validation/ValidationException.scala
```diff
@@ -15,7 +15,8 @@
 
 package za.co.absa.enceladus.utils.validation
 
-class ValidationException(val msg: String, val errors: Seq[String]) extends Exception(msg)
+class ValidationException(val msg: String, val errors: Seq[String])
+  extends Exception(s"$msg Due to errors: ${errors.mkString(",")}")
```
benedeki, line 19: **For discussion:** The `ValidationException` is used in `StandardizationInterpreter.validateSchemaAgainstSelfInconsistencies` where logging is done as well. I am not sure, it's a good idea to include the errors in the message here as well. 🤔

dk1844:
> Usability wise it is a pass from me. I run some tests using different encodings and types it is applied to. Just one question and this is general, about how are defaults supposed to work? Are we supposed to get casting error together with a default applied?
> 
> ```
> last_name                                  |errCol
> [61 6C 66 61 20 61 6C 66 61 20 61 6C 66 61]|[[stdNullError, E00002, Standardization Error - Null detected in non-nullable attribute, last_name, [null], []]]
> ```
> 
> Schema:
> 
> ```
> {
>   "name" : "last_name",
>   "type" : "binary",
>   "nullable" : false,
>   "metadata" : { "encoding" : "none", "default" : "alfa alfa alfa" }
> }
> ```

Yes, I believe so. It was explained to me in this way, that the "default" can actually be thought of as a _fallback value_ that gets used in case that standardized value cannot.

So, in my understanding what you are demonstrating is a prime example of the desired behavior - there would be a `null` in a non-nullable column, thus `null` cannot be assigned and the default value is used instead and the record is marked with the respective error.

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
```diff
@@ -52,6 +53,7 @@ import scala.util.{Random, Try}
   *         NumericParser !
   *         StringParser !
   *         BooleanParser !
+ *        BinaryParser !
```
Zejnilovic, line 56:
```suggestion
  *       BinaryParser !
```

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
```diff
@@ -407,6 +409,34 @@ object TypeParser {
...
+                                        isArrayElement: Boolean)
+                                       (implicit defaults: Defaults) extends PrimitiveParser[Array[Byte]] {
+    override protected def assemblePrimitiveCastLogic: Column = {
+      origType match {
+        case BinaryType => column
+        case StringType =>
+          val encoding = field.structField.getMetadataString(MetadataKeys.Encoding).map(_.toLowerCase)
+
+          encoding match {
+            case Some(MetadataValues.Encoding.Base64) => unbase64(column)
```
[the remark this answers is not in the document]
dk1844: Great question, took some finding out:

For invalid default value, a validation exception is thrown (there is a test case for it, too). 
For the data, where the we use `unbase64` spark function, any invalid base64 characters are ignored:

unbase64-ignores-invalid-base64-chars

Am I correct to assume this is unwanted and the desired outcome should be a standardization error instead?

utils/src/main/scala/za/co/absa/enceladus/utils/types/TypedStructField.scala
```diff
@@ -178,6 +182,25 @@ object TypedStructField {
...
+  // BinaryTypeStructField
+  final class BinaryTypeStructField private[TypedStructField](structField: StructField, encoding: Option[String])
+                                                             (implicit defaults: Defaults)
+    extends TypedStructFieldTagged[Array[Byte]](structField) {
+
+    // used to convert the default value from metadata's [[MetadataKeys.DefaultValue]]
+    override protected def convertString(string: String): Try[Array[Byte]] = {
+      encoding.map(_.toLowerCase) match {
+        case Some(MetadataValues.Encoding.Base64) => Try(Base64.getDecoder.decode(string))
+        case Some(MetadataValues.Encoding.None) | None => Try{string.getBytes} // use as-is
```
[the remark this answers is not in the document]
dk1844: True, it should not. Changed as suggested.

utils/src/main/scala/za/co/absa/enceladus/utils/types/TypedStructField.scala
```diff
@@ -178,6 +182,25 @@ object TypedStructField {
...
+        case _ => throw new IllegalStateException(s"Unsupported encoding for Binary field ${structField.name}: '${encoding.get}'")
```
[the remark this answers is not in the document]
dk1844: True, it makes more sense and it might be also probably a bit more efficient. Thanks.

utils/src/main/scala/za/co/absa/enceladus/utils/types/TypedStructField.scala
```diff
@@ -134,6 +136,8 @@ object TypedStructField {
   def apply(structField: StructField)(implicit defaults: Defaults): TypedStructField = {
     structField.dataType match {
       case _: StringType    => new StringTypeStructField(structField)
+      case _: BinaryType    =>
+        new BinaryTypeStructField(structField, structField.getMetadataString(MetadataKeys.Encoding))
```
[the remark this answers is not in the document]
dk1844: Makes sense, done.

utils/src/main/scala/za/co/absa/enceladus/utils/validation/ValidationException.scala
dk1844: In cases where the exception is expected, it is easy to query the errors. On the other hand, when it is the underlying case at the top of the stack to cause the process to exit (specifically without direct means to debug), it is much less convenient.

I have summed it up in https://github.com/AbsaOSS/enceladus/pull/1370#issuecomment-638891307

Are there any other ideas on the topic or maybe a middle ground?

utils/src/main/scala/za/co/absa/enceladus/utils/validation/field/BinaryFieldValidator.scala
dk1844: To simplify processing, holding a non-trivial call's return value in under a succinct name.

If the question is aimed more if I don't use the `field.defaultValueWithGlobal` as the default value, I do that deliberately -- I am just checking the explicit default instead that may be necessary to be decoded. We do not expect anything like that from the built-in default (only String/StringType values get decoded anyway).

utils/src/main/scala/za/co/absa/enceladus/utils/validation/field/BinaryFieldValidator.scala
dk1844: To me, this makes things more clear and in theory, we could more validations errors from each, too.

It actually seems to me that `tryToValidationIssues` would be better final anyway and the client should override mostly the `validate` method. It does not seem worth hacking into `tryToValidationIssues`, when it can be used for the general processing as it is now and special validation cases can be handled separately and result in specific validation errors.

utils/src/main/scala/za/co/absa/enceladus/utils/validation/field/BinaryFieldValidator.scala
dk1844: Related to the https://github.com/AbsaOSS/enceladus/pull/1370#discussion_r436692778 .

utils/src/main/scala/za/co/absa/enceladus/utils/validation/field/BinaryFieldValidator.scala
dk1844: Redone, it's just an object now. I have seen it with `NumericFieldValidator` and the likes, so I just followed the suit.

utils/src/main/scala/za/co/absa/enceladus/utils/validation/field/BinaryFieldValidator.scala
dk1844: I have rewritten the logic, hope it is clearer.

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
dk1844: Thanks, fixed.

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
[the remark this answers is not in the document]
benedeki: Thank you for testing it. Yes, the default value I expected to behave like this, and it's the correct way.
> 
> 
> Am I correct to assume this is unwanted and the desired outcome should be a standardization error instead?

Yes (if by error you mean entry in `errCol`), that would be the fully correct way. To do it, probably only an UDF is the answer? There probably the `BinaryTypeStructField.stringToTyped` would be nice to use.
If we keep it like this if could result in data loss.

utils/src/main/scala/za/co/absa/enceladus/utils/types/TypedStructField.scala
```diff
@@ -183,16 +184,18 @@ object TypedStructField {
   }
 
   // BinaryTypeStructField
-  final class BinaryTypeStructField private[TypedStructField](structField: StructField, encoding: Option[String])
+  final class BinaryTypeStructField private[TypedStructField](structField: StructField)
                                                              (implicit defaults: Defaults)
     extends TypedStructFieldTagged[Array[Byte]](structField) {
+    val normalizedEncoding = structField.getMetadataString(MetadataKeys.Encoding).map(_.toLowerCase)
```
benedeki, line 190: Missing type for public property.

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
Zejnilovic: If fixes are so easy/straight forward then just go ahead and resolve the convos. 👍

utils/src/main/scala/za/co/absa/enceladus/utils/types/TypedStructField.scala
dk1844: Thanks, added.

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
```diff
@@ -422,9 +422,12 @@ object TypeParser {
       origType match {
         case BinaryType => column
         case StringType =>
-          val validationIssues: Seq[ValidationIssue] = field.validate()
-          if (validationIssues.nonEmpty) {
-            throw new IllegalStateException(s"There are validation issues, cannot continue parsing binary field $field: $validationIssues")
+          val (validationErrors, validationWarnings) = field.validate().span(_.isInstanceOf[ValidationError])
```
benedeki, line 425: Sorry missed that in the previous review.
The validation is done before getting here in the Standardization flow. No point to doing it here again.

utils/src/main/scala/za/co/absa/enceladus/utils/validation/field/BinaryFieldValidator.scala
```diff
@@ -24,32 +24,27 @@ import za.co.absa.enceladus.utils.validation.{ValidationError, ValidationIssue}
...
-    (for {
-      enc <- encoding if enc == MetadataValues.Encoding.Base64
-      value <- defaultValue
-    } yield {
-      // only base64 and explicit default value could result in validation issues here:
-      Try {
-        Base64.getDecoder.decode(value)
+    (encoding, defaultValue) match {
+      case (Some(MetadataValues.Encoding.Base64), Some(encodedValue)) => Try {
+        Base64.getDecoder.decode(encodedValue)
```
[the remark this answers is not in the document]
benedeki: Still there, and as there are other changes to do... 😉

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
```diff
@@ -35,7 +35,7 @@ import za.co.absa.enceladus.utils.typeClasses.{DoubleLike, LongLike}
 import za.co.absa.enceladus.utils.types.TypedStructField._
 import za.co.absa.enceladus.utils.types.{Defaults, TypedStructField}
 import za.co.absa.enceladus.utils.udf.{UDFBuilder, UDFLibrary, UDFNames}
-import za.co.absa.enceladus.utils.validation.ValidationIssue
+import za.co.absa.enceladus.utils.validation.{ValidationError, ValidationIssue}
```
benedeki, line 38: Some unused imports.

dk1844:
> Yes (if by error you mean entry in errCol), that would be the fully correct way. To do it, probably only an UDF is the answer? There probably the BinaryTypeStructField.stringToTyped would be nice to use.
If we keep it like this if could result in data loss.

@benedeki, I have actual created two version of this:
v1: https://github.com/AbsaOSS/enceladus/commit/58d887e5ef27ab07eabb82a91cbbd60a4810ef56
v2: https://github.com/AbsaOSS/enceladus/commit/6e9f3fc45ded2e75a063b9d498f6463ac94a6b6e

And I have trouble deciding. The first version does exactly what you suggest - using a custom UDF with `stringToTyped` within. It works, but it seems quite horrific to me in terms of UDFs (dragging in the `(Binary)Type(d)StructField` hierarchy). In my mind, UDFs should be as small and simple as possible (not only because of problematic optimization, but also because it type-safety in compile-time is inadequate at best.)

The second version does the same, but the UDF is global and standalone. It is true, that there are no ties to `stringToTyped`, on the other hand, it is so clear in what it does -- in my mind, this is better even at the price of slight redundancy.

So, unless there I made the `stringToTyped`-version unnecessarily complicated, I would probably prefer the latter.

dk1844:
Release notes [added](https://github.com/AbsaOSS/enceladus/issues/1235#issuecomment-642449439).

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
```diff
@@ -422,23 +420,16 @@ object TypeParser {
...
-                column.cast(field.dataType) // use as-is
-              case _ => throw new IllegalStateException(s"Unsupported encoding for Binary field ${field.structField.name}:" +
-                s" '${field.normalizedEncoding.get}'")
-            }
+          // already validated in Standardization
+          field.normalizedEncoding match {
+            case Some(MetadataValues.Encoding.Base64) => callUDF(UDFNames.binaryUnbase64, column)
+            case Some(MetadataValues.Encoding.None) | None =>
+              if (field.normalizedEncoding.isEmpty) {
+                logger.info(s"Binary field ${field.structField.name} does not have encoding setup in metadata. Reading as-is.")
```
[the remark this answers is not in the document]
dk1844: True, that would be more suitable. Thanks, applied.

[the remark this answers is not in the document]
benedeki: It was really a question. 😄 But good as it is now, while wasn't wrong before 😉

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/interpreter/stages/TypeParser.scala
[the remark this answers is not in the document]
dk1844: Solved via: https://github.com/AbsaOSS/enceladus/pull/1370#issuecomment-641907316 and subsequent.

PR #1376, opened 2020-06-04
1375: Template for Epic

benedeki:


.github/ISSUE_TEMPLATE/epic.md
```diff
@@ -0,0 +1,20 @@
...
+pipelines: Epics
+---
+
+## Background
+A clear and concise intro into the situation.
+
+## Goal
+The goal that epic wants to achieve.
+
+## Expected Task List [Optional]
```
Zejnilovic, line 14: I think that this is not needed since Github/Zenhub will automatically create the list from the issues connected to the epic.

.github/ISSUE_TEMPLATE/epic.md
```diff
@@ -0,0 +1,20 @@
...
+A list of expected issues that will be needed to achieve this Epic
+1. 
+2. 
+3. 
+
+**NB! Don't forget to assign the Epic into a Milestone**
```
Zejnilovic, line 20: I don't think this is necessary. We have no hard requirements and deadlines.

Zejnilovic: I mostly agree with you. What troubles me, is that we currently use milestones as a "release planner" as well. And you can have only on milestone per Epic meaning that any Epic connected to a Milestone of a theme, cannot be added to a milestone of a release.

.github/ISSUE_TEMPLATE/epic.md
Zejnilovic: Sure

.github/ISSUE_TEMPLATE/epic.md
benedeki: Do we? 😉 
We had a milestone 3.0.0 but there are literally a handful if items assigned. Also there is Spline 0.5.x and Rest API, the later probably will not get there when we finally release it. So Release based milestones are little big and stretched.
Finally we can have still major release milestones. After all the major releas changes are based on breaking features.
So to conclude, this doesn't prevent them just makes it little bit more granular, to make it more useful IMO.

Zejnilovic: Small one: I still don't like having lines there that are made only to be deleted.

Issue #1372, opened 2020-06-03
An exception at the end of a job when sending control info to Kafka


yruslan:
## Describe the bug
This exception is thrown by the built-in plugin that sends INFO files to Kafka.

This is not deterministic, it happens only from time to time. I've seen it happen on Standardization and Conformance.

```
20/06/03 09:26:35 ERROR ControlInfoProducerKafka: Failed to send control measurements to Kafka.
org.apache.kafka.common.KafkaException: Producer closed while send in progress
	at org.apache.kafka.clients.producer.internals.RecordAccumulator.append(RecordAccumulator.java:200)
	at org.apache.kafka.clients.producer.KafkaProducer.doSend(KafkaProducer.java:907)
	at org.apache.kafka.clients.producer.KafkaProducer.send(KafkaProducer.java:846)
	at org.apache.kafka.clients.producer.KafkaProducer.send(KafkaProducer.java:733)
	at za.co.absa.enceladus.plugins.builtin.common.mq.kafka.ControlInfoProducerKafka$$anonfun$send$1.apply(ControlInfoProducerKafka.scala:76)
	at za.co.absa.enceladus.plugins.builtin.common.mq.kafka.ControlInfoProducerKafka$$anonfun$send$1.apply(ControlInfoProducerKafka.scala:76)
	at scala.concurrent.impl.Future$PromiseCompletingRunnable.liftedTree1$1(Future.scala:24)
	at scala.concurrent.impl.Future$PromiseCompletingRunnable.run(Future.scala:24)
	at scala.concurrent.impl.ExecutionContextImpl$AdaptedForkJoinTask.exec(ExecutionContextImpl.scala:121)
	at scala.concurrent.forkjoin.ForkJoinTask.doExec(ForkJoinTask.java:260)
	at scala.concurrent.forkjoin.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1339)
	at scala.concurrent.forkjoin.ForkJoinPool.runWorker(ForkJoinPool.java:1979)
	at scala.concurrent.forkjoin.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:107)
```

## To Reproduce
Run Standardization or Conformance on a cluster

## Expected behaviour
The job should wait for the control metrics message to be fully sent before closing the producer. No errors or exceptions should happen during normal operation.

yruslan:
We need to confirm if it happens in `2.6.0` as well. It might be already fixed.

[...]

PR #1384, opened 2020-06-10
#1372 Fix timing issue when send() happened after close()

yruslan:
This happened because conversion from Java future to Scala one happens
asynchronously, so close() had a chance to execute before send()


yruslan:
Rebased against latest `develop`.

Issue #1299, opened 2020-04-23
Upload schema from a secure schema registry


yruslan:
## Background
This is requested by our users as something that would be very helpful to them.

## Feature
Upload schema from a secure schema registry.

@kevinwallimann gave a great idea how we can simplify all these things.

We should add support for specifying the subject of a schema registry.
The base URL of schema server should be configured in Menas in `application.properties`

User should provide:
- [x] A subject in the schema registry. ~For the naming strategy based on the topic name it will be `topic_name`-`value`.~
- ~[ ] An optional key subject might be specified (e.g. `topic_name`-`key`). If it is specified, the schema of the key should be merged with the value schema.~
- [x] if the provided `topic_name` does not exist,  there should be an automatic (behind the scenes) fallback to `topic_name-value`. If even that does not succeed, the schema loading fails.
 
Menas should:
* Upload the schema from the schema registry.

dk1844:
Release notes
Schema can be updated by specifying schema-name (-value will be appended, optionally also -key). Secure schema registry is supported, too.

PR #1405, opened 2020-06-19
Feature/1299 schema from secure schema registry

dk1844:
- by specifying a subject name, schema can be loaded from schema repository
  - if the schema is not found by the subject name, `-value` is appended to the subject name and retried.
  - the latest version is always loaded

Schema repository is determined by new config field: 
```
menas.schemaRegistryBaseUrl=https://example-schema-repository:8081
```
If the key is not present, the UI option is not shown.

If the schema repository is secured, the following configuration fields must also be populated:
```config
javax.net.ssl.trustStore=/path/to/truststore.jks
javax.net.ssl.trustStorePassword=somePassword
javax.net.ssl.keyStore=/path/to/keystore.jks
javax.net.ssl.keyStorePassword=somePassword
```

## UI preview
key-gone-fallback-in

Tested manually against a known secure schema repository.


menas/src/main/scala/za/co/absa/enceladus/menas/services/SchemaRegistryService.scala
```diff
@@ -0,0 +1,85 @@
...
+   * @return
+   */
+  def loadSchemaByTopicName(topicName: String): SchemaResponse = {
+    val schemaUrl = SchemaRegistryService.getLatestSchemaUrl(schemaRegistryBaseUrl, topicName)
+    loadSchemaByUrl(schemaUrl)
+  }
+
+  /**
+   * Loading the schema by a full URL
+   * @param remoteUrl
```
Zejnilovic, line 54: Missing description

menas/src/main/scala/za/co/absa/enceladus/menas/services/SchemaRegistryService.scala
```diff
@@ -0,0 +1,85 @@
...
+      val connection = url.openConnection()
+      val mimeType = SchemaController.avscContentType // only AVSC is expected to come from the schema registry
+      val fileStream = Source.fromInputStream(connection.getInputStream)
+      val fileContent = fileStream.mkString
+      fileStream.close()
+
+      SchemaResponse(fileContent, mimeType, url)
+    } catch {
+      case NonFatal(e) =>
+        throw RemoteSchemaRetrievalException(SchemaType.Avro, s"Could not retrieve a schema file from $remoteUrl. " +
```
Zejnilovic, line 69: This `$remoteUrl` might show to the user more info than needed/wanted. Maybe show them just the topic?

utils/src/main/scala/za/co/absa/enceladus/utils/schema/SchemaUtils.scala
```diff
@@ -586,4 +586,59 @@ object SchemaUtils {
     }
   }
 
+  /**
+   * Combines schemas into one. Wrapper for [[SchemaUtils#combineStructFieldLists]]
+   *
+   * @param firstType
```
Zejnilovic, line 592: missing descriptiions

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/SchemaController.scala
```diff
@@ -59,36 +59,54 @@ class SchemaController @Autowired()(
...
+    val schemaType: SchemaType.Value = SchemaType.fromOptSchemaName(format)
+
+    val valueSchemaResponse = schemaRegistryService.loadSchemaByTopicName(s"$topicStem-value")
+    val valueSparkStruct = SchemaParser.getFactory(sparkMenasConvertor).getParser(schemaType).parse(valueSchemaResponse.fileContent)
+
+    val combinedStructType = if (mergeWithKey) {
+      val keySchemaResponse = schemaRegistryService.loadSchemaByTopicName(s"$topicStem-key")
+      val keySparkStruct = SchemaParser.getFactory(sparkMenasConvertor).getParser(schemaType).parse(keySchemaResponse.fileContent)
+
+      SchemaUtils.combineStructTypes(valueSparkStruct, keySparkStruct)
```
yruslan, line 94: Maybe the key goes first?

utils/src/main/scala/za/co/absa/enceladus/utils/implicits/StringImplicits.scala
```diff
@@ -164,5 +164,32 @@ object StringImplicits {
...
+     * "abc" / "/123" -> "abc/123"
+     * "abc/" / "/123" -> "abc/123",
+     * but:
+     * "file:///" / "path" -> "file:///path",
+     * }}}
+     *
+     * @param another
+     * @return
+     */
+    def /(another: String): String = { // scalastyle:ignore method.name
```
yruslan, line 191: 👍  Nice

menas/src/main/resources/application.properties.template
```diff
@@ -144,3 +144,13 @@ menas.oozie.splineMongoURL=mongodb://localhost:27017
 #Provide optional spark config options to be passed onto standardization and conformance jobs
 #menas.oozie.extraSparkConfigs={"spark.eventLog.dir" : "hdfs:///spark2-history/", \
 #"spark.history.fs.logDirectory": "hdfs:///spark2-history" }
+
+#----------- Schema Registry
+# URL of schema registry:
+menas.schemaRegistryBaseUrl=https://schemaRegistryHostname:8081
```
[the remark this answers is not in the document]
dk1844: OK, could not hurt.

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/SchemaController.scala
dk1844: Well, that depends. The task definition states 
> _An optional key subject might be specified (e.g. topic_name-key). If it is specified, the schema of the key should be merged with the value schema._

I understand it in the sense that if there is a name-clash (e.g. `field123` is contained in both key and value schema), the value's column should take precedence in case we cannot merge them somehow (both being struct). This behavior is reflected in the parameter order when calling `SchemaUtils.combineStructTypes`. It is quite easy to switch it for the key fields to take precedence. Should we switch so that key fields take precedence instead?

menas/src/main/scala/za/co/absa/enceladus/menas/services/SchemaRegistryService.scala
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+}
+
+object SchemaRegistryService {
+
+  import za.co.absa.enceladus.utils.implicits.StringImplicits._
+
+  def getLatestSchemaUrl(baseUrl: String, topicName: String): String =
+    baseUrl / "subjects" / topicName / "versions/latest/schema"
```
kevinwallimann, line 81: Just leaving a comment about terminology, since I got a notification about this issue.
I think you confused topic with subject. Topic is a Kafka concept, while Subject is a Schema Registry concept. See also: https://docs.confluent.io/current/schema-registry/schema_registry_tutorial.html#terminology-review
The API reference uses "subjectName" and not "topicName" https://docs.confluent.io/current/schema-registry/develop/api.html#get--subjects-(string-%20subject)-versions-(versionId-%20version)-schema

In the Post request, the param is named "topicStem" which should probably be "subjectStem"

menas/src/main/scala/za/co/absa/enceladus/menas/services/SchemaRegistryService.scala
dk1844: What kind of unwanted information do you imagine it can expose? The security here is done via the keystore/trustStore files (that do not get exposed). I do not expect any type of https://username:password@domain kind of configurations (this specific case could be addressed if we really wanted to)

Hmm, I actually find this way beneficial for the user. Given the fact that they can already put in the URL themselves, this error would give them insight into what URL has been tried and they can both check the SR if they have the means to do that (is the machine up? is the service up? is the path/topic-name correct?) and they can use this information to adjust the URL and use it in the "by URL remote loading" thereafter. The presence of the resulting URL can also help significantly with troubleshooting a deployed solution.

**I was actually thinking the opposite - that we might want to show them what the constructed URL (by the topic name stem) will be so that there is total visibility of what the tool aims to do.**

I see your point, but since we primarily aim to aid those using the tool in-house, more information available should be helpful, not hurtful.

--- 
One slight problem here is that the `loadSchemaByUrl` here is used for both cases - loading from URL and loading by topic-name stem, but that is the least of the concerns - that can I can easily adjust.

Can you confirm/refuse my "helpful-visibility-into-things" hypothesis, please?

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/SchemaController.scala
dk1844: Or we could make it changeable via a configuration property (which topic should take precedence on name clashes (`-key` vs `-value`)) - that would be the most elegant solution.

Provided that both scenarios actually make sense, which I am not completely sure of THB.

utils/src/main/scala/za/co/absa/enceladus/utils/schema/SchemaUtils.scala
dk1844: Added, thanks.

menas/src/main/scala/za/co/absa/enceladus/menas/services/SchemaRegistryService.scala
dk1844: Added, thanks.

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/SchemaController.scala
```diff
@@ -59,36 +59,54 @@ class SchemaController @Autowired()(
...
-      fileContent = fileContent.getBytes,
-      fileMIMEType = mimeType)
+      filename = schemaResponse.url.getFile,
+      fileContent = schemaResponse.fileContent.getBytes,
+      fileMIMEType = schemaResponse.mimeType)
 
     uploadSchemaToMenas(principal.getUsername, menasFile, sparkStruct, schemaType)
   }
 
+  @PostMapping(Array("/topic"))
```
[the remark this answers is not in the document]
dk1844: OK, could be.

menas/src/main/scala/za/co/absa/enceladus/menas/services/SchemaRegistryService.scala
```diff
@@ -0,0 +1,85 @@
...
+import za.co.absa.enceladus.utils.config.SecureConfig
+
+import scala.io.Source
+import scala.util.control.NonFatal
+
+@Service
+class SchemaRegistryService @Autowired()() {
+
+  @Value("${menas.schemaRegistryBaseUrl}")
+  val schemaRegistryBaseUrl: String = ""
```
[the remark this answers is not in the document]
dk1844: OK, agreed to change as suggested (the underlying reason: we want to support Menas without a dedicated schema registry running)

menas/src/main/scala/za/co/absa/enceladus/menas/services/SchemaRegistryService.scala
dk1844: Yes. You are completely correct and this is quite eye-opening :)

I will address this as well after I deal with https://github.com/AbsaOSS/enceladus/pull/1405#discussion_r443436688

menas/src/main/scala/za/co/absa/enceladus/menas/services/SchemaRegistryService.scala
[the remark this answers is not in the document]
dk1844: OK, done. I hope I came up with a solution similar to the one you probably had in mind.

Since the configuration is only on backend, the backend exposes an availability endpoint that signifies what schema-update options are available. Based on this, the UI starts with the by-topicName upload disabled, queries the endpoint to enable/keep disabled the newly added option.

Looks like this:
schema-registry-disabled
schema-registry-enabled-by-availability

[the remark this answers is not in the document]
Zejnilovic: I like it. I like the API behind it as well, will give us options for this to be configurable as desired.

menas/src/main/resources/application.properties.template
```diff
@@ -144,3 +144,14 @@ menas.oozie.splineMongoURL=mongodb://localhost:27017
 #Provide optional spark config options to be passed onto standardization and conformance jobs
 #menas.oozie.extraSparkConfigs={"spark.eventLog.dir" : "hdfs:///spark2-history/", \
 #"spark.history.fs.logDirectory": "hdfs:///spark2-history" }
+
+#----------- Schema Registry
+# URL of schema registry [optional]:
+# if omitted, the ability to load schmema by topic name will be hidden.
```
Zejnilovic, line 150:
```suggestion
# if omitted, the ability to load schema by topic name will be hidden.
```

dk1844: Thanks for catching this 👍

dk1844: Fixed.

menas/src/main/scala/za/co/absa/enceladus/menas/services/SchemaRegistryService.scala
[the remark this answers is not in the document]
dk1844: After a discussion with @Zejnilovic, the API was changed to only include the repository option.

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/SchemaController.scala
dk1844: After a discussion with @kevinwallimann and @yruslan, the combine-functionality was removed.

menas/src/main/scala/za/co/absa/enceladus/menas/services/SchemaRegistryService.scala
dk1844: The terminology has been changed to mention schema registry _subject_. Thank you.

dk1844:
Description and [release notes](https://github.com/AbsaOSS/enceladus/issues/1299#issuecomment-646846433) adjusted to the requirement change.

Zejnilovic:
- [x] code reviewed
- [x] pulled
- [x] built
- [x] run - to the limit that I don't have a test env

utils/src/main/scala/za/co/absa/enceladus/utils/implicits/StringImplicits.scala
AdrianOlosutean: Really like this

menas/src/main/scala/za/co/absa/enceladus/menas/services/SchemaRegistryService.scala
```diff
@@ -0,0 +1,100 @@
...
+      val mimeType = SchemaController.avscContentType // only AVSC is expected to come from the schema registry
+      val fileStream = Source.fromInputStream(connection.getInputStream)
+      val fileContent = fileStream.mkString
+      fileStream.close()
+
+      SchemaResponse(fileContent, mimeType, url)
+    } catch {
+      case NonFatal(e) =>
+        throw RemoteSchemaRetrievalException(SchemaType.Avro, s"Could not retrieve a schema file from $remoteUrl. " +
+          s"Please check the correctness of the URL and a presence of the schema at the mentioned endpoint", e)
```
AdrianOlosutean, line 83:
```suggestion
          s"Please check the correctness of the URL and schema presence at the mentioned endpoint", e)
```

utils/src/main/scala/za/co/absa/enceladus/utils/implicits/StringImplicits.scala
kevinwallimann: I like this too. Maybe bring it up in the CQC to see if it's useful for other projects. If so, it would be a good candidate for the absa commons

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/SchemaController.scala
```diff
@@ -163,6 +181,15 @@ class SchemaController @Autowired()(
...
     }
   }
+
+  @GetMapping(path = Array("/availability"))
+  @ResponseStatus(HttpStatus.OK)
+  def getAvailability(): CompletableFuture[SchemaApiAvailability] = {
+    val registryAvailable = schemaRegistryService.schemaRegistryBaseUrl.isDefined
+
+    // the ability to upload or input a remoteUrl is always enabled, so it is not part of the availability endpoint at the moment
+    Future { SchemaApiAvailability(registry = registryAvailable)}
```
yruslan, line 191: This does not need to be executed in a separate thread since the value is already known.
```suggestion
    Future.successful(SchemaApiAvailability(registry = registryAvailable))
```

menas/src/main/scala/za/co/absa/enceladus/menas/models/SchemaApiAvailability.scala
```diff
@@ -0,0 +1,18 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.enceladus.menas.models
+
+case class SchemaApiAvailability(registry: Boolean)
```
yruslan, line 18: As you've mentioned this could be generalized for getting availability of potentially many other API features.
Something like `case class RestApiFeatures(isSchemaRegistryAvailable: Boolean)`.

It is a not strong suggestion. Please, feel free to leave it as is if you prefer.

utils/src/main/scala/za/co/absa/enceladus/utils/config/ConfigUtils.scala
```diff
@@ -67,5 +67,34 @@ object ConfigUtils {
     }
   }
 
+  implicit class ConfigImplicits(config: Config) {
```
yruslan, line 70: 👍  I like this class. Maybe we should move methods from `ConfigReader` here (not necessarily in this PR).

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/SchemaController.scala
```diff
@@ -163,6 +181,15 @@ class SchemaController @Autowired()(
         throw notFound()
     }
   }
+
+  @GetMapping(path = Array("/availability"))
```
yruslan, line 185: Maybe `/features`.

Availability sounds like something related to 'high availability' and brings expectation of something being either 'available' or 'down'. But in this case the availability of a feature is related to the configuration of the service, not to its operation.

utils/src/main/scala/za/co/absa/enceladus/utils/config/ConfigUtils.scala
dk1844: Sure, it makes sense to do something like that in the future. We will need to discuss the specifics, though, because the `ConfigImplicits` uses explicit from the caller's perspective while the `ConfigReader` uses an internally-obtained one.

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/SchemaController.scala
dk1844: True, changed.

menas/src/main/scala/za/co/absa/enceladus/menas/models/SchemaApiAvailability.scala
dk1844: While I agree with the `-Features` part, for now, the API is specific for the Schema manipulation, I fear that `RestApi-` is too broad for the usage we have now. We can definitely rename it when it grows and gets used elsewhere.

So, I went with a middle option: `SchemaApiFeature`.

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/SchemaController.scala
dk1844: I completely agree with this point, changed.

AdrianOlosutean:
> * [x]  code reviewed
> * [ ]  pulled
> * [ ]  built
> * [ ]  run

menas/src/main/resources/application.properties.template
```diff
@@ -144,3 +144,14 @@ menas.oozie.splineMongoURL=mongodb://localhost:27017
...
 #menas.oozie.extraSparkConfigs={"spark.eventLog.dir" : "hdfs:///spark2-history/", \
 #"spark.history.fs.logDirectory": "hdfs:///spark2-history" }
+
+#----------- Schema Registry
+# URL of schema registry [optional]:
+# if omitted, the ability to load schema by subject name from the schema repository will be hidden.
+menas.schemaRegistryBaseUrl=https://localhost:8081
+
+# When using secure schema registry, following paths and passwords must be specified
+javax.net.ssl.trustStore=/path/to/truststore.jks
```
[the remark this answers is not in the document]
dk1844: Already merged it a couple of minutes ago, but I can create a PR with changes like these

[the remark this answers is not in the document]
dk1844: -> https://github.com/AbsaOSS/enceladus/pull/1430

PR #1425, opened 2020-07-01
#1378 Info file templates

### LINKED ISSUE #1378: Make first checkpoint in _INFO file optional
## Background
Generating the initial _INFO file causes problem and consume too much resources for some teams. And the value of such control measures is questionable. Consider an option to simplify the process, make the original input optional.

## Feature
Details to be specified

## Related to
* [Atum task](https://github.com/AbsaOSS/atum/issues/35)

AdrianOlosutean:
INFO file minimal and recommended templates in examples.
Closes #1378

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -27,12 +27,11 @@
 * `<Filename>` - source file name
 * `<Version>` - version of the source file (integer)
 * `<Date>` - date of the input processing
-* `<ProcessStartTimeX>` - Start of the processing; String timestamp in format ‘dd-MM-yyyy HH:mm:ss’ or ‘dd-MM-yyyy HH:mm:ss ZZZ’(X in the name standing for Raw or Source)
-* `<ProcessEndTimeX>` - End of the processing; String timestamp in format ‘dd-MM-yyyy HH:mm:ss’ or ‘dd-MM-yyyy HH:mm:ss ZZZ’(X stands for Raw or Source)
 
 ###Optionally:
-* `<RawRecordCount>` - Number of records in Raw state
-* `<SourceRecordCount>` - Number of records in Source state
+* `<ProcessStartTime>` - Start of the processing; String timestamp in format ‘dd-MM-yyyy HH:mm:ss’ or ‘dd-MM-yyyy HH:mm:ss ZZZ’(doesn't have to be the same for Raw and Source)
+* `<ProcessEndTime>` - End of the processing; String timestamp in format ‘dd-MM-yyyy HH:mm:ss’ or ‘dd-MM-yyyy HH:mm:ss ZZZ’(doesn't have to be the same for Raw and Source)
+* `<RecordCount>` - Number of records in Raw/Source state which should be the same for Raw and Source
 
 [Template for bare minimum required data](examples/info_files/_INFO_file_template_minimal.json)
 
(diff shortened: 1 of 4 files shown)


examples/info_files/_INFO_file_template_recommended.json
```diff
@@ -0,0 +1,43 @@
...
+		"historyType": "Snapshot",
+		"dataFilename": "<Filename>",
+		"sourceType": "<SourceType>",
+		"version": "<Version>",
+		"informationDate": "<Date>"
+	},
+	"checkpoints": [
+    {
+      "name": "Source",
+      "processStartTime": "01-01-1970 00:00:00 +0000",
```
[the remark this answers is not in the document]
Zejnilovic: Same for the minimal

examples/info_files/_INFO_file_template_minimal.json
```diff
@@ -11,17 +11,17 @@
 	"checkpoints": [
     {
       "name": "Source",
-      "processStartTime": "01-01-1970 00:00:00 +0000",
-      "processEndTime": "01-01-1970 0:00:00 +0000",
+      "processStartTime": "<ProcessStartTime>",
```
benedeki, line 14: I would make the minimal simpler, there are no data anyway. So that's why I think _"01-01-1970 00:00:00 +0000"_ is better here.

examples/info_files/_INFO_file_template_recommended.json
```diff
@@ -11,33 +11,33 @@
 	"checkpoints": [
     {
       "name": "Source",
-      "processStartTime": "01-01-1970 00:00:00 +0000",
-      "processEndTime": "01-01-1970 0:00:00 +0000",
+      "processStartTime": "<ProcessStartTime>",
```
benedeki, line 14: **For discussion:** I wonder if the same field name for the timestamps in `Source` and `Raw` doesn't imply they have to be the same? 🤔

examples/README.md
```diff
@@ -24,12 +25,18 @@
 ### Fields (possibly) changed for every run
 
 * `<Filename>` - source file name
-* `<Version>` - version of the source file
+* `<Version>` - version of the source file (integer)
 * `<Date>` - date of the input processing
+* `<ProcessStartTime>` - Start of the processing; String timestamp in format ‘dd-MM-yyyy HH:mm:ss’ or ‘dd-MM-yyyy HH:mm:ss ZZZ’
+* `<ProcessEndTime>` - End of the processing; String timestamp in format ‘dd-MM-yyyy HH:mm:ss’ or ‘dd-MM-yyyy HH:mm:ss ZZZ’
+
+Optionally:
```
benedeki, line 33: Would make it same level as the other fields categories (###)

examples/info_files/_INFO_file_template_recommended.json
AdrianOlosutean: I will slightly change it

examples/README.md
```diff
@@ -12,6 +12,32 @@
                       limitations under the License.
 
 # Examples of Enceladus usage
+## Templates for _INFO files
+#### Templates which illustrate the way to write _INFO files, one with the minimally necessary fields and one with the recommended fields to add
```
Zejnilovic, line 16: This can be ignored but by the Markdown style guide, every level of heading need to be surrounded by a new line. In the case of two headings after each other, one new line is enough.

Zejnilovic: Now that I focus on it, why is everything a heading?

benedeki: Oh, I missed that. Good point.

AdrianOlosutean: Line 16 doesn't need to be a heading indeed

PR #1428, opened 2020-07-01
1422 and 1423 Remove HDFS and Oozie from Menas

### LINKED ISSUE #1422: Remove hard dependency on HDFS from Menas
## Background
We want to go into the cloud where this might not be possible, or we might choose different storage.

## Feature
Remove hard dependency on HDFS from Menas
- it is not needed to run menas
- HDFS browsers in menas are disabled

### LINKED ISSUE #1423: Remove Oozie from Menas
## Background
We will be using a different scheduler then Oozie in the cloud.

## Feature
Remove Oozie from Menas

Zejnilovic:
Removing HDFS and Oozie from Menas, dao and data-model

All this is for the containerization and moving Menas to cloud

Closes #1422 
Closes #1423


data-model/src/main/scala/za/co/absa/enceladus/model/menas/HDFSFolder.scala
```diff
@@ -15,9 +15,9 @@
 
 package za.co.absa.enceladus.model.menas
 
-case class HDFSFolder
+case class HDFSFolder2
```
AdrianOlosutean, line 18: Why HDFSFolder2? Shouldn't it be deleted?

Zejnilovic: Good catch

menas/ui/css/style.css
```diff
@@ -82,17 +82,3 @@ html, body {
...
-  margin-top: -1rem !important;
-  position: absolute !important;
-}
-
-.scheduleMenuHBox {
-  height: 2rem !important;
-  margin-top: 1rem !important;
-}
-
-h5[id$=datasetDetailView--scheduleTimingTitle] {
```
benedeki, line 96: `h5` can go too?

data-model/src/main/scala/za/co/absa/enceladus/model/menas/HDFSFolder.scala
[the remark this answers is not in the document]
Zejnilovic: This was already removed

data-model/src/main/scala/za/co/absa/enceladus/model/Dataset.scala
```diff
@@ -44,8 +43,7 @@ case class Dataset(
   dateDisabled: Option[ZonedDateTime] = None,
   userDisabled: Option[String]        = None,
   conformance:  List[ConformanceRule],
-  parent:       Option[MenasReference] = None,
-  schedule:     Option[OozieSchedule] = None) extends VersionedModel with Auditable[Dataset] {
```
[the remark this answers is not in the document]
Zejnilovic: should be ignored

menas/ui/css/style.css
Zejnilovic: It is H5 for a scheduler

Zejnilovic:
> What about changing the project version in the main pom to make the build recognizable from regular branch?

Smart. Added AWS qualifier. Adheres to semver.

PR #1429, opened 2020-07-02
1424 Add Menas Dockerfile

### LINKED ISSUE #1424: Create a Dockerfile for Menas
## Background
We want Menas to run on EKS, we need Menas dockerized

## Feature
Create a Dockerfile for Menas

Zejnilovic:
Closes #1424

### DIFF
--- a/menas/Dockerfile
+++ b/menas/Dockerfile
@@ -1,0 +1,23 @@
+# To build, first build menas and be in enceladus/menas folder. Then run:
+# $> docker build -t menas .
+# Run image using:
+# $> docker run \
+#       -e JAVA_OPTS=' \
+#           -Dmenas.mongo.connection.string=mongodb://host.docker.internal:27017 \
+#           -Dmenas.mongo.connection.database=menas \
+#       -p 8080:8080 \
+#       menas
+# You can skip the environment properties if you set up your own properties before build
+FROM tomcat:9-jdk8
+
+LABEL \
+    vendor="AbsaOSS" \
+    copyright="2020 ABSA Group Limited" \
+    license="Apache License, version 2.0" \
+    name="Menas"
+
+EXPOSE 8080
+EXPOSE 8009
+
+RUN rm -rf /usr/local/tomcat/webapps/*
+COPY target/*.war /usr/local/tomcat/webapps/ROOT.war


Zejnilovic:
This is a 1st version PoC. In the next version, as a next step, we will be solving how to containerize without such a huge dependency on properties so we do not kill the portability that containers offer.

menas/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+    vendor="AbsaOSS" \
+    copyright="2020 ABSA Group Limited" \
+    license="Apache License, version 2.0" \
+    name="Menas"
+
+EXPOSE 8080
+EXPOSE 8009
+
+RUN rm -rf /usr/local/tomcat/webapps/*
+COPY target/*.war /usr/local/tomcat/webapps/ROOT.war
```
AdrianOlosutean, line 23: I was wondering if you could try to package it as jar and just run the built jar instead of depending on Tomcat. If it's not a low hanging fruit, then it can be delayed

Zejnilovic: As far as I know, a fat jar does not allow for any post-build server configuration. That would also mean problems with logging and persisting these. And in the end, we end up with a containerized application that is not portable. 

Fat jar might be interesting to run a service ASAP for a PoC connection. But for it to be deployed, scaled and properly configured you will always choose a war. You don't want to have to rebuild/restart to change one or two values.

menas/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+LABEL \
+    vendor="AbsaOSS" \
+    copyright="2020 ABSA Group Limited" \
+    license="Apache License, version 2.0" \
+    name="Menas"
+
+EXPOSE 8080
+EXPOSE 8009
+
+RUN rm -rf /usr/local/tomcat/webapps/*
```
lokm01, line 22: Do we want to remove all webapps? Doesn't this include the tomcat UIs?

Zejnilovic: There is no need for any extra management through UI since it is a single tomcat app. But if you want it there, we can leave it there. This would mean playing with the URLs as well

lokm01: If we figure out way to automatically ship logs somewhere central, I'd say you're right. We could put in fluentd agent to forward the logs into S3 or cloudwatch in future. For now I'd say it may be useful to keep the management UIs for POC purposes, but I'm happy to be challenged on this :)

Zejnilovic: None of the original `webapps` have any implicit or explicit correlation with either Catalina(servlet container), Jasper (JSP Engine) or any other system component of Tomcat. So unless we need to have a management tool UI, there is no reason to keep them.
And honestly, I don't see the logs connection.

lokm01: Happy, I'm not sure what's on those UIs tbh. Had a hope that it could provide some useful info/tools for debugging. Even if the online war upload.

benedeki: For final version I would be for all removal. For POC maybe no need to bother?

Zejnilovic: > For final version I would be for all removal. For POC maybe no need to bother?

PoC wise, this is easier for me. Later we might want to use `tomcat:9-jdk8-openjdk-slim` image which produces half of the image size.

PR #1430, opened 2020-07-07
#1299 application.properties.template cleanup

dk1844:
Just a template cleanup suggested in https://github.com/AbsaOSS/enceladus/pull/1405#discussion_r450649393.

### DIFF
--- a/menas/src/main/resources/application.properties.template
+++ b/menas/src/main/resources/application.properties.template
@@ -147,8 +147,8 @@ menas.oozie.splineMongoURL=mongodb://localhost:27017
 
 #----------- Schema Registry
 # URL of schema registry [optional]:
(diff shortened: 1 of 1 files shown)


menas/src/main/resources/application.properties.template
```diff
@@ -151,7 +151,7 @@ menas.oozie.splineMongoURL=mongodb://localhost:27017
 menas.schemaRegistryBaseUrl=https://localhost:8081
```
benedeki, line 151: Discussed with @dk1844 if this should be commented as well (me not sure what is better). Leaving up to him 😉

dk1844: After due consideration, I have commented it out as suggested by @benedeki.

PR #1438, opened 2020-07-10
#1437 Implement FillNull Conformance Rule

### LINKED ISSUE #1437: Implement FillNulls Conformance Rule
## Background
Implement FillNulls Conformance Rule

## Feature
Implement Conformance Rule that takes a column and replaces all the null values with a literal.

Zejnilovic:
New Conformance rule FillNulls that goes through a column and all nulls replaces with literals. Produces new column. Does not overwrite.

image

Closes #1437 

RN: New Conformance Rule - FillNulls that creates a new column out of a source column and replaces all the nulls with a literal specified.


data-model/src/main/scala/za/co/absa/enceladus/model/conformanceRule/package.scala
```diff
@@ -121,6 +122,14 @@ package object conformanceRule {
     override def withUpdatedOrder(newOrder: Int): ConformanceRule = copy(order = newOrder)
   }
 
+  case class FillNullsConformanceRule(order: Int,
+                                      controlCheckpoint: Boolean,
+                                      outputColumn: String,
+                                      inputColumn: String,
```
dk1844, line 128: It seems a little confusing that even though many of the specific conformance rules have some common parameters, their order varies. One time, it is `{order, outputColumn, controlCheckpoint, ...}`, another time it is `{order, controlCheckpoint, outputColumn, ...}`. Maybe the API would be more usable if it followed one suit.

data-model/src/main/scala/za/co/absa/enceladus/model/conformanceRule/package.scala
```diff
@@ -121,6 +122,14 @@ package object conformanceRule {
...
+                                      value: String) extends ConformanceRule {
+    override def withUpdatedOrder(newOrder: Int): ConformanceRule = copy(order = newOrder)
```
dk1844, line 130: Probably not important, but I'll ask anyway:

Why do you strip the specific type here? The trait implementation would, of course, be satisfied if `FillNullsConformanceRule` was used as the return type and the object would not need to be cast if used further.

data-model/src/main/scala/za/co/absa/enceladus/model/conformanceRule/package.scala
Zejnilovic: I copied one and enhanced it. Did not think about it. I can fix it.

data-model/src/main/scala/za/co/absa/enceladus/model/conformanceRule/package.scala
Zejnilovic: If I understand correctly, means can only provide you with a string and we don't care about types until we get into conformance, where we cast it at that point of time.

PR #1445, opened 2020-07-16
#1400 Helper scripts run custom JAR

### LINKED ISSUE #1400: Improve helper scripts to run custom JARs
## Background
Currently, the JAR corresponding to Standardization and Conformance jobs is configured in `enceladus_env.sh`. For development and testing it would be very useful to override the location of the JAR so that SNAPSHOT versions be easier to test.

## Feature
Check an environment variable `SPARK_JOSB_JAR` (the name is deliberately very short) in helper scripts. If it is defined, use the specified JAR instead of ones defined in `enceladus_env.sh`.

The workflow of a dev will be like this:
1. Scp a snapshot jar to the user's home folder.
2. Login to the env.
3. Run `export SPARK_JOBS_JAR=/path/to/jars/in/the/home/folder/spark-jobs.jar`
4. Run `run_standardization.sh` or `run_conformance.sh` as usual. The custom JAR will be used.

AdrianOlosutean:
Closes #1400


scripts/bash/enceladus_env.template.sh
```diff
@@ -26,7 +26,7 @@ SPARK_SUBMIT="$SPARK_HOME/bin/spark-submit"
 
 HDP_VERSION="2.7.3"
 
-SPARK_JOBS_JAR="enceladus-spark-jobs.jar"
+SPARK_JOBS_JAR="${SPARK_JOBS_JAR:="enceladus-spark-jobs.jar"}"
```
Zejnilovic, line 29: Generally, I like the solution and it will work. but then this doesn't seem like a plain environment script.

How about introducing `SPARK_JOBS_JAR_OVERRIDE` or something that is then used in `run_standardization.sh` like `export JAR=${SPARK_JOBS_JAR_OVERRIDE:=$SPARK_JOBS_JAR}`

Basically using your solution but in a lower place.

AdrianOlosutean: So then it would be modified in more places?

Zejnilovic: I don't want evaluation and user overriding in ENV script. Yes, it would be in more places. 2 places vs 1. `run_standardization.sh` and  `run_conformance.sh`. Or you can try and do it in `run_enceladus.sh`.

Now that I am looking into `run_enceladus.sh` I see 

```shell
    --jar)
    JAR="$2"
    ;;
```
What does it do? Does it do the same?

AdrianOlosutean: Yes, so I think it's about modifying `run_standardization.sh` and `run_conformance.sh` to include this override.

scripts/bash/run_conformance.sh
```diff
@@ -18,7 +18,7 @@ SRC_DIR=`dirname "$0"`
 source ${SRC_DIR}/enceladus_env.sh
 
 export CLASS=${CONF_CLASS}
-export JAR=${SPARK_JOBS_JAR}
+export JAR=${SPARK_JOBS_JAR:=$SPARK_JOBS_JAR_OVERRIDE}
```
Zejnilovic, line 21: for it to be an override you would have to do
```suggestion
export JAR=${SPARK_JOBS_JAR_OVERRIDE:=$SPARK_JOBS_JAR}
```

yruslan: `--jar` allows you to override the jar location from the command line. There is also `--class`, but there is a bug in the script and it doest it wrong. You can fix `--class` if you want.

AdrianOlosutean: How about the user having to specify `SPARK_JOBS_JAR_OVERRIDE` so it is a bit more explicit?

yruslan: Left a comment: https://github.com/AbsaOSS/enceladus/issues/1400#issuecomment-659971243

scripts/bash/run_conformance.sh
```diff
@@ -18,7 +18,7 @@ SRC_DIR=`dirname "$0"`
 source ${SRC_DIR}/enceladus_env.sh
 
 export CLASS=${CONF_CLASS}
-export JAR=${SPARK_JOBS_JAR}
+export JAR=${SPARK_JOBS_JAR_OVERRIDE:-$SPARK_JOBS_JAR}
```
benedeki, line 21: I would move this to `run_enceladus.sh` so it's actually present in the code only once.

The separation is due to historical reasons, when **Standardization** and **Conformance** were in different jar files.

PR #1454, opened 2020-07-24
Fix pagination

Zejnilovic:
The previous fix worked with the logic of GitHub pages and did not work with the logic of paginator. 

This should be the fix. Sadly Github does not provide environment to test these on.


Zejnilovic:
I basically reverted the previous commit and added base URLs

PR #1462, opened 2020-07-28
#1461 FillNulls now uses simple literal cast

### LINKED ISSUE #1461: Upgrade RuelInterpreter.inferStrictestType
## Feature
Upgrade RuelInterpreter.inferStrictestType

Zejnilovic:
This is just a feasibility showcase. Will add tests with a date type if everyone likes this as the hotfix.

Closes #1461 

Release Notes: FillNulls now supports simple casts to all supported data formats.

**Edit1:** Test Added. Please review


spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/interpreter/rules/RuleInterpreter.scala
```diff
@@ -76,6 +80,46 @@ trait RuleInterpreter {
...
   }
 
+  /**
+   * Preforms a simple type cast to input base on the DataType
+   * @param input Value to be casted
+   * @param dataType DataType of the value to be casted to
+   * @return Returns Column representation of the newly casted value
+   */
+  def simpleLiteralCast(input: String, dataType: DataType): Column = {
+    Try({
```
benedeki, line 90: I would make this `Try()` into a function returning `Try[Column]`. Then it can be used in `inferStrictestType` too, by smartly cyclining through the types and we got consistent behavior from both `inferStrictestType` and `simpleLiteralCast`

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/interpreter/rules/RuleInterpreter.scala
```diff
@@ -76,6 +80,46 @@ trait RuleInterpreter {
...
+      dataType match {
+        case _: ByteType =>
+          lit(input.toByte)
```
benedeki, line 93: Mind the under-/overflow in the integer types.

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/interpreter/rules/RuleInterpreter.scala
Zejnilovic: I am not sure I get what you mean

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/interpreter/rules/RuleInterpreter.scala
Zejnilovic: Again what do you mean? I should be doing checks for under/overflows?

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/interpreter/rules/FillNullsRuleInterpreter.scala
```diff
@@ -41,22 +44,33 @@ case class FillNullsRuleInterpreter(rule: FillNullsConformanceRule) extends Rule
       rule.outputColumn
     )
 
+    val dataType = SchemaUtils.getFieldType(rule.inputColumn, df.schema).get
+    val default: Column = simpleLiteralCast(rule.value, dataType) match {
```
yruslan, line 48: Sorry, I might be out of context. But why not
```scala
val default: Column = lit(rule.value).cast(dataType)
```

Zejnilovic: We do not support all DataTypes, officially. But they might pass through standardization. So from this point of view, it is to be on the safe side.

Also, there were ideas of passing in some extra metadata (to almost all rules) and this was what I was left with.

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/interpreter/rules/RuleInterpreter.scala
benedeki: It's just a suggestion, how to make `lit` behave similarly (consistently) as `fillNulls`. Doesn't affect you solution, that does exactly what it needs.
1) Makte the `Try(...)` expression into a private function
2) Call the function in your new `simpleLiteralCast`
3) In `inferStrictestType` in a cycle of calls to this function with different datatypes and `Try`'s recover find the strictest type applicable.

Bonus for 3* - it can be done smart, if it fails on Double, no point to try on Float etc.

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/interpreter/rules/FillNullsRuleInterpreter.scala
```diff
@@ -41,22 +44,33 @@ case class FillNullsRuleInterpreter(rule: FillNullsConformanceRule) extends Rule
...
-      conformRootField(df)
+      conformRootField(df, default)
     }
   }
 
-  private def conformNestedField(df: Dataset[Row])(implicit spark: SparkSession): Dataset[Row] = {
+  private def conformNestedField(df: Dataset[Row], default: Column)(implicit spark: SparkSession): Dataset[Row] = {
     df.nestedWithColumnExtended(rule.outputColumn, getField => when(
-      getField(rule.inputColumn).isNull, inferStrictestType(rule.value)
+      getField(rule.inputColumn).isNull, default
```
yruslan, line 67: Alternatively, there is `coalesce()` in Spark that does the same thing, just more concise code.

spark-jobs/src/test/scala/za/co/absa/enceladus/conformance/interpreter/rules/FillNullsRuleSuite.scala
```diff
@@ -40,8 +40,17 @@ class FillNullsRuleSuite extends FunSuite with SparkTestBase with TestRuleBehavi
...
+    order = 2,
+    outputColumn = "date2",
+    controlCheckpoint = false,
+    inputColumn = "date",
+    value = "1900-05-05"
+  )
+
   private val literalRulesList1 = List(literalRule)
   private val literalRulesList2 = List(literalRule, literalArrayRule)
+  private val literalRulesList3 = List(literalRule, literalArrayRule, literalDateRule)
```
benedeki, line 53: **Tiny:** You changed the test names not to refer to literal. I think these `val`s suffer from same. But they are private...

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/interpreter/rules/RuleInterpreter.scala
Zejnilovic: I don't want to rewrite lit here and broaden the scope of this hot-feature. But I think I get what you mean.

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/interpreter/rules/FillNullsRuleInterpreter.scala
Zejnilovic: Sorry missed this earlier. Implemented

spark-jobs/src/test/scala/za/co/absa/enceladus/conformance/interpreter/rules/FillNullsRuleSuite.scala
Zejnilovic: Done

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/interpreter/rules/FillNullsRuleInterpreter.scala
```diff
@@ -44,15 +45,14 @@ case class FillNullsRuleInterpreter(rule: FillNullsConformanceRule) extends Rule
       rule.outputColumn
     )
 
-    val dataType = SchemaUtils.getFieldType(rule.inputColumn, df.schema).get
+    val dataType: DataType = SchemaUtils.getFieldType(rule.inputColumn, df.schema).get
```
benedeki, line 48: **Tiny (not worth its on commit unless other changes: ** `val targetDataType` would be an explanatory name, IMO

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/interpreter/rules/RuleInterpreter.scala
benedeki: Fair enough, you are right.
That's why I am not fond of hot features.

PR #1470, opened 2020-07-30
1469 Fix hyperconformance not getting default version

Zejnilovic:


spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/HyperConformance.scala
```diff
@@ -128,6 +128,7 @@ object HyperConformance extends StreamTransformerFactory with HyperConformanceAt
       datasetName = conf.getString(datasetNameKey),
       datasetVersion = conf.getInt(datasetVersionKey),
       reportDate = new SimpleDateFormat(ReportDateFormat).format(new Date()),
+      reportVersion = Option(getReportVersion(conf)),
```
benedeki, line 131:
```suggestion
      reportVersion = Option(getReportVersion(conf)),
      performanceMetricsFile = None,
      folderPrefix = None,
      persistStorageLevel = None,
```
Perhaps add the other fields that were there originally (I know they equal to default)

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/HyperConformance.scala
```diff
@@ -128,6 +128,7 @@ object HyperConformance extends StreamTransformerFactory with HyperConformanceAt
       datasetName = conf.getString(datasetNameKey),
       datasetVersion = conf.getInt(datasetVersionKey),
       reportDate = new SimpleDateFormat(ReportDateFormat).format(new Date()),
```
benedeki, line 130:
```suggestion
      reportDate = new SimpleDateFormat(ReportDateFormat).format(new Date()),  // Still need a report date for mapping table patterns
```
I think the comment was useful there.

PR #1475, opened 2020-08-05
1474 Fix conf not logging at the start

### LINKED ISSUE #1474: logEffectiveConfigProps not being used
## Describe the bug
ConfigReader.logEffectiveConfigProps

## Expected behaviour
ConfigReader.logEffectiveConfigProps. should be used in the beginning of Standardization & Conformance

Zejnilovic:
Returns the `logEffectiveConfigProps` 

Fixes #1474 

RN: Fixes properties logging at the start of each spark job

### DIFF
--- a/spark-jobs/src/main/scala/za/co/absa/enceladus/common/CommonJobExecution.scala
+++ b/spark-jobs/src/main/scala/za/co/absa/enceladus/common/CommonJobExecution.scala
@@ -56,10 +56,7 @@ trait CommonJobExecution {
 
   protected val log: Logger = LoggerFactory.getLogger(this.getClass)
   protected val conf: Config = ConfigFactory.load()
-
-  private val confReader: ConfigReader = new ConfigReader(conf)
-  confReader.logEffectiveConfigProps(Constants.ConfigKeysToRedact)
-
+  protected val confReader: ConfigReader = new ConfigReader(conf)
   protected val menasBaseUrls: List[String] = MenasConnectionStringParser.parse(conf.getString("menas.rest.uri"))
 
   protected def obtainSparkSession[T](jobName: String)(implicit cmd: JobConfigParser[T]): SparkSession = {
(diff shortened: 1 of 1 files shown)


spark-jobs/src/main/scala/za/co/absa/enceladus/common/CommonJobExecution.scala
```diff
@@ -56,6 +56,10 @@ trait CommonJobExecution {
 
   protected val log: Logger = LoggerFactory.getLogger(this.getClass)
   protected val conf: Config = ConfigFactory.load()
+
+  private val confReader: ConfigReader = new ConfigReader(conf)
+  confReader.logEffectiveConfigProps(Constants.ConfigKeysToRedact)
```
AdrianOlosutean, line 61: I think this should be in prepareJob

Zejnilovic: why not

benedeki:
> 
> 
> I think it's an inappropriate place to put this logic.
> I think it should be somewhere around `StandardizationConfig`/`ConformanceConfig`/`StandardizationConformanceConfig.getFromArguments(args)`
> Or maybe even right after that call in the `main` functions.
> Happy to discuss...

OK the part of `getFromArguments` is nonsense (sorry messed up configuration and argument, both named config). But about the logging done in `obtainSparkSession` not belonging there, I still insist on.

Zejnilovic:
> > I think it's an inappropriate place to put this logic.
> > I think it should be somewhere around `StandardizationConfig`/`ConformanceConfig`/`StandardizationConformanceConfig.getFromArguments(args)`
> > Or maybe even right after that call in the `main` functions.
> > Happy to discuss...
> 
> OK the part of `getFromArguments` is nonsense (sorry messed up configuration and argument, both named config). But about the logging done in `obtainSparkSession` not belonging there, I still insist on.

`StandardizationConfig` and other configs are scopts. This is application conf. So it makes sense to put it where it is loaded. `StandardizationConfig` knows nothing about app confs

benedeki:
> 
> 
> > > I think it's an inappropriate place to put this logic.
> > > I think it should be somewhere around `StandardizationConfig`/`ConformanceConfig`/`StandardizationConformanceConfig.getFromArguments(args)`
> > > Or maybe even right after that call in the `main` functions.
> > > Happy to discuss...
> > 
> > 
> > OK the part of `getFromArguments` is nonsense (sorry messed up configuration and argument, both named config). But about the logging done in `obtainSparkSession` not belonging there, I still insist on.
> 
> `StandardizationConfig` and other configs are scopts. This is application conf. So it makes sense to put it where it is loaded. `StandardizationConfig` knows nothing about app confs

No argue about that (that why I corrected myself on the (half) original suggestion)
But my complain still stands. That logging shouldn't be in the function `obtainSparkSession`.

PR #1485, opened 2020-08-13
#1393 Add new fixed width project

### LINKED ISSUE #1393: Conflicting dependency on 'enceladus-utils:1.7.0'
## Describe the bug
Currently, `spark-jobs` depends on `enceladus-fixedWidth:1.7.0` which depends on 'enceladus-utils:1.7.0'. This creates a class path conflict between current `utils` module and the old `enceladus-utils` which have the same namespace of packages.

## To Reproduce
Inspect the dependencies of `spark-jobs`.

Or (worked for me, but might not always work since class name resolution might be non-deterministic)

Try to go to the source code of `SparkTestBase` from any unit test suite that uses it.

## Expected behaviour
The `spark-jobs` module should depend only on one version of `utils`.

## Screenshots
Screenshot 2020-06-12 at 09 42 24

## Proposed solution
We can do either of these:
- Remove `enceladus-fixedWith` dependency if nobody uses it.
- Revive `fixed-width` (yes, with the new name) in `enceladus` project.
- Revive `fixed-width` in a separate repository. This might require removing its dependency on Enceladus utils.

Zejnilovic:
Implementation of new FixedWidth project in Enceladus

Added charset, treatEmptyValuesAsNulls and nullValue as spark-job switches

Fixes #1393 

RN: Update version of FIxedWidth and add charset, treatEmptyValuesAsNulls and nullValue options.

### DIFF
--- a/spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/config/StandardizationConfig.scala
+++ b/spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/config/StandardizationConfig.scala
@@ -78,7 +78,9 @@ case class StandardizationConfig(rawFormat: String = "xml",
   override def withPerformanceMetricsFile(value: Option[String]): StandardizationConfig = copy(performanceMetricsFile = value)
   override def withFolderPrefix(value: Option[String]): StandardizationConfig = copy(folderPrefix = value)
   override def withPersistStorageLevel(value: Option[StorageLevel]): StandardizationConfig = copy(persistStorageLevel = value)
-  override def withFixedWidthTreatEmptyValuesAsNulls(value: Option[Boolean]): StandardizationConfig = copy(fixedWidthTreatEmptyValuesAsNulls = value)
+  override def withFixedWidthTreatEmptyValuesAsNulls(value: Option[Boolean]): StandardizationConfig = {
+    copy(fixedWidthTreatEmptyValuesAsNulls = value)
+  }
   override def withFixedWidthNullValue(value: Option[String]): StandardizationConfig = copy(fixedWidthNullValue = value)
 }
 
--- a/spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/config/StandardizationConfig.scala
+++ b/spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/config/StandardizationConfig.scala
@@ -37,7 +37,7 @@ case class StandardizationConfig(rawFormat: String = "xml",
                                  csvQuote: Option[String] = None,
                                  csvEscape: Option[String] = None,
                                  cobolOptions: Option[CobolOptions] = None,
-                                 fixedWidthTrimValues: Option[Boolean] = Some(false),
+                                 fixedWidthTrimValues: Option[Boolean] = None,
                                  rawPathOverride: Option[String] = None,
                                  failOnInputNotPerSchema: Boolean = false,
                                  datasetName: String = "",
@@ -49,7 +49,9 @@ case class StandardizationConfig(rawFormat: String = "xml",
                                  folderPrefix: Option[String] = None,
                                  persistStorageLevel: Option[StorageLevel] = None,
                                  credsFile: Option[String] = None,
-                                 keytabFile: Option[String] = None)
+                                 keytabFile: Option[String] = None,
+                                 fixedWidthTreatEmptyValuesAsNulls: Option[Boolean] = None,
+                                 fixedWidthNullValue: Option[String] = None)
   extends StandardizationConfigParser[StandardizationConfig]{
   override def withRawFormat(value: String): StandardizationConfig = copy(rawFormat = value)
   override def withCharset(value: Option[String]): StandardizationConfig = copy(charset = value)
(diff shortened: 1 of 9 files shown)


spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/config/StandardizationConfigParser.scala
```diff
@@ -129,20 +133,28 @@ object StandardizationConfigParser {
         config.withRawPathOverride(Some(value)))
         .text("override the path of the raw data (used internally for performance tests)"),
 
+      opt[Boolean]("empty-values-as-nulls").optional()
+        .action((value, config) => config.withFixedWidthTreatEmptyValuesAsNulls(Some(value)))
+        .text("For FixedWidth file format. Treats empty values as null. Default is false"),
```
dk1844, line 138: The only unclear thing to me is where the defaults are coming from.

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/config/StandardizationConfigParser.scala
```diff
@@ -129,20 +133,28 @@ object StandardizationConfigParser {
...
+
+      opt[String]("null-value").optional()
+        .action((value, config) => config.withFixedWidthNullValue(Some(value)))
+        .text("""For FixedWidth file format. Sets the representation of a null value. Defaults is ""."""),
```
dk1844, line 142: dtto

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/config/StandardizationConfigParser.scala
AdrianOlosutean: Does it make sense to just have it as Boolean instead of Option[Boolean]?

Zejnilovic: The defaults are defaults of the FixedWidth file format. Same as CSV Delimiter, header, etc.

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/config/StandardizationConfig.scala
```diff
@@ -76,6 +78,8 @@ case class StandardizationConfig(rawFormat: String = "xml",
   override def withPerformanceMetricsFile(value: Option[String]): StandardizationConfig = copy(performanceMetricsFile = value)
   override def withFolderPrefix(value: Option[String]): StandardizationConfig = copy(folderPrefix = value)
   override def withPersistStorageLevel(value: Option[StorageLevel]): StandardizationConfig = copy(persistStorageLevel = value)
+  override def withFixedWidthTreatEmptyValuesAsNulls(value: Option[Boolean]): StandardizationConfig = copy(fixedWidthTreatEmptyValuesAsNulls = value)
```
benedeki, line 81: **Minor:** Line too long

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization_conformance/config/StandardizationConformanceConfig.scala
```diff
@@ -82,6 +84,8 @@ case class StandardizationConformanceConfig(datasetName: String = "",
   override def withFixedWidthTrimValues(value: Option[Boolean]): StandardizationConformanceConfig = copy(fixedWidthTrimValues = value)
   override def withRawPathOverride(value: Option[String]): StandardizationConformanceConfig = copy(rawPathOverride = value)
   override def withFailOnInputNotPerSchema(value: Boolean): StandardizationConformanceConfig = copy(failOnInputNotPerSchema = value)
+  override def withFixedWidthTreatEmptyValuesAsNulls(value: Option[Boolean]): StandardizationConformanceConfig = copy(fixedWidthTreatEmptyValuesAsNulls = value)
```
benedeki, line 87: **Minor:** Line too long

spark-jobs/src/test/scala/za/co/absa/enceladus/standardization/StandardizationFixedWidthSuite.scala
```diff
@@ -30,7 +30,7 @@ import za.co.absa.enceladus.utils.implicits.DataFrameImplicits.DataFrameEnhancem
 import za.co.absa.enceladus.utils.udf.UDFLibrary
 
 class StandardizationFixedWidthSuite extends FunSuite with SparkTestBase with MockitoSugar{
-  private implicit val udfLibrary:UDFLibrary = new UDFLibrary()
+  private implicit val udfLibrary: UDFLibrary = new UDFLibrary()
   private val log: Logger = LoggerFactory.getLogger(this.getClass)
```
benedeki, line 34: **Original code issue:** In case if you would be fixing the long lines or anything else, the `val log: Logger` is not used.

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/config/StandardizationConfig.scala
Zejnilovic: This is a good place for the exception we allowed ourselves. We said on CQC that a line shouldn't be longer than 120. But if it is justifiable it is ok to go to 140. Now, remember I fought for numbers 80 and 100. But here I think it is justifiable to have it longer since it does not have any "logic". It is just a `copy` method adding that one parameter, 20 times repeating, for different parameters.

Zejnilovic: I take it back, I was on a wrong branch and didn't see this one goes to 150... 🤦

benedeki: Also in such case it's fine to add a scalastyle ignore rule.

PR #1489, opened 2020-08-18
Merge spline 0.5.3 into aws-poc

Zejnilovic:
So this is just a merge of branch `feature/735-integrate-spline-0-4-x` into `aws-poc`, I am not sure how well it is going to work but for the purposes of PoC it should be enough.


menas/src/test/resources/application.properties
```diff
@@ -68,7 +68,9 @@ menas.environment=localhost
 
 menas.spark.master=local[1]
 
-spline.urlTemplate=http://localhost:8080/spline/dataset/lineage/_search?path=%s&application_id=%s
+# Avoiding the http(s) protocol will cause that the same protocol will be used as for Menas itself, avoiding issues
+menas.lineage.readApiUrl=//localhost:8080/spline-gw/consumer
+lineageService=//localhost:8080/spline-ui/app/lineage-overview?executionEventId=%s
```
benedeki, line 73: I am not sure about this configuration 🤔

spark-jobs/src/main/scala/za/co/absa/enceladus/common/CommonJobExecution.scala
```diff
@@ -86,12 +86,12 @@ trait CommonJobExecution {
     dao.authenticate()
     val dataset = dao.getDataset(cmd.datasetName, cmd.datasetVersion)
     val reportVersion = getReportVersion(cmd, dataset)
-    val pathCfg = getPathConfig(cmd, dataset, reportVersion)
+    val pathCfg: PathConfig = getPathConfig(cmd, dataset, reportVersion)
```
benedeki, line 89: Not a problem, just curious - why did you add the type declaration? :-)

benedeki (approved):
In fast review it looks good. Will look at it in more detail on Friday, if still relevant.

spark-jobs/src/main/scala/za/co/absa/enceladus/common/CommonJobExecution.scala
Zejnilovic: Good question. Might be a misclick.

menas/src/test/resources/application.properties
Zejnilovic: It shouldn't be there.

benedeki (approved):

PR #1506, opened 2020-08-28
1480 Add configurable original data mutability check

### LINKED ISSUE #1480: It's possible to delete the original column with Drop Conformance Rule
## Describe the bug
By the original paradigm it should not be possible to modify the original data within Conformance. But it is possible to delete the original columns using Drop Conformance Rule

## To Reproduce
1. Go to Menas -> Conformance Rules
2. Create a new conformance rule
3. Choose Drop Conformance Rule and enter an original dataset column as its parameter
4. Save
5. Execute Conformance
6 It passes

## Expected behaviour
It should not be possible to drop the original column

Zejnilovic:
RN: Add a check for DropConformanceRule that it is not trying to drop a column from the original schema. If this is desired user can use option `conformance.allowDataFrameMutability` and set it to true. In that case, it will only print a warning.

Fixes #1480


lokm01 (approved):

spark-jobs/src/main/resources/reference.conf
```diff
@@ -95,3 +95,6 @@ timezone="UTC"
 # Optional security settings
 #kafka.security.protocol="SASL_SSL"
 #kafka.sasl.mechanism="GSSAPI"
+
+# Optional - allows original dataframe columns to be dropped
+conformance.allowDataFrameMutability=false
```
AdrianOlosutean, line 100: How about calling it allowOriginalColumnDrop? I'm not so sure about the mutability keyword

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/interpreter/DynamicInterpreter.scala
```diff
@@ -65,6 +65,13 @@ object DynamicInterpreter {
     conformedDf
   }
 
+  def findRulesBreakingImmutability(steps: List[ConformanceRule], schema: StructType): Seq[DropConformanceRule] = {
```
AdrianOlosutean, line 68:
```suggestion
  def findOriginalColumnsDropRules(steps: List[ConformanceRule], schema: StructType): Seq[DropConformanceRule] = {
```

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/interpreter/DynamicInterpreter.scala
```diff
@@ -110,6 +120,27 @@ object DynamicInterpreter {
     optimizerTimeTracker.cleanupWorkaroundDf(conformedDf)
   }
 
+  private def checkMutabilityNotViolated(inputDf: DataFrame, steps: List[ConformanceRule])
```
AdrianOlosutean, line 123:
```suggestion
  private def validateDropRules(inputDf: DataFrame, steps: List[ConformanceRule])
```

spark-jobs/src/main/resources/reference.conf
Zejnilovic: I would rather keep the naming general. 

What if we do this for other rules? We would have to go with having multiple parameters for each rule like `allowOriginalColumnDrop`, `allowOriginalColumnThisRule`, `allowOriginalColumnThatRule`. Because if we have `allowOriginalColumnDrop`, we cannot replace it just like that unless we increase a major version.

dk1844: I think both comments have merit. I am thinking about the granularity we want to be setting. If indeed the only thing to set is to mutate or not (I'd go with the currently suggested solution of a single field that cover everything). If however, we wanted to be allowing e.g. renaming, dropping, changing columns scheme, and other variants, it would make sense to be more specific. 

But we could always introduce more granularity later.

Adrian has a point that the name should probably contain the word _original_, because it is only the original columns that are (not) restricted from being modified. So my guess is just one field and maybe calling it `allowOriginalColumnsModification` or `allowOriginalColumnsMutability`?

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/interpreter/DynamicInterpreter.scala
dk1844: the method name should probably follow the suit of the final configuration field name, (_original_ should be included, _drop_ maybe not). `findOriginalColumnsModificationRules`?

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/interpreter/DynamicInterpreter.scala
```diff
@@ -65,6 +65,13 @@ object DynamicInterpreter {
     conformedDf
   }
 
+  def findRulesBreakingImmutability(steps: List[ConformanceRule], schema: StructType): Seq[DropConformanceRule] = {
+    steps.flatMap {
+      case rule: DropConformanceRule => if (SchemaUtils.fieldExists(rule.outputColumn, schema)) Some(rule) else None
```
dk1844, line 70:
```suggestion
      case rule: ConformanceRule => if (SchemaUtils.fieldExists(rule.outputColumn, schema)) Some(rule) else None
```
My thinking here is that any (existing or future) conformance rule may be in theory breaking the immutability of the original columns, not specifically the `DropConformanceRule` in general (even though maybe it is true now). Thus, the problematic rules could be defined more broadly and in a future-proof way, so this is my attempt to do so.

I am basically hinting [Uncle Bob's Avoid Switch Statement clause](https://medium.com/@myoungsokang_36227/uncle-bob-says-avoid-switch-statements-63d94e3b59e9).

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/interpreter/DynamicInterpreter.scala
dk1844: The method name is one thing to consider; I'd maybe only pass `inputSchema: StructType`, because the DataFrame is actually not needed at all.

dk1844 (approved):

spark-jobs/src/main/resources/reference.conf
Zejnilovic: I like the `originalColumn`  paired it with `mutability` since that is what we use in our talks

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/interpreter/DynamicInterpreter.scala
Zejnilovic: Made it private. Used naming from previous comment

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/interpreter/DynamicInterpreter.scala
Zejnilovic: transformed into anonymous function

AdrianOlosutean (approved):
LGTM

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/interpreter/DynamicInterpreter.scala
```diff
@@ -110,6 +120,27 @@ object DynamicInterpreter {
     optimizerTimeTracker.cleanupWorkaroundDf(conformedDf)
   }
 
+  private def checkMutabilityNotViolated(schema: StructType, steps: List[ConformanceRule])
```
benedeki, line 123: *Rant/thought:* I see a basis for Conformance rules validation here. An API, that the UI would call to indicate that the set of rules is:
* executable
* warn it might not work (based on configuration)
* not correct at all

The whole thing is out of scope of this task, of course. But would be nice to have it in mind 😉

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/interpreter/DynamicInterpreter.scala
```diff
@@ -110,6 +120,27 @@ object DynamicInterpreter {
...
+                                        (implicit ictx: InterpreterContext): Unit = {
+    val rulesInViolation = findOriginalColumnsModificationRules(steps, schema)
+
+    if (rulesInViolation.nonEmpty) {
+      val violationsString = rulesInViolation.map(rule =>
+        s"Rule number ${rule.order} - ${rule.getClass.getSimpleName}"
+      ).mkString("\n")
+      if (ictx.featureSwitches.allowOriginalColumnsMutability) {
+        log.warn(
```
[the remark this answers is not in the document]
Zejnilovic: The setting is in reference or app conf. It is not a CLI option. User might not be aware

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/interpreter/DynamicInterpreter.scala
Zejnilovic: We are now talking UI validations. We even have an issue for it. 

If we validate on creation, there is no need for call from spark-jobs to check

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/ConformancePropertiesProvider.scala
```diff
@@ -55,6 +56,12 @@ class ConformancePropertiesProvider {
     enabled
   }
 
+  private def isOriginalColumnsMutabilityEnabled[T]()(implicit cmd: ConformanceConfigParser[T]): Boolean = {
+    val enabled = conf.getOptionBoolean(allowOriginalColumnsMutabilityKey).getOrElse(false)
+    log.info(s"DataFrame mutability enabled = $enabled")
```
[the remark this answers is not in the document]
Zejnilovic: F convention of other methods already here

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/interpreter/DynamicInterpreter.scala
benedeki: My point-of-view it will always need to be done before executing Conformance. The UI can be bypassed, the rules behavior changed or schema altered.
But as mentioned before out of scope of this. 🐱

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/interpreter/DynamicInterpreter.scala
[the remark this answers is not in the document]
benedeki: I think it's still OK. Because presumably done by higher authority, UI, when implemented validation, can still warn about that.
The way how it is done, it's not possibly to have rule modifying an original column without a warning or worse. For users (not Absa) who are OK with such behavior, that's a false flag.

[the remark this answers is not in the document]
Zejnilovic: This is spark-jobs application conf. UI can not warn them about this being on or off. This would take us to a whole another topic of having a REST endpoint with the config in it.

And it is hardly a false flag. It is a warning ([def][log4jwarn]). It is meant to be there if users are doing something dangerous and potentially harmful and something that we do not fully test. When we implement a full suite of Unit or other tests, let's change this.

[log4jwarn]: https://logging.apache.org/log4j/1.2/apidocs/org/apache/log4j/Level.html#WARN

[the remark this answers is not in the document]
benedeki: I don't want this to be a blocking issue. 😃 
But my point is, if the admin of the application says (configures) it's ok to modify the original columns, I still believe the warning is unjustified. Unless we would have three levels of settings 😉 
To explain my UI remark: exactly because UI wouldn't know about the eventual setting of _SpakJobs_, the validation there (once we have it) should allow such a rule, but warn about potential problem.

benedeki (approved):
- [x] code reviewed
- [x] pulled
- [x] built
- [x] run

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/interpreter/DynamicInterpreter.scala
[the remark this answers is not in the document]
Zejnilovic: Other team members? @dk1844 @AdrianOlosutean @HuvarVer ? 

Do you have any thoughts or do you not care in this instance?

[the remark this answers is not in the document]
AdrianOlosutean: I would agree to remove the warning now and add it to UI, which can be done in another task

PR #1511, opened 2020-08-31
#1510 Help and showing jar version with different approach

### LINKED ISSUE #1508: --help throws an error
## Describe the bug
Running `--help` throws an error plus it prints help. Looks very misleading.

## To Reproduce

```shell
spark-submit --class za.co.absa.enceladus.standardization.StandardizationJob /enceladus/spark-jobs/target/spark-jobs-2.13.0-SNAPSHOT.jar --help
Error: Unknown option --help
Error: Missing option --raw-format
Error: Missing option --dataset-name
Error: Missing option --dataset-version
Error: Missing option --report-date
Error: No authentication method specified (e.g. --menas-auth-keytab)
Job Parameters
Usage: Standardization Job [options]

  -f, --raw-format <value>
                           format of the raw data (csv, xml, parquet, fixed-width, etc.)
  --charset <value>        use the specific charset (default is UTF-8)
  --row-tag <value>        use the specific row tag instead of 'ROW' for XML format
  --delimiter <value>      use the specific delimiter instead of ',' for CSV format
  --csv-quote <value>      use the specific quote character for creating CSV fields that may contain delimiter character(s) (default is '"')
  --csv-escape <value>     use the specific escape character for CSV fields (default is '\')
  --header <value>         use the header option to consider CSV header
  --trimValues <value>     use --trimValues option to trim values in  fixed width file
  --strict-schema-check <value>
                           use --strict-schema-check option to fail or proceed over rows not adhering to the schema (with error in errCol)
  --copybook <value>       Path to a copybook for COBOL data format
  --is-xcom <value>        Does a mainframe file in COBOL format contain XCOM record headers
  --cobol-is-text <value>  Specifies if the mainframe file is ASCII text file
  --cobol-encoding <value>
                           Specify encoding of mainframe files (ascii or ebcdic)
  --cobol-trimming-policy <value>
                           Specify string trimming policy for mainframe files (none, left, right, both)
  --empty-values-as-nulls <value>
                           For FixedWidth file format. Treats empty values as null. Default is false
  --null-value <value>     For FixedWidth file format. Sets the representation of a null value. Defaults is "".
  -D, --dataset-name <value>
                           Dataset name
  -d, --dataset-version <value>
                           Dataset version
  -R, --report-date <value>
                           Report date in 'yyyy-MM-dd' format
  -r, --report-version <value>
                           Report version. If not provided, it is inferred based on the publish path (it's an EXPERIMENTAL feature)
  --menas-auth-keytab <value>
                           Path to keytab file used for authenticating to menas
  --performance-file <value>
                           Produce a performance metrics file at the given location (local filesystem)
  --folder-prefix <value>  Adds a folder prefix before the infoDateColumn
  --persist-storage-level <value>
                           Specifies persistence storage level to use when processing data. Spark's default is MEMORY_AND_DISK.
Exception in thread "main" za.co.absa.enceladus.common.config.ConfigError: Command line parameters error
	at za.co.absa.enceladus.standardization.config.StandardizationConfig$.tryFromArguments(StandardizationConfig.scala:91)
	at za.co.absa.enceladus.standardization.config.StandardizationConfig$.getFromArguments(StandardizationConfig.scala:94)
	at za.co.absa.enceladus.standardization.StandardizationJob$.main(StandardizationJob.scala:32)
	at za.co.absa.enceladus.standardization.StandardizationJob.main(StandardizationJob.scala)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.apache.spark.deploy.JavaMainApplication.start(SparkApplication.scala:52)
	at org.apache.spark.deploy.SparkSubmit.org$apache$spark$deploy$SparkSubmit$$runMain(SparkSubmit.scala:845)
	at org.apache.spark.deploy.SparkSubmit.doRunMain$1(SparkSubmit.scala:161)
	at org.apache.spark.deploy.SparkSubmit.submit(SparkSubmit.scala:184)
	at org.apache.spark.deploy.SparkSubmit.doSubmit(SparkSubmit.scala:86)
	at org.apache.spark.deploy.SparkSubmit$$anon$2.doSubmit(SparkSubmit.scala:920)
	at org.apache.spark.deploy.SparkSubmit$.main(SparkSubmit.scala:929)
	at org.apache.spark.deploy.SparkSubmit.main(SparkSubmit.scala)
2020-08-31 10:26:24,610 INFO util.ShutdownHookManager: Shutdown hook called
2020-08-31 10:26:24,611 INFO util.ShutdownHookManager: Deleting directory /private/var/folders/52/m1m_2cf539gcfvjmr2_9tytr0000gp/T/spark-ded2f06f-3eb0-4afb-ba15-90244f699de7
```

## Expected behaviour
Print only --help when called for it. Do not print issue.

### LINKED ISSUE #1509: Print JAR version on --help options
## Background
When people run --help command they should see a Standardization/Conformance version as well

## Feature
when --help is invoked print something like

```
Standardization version 10.20.30
Job Parameters
Usage: Standardization Job [options]
```

### LINKED ISSUE #1510: Enceladus - SparkJobs version should not be overrideable
## Background
The version of the project is injected from Maven pom file via `application.conf`/`application.properties`.

## Feature
The way how it is done, it is possible to override the version via command line. Would be better to make it hard-coded without the possibility to be influenced by the user.

## Proposed Solution
Check for inspiration https://stackoverflow.com/questions/3697449/retrieve-version-from-maven-pom-xml-in-code/3697482

### LINKED ISSUE #1530: Menas version should not be overrideable
## Background
The version of the project display in UX as well is injected from Maven pom file via `application.properties`.

## Feature
The way how it is done, it is possible to override the version via command line. Would be better to make it hard-coded without the possibility to be influenced by the user.
Menas already displays the version. Only the way how it is obtained need to be changed.

## Proposed Solution
Check for inspiration https://stackoverflow.com/questions/3697449/retrieve-version-from-maven-pom-xml-in-code/3697482

AdrianOlosutean:
Fixes #1508 
Closes #1509 
Closes #1510 
Closes #1530


spark-jobs/src/main/scala/za/co/absa/enceladus/common/config/JobConfigParser.scala
```diff
@@ -51,7 +51,10 @@ object JobConfigParser {
   def jobConfigParser[R <: JobConfigParser[R]]: OParser[_, R] = {
     val builder = OParser.builder[R]
     import builder._
-    OParser.sequence(head("Job Parameters"),
+    OParser.sequence(
+
+      head(s"Enceladus ${this.getClass.getPackage.getImplementationVersion}"),
```
Zejnilovic, line 56:
```suggestion
      head(s"Enceladus ${this.getClass.getPackage.getImplementationVersion}v"),
```

or 

```suggestion
      head(s"Enceladus, v. ${this.getClass.getPackage.getImplementationVersion}"),
```

or something similar

spark-jobs/src/main/resources/reference.conf
```diff
@@ -20,9 +20,6 @@
 
 # IMPORTANT. Add all sensitive configuration keys to Constants.ConfigKeysToRedact.
 
-# Enceladus version (this is populated automatically during the build).
-enceladus.version=${project.version}
```
Zejnilovic, line 24: There are usages of this property leading to messages like 

```
2020-08-31 20:55:41,519 INFO standardization.StandardizationJob$: Enceladus version N/A
```

Look into `utils/src/main/scala/za/co/absa/enceladus/utils/general/ProjectMetadataTools.scala`

dk1844: @Zejnilovic , so your suggestion is to change the implementatoin of `ProjectMetadataTools.getEnceladusVersion`, too?

spark-jobs/src/main/scala/za/co/absa/enceladus/common/config/JobConfigParser.scala
dk1844: I'd be a fan of the latter.

spark-jobs/src/main/resources/reference.conf
Zejnilovic: Now, this is the hard question because `ProjectMetadataTools` is in `utils`. Either change:
- the `ProjectMetadataTools`
- or the `CommonJobExecution.obtainSparkSession`'s `enceladusVersion`

I would go with the later. 

Actually have something like `ProjectMetadataTools` in spark-jobs and use the maven plugin there would be probably better. But it is also sed in PerformanceMetricTools

AdrianOlosutean: I would say it's better to have it only one way, so modify `ProjectMetadataTools`. What do you say about this?

AdrianOlosutean:
@Zejnilovic can you double check if this version works properly?

Zejnilovic:
> @Zejnilovic can you double check if this version works properly?

This works but:

- why does `--help` take so much time  `2.48s user 0.32s system 24% cpu 11.330 total` - used `time` in shell
- `PerformanceMetricTools.scala` line 76 still uses `ProjectMetadataTools.getEnceladusVersion`
- what happens with `ProjectMetadataTools.getEnceladusVersion`?
- `CommonJobExecution` has a non-used import on line 38

benedeki (commented):
Already addressed more tickets, so please do the Menas part too (#1530)
And connect this PR to the Epic instead (#1519)

Finally I noticed the other of the tickets are not even _In progress_. Please try to set these for better overview.

spark-jobs/src/main/resources/reference.conf
Zejnilovic: There is no problem. Just possibility of breaking something that relies on it.

benedeki: Not sure I follow how (doesn't mean it's not the case).
On the other hand I see, it can be nicely leverage in Menas too.

spark-jobs/src/main/scala/za/co/absa/enceladus/common/config/JobConfigParser.scala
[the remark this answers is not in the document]
AdrianOlosutean: Actually no

AdrianOlosutean:
I tried updating `UserInfoController`, but it seems to fail like that. It sems to work the way it currently is. Does anyone have any ideas or suggestions?

utils/src/main/scala/za/co/absa/enceladus/utils/general/ProjectMetadata.scala
```diff
@@ -15,15 +15,8 @@
...
+object ProjectMetadata {
 
-object ProjectMetadataTools {
-  private val conf: Config = ConfigFactory.load()
-
-  private val enceladusVersionKey = "enceladus.version"
-
-  /** Returns Enceladus version from the application configuration properties */
-  def getEnceladusVersion: String = conf.getOptionString(enceladusVersionKey).getOrElse("N/A")
+  def enceladusVersion: String = getClass.getPackage.getImplementationVersion
```
benedeki, line 20: Why not `val`? I think it would be justified here. 🤔

benedeki (commented):
Please clear the `reference.conf` and `application.properties` from the version entries.

dk1844 (dismissed):

benedeki (changes requested):
Built and tested the `SparkJobs` part.
Help works now and shows the version. The version is not override-able anymore via config.
That closes #1508, #1509 and #1510.

Please look into #1530 as well, to solve it completely.

utils/src/main/scala/za/co/absa/enceladus/utils/general/ProjectMetadata.scala
```diff
@@ -15,8 +15,10 @@
...
-object ProjectMetadata {
-
-  val enceladusVersion: String = getClass.getPackage.getImplementationVersion
+trait ProjectMetadata {
 
+  def enceladusVersion(): String = {
+    getClass.getPackage.getImplementationVersion
+  }
 }
+object ProjectMetadata extends ProjectMetadata
```
Zejnilovic, line 24: Why implement a selfless trait pattern? If this is not used as a mixin it will output version of the utils module, will it not?

spark-jobs/src/main/scala/za/co/absa/enceladus/common/CommonJobExecution.scala
```diff
@@ -59,7 +59,7 @@ trait CommonJobExecution {
   protected val menasBaseUrls: List[String] = MenasConnectionStringParser.parse(conf.getString("menas.rest.uri"))
 
   protected def obtainSparkSession[T](jobName: String)(implicit cmd: JobConfigParser[T]): SparkSession = {
-    val enceladusVersion = ProjectMetadata.enceladusVersion
+    val enceladusVersion = ProjectMetadata.enceladusVersion()
```
Zejnilovic, line 62: Doesn't this stop providing correct number if utils ever become a different version? Why not just extend the trait with `ProjectMetadata`?

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/UserInfoController.scala
```diff
@@ -16,30 +16,27 @@
...
 import za.co.absa.enceladus.model.user.UserInfo
+import za.co.absa.enceladus.utils.general.ProjectMetadata
 
 @RestController
 @RequestMapping(Array("/api/user"))
-class UserInfoController extends BaseController {
-
-  @Value("${menas.version}")
-  val menasVersion: String = ""
+class UserInfoController extends BaseController with ProjectMetadata {
```
Zejnilovic, line 28: You have both import and extends of ProjectMetadata

utils/src/main/scala/za/co/absa/enceladus/utils/general/ProjectMetadata.scala
```diff
@@ -15,8 +15,10 @@
 
 package za.co.absa.enceladus.utils.general
 
-object ProjectMetadata {
-
-  val enceladusVersion: String = getClass.getPackage.getImplementationVersion
+trait ProjectMetadata {
 
+  def enceladusVersion(): String = {
```
benedeki, line 20: **For consideration:** It's utils, kind of shared module. What about calling it `projectVersion`
(No a change request, it's really just an idea. 😉 )

utils/src/main/scala/za/co/absa/enceladus/utils/performance/PerformanceMetricTools.scala
```diff
@@ -24,7 +24,7 @@ import za.co.absa.enceladus.utils.fs.FileSystemVersionUtils
 import za.co.absa.enceladus.utils.general.ProjectMetadata
 import za.co.absa.enceladus.utils.schema.SchemaUtils
 
-object PerformanceMetricTools {
+object PerformanceMetricTools extends ProjectMetadata {
```
benedeki, line 27: Do this have to extend the `trait` (cake pattern)? Cannot it use the object?

utils/src/main/scala/za/co/absa/enceladus/utils/general/ProjectMetadata.scala
```diff
@@ -15,8 +15,10 @@
 
 package za.co.absa.enceladus.utils.general
 
-object ProjectMetadata {
-
-  val enceladusVersion: String = getClass.getPackage.getImplementationVersion
+trait ProjectMetadata {
```
benedeki, line 18: Please add a comment, why the trait exists - that the object in some cases does not work (_Spring_)

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/UserInfoController.scala
benedeki, line 28: Please add a comment why using `trait` instead of the object.

benedeki (changes requested):
Looks good, what remains are cosmetic changes. 👍

utils/src/main/scala/za/co/absa/enceladus/utils/general/ProjectMetadata.scala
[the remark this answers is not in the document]
AdrianOlosutean: I thought it would be good to emphasize that what runs is somewhat like a procedure

AdrianOlosutean: I agree it's a better name

AdrianOlosutean:
I'm not sure exactly, but I noticed that now it takes a bit of time to start-up when running the jar. Can anyone try it too and confirm if it's the case?

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/UserInfoController.scala
AdrianOlosutean: Yes, you need to import it even if you extend it

Zejnilovic: I thought I saw `._`

spark-jobs/src/main/scala/za/co/absa/enceladus/common/CommonJobExecution.scala
AdrianOlosutean: The idea is to extend only in Menas. I would say(although not 100% sure) that it will not provide a different result cause the result of that function is dependent on the menas pom configuration, it mostly just calls that code

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/UserInfoController.scala
benedeki: No relevant anymore.

utils/src/main/scala/za/co/absa/enceladus/utils/general/ProjectMetadata.scala
benedeki: No relevant anymore.

scripts/bash/run_standardization_conformance.sh
```diff
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-SRC_DIR=`dirname "$0"`
+SRC_DIR=`dirname "$0"
```
benedeki, line 16: Why this change?

AdrianOlosutean: Looks like a leftover

benedeki (approved):
- [x] code reviewed
- [x] pulled
- [x] built
- [x] run

Run all three Spark jobs with `--help` parameter, tried to override the version. Run Menas

PR #1517, opened 2020-09-08
1482 Strict Parsing for Decimal

### LINKED ISSUE #1482: Add a strict parsing mode to Decimal type.
## Background
When a string is parsed to Decimal type, in case the string contains more decimal places then the Decimal's scale parameter the extra decimal places are trimmed.

## Feature
Implement a strict parsing option - triggered by metadata entry, defaulting to `false` - that will generate an error in `ErrCol` in case the decimal places overflow the scale of the define Decimal type.

## Example
Under strict rule the parsing would be as follows:
Decimal(10,2)
"123.45" -> 123.45, no error in `ErrCol`
"1234,567-> 1234.56, error in `ErrCol`

AdrianOlosutean:
Closes #1482


utils/src/main/scala/za/co/absa/enceladus/utils/types/TypedStructField.scala
```diff
@@ -349,14 +349,21 @@ object TypedStructField {
...
       }
 
     override def validate(): Seq[ValidationIssue] = {
-      NumericFieldValidator.validate(this)
+      DecimalFieldValidator.validate(this)
+    }
+
+    override protected def convertString(string: String): Try[BigDecimal] = {
+      if (strictParsing && scale < string.split("\\.")(1).length)
+          Failure(new IllegalArgumentException(s"Strict parsing does not allow $string"))
```
Zejnilovic, line 365: This message will need an update. It is not strict parsing that doesn't allow random values. It is the `s"Mismatch between the scale of the data and the schema defined scale resulting in loss of precission"` or something like this

AdrianOlosutean: True, the message should be more explanatory. How about `$string exceeds the defined scale limit in the schema`

Zejnilovic: Even better

utils/src/main/scala/za/co/absa/enceladus/utils/types/TypedStructField.scala
```diff
@@ -349,16 +349,25 @@ object TypedStructField {
...
+
+    override def needsUdfParsing: Boolean = strictParsing || super.needsUdfParsing
 
     override def validate(): Seq[ValidationIssue] = {
-      NumericFieldValidator.validate(this)
+      DecimalFieldValidator.validate(this)
     }
 
+    override protected def convertString(string: String): Try[BigDecimal] =
+      parser.flatMap(parser => parser.parse(string)).flatMap(_ => super.convertString(string))
```
benedeki, line 369: Why is this needed?

benedeki (commented):
Also, please can you prepare a gh-pages update on the new metadata, please.

utils/src/main/scala/za/co/absa/enceladus/utils/types/TypedStructField.scala
AdrianOlosutean: True, it's redundant

utils/src/main/scala/za/co/absa/enceladus/utils/types/parsers/DecimalParser.scala
```diff
@@ -16,11 +16,15 @@
...
+
 import za.co.absa.enceladus.utils.numeric.NumericPattern
 
+import scala.util.{Failure, Success, Try}
+
 class DecimalParser(override val pattern: NumericPattern,
                     override val min: Option[BigDecimal],
-                    override val max: Option[BigDecimal])
+                    override val max: Option[BigDecimal],
+                    val maxScale: Option[Int])
```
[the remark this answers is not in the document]
AdrianOlosutean: You mean `None`?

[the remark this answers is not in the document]
benedeki: Yup, sorry.

HuvarVer (approved):
I think useablility is good enough. Don't forget to update the documentation.
Done tests of new feature and smoke tests of UI, done e2e tests by Hermese 0.2.2

utils/src/main/scala/za/co/absa/enceladus/utils/types/TypedStructField.scala
```diff
@@ -349,16 +349,25 @@ object TypedStructField {
...
       DecimalTypeStructField.maxPossible(dataType)
     ){
+    val strictParsing: Boolean = getMetadataStringAsBoolean(MetadataKeys.StrictParsing).getOrElse(false)
 
-    override val parser: Try[DecimalParser] =
+    override val parser: Try[DecimalParser] = {
+      val maxScale = if(strictParsing) Some(scale) else None
       pattern.map { patternOpt =>
-        DecimalParser(patternOpt.getOrElse(NumericPattern(defaults.getDecimalSymbols)), Option(typeMin), Option(typeMax))
+        val pattern: NumericPattern = patternOpt.getOrElse(NumericPattern(defaults.getDecimalSymbols))
```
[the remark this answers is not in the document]
AdrianOlosutean: Yes, and otherwise the line would be too long

utils/src/main/scala/za/co/absa/enceladus/utils/types/TypedStructField.scala
```diff
@@ -366,7 +366,7 @@ object TypedStructField {
     }
 
     override protected def convertString(string: String): Try[BigDecimal] =
-      parser.flatMap(parser => parser.parse(string)).flatMap(_ => super.convertString(string))
+      parser.flatMap(parser => parser.parse(string))
```
benedeki, line 369: This equals to inherited code. No need to override. 😉

benedeki (approved):

HuvarVer (approved):

Issue #715, opened 2019-07-29
When updating schema on dataset there is no check for collisions with conformance rules


Zejnilovic:
## Describe the bug
If I have a schema with columns `a`,`b`,`c`, create a column `d` (string) using conformance rule and then update schema to contain `d` (timestamp) column of different kind and change it for the dataset, there is nothing stopping me and the mistake will be only evident in the middle of conformance when it tells me there already is column `d`.

## To Reproduce
Steps to reproduce the behavior OR commands run:
1. Have dataset `X` with columns `a`,`b`,`c`
2. Add conformance rule that creates columns `d`
3. Change schema of dataset `X` to have column `d` from raw. 
4. Run standardization and conformance
5. See error from conformance in console/logger/STDOUT

## Expected behaviour
When changing/updating schema check if it is possible depending on output columns of conformance rules.

GeorgiChochov:
Yes, but we shouldn't disallow the schema from being changed. Let's add a warning on the dataset saying that it's in a `non-runnable` state.

Zejnilovic:
Now we found out that this could lead to almost a cyclic issue where no rule can be deleted because rule dependency validation fails when trying to validate.

PR #1533, opened 2020-09-22
1518 Basic export/import feature

### LINKED ISSUE #1518: Implement an API for Single Entity Export and Import
## Background
This feature would help people when transitioning from one environment to another.

## Feature
Implement an API for single Entity Export and Import. 

- This entity should be version-less, holding just type and name. Version should be the next new version so that the unique combination of Name and Version is preserved.
- Single means that the entity should not take any other linked entities together with it, but it should carry the list of entities that need to be updated.
- The feature should export only one version, leaving parent information behind.
- On import, there should be a validation of new links
- Attachments should not be exported. **Handle export button for such entities**

### API Endpoints
_JSON examples can be view bellow_

- Export `.../api/{entity}/export/{name}/{version}`
  - returns the JSON payload as seen bellow
- Import `.../api/{entity}/import/`
  - Accepts the JSON payload as seen bellow
  - returns JSON descriptor of the newly created entity

### JSON Payload

There are 3 top-level keys here:

- `entity` represents the entity as it is in the mongo database
- `old_env_dependencies` represents the list of the dependencies collected from the `entity` that will need linking on the new environment.
- `new_env_dependencies` represents the list of new dependency links on the new environment and these will be used to create the entity with all links intact.

Keys in `old_env_dependencies` and `new_env_dependencies` are:

- `type` which represents the type of dependency. This is either schema or mapping_table
- `name` is the name of the entity
- `version` is the version of the entity
- `order` is mainly an index of the conformance rule corresponding to an attribute of the same name in conformance rule. _To follow the robustness principle we should probably allow schema to have an order equaling to 0 (or anything else)_

The idea is that on **Export** people would receive the `entity` and `old_env_dependencies` filled out and `new_env_dependencies` empty. This forces a thoughtful usage of this feature and users would have to think if the entities are already on the new environment or not.

Old and new dependency arrays do not have to be sorted. For the type schema there should be only one and mapping tables are made unique thanks to the order which should comply with the order of conformance rules.

On Import people would send the JSON payload as is with updated `new_env_dependencies` and receive a JSON descriptor of the newly created entity. Mainly name, version and type of an entity.

##### JSON Payload

```json
{
  "entity": {
    "full_entity": "like_in_mongo"
  },
  "old_env_dependencies" : [
    {"type": "schema", "name": "schemaX", "version": 10},
    {"type": "mapping_table", "name": "MTZ", "version": 3, "order": 0}
  ],
  "new_env_dependencies": [
    {"type": "schema", "name": "schemaX", "version": 2},
    {"type": "mapping_table", "name": "MTZ", "version": 2, "order": 0}
  ],
}
```

##### JSON Returned from import

```json
{ 
  "name": "Alfa", 
  "version": 10, 
  "type": "Dataset"
}
```

## Extra comments

- What happens if we have more conformance rules with linked entities like mapping table. Can we do this more generally? Conformance rules should be able to report that they have a link.
- Some Metadata of the data model or export feature version could be needed for validation.
- Not sure I like the names `old_env_dependencies` and `new_env_dependencies`

Zejnilovic:
- to export item `GET /menas/api/{entity}/exportItem/{name}/{version}`
- to export latestet `GET /menas/api/{entity}/exportItem/{name}`
- to import item `POST /menas/api/{entity}/importItem`

Entities:
- dataset
- schema
- mapping table

Validations
- validate schema
  - name 
  - no parent passed
- validate mapping table
  - name 
  - no parent passed
  - passed schema exists
  - passed default values input columns exist in the schema
- validate dataset
  - name 
  - no parent passed
  - passed schema exists
  - passed mapping tables exist
  - transitive schema -> conformance rules operate only on fields that exist

Fixes #1518


menas/src/main/scala/za/co/absa/enceladus/menas/services/DatasetService.scala
```diff
@@ -102,4 +102,10 @@ class DatasetService @Autowired() (datasetMongoRepository: DatasetMongoRepositor
     }
   }
 
+  override def importSingleItem(item: Dataset, username: String): Future[Option[Dataset]] = {
+    getLatestVersionValue(item.name).flatMap {
+      case Some(version) => update(username, item.copy(version = version))
+      case None => super.create(item.copy(version = 1), username)
+    }
+  }
```
Zejnilovic, line 110: This here bugs me. This is something that is in `DatasetService`, `SchemaService` and `MappingTableService` but cannot be abstracted to `VersionedModelService` because of the usage of method `copy` which need implementation of `C` like `Dataset`.  Not sure what to do.

Any ideas on how to do it?

Zejnilovic:
The validation of transitive schema will be very important for #715

menas/src/main/scala/za/co/absa/enceladus/menas/services/DatasetService.scala
```diff
@@ -102,4 +102,150 @@ class DatasetService @Autowired() (datasetMongoRepository: DatasetMongoRepositor
...
+                                       validations: Future[Validation]): Future[Validation] = {
+    val maybeFields = maybeSchema.map { case Some(x) => x.fields.flatMap(f => f.getAllChildren :+ f.getAbsolutePath()).toSet }
+    val accumulator = Tuple2(validations, maybeFields)
+
+    conformanceRules.foldLeft(accumulator) { case ((validation, currentColumns), conformanceRule) =>
+      conformanceRule match {
+        case cr: CastingConformanceRule =>
+          validateInAndOut(validation, currentColumns, cr)
+        case cr: NegationConformanceRule =>
+          validateInAndOut(validation, currentColumns, cr)
```
Zejnilovic, line 154: Should these methods be directly on the Conformance Rules?

dk1844:
Is there a documentation for this REST API somewhere? (or planned in gh-pages?) It could come quite handy for the users, I'd say.

Zejnilovic:
> Is there a documentation for this REST API somewhere? (or planned in gh-pages?) It could come quite handy for the users, I'd say.

There is none yet because we do not have validations on the API side of the things meaning they could leave something in an inconsistent state. That is why we generally frown on usage of the API. 

Also not sure if everything is restful and up to standards. Audit of the API is one of the tasks in the backlog. After that, we can do propper documentation. 

- I would not like to add only export/import documentation because then they will ask for more (I know I would)
- I would not like to add everything right now to the documentation as the API needs the audit and validations

data-model/src/test/scala/za/co/absa/enceladus/model/DatasetTest.scala
```diff
@@ -0,0 +1,45 @@
...
+
+  private val dataset = Dataset(name = "Test DS",
+    version = 1,
+    hdfsPath = "newPath",
+    hdfsPublishPath = "newPublishPath",
+    schemaName = "newSchema",
+    schemaVersion = 1,
+    conformance = List(conformanceRule1))
+
+  private val jsonDataset = """{"name":"Test DS","hdfsPath":"newPath","hdfsPublishPath":"newPublishPath",""" +
```
dk1844, line 37:
```suggestion
  private val expectedDataset = """{"name":"Test DS","hdfsPath":"newPath","hdfsPublishPath":"newPublishPath",""" +
```

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/VersionedModelController.scala
```diff
@@ -98,6 +98,29 @@ abstract class VersionedModelController[C <: VersionedModel with Product with Au
...
+  }
+
+  @GetMapping(Array("/exportItem/{name}"))
+  @ResponseStatus(HttpStatus.OK)
+  def exportLatestEntity(@PathVariable name: String): CompletableFuture[String] = {
+    versionedModelService.exportLatestItem(name)
+  }
+
+  @PostMapping(Array("/importItem"))
+  @ResponseStatus(HttpStatus.CREATED)
```
dk1844, line 114: Is the L[ocation header sent](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/201) when the item is imported? Is it done somehow behind the scenes or is it missing?

data-model/src/main/scala/za/co/absa/enceladus/model/MappingTable.scala
```diff
@@ -76,4 +80,21 @@ case class MappingTable(name: String,
...
   }
+
+  override def exportItem(): String = {
+    val mapper = new ObjectMapper()
+    mapper.registerModule(DefaultScalaModule)
+
+    val root: ObjectNode = mapper.createObjectNode()
+    val defaultMappingValueJsonList: ArrayNode = mapper.valueToTree(defaultMappingValue.toArray)
+
+    root.put("name", name)
```
dk1844, line 91: Since you are already using the jackson object mapper, is there a reason why you do the marshalling manually instead of  using Jackson's `toJson()` methods (while marking some of the fields you want to omit from marshalling by the means of `@JsonIgnoreProperties` and similar)?

Hinting something along the lines of [1](https://coderwall.com/p/o--apg/easy-json-un-marshalling-in-scala-with-jackson) and [2](https://stackoverflow.com/questions/25654811/jsonignore-serialising-scala-case-class-property-using-jackon-and-json4s)

dk1844 (dismissed):
I like your architectural design of the feature 👍

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/VersionedModelController.scala
Zejnilovic: It should be behind the scenes

dk1844:
> There is none yet because we do not have validations on the API side of the things meaning they could leave something in an inconsistent state. That is why we generally frown on usage of the API.
> 
> Also not sure if everything is restful and up to standards. Audit of the API is one of the tasks in the backlog. After that, we can do propper documentation.

I understand. That is why I suggested to put it in the backlog, too.

> * I would not like to add only export/import documentation because then they will ask for more (I know I would)
Sorry, but I have a bit hard time agreeing to this reasoning, even though I understand the underlying motivation. Even with running a risk of new feature requests, I believe documenting is worth it nonetheless.

data-model/src/main/scala/za/co/absa/enceladus/model/MappingTable.scala
Zejnilovic: If I am not wrong we use full (all attribute) JSON SerDe as it is tested here [JsonSerializerSuite](https://github.com/AbsaOSS/enceladus/blob/develop/dao/src/test/scala/za/co/absa/enceladus/dao/rest/JsonSerializerSuite.scala) for other (presumably mongo related) purposes. 

Now how I understood the ignore, it is applied `case class` wide and cannot be specified specially for one method.

Zejnilovic:
> Sorry, but I have a bit hard time agreeing to this reasoning, even though I understand the underlying motivation. Even with running a risk of new feature requests, I believe documenting is worth it nonetheless.

That's fine. I can add it in a stand-alone pr.

data-model/src/main/scala/za/co/absa/enceladus/model/MappingTable.scala
dk1844: Yes, that makes sense. That is the point I was missing. 👍

data-model/src/test/scala/za/co/absa/enceladus/model/DatasetTest.scala
Zejnilovic: done for schema and mapping table as well

dk1844 (approved):

benedeki (commented):
Work in progress. Haven't one through all yet

data-model/src/test/scala/za/co/absa/enceladus/model/DatasetTest.scala
```diff
@@ -0,0 +1,45 @@
...
+ * limitations under the License.
+ */
+
+package za.co.absa.enceladus.model
+
+import org.scalatest.FunSuite
+import za.co.absa.enceladus.model.conformanceRule.LiteralConformanceRule
+
+class DatasetTest extends FunSuite {
+  private val conformanceRule1 = LiteralConformanceRule(
```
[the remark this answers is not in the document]
Zejnilovic: Because they are not scoped to the test. They can be reused later.

data-model/src/test/scala/za/co/absa/enceladus/model/DatasetTest.scala
```diff
@@ -0,0 +1,45 @@
...
+  private val dataset = Dataset(name = "Test DS",
+    version = 1,
+    hdfsPath = "newPath",
+    hdfsPublishPath = "newPublishPath",
+    schemaName = "newSchema",
+    schemaVersion = 1,
+    conformance = List(conformanceRule1))
+
+  private val expectedDataset = """{"name":"Test DS","hdfsPath":"newPath","hdfsPublishPath":"newPublishPath",""" +
+  """"schemaName":"newSchema","schemaVersion":1,"conformance":[{"_t":"LiteralConformanceRule","order":0,""" +
```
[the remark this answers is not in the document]
Zejnilovic: Yes?

data-model/src/test/scala/za/co/absa/enceladus/model/DatasetTest.scala
```diff
@@ -0,0 +1,45 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.enceladus.model
+
+import org.scalatest.FunSuite
+import za.co.absa.enceladus.model.conformanceRule.LiteralConformanceRule
+
+class DatasetTest extends FunSuite {
```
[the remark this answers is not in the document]
Zejnilovic: In this module, tests were named with `Test` ending

menas/src/main/scala/za/co/absa/enceladus/menas/services/DatasetService.scala
```diff
@@ -102,4 +102,150 @@ class DatasetService @Autowired() (datasetMongoRepository: DatasetMongoRepositor
     }
   }
 
+  override def importItem(item: Dataset, username: String): Future[Option[Dataset]] = {
+    getLatestVersionValue(item.name).flatMap {
+      case Some(version) => update(username, item.copy(version = version))
```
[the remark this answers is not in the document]
Zejnilovic: `VersionedMongoRepository#update` on line "computes" new version. (doing the +1)

dk1844 (approved):

utils/src/main/scala/za/co/absa/enceladus/utils/general/ProjectMetadata.scala
```diff
@@ -16,8 +16,11 @@
 package za.co.absa.enceladus.utils.general
 
 trait ProjectMetadata {
-
   def projectVersion: String = {
     getClass.getPackage.getImplementationVersion
   }
+
+  def projectMajorVersion: Int = {
```
AdrianOlosutean, line 23: Would a check make sense here?

Zejnilovic: In the next commit, this will be deleted/reverted

data-model/src/main/scala/za/co/absa/enceladus/model/conformanceRule/package.scala
```diff
@@ -40,6 +40,11 @@ package object conformanceRule {
     val controlCheckpoint: Boolean
 
     def withUpdatedOrder(newOrder: Int): ConformanceRule
+
+    def hasEntityConnected: Boolean = false
+    def maybeConnectedEntityName: Option[String] = None
```
AdrianOlosutean, line 45: I would prefer not to name it as maybe.... Why not just like connectedEntityName?

AdrianOlosutean (commented):
Good work. I have a few questions and discussion points

data-model/src/main/scala/za/co/absa/enceladus/model/conformanceRule/package.scala
Zejnilovic: I think maybe signifies that it is an option and I think I got it on CQC.

data-model/src/main/scala/za/co/absa/enceladus/model/Exportable.scala
```diff
@@ -0,0 +1,31 @@
...
+import com.fasterxml.jackson.module.scala.DefaultScalaModule
+
+trait Exportable {
+  @JsonIgnore
+  protected def objectMapperBase: ObjectMapper = new ObjectMapper()
+    .registerModule(DefaultScalaModule)
+    .configure(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS, false)
+
+  @JsonIgnore
+  protected def objectMapperRoot: ObjectNode = objectMapperBase.createObjectNode()
```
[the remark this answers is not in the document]
Zejnilovic: Not sure I follow

menas/src/main/scala/za/co/absa/enceladus/menas/services/DatasetService.scala
Zejnilovic: I have been thinking about (trying a bit), to move it to conformance rules. But since it is built on top of features it does not make sense to pull it there. But to move it to a different file in the Menas maybe Menas/Services might do the trick of improving readability.

Maybe I could define on the ConformanceRules which type of validation they use, the same as "connected entities"?

data-model/src/main/scala/za/co/absa/enceladus/model/conformanceRule/package.scala
AdrianOlosutean: My view is that a type itself is also a part of the documentation and this style reminds me of Hungarian notation. @dk1844 and @benedeki , what are your opinions on this?

data-model/src/main/scala/za/co/absa/enceladus/model/Exportable.scala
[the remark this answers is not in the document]
AdrianOlosutean: I'm trying to suggest if you could avoid doing like `objectItemMapper.put` for each field and just put the whole case class instance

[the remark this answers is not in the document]
Zejnilovic: Yes, the "auto" SerDe is already being used for a different use case. Representing data in Mongo. If I did not have puts here, I would have "gets" in the import method.

menas/src/main/scala/za/co/absa/enceladus/menas/services/VersionedModelService.scala
```diff
@@ -72,6 +72,66 @@ abstract class VersionedModelService[C <: VersionedModel with Product with Audit
...
+  private[services] def validateSingleImport(item: C, metadata: Map[String, String]): Future[Validation] = {
+    val validation = Validation()
+      .withErrorIf(!hasValidNameChars(item.name), "item.name", s"name '${item.name}' contains unsupported characters")
+      .withErrorIf(item.parent.isDefined, "item.parent", "parent should not be defined on import")
+    val withMetadataValidation = validateMetadata(validation, metadata)
+    Future(withMetadataValidation)
+  }
+
+  private[services] def validateMetadata(validation: Validation, metadata: Map[String, String]): Validation = {
+    val exportVersionErrorMessage =
```
benedeki, line 111: I would make this an inner function used as a parameter into `withErrorIf`.
`def exportVersionErrorMessage(version: String).` Then perhaps the ugly `.stripMargin` won't be needed even.

menas/src/main/scala/za/co/absa/enceladus/menas/services/VersionedModelService.scala
```diff
@@ -72,6 +72,66 @@ abstract class VersionedModelService[C <: VersionedModel with Product with Audit
...
+        if (validation.isValid()) {
+          importItem(item, username)
+        } else {
+          throw ValidationException(validation)
+        }
+      }
+    } yield result
+  }
+
+  private[services] def validateSingleImport(item: C, metadata: Map[String, String]): Future[Validation] = {
```
benedeki, line 102: Why is the return type `Future[]`? IMHO just unnecessary complicates things all around.

benedeki (commented):
Happy to discuss for details in a session.

data-model/src/test/scala/za/co/absa/enceladus/model/DatasetTest.scala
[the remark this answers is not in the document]
Zejnilovic: Yes, maximally scoped in runtime code. But for tests, I think this is ok.

menas/src/main/scala/za/co/absa/enceladus/menas/services/VersionedModelService.scala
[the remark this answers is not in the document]
Zejnilovic: It was done like this previously if I remember correctly. I can switch it to protected.

menas/src/main/scala/za/co/absa/enceladus/menas/services/DatasetService.scala
```diff
@@ -102,4 +101,150 @@ class DatasetService @Autowired() (datasetMongoRepository: DatasetMongoRepositor
...
+        case cr: SparkSessionConfConformanceRule =>
+          (validateOutputColumn(validation, currentColumns, cr.outputColumn), currentColumns.map(f => f + cr.outputColumn))
+        case cr: DropConformanceRule => (
+          for {
+            f <- currentColumns
+            v <- validation
+          } yield {
+            v.withErrorIf(
+              !f.contains(cr.outputColumn),
+              "item.conformanceRules",
```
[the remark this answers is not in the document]
Zejnilovic: It is a path to the "wrong" thing in the JSON sent as a payload as part of the REST API Call

menas/src/main/scala/za/co/absa/enceladus/menas/services/VersionedModelService.scala
Zejnilovic: When you look at the overrides of this method, you can see it cannot return non-Future validations.

data-model/src/main/scala/za/co/absa/enceladus/model/conformanceRule/package.scala
dk1844: I agree with the guys that just simple `connectedEntityName: Option[String]` would suffice.

A last resort scenario could be `optConnectedEntityName`, only for cases when you have an optional variant and then a defined variant, and you somehow need both to work together. Further in code, you work with validations that are actually of type `Future[Validation]`, but you do not normally include the `future` word in the field/method name either.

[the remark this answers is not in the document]
dk1844: I was about to suggest a similar thing - having the `hasEntityConnected` be derived from the connected entity - by checking if it is defined or not (in the original case of `Option`)

data-model/src/test/scala/za/co/absa/enceladus/model/DatasetTest.scala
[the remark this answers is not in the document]
dk1844: It's true that in tests, you can e.g. afford to rely on implicit conversion methods that you wouldn't normally dare to put in the production code 😆

menas/src/main/scala/za/co/absa/enceladus/menas/services/DatasetService.scala
```diff
@@ -102,4 +101,150 @@ class DatasetService @Autowired() (datasetMongoRepository: DatasetMongoRepositor
...
+    val validationsWithSchema = validateSchema(item.schemaName, item.schemaVersion, validations, maybeSchema)
+    val validationWithConnected = validateConnectedEntities(confRulesWithConnectedEntities, validationsWithSchema)
+    validateConformanceRules(item.conformance, maybeSchema, validationWithConnected)
+  }
+
+  private def validateConnectedEntities(confRulesWithConnectedEntities: List[ConformanceRule],
+                                    validationWithSchema: Future[Validation]): Future[Validation] = {
+    confRulesWithConnectedEntities.foldLeft(validationWithSchema) { (acc, cr) =>
+      val maybeConnectedEntity = cr.maybeConnectedEntityType.get match {
+        case "mappingTable" => datasetMongoRepository.getConnectedMappingTable(
```
dk1844, line 125: This part of code seems fragile. At the very least, there should be a `case _ =>`, otherwise all future connected entity type would be looked up among mapping tables. If you make the entity type enum, you'll get exhaustion check here :) 

Another solution could probably have mappings for the appropriate connected types to their their validation routines.

menas/src/main/scala/za/co/absa/enceladus/menas/services/DatasetService.scala
```diff
@@ -102,4 +101,150 @@ class DatasetService @Autowired() (datasetMongoRepository: DatasetMongoRepositor
...
+      }
+
+      val issueMsg =
+        s"""Connected ${cr.maybeConnectedEntityType.get} ${cr.maybeConnectedEntityName.get}
+           | v${cr.maybeConnectedEntityVersion.get} could not be found""".stripMargin.replaceAll("[\\r\\n]", "")
+
+      for {
+        ce <- maybeConnectedEntity
+        accValidations <- acc
+      } yield accValidations.withErrorIf(ce.isEmpty, s"item.${cr.maybeConnectedEntityType.get}", issueMsg)
```
dk1844, line 138:
```suggestion
      } yield ce.map{entity=>accValidations.withError(s"item.${cr.maybeConnectedEntityType.get}", issueMsg)}.getOrElse(accValidations)```

I don't see the added value of `withErrorIf` when you have to blindly call `Option.get` (I know that it is "by design" guarded by the `key` and `error` params being `=>` by name, but still, I don't think it read nicely.

File: data-model/src/main/scala/za/co/absa/enceladus/model/Dataset.scala [hunk 18]
```diff
@@ -17,6 +17,10 @@ package za.co.absa.enceladus.model
 
 import java.time.ZonedDateTime
 
+import com.fasterxml.jackson.databind.{ObjectMapper, SerializationFeature}
+import com.fasterxml.jackson.databind.node.{ArrayNode, ObjectNode}
+import com.fasterxml.jackson.module.scala.DefaultScalaModule
+import com.fasterxml.jackson.module.scala.experimental.ScalaObjectMapper
```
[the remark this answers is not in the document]
Zejnilovic: Fixed

data-model/src/main/scala/za/co/absa/enceladus/model/Dataset.scala
```diff
@@ -106,4 +112,23 @@ case class Dataset(
           AuditFieldName("schedule", "Schedule"))) ++
         super.getSeqFieldsAudit(newRecord, AuditFieldName("conformance", "Conformance rule")))
   }
+
+  override def exportItem(): String = {
+    val mapper = new ObjectMapper()
+      .registerModule(DefaultScalaModule)
+      .configure(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS, false)
+
+    val root: ObjectNode = mapper.createObjectNode()
```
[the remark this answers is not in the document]
Zejnilovic: Fixed

data-model/src/main/scala/za/co/absa/enceladus/model/Dataset.scala
```diff
@@ -106,4 +112,23 @@ case class Dataset(
           AuditFieldName("schedule", "Schedule"))) ++
         super.getSeqFieldsAudit(newRecord, AuditFieldName("conformance", "Conformance rule")))
   }
+
+  override def exportItem(): String = {
+    val mapper = new ObjectMapper()
+      .registerModule(DefaultScalaModule)
+      .configure(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS, false)
```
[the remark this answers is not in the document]
Zejnilovic: Fixed

data-model/src/main/scala/za/co/absa/enceladus/model/SchemaField.scala
```diff
@@ -35,4 +36,9 @@ case class SchemaField
   def getAbsolutePath(): String = {
     if(path.isEmpty) name else s"${path}.${name}"
```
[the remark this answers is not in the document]
Zejnilovic: Fixed

menas/src/main/scala/za/co/absa/enceladus/menas/models/Validation.scala
```diff
@@ -32,4 +32,7 @@ case class Validation (errors: Map[String, List[String]] = Map()) {
     this.copy(errors = errors + (key -> (error :: errors.getOrElse(key, Nil))))
   }
 
+  def withErrorIf(condition: Boolean, key: String, error: String): Validation = {
```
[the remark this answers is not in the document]
Zejnilovic: Fixed

menas/src/main/scala/za/co/absa/enceladus/menas/repositories/DatasetMongoRepository.scala
```diff
@@ -112,4 +125,6 @@ class DatasetMongoRepository @Autowired()(mongoDb: MongoDatabase)
       .toFuture()
   }
 
+
```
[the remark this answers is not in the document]
Zejnilovic: Fixed

menas/src/main/scala/za/co/absa/enceladus/menas/services/DatasetService.scala
```diff
@@ -16,17 +16,17 @@
 package za.co.absa.enceladus.menas.services
 
 import scala.concurrent.Future
-
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.stereotype.Service
-
+import za.co.absa.enceladus.menas.exceptions.ValidationException
```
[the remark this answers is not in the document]
Zejnilovic: Fixed

menas/src/main/scala/za/co/absa/enceladus/menas/services/DatasetService.scala
```diff
@@ -16,17 +16,17 @@
...
+import za.co.absa.enceladus.menas.models.Validation
 import za.co.absa.enceladus.menas.repositories.DatasetMongoRepository
 import za.co.absa.enceladus.menas.repositories.OozieRepository
-import za.co.absa.enceladus.model.Dataset
-import za.co.absa.enceladus.model.UsedIn
-import za.co.absa.enceladus.model.conformanceRule.ConformanceRule
+import za.co.absa.enceladus.model.{Dataset, Schema, SchemaField, UsedIn}
```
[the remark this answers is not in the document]
Zejnilovic: Fixed

utils/src/main/scala/za/co/absa/enceladus/utils/general/ProjectMetadata.scala
Zejnilovic: Fixed

data-model/src/main/scala/za/co/absa/enceladus/model/conformanceRule/package.scala
Zejnilovic: Fixed

menas/src/main/scala/za/co/absa/enceladus/menas/services/VersionedModelService.scala
```diff
@@ -72,6 +72,66 @@ abstract class VersionedModelService[C <: VersionedModel with Product with Audit
...
+                                       schemaVersion: Int,
+                                       validations: Future[Validation],
+                                       maybeSchema: Future[Option[Schema]]): Future[Validation] = {
+    for {
+      schema <- maybeSchema
+      validation <- validations
+    } yield validation.withErrorIf(
+      schema.isEmpty,
+      "item.schema",
+      s"schema ${schemaName} v${schemaVersion} defined for the dataset could not be found"
```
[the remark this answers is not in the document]
Zejnilovic: Fixed

menas/src/main/scala/za/co/absa/enceladus/menas/services/VersionedModelService.scala
```diff
@@ -19,20 +19,20 @@ import org.mongodb.scala.result.UpdateResult
...
 import za.co.absa.enceladus.model.menas.audit._
+import za.co.absa.enceladus.model.ModelVersion
 
 import scala.concurrent.Future
-
 import com.mongodb.MongoWriteException
 
-abstract class VersionedModelService[C <: VersionedModel with Product with Auditable[C]](versionedMongoRepository: VersionedMongoRepository[C])
-  extends ModelService(versionedMongoRepository) {
+//noinspection ScalaStyle
```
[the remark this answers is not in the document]
Zejnilovic: Fixed

data-model/src/main/scala/za/co/absa/enceladus/model/conformanceRule/package.scala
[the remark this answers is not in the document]
Zejnilovic: I tried something. New seal trait and case classes made from it. Search for `ConnectedEntity`

menas/src/main/scala/za/co/absa/enceladus/menas/services/VersionedModelService.scala
```diff
@@ -72,6 +72,66 @@ abstract class VersionedModelService[C <: VersionedModel with Product with Audit
...
+
+  private[services] def validateSingleImport(item: C, metadata: Map[String, String]): Future[Validation] = {
+    val validation = Validation()
+      .withErrorIf(!hasValidNameChars(item.name), "item.name", s"name '${item.name}' contains unsupported characters")
+      .withErrorIf(item.parent.isDefined, "item.parent", "parent should not be defined on import")
+    val withMetadataValidation = validateMetadata(validation, metadata)
+    Future(withMetadataValidation)
+  }
+
+  private[services] def validateMetadata(validation: Validation, metadata: Map[String, String]): Validation = {
```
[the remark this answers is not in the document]
Zejnilovic: Fixed. New method on `Validation`

menas/src/main/scala/za/co/absa/enceladus/menas/services/VersionedModelService.scala
Zejnilovic: Fixed

menas/src/main/scala/za/co/absa/enceladus/menas/services/DatasetService.scala
Zejnilovic: I tried something new. Please review

menas/src/main/scala/za/co/absa/enceladus/menas/services/DatasetService.scala
```diff
@@ -102,4 +101,147 @@ class DatasetService @Autowired() (datasetMongoRepository: DatasetMongoRepositor
...
+    confRulesWithConnectedEntities.foldLeft(validations) { (acc, cr) =>
+      val connectedEntities = cr.connectedEntities.map {
+        case mt: ConnectedMappingTable => (mt, datasetMongoRepository.getConnectedMappingTable(mt.name, mt.version))
+      }
+
+      connectedEntities.foldLeft(acc) { case (v, (entityDef, entityDBInstance)) =>
+        for {
+          instance <- entityDBInstance
+          accValidations <- v
+        } yield accValidations.withErrorIf(instance.isEmpty, s"item.${entityDef.kind}", standardizedErrMessage(entityDef))
```
Zejnilovic, line 134: Seems like I don't have enough deep understanding on for comprehension. 
Why would this not work?
image

data-model/src/main/scala/za/co/absa/enceladus/model/MappingTable.scala
```diff
@@ -76,4 +78,19 @@ case class MappingTable(name: String,
...
+  override def exportItem(): String = {
+    val defaultMappingValueJsonList: ArrayNode = objectMapperBase.valueToTree(defaultMappingValue.toArray)
+
+    val objectItemMapper = objectMapperRoot.`with`("item")
+
+    objectItemMapper.put("name", name)
+    description.map(d => objectItemMapper.put("description", d))
+    objectItemMapper.put("hdfsPath", hdfsPath)
+    objectItemMapper.put("schemaName", schemaName)
+    objectItemMapper.put("schemaVersion", schemaVersion)
```
benedeki, line 91: Comment on this got burried, I'm afraid. What's the point of the version, when there probably won't be any relevancy between the versions of export system and import system.
Although I might see some point. Would like to discuss. 😉

data-model/src/test/scala/za/co/absa/enceladus/model/DatasetTest.scala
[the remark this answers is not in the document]
benedeki: Final word on that, but don't feel pressured  :to move it. :-)
If the would to be shared eventually, it's easy to move them out. This is speculative anticipation. But really no big deal.  ;-)

menas/src/main/scala/za/co/absa/enceladus/menas/services/VersionedModelService.scala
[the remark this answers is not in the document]
benedeki: Up to you. I find the `protected` more appropriate, but it's affected by my personal bias 😉

benedeki: I still don't see. Can we have a call on that?

menas/src/main/scala/za/co/absa/enceladus/menas/services/DatasetService.scala
benedeki: I am generally confused by this part of code. Another topic for discussion, please?

data-model/src/main/scala/za/co/absa/enceladus/model/MappingTable.scala
Zejnilovic: - Double-checking if what you exported has what you need. 
- Calling an export on dataset XYZ version 3 and then getting the schema that you need by extracting the name and version from the JSON
- Default value

data-model/src/main/scala/za/co/absa/enceladus/model/conformanceRule/ConnectedEntity.scala
```diff
@@ -0,0 +1,26 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.enceladus.model.conformanceRule
+
+sealed trait ConnectedEntity{
+  val name: String
+  val version: Int
+  val kind: String
```
dk1844, line 21: I think this is most usually dubbed as
```
`type`
``` 
(backticks needed as it is a reserved word, but I can live with `kind` as well  :) This approach is definitely an improvement over the previous state IMHO.

data-model/src/main/scala/za/co/absa/enceladus/model/conformanceRule/package.scala
```diff
@@ -80,6 +82,10 @@ package object conformanceRule {
                                     outputColumn: String,
                                     isNullSafe: Boolean = false) extends ConformanceRule {
     override def withUpdatedOrder(newOrder: Int): MappingConformanceRule = copy(order = newOrder)
+
+    override def connectedEntities: Seq[ConnectedEntity] = Seq(
+      ConnectedMappingTable(mappingTable, mappingTableVersion)
```
dk1844, line 87:
```suggestion
      ConnectedMappingTable(mappingTableName, mappingTableVersion)
```

menas/src/main/scala/za/co/absa/enceladus/menas/services/DatasetService.scala
dk1844: I think you cannot directly mix-and-match different kind of functors in a single for comprehension. You might need a monad transformer to do this. I actually find [this article on the topic](https://blog.buildo.io/monad-transformers-for-the-working-programmer-aa7e981190e7) both equally entertaining and educative.

menas/src/main/scala/za/co/absa/enceladus/menas/services/DatasetService.scala
dk1844: This is now marked as outdated, but has not been addressed, I think.

data-model/src/main/scala/za/co/absa/enceladus/model/conformanceRule/package.scala
Zejnilovic: Why this change?

menas/src/main/scala/za/co/absa/enceladus/menas/services/DatasetService.scala
dk1844: Looks much better. When adding a new connected entity type, you will still need to add cases here, but at least you will be notified about this during compilation because sealed trait should be checked for exhaustion. So fine by me.

menas/src/main/scala/za/co/absa/enceladus/menas/services/DatasetService.scala
Zejnilovic: Correct. I know about it.

data-model/src/main/scala/za/co/absa/enceladus/model/conformanceRule/package.scala
dk1844: Because the field actually signifies the name of the mapping table. `mappingTable` might suggest that it could be an object with subfields or something similar. So I am just suggesting for the field name to change slightly.

The importance of this suggestion is low, so don't be concerned with this too much.

Zejnilovic: While I do 100% agree with you. This is a name of a field which did not change since 1.0.0 and is automatically deserialized. This change would mean we need to update our data model as well and have a migration script for mongo.

menas/src/main/scala/za/co/absa/enceladus/menas/services/DatasetService.scala
Zejnilovic: Fixed

menas/src/main/scala/za/co/absa/enceladus/menas/services/DatasetService.scala
benedeki: Let's do it in the future. Good enough for now.

data-model/src/test/scala/za/co/absa/enceladus/model/DatasetTest.scala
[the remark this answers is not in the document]
benedeki: Agreed to keep.

menas/src/main/scala/za/co/absa/enceladus/menas/services/VersionedModelService.scala
benedeki: I was wrong on getting rid of the `Future`, sorry. Of course `Await` makes no sense here.

menas/src/main/scala/za/co/absa/enceladus/menas/services/DatasetService.scala
```diff
@@ -102,4 +113,177 @@ class DatasetService @Autowired() (datasetMongoRepository: DatasetMongoRepositor
...
+      }
```
benedeki, line 160: I took the liberty to suggest an alternate solution for the whole function. Has the advantage, that if there are more of the same connected entities, there will be only one validation for all of them.
```suggestion
    val allConnectedEntites: Set[ConnectedEntity] = confRulesWithConnectedEntities.foldLeft(Set.empty[ConnectedEntity]) { (acc, cr) =>
      acc ++ cr.connectedEntities.toSet
    }
    
    allConnectedEntites.foldLeft(Future(Validation())) { (accValidations, entityDef) =>
      entityDef match {
        case mt: ConnectedMappingTable =>
          val entityDBInstance = datasetMongoRepository.getConnectedMappingTable(mt.name, mt.version)
          for {
            instance <- entityDBInstance
            validations <- accValidations
          } yield validations.withErrorIf(instance.isEmpty, s"item.${entityDef.kind}", standardizedErrMessage(entityDef))
      }
    }
```

menas/src/main/scala/za/co/absa/enceladus/menas/services/DatasetService.scala
```diff
@@ -137,111 +162,128 @@ class DatasetService @Autowired() (datasetMongoRepository: DatasetMongoRepositor
...
+                                       maybeSchema: Future[Option[Schema]]): Future[Validation] = {
+
+    val maybeFields = maybeSchema.map {
+      case Some(x) => x.fields.flatMap(f => f.getAllChildren :+ f.getAbsolutePath).toSet
+      case None => Set.empty[String]
+    }
+    val accumulator = RuleValidationsAndFields(Seq.empty[Future[Validation]], maybeFields)
 
-    conformanceRules.foldLeft(accumulator) { case ((validation, currentColumns), conformanceRule) =>
+    val ruleValidationsAndFields = conformanceRules.foldLeft(accumulator) { case (validationsAndFields, conformanceRule) =>
```
benedeki, line 173: **Just a note:** This part of code is out of scope of this class, IMO. But it's good enough now.
Time to refactor is, when the validation is applied to standard dataset edits too.

benedeki (changes requested):

menas/src/main/scala/za/co/absa/enceladus/menas/services/MappingTableService.scala
```diff
@@ -79,13 +79,15 @@ class MappingTableService @Autowired() (mappingTableMongoRepository: MappingTabl
...
     val maybeSchema = datasetMongoRepository.getConnectedSchema(item.schemaName, item.schemaVersion)
 
-    val validations = super.validateSingleImport(item, metadata)
-    val validationsWithSchema = validateSchema(item.schemaName, item.schemaVersion, validations, maybeSchema)
-    validateDefaultValues(item, maybeSchema, validationsWithSchema)
+    for {
+      validationBase <- super.validateSingleImport(item, metadata)
+      validationSchema <- validateSchema(item.schemaName, item.schemaVersion, maybeSchema)
+      validationDefaultValues <- validateDefaultValues(item, maybeSchema)
+    } yield validationBase.merge(validationSchema).merge(validationDefaultValues)
```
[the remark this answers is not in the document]
Zejnilovic: Fixed

menas/src/main/scala/za/co/absa/enceladus/menas/services/DatasetService.scala
```diff
@@ -112,17 +124,30 @@ class DatasetService @Autowired() (datasetMongoRepository: DatasetMongoRepositor
...
-    val validations = super.validateSingleImport(item, metadata)
-    val validationsWithSchema = validateSchema(item.schemaName, item.schemaVersion, validations, maybeSchema)
-    val validationWithConnected = validateConnectedEntitiesExistence(confRulesWithConnectedEntities, validationsWithSchema)
-    validateConformanceRules(item.conformance, maybeSchema, validationWithConnected)
+    for {
+      validationBase <- super.validateSingleImport(item, metadata)
+      validationSchema <- validateSchema(item.schemaName, item.schemaVersion, maybeSchema)
+      validationConnectedEntities <- validateConnectedEntitiesExistence(confRulesWithConnectedEntities)
+      validationConformanceRules <- validateConformanceRules(item.conformance, maybeSchema)
+    } yield validationBase.merge(validationSchema).merge(validationConnectedEntities).merge(validationConformanceRules)
```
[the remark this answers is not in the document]
Zejnilovic: Fixed

menas/src/main/scala/za/co/absa/enceladus/menas/services/MappingTableService.scala
```diff
@@ -79,13 +79,15 @@ class MappingTableService @Autowired() (mappingTableMongoRepository: MappingTabl
...
   }
 
-  private def validateDefaultValues(item: MappingTable, maybeSchema: Future[Option[Schema]], validationsWithSchema: Future[Validation]) = {
-    item.defaultMappingValue.foldLeft(validationsWithSchema) { (acc, defaultValue) =>
+  private def validateDefaultValues(item: MappingTable, maybeSchema: Future[Option[Schema]]) = {
```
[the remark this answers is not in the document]
Zejnilovic: Fixed

menas/src/main/scala/za/co/absa/enceladus/menas/services/MappingTableService.scala
```diff
@@ -68,4 +69,34 @@ class MappingTableService @Autowired() (mappingTableMongoRepository: MappingTabl
...
+        schema <- maybeSchema
+        accValidations <- acc
+      } yield {
+        accValidations.withErrorIf(
+          schema.exists(s => !s.fields.exists(_.getAbsolutePath == defaultValue.columnName)),
+          "item.defaultMappingValue",
+          s"Cannot fiend field ${defaultValue.columnName} in schema")
+      }
+    }
+  }
```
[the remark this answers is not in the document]
Zejnilovic: Fixed

menas/src/main/scala/za/co/absa/enceladus/menas/services/DatasetService.scala
```diff
@@ -112,17 +124,30 @@ class DatasetService @Autowired() (datasetMongoRepository: DatasetMongoRepositor
...
+        connectedEntity match {
+          case mt: ConnectedMappingTable =>
+            for {
+              instance <- datasetMongoRepository.getConnectedMappingTable(mt.name, mt.version)
+              accValidations <- validation
+            } yield instance
+              .map { _ => accValidations.withError(s"item.${mt.kind}", standardizedErrMessage(mt))}
+              .getOrElse(accValidations)
+        }
+      }
```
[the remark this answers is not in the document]
Zejnilovic: Fixed

menas/src/main/scala/za/co/absa/enceladus/menas/services/DatasetService.scala
```diff
@@ -112,17 +124,30 @@ class DatasetService @Autowired() (datasetMongoRepository: DatasetMongoRepositor
...
-    confRulesWithConnectedEntities.foldLeft(validations) { (acc, cr) =>
+    confRulesWithConnectedEntities.foldLeft(Future(Validation())) { (acc, cr) =>
+      cr.connectedEntities.foldLeft(acc) { (validation, connectedEntity)  =>
+        connectedEntity match {
+          case mt: ConnectedMappingTable =>
+            for {
+              instance <- datasetMongoRepository.getConnectedMappingTable(mt.name, mt.version)
+              accValidations <- validation
+            } yield instance
+              .map { _ => accValidations.withError(s"item.${mt.kind}", standardizedErrMessage(mt))}
```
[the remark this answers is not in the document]
Zejnilovic: Fixed

menas/src/main/scala/za/co/absa/enceladus/menas/services/DatasetService.scala
```diff
@@ -102,4 +113,177 @@ class DatasetService @Autowired() (datasetMongoRepository: DatasetMongoRepositor
...
+              .getOrElse(accValidations)
+        }
+      }
+
+      val connectedEntities = cr.connectedEntities.map {
+        case mt: ConnectedMappingTable => (mt, datasetMongoRepository.getConnectedMappingTable(mt.name, mt.version))
+      }
+
+      connectedEntities.foldLeft(acc) { case (v, (entityDef, entityDBInstance)) =>
```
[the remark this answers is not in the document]
Zejnilovic: Fixed

menas/src/main/scala/za/co/absa/enceladus/menas/services/DatasetService.scala
Zejnilovic: Fixed

menas/src/main/scala/za/co/absa/enceladus/menas/services/DatasetService.scala
Zejnilovic: Also, Validations need to be moved to Utils. Then this might be moved to DataModel.

benedeki (dismissed):
- [x] code reviewed
- [x] pulled
- [x] built
- [ ] run

Haven't run it yet to test it.

data-model/src/main/scala/za/co/absa/enceladus/model/conformanceRule/package.scala
dk1844: 🤷‍♂️

menas/src/main/scala/za/co/absa/enceladus/menas/services/VersionedModelService.scala
```diff
@@ -285,8 +285,8 @@ abstract class VersionedModelService[C <: VersionedModel with Product with Audit
     }
   }
 
-  private[services] def hasWhitespace(name: String): Boolean = !name.matches("""\w+""")
-  private[services] def hasValidNameChars(name: String): Boolean = name.matches("""[a-zA-Z0-9._-]+""")
+  private[services] def hasWhitespace(name: String): Boolean = name != None.orNull && !name.matches("""\w+""")
```
dk1844, line 288:
```suggestion
  private[services] def hasWhitespace(name: String): Boolean = Option(name).exists( definedName => !definedName.matches("""\w+"""))
```
and the same below.

menas/src/main/scala/za/co/absa/enceladus/menas/services/DatasetService.scala
```diff
@@ -112,136 +124,159 @@ class DatasetService @Autowired() (datasetMongoRepository: DatasetMongoRepositor
...
-        for {
-          instance <- entityDBInstance
-          accValidations <- v
-        } yield accValidations.withErrorIf(instance.isEmpty, s"item.${entityDef.kind}", standardizedErrMessage(entityDef))
+    allConnectedEntities.foldLeft(Future(Validation())) { (accValidations, entityDef) =>
+      entityDef match {
+        case mt: ConnectedMappingTable =>
+          val entityDBInstance = datasetMongoRepository.getConnectedMappingTable(mt.name, mt.version)
+          for {
+            instance <- entityDBInstance
```
dk1844, line 150:
```suggestion
            instance <- entityDbInstance
```

dk1844 (approved):

AdrianOlosutean (approved):
Ran it, looks good now

menas/src/main/scala/za/co/absa/enceladus/menas/services/VersionedModelService.scala
benedeki: I like Daniel's suggestion.

Zejnilovic: I thought about this and found it to be too hard to immediately understand what it protects from.

dk1844: In my eyes, your solution's downsides are:
 - `None.orNull` - the reader has to realize this is just `null`, but you are avoiding it for a scala-style warning, even though I'd find it more readable. But you would have to add a scala-style ignore comment, etc.
 - realization of the short-circuiting of the `&&` operator.

Both things are well-established phenomenons but they may require some thought. 

On the other hand, my solution only requires one non-trivial realization that `Option(null)` yields `None`. Again, some thought/experience, but otherwise I find it more easily readable.

Of course, your choice. 😄

dk1844 (dismissed):

menas/src/main/scala/za/co/absa/enceladus/menas/services/VersionedModelService.scala
Zejnilovic: Fixed

menas/src/main/scala/za/co/absa/enceladus/menas/services/DatasetService.scala
Zejnilovic: Fixed

Zejnilovic:
Just added 2 unit tests and applied comments

benedeki (approved):

data-model/src/main/scala/za/co/absa/enceladus/model/conformanceRule/ConnectedEntity.scala
benedeki: Because `type` is a reserved word, I think `kind` is better here.

Issue #1536, opened 2020-09-30
Dataset - Monitoring - Run Details - shows wrong end time


Zejnilovic:
## Describe the bug
Monitoring (Dataset tab) shows run end time with presumably bad timezone.

## To Reproduce
Steps to reproduce the behaviour OR commands run:
1. Run full std and conf
2. Check Run 
3. Check Monitoring tab
4. Compare end times

## Expected behaviour
Either everything should be timezone or nothing. Ideally, it should be as is in _INFO file

### Monitoring
![image](https://user-images.githubusercontent.com/17467768/94660883-95a15f00-0306-11eb-819a-633bd588d176.png)
### Run Detail
![image](https://user-images.githubusercontent.com/17467768/94660935-a5b93e80-0306-11eb-82ce-5433b405d328.png)

DzMakatun:
Release note:
Fixed a bug in shown time in Monitoring. Both 'Run Start" and 'Latest checkpoint End time' are now shown in browser's timezone.

PR #1537, opened 2020-09-30
#1536 Add timezone parsing for last checkpoint end DateTime

DzMakatun:
Fixes the bug in Monitoring UI. Previously the end time of the last checkpoint of a run was parsed without timezone offset. As a result, in contrast to 'startDateTime' of a run, it was displayed incorrectly and created confusion. Now both dates are parsed with respect to timezone. The UI displays both DateTimes in the browser timezone.

IMPORTANT NOTE:
Menas time measurements (job start, end, checkpoint times) are subject to certain errors. It is because the time is measured in the driver process of a Spark Job and does not account for other asynchronous processes within the same job. For example, the start and end times do not account for job startup and shutdown overheads. As a result, the metrics do not present exact job timing and should be considered for illustrative  purposes only. **For reliable performance evaluation the timings, e.g. job duration, should be taken from external services, such as Spark History or YARN.**


menas/ui/service/MonitoringService.js
```diff
@@ -337,8 +337,8 @@ var MonitoringService = new function() {
       name: original["name"],
       order: original["order"],
       workflowName: original["workflowName"],
-      //startDate: moment(original["processStartTime"], "DD-MM-YYYY HH:mm:ss").toDate(),
-      endDate: moment(original["processEndTime"], "DD-MM-YYYY HH:mm:ss").toDate()
+      //startDate: moment.parseZone(original["processStartTime"], "DD-MM-YYYY HH:mm:ss ZZ").toDate(),
```
Zejnilovic, line 340: Where does the `startDate` come from?

DzMakatun: in the  commented line the 'startDate' is processStartTime of the last checkpoint.
the 'startDateTime' of the entire job is being casted inside mongo query in order to filter and sort the output

benedeki: Isn't that a possible source of discrepancy (as this bug shows), that each is done at different places?
Cannot it be done all at one place?

DzMakatun: @benedeki:
 - All the date/time fields are currently stored as str in Mongo
- When we query the runs, we have to cast infodate and start time of a job to Mongo's ISODate in order to do filtering, sorting and grouping within the monitoring query. However, it is a costly operation, therefore we only cast the fields that are required.  
- In Web UI, we need to cast date/time fields to DateTime. For those 2 fields we do cast ISODate -> DateTime. For the rest: str - > DateTime. Currently, there is only one later field in use, but more can be added when monitoring evolves.

benedeki: And I guess the original string for start time is not available here anymore to use?
I would consider this a hotfix then. But still a possible source of future problems.

DzMakatun: @benedeki, in principal it is possible to duplicate those fields, to have them as both as ISODate and original String. However, it may increase discrepancy even more. What do you think?

benedeki: Not sure. For now let's keep it.
Btw, whys is the field in Mongo a string and not a timestamp?

benedeki (approved):
Looks good as a fix.
(Haven't run it though to verify)

menas/ui/service/MonitoringService.js
Zejnilovic: > Btw, whys is the field in Mongo a string and not a timestamp?

This is _INFO file transformed into a Run object. JSON does not have a date/time type. So it just seems it was never parsed when moved to Run object

Zejnilovic (approved):
Code revied, built, checked UI

PR #1544, opened 2020-10-05
Feature/1496 SSR key/cert select

dk1844:
The schema registry connection is now explicitly supplying its SSL context created from the in-config referenced trustStore and keyStore JKS files without relying on the automatic mechanism from the environment variables.

Tested on Menas-dev on premise where the problems manifesting with the current `develop` version, the issue seems to be resolved - a schema from the SSR can be actually loaded. Big thanks to @hamc17 who was willing to deploy the test artifacts to the environment which I do not have access to, presently.

There is a standing question if this approach, albeit usable in this case, is the general way to go for similar cases. The choice generally goes down to:
 - multiple smaller key/trust stores and picking up what is needed (this PRs approach)
 - having one bigger key/trust store and using aliasing to pick the keys to be used (I expect this is possible, too)

There are some warnings for cases where trustStore/keyStore is not set, these warning can be suppressed by `menas.schemaRegistry.warnUnsecured=false` (default: `true`) in Menas's config.

### DIFF
--- a/menas/src/main/scala/za/co/absa/enceladus/menas/services/SchemaRegistryService.scala
+++ b/menas/src/main/scala/za/co/absa/enceladus/menas/services/SchemaRegistryService.scala
@@ -15,30 +15,33 @@
 
 package za.co.absa.enceladus.menas.services
 
+import java.io.FileInputStream
 import java.net.URL
+import java.security.KeyStore
 
 import com.typesafe.config.ConfigFactory
-import javax.net.ssl.{HostnameVerifier, HttpsURLConnection, KeyManagerFactory, SSLContext, SSLSession, TrustManagerFactory}
+import javax.net.ssl.{HttpsURLConnection, KeyManagerFactory, SSLContext, TrustManagerFactory}
 import org.slf4j.LoggerFactory
-import org.springframework.beans.factory.annotation.{Autowired, Value}
+import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.stereotype.Service
 import za.co.absa.enceladus.menas.controllers.SchemaController
 import za.co.absa.enceladus.menas.models.rest.exceptions.RemoteSchemaRetrievalException
 import za.co.absa.enceladus.menas.services.SchemaRegistryService._
 import za.co.absa.enceladus.menas.utils.SchemaType
-import za.co.absa.enceladus.utils.config.{ConfigUtils, SecureConfig}
 import za.co.absa.enceladus.utils.config.ConfigUtils.ConfigImplicits
+import za.co.absa.enceladus.utils.config.SecureConfig
 
-import scala.collection.JavaConverters._
 import scala.io.Source
 import scala.util.control.NonFatal
 
 @Service
 class SchemaRegistryService @Autowired()() {
 
-  private val logger  = LoggerFactory.getLogger(this.getClass)
-
+  private val logger = LoggerFactory.getLogger(this.getClass)
   private val config = ConfigFactory.load()
+
+  private val defaultStoreType = "JKS"
+  private val defaultSslContextProtocol = "TLS"
 
   if (SecureConfig.hasKeyStoreProperties(config) && SecureConfig.hasTrustStoreProperties(config)) {
     //SecureConfig.setKeyStoreProperties(config)
@@ -66,27 +69,46 @@ class SchemaRegistryService @Autowired()() {
     }
   }
 
-  def logAndGetKeyStoreAliases(): Seq[(String, Boolean)] = {
-    import java.security.KeyStore
-    val ks = KeyStore.getInstance("JKS")
+  /**
+   * JKS based TrustManagerFactory option based on [[SecureConfig.Keys.javaxNetSslTrustStore]] and
+   * [[SecureConfig.Keys.javaxNetSslTrustStorePassword]] presence in [[config]]
+   */
+  lazy val trustManagerFactory: Option[TrustManagerFactory] = {
+    if (SecureConfig.hasTrustStoreProperties(config)) {
+      val trustStore = KeyStore.getInstance(defaultStoreType)
 
-    val ksInputStream = new java.io.FileInputStream(config.getString(SecureConfig.Keys.javaxNetSslKeyStore))
-    ks.load(ksInputStream, config.getString(SecureConfig.Keys.javaxNetSslKeyStorePassword).toCharArray)
-    ksInputStream.close()
+      val tsInputStream = new FileInputStream(config.getString(SecureConfig.Keys.javaxNetSslTrustStore))
+      trustStore.load(tsInputStream, config.getString(SecureConfig.Keys.javaxNetSslTrustStorePassword).toCharArray)
+      tsInputStream.close()
 
-    val aliases = ks.aliases().asScala.toSeq
-    logger.info(s"aliases in the keystore:\n ${aliases.mkString("\n ")}") // printing aliases found
+      val tmf = TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm)
+      tmf.init(trustStore)
 
-    val keyEntries = ks.aliases.asScala.map(alias => (alias, ks.isKeyEntry(alias))).toList
-    logger.info(s"keystore aliases - key info:")
-      keyEntries.foreach { case (alias, isKeyEntry) =>
-        logger.info(s"$alias\t\t${ if(isKeyEntry) "has key" else "no key" }")
+      Some(tmf)
+    } else {
+      None
     }
+  }
 
-    // val certs = ks.aliases().asScala.map(ks.getCertificate).toList
-    // val keys = ks.aliases().asScala.map(alias => ks.getKey(alias, config.getString(SecureConfig.Keys.javaxNetSslKeyStorePassword).toCharArray)).toList
+  /**
+   * JKS based KeyManagerFactory option based on [[SecureConfig.Keys.javaxNetSslKeyStore]] and
+   * [[SecureConfig.Keys.javaxNetSslKeyStorePassword]] presence in [[config]]
+   */
+  lazy val keyManagerFactory: Option[KeyManagerFactory] = {
+    if (SecureConfig.hasKeyStoreProperties(config)) {
+      val ks = KeyStore.getInstance(defaultStoreType)
 
-    keyEntries
+      val ksInputStream = new FileInputStream(config.getString(SecureConfig.Keys.javaxNetSslKeyStore))
+      ks.load(ksInputStream, config.getString(SecureConfig.Keys.javaxNetSslKeyStorePassword).toCharArray)
+      ksInputStream.close()
+
+      val kmf = KeyManagerFactory.getInstance(KeyManagerFactory.getDefaultAlgorithm)
+      kmf.init(ks, config.getString(SecureConfig.Keys.javaxNetSslKeyStorePassword).toCharArray)
+
+      Some(kmf)
+    } else {
+      None
+    }
   }
 
   /**
@@ -96,31 +118,34 @@ class SchemaRegistryService @Autowired()() {
    * @return `SchemaResponse` object containing the obtained schema.
    */
   def loadSchemaByUrl(remoteUrl: String): SchemaResponse = {
-    val ksAliases = logAndGetKeyStoreAliases
     try {
+      if (keyManagerFactory.isEmpty) {
+        logger.warn("keyManagerFactory is not, secure schema registry integration may not function properly. " +
+          s"Set ${SecureConfig.Keys.javaxNetSslKeyStore} and ${SecureConfig.Keys.javaxNetSslKeyStorePassword} to fix the issue.")
+      }
+      if (trustManagerFactory.isEmpty) {
+        logger.warn("trustManagerFactory is not, secure schema registry integration may not function properly. " +
+          s"Set ${SecureConfig.Keys.javaxNetSslTrustStore} and ${SecureConfig.Keys.javaxNetSslTrustStorePassword} to fix the issue.")
+      }
 
-
-      import java.security.KeyStore
-      val ks = KeyStore.getInstance("JKS")
-
-      val ksInputStream = new java.io.FileInputStream(config.getString(SecureConfig.Keys.javaxNetSslKeyStore))
-      ks.load(ksInputStream, config.getString(SecureConfig.Keys.javaxNetSslKeyStorePassword).toCharArray)
-      ksInputStream.close()
-
-
-      val tmf = TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm)
-      tmf.init(ks)
-
-      val kmf = KeyManagerFactory.getInstance(KeyManagerFactory.getDefaultAlgorithm)
-      kmf.init(ks, config.getString(SecureConfig.Keys.javaxNetSslKeyStorePassword).toCharArray)
-
-      val ctx = SSLContext.getInstance("TLS")
-      ctx.init(kmf.getKeyManagers, tmf.getTrustManagers, null)
-      val sslFactory = ctx.getSocketFactory
+      val ctx = SSLContext.getInstance(defaultSslContextProtocol)
+      ctx.init(
+        keyManagerFactory.map(_.getKeyManagers).orNull,
+        trustManagerFactory.map(_.getTrustManagers).orNull,
+        null // scalastyle:ignore null - Java API
+      )
 
       val url = new URL(remoteUrl)
-      val connection: HttpsURLConnection = url.openConnection().asInstanceOf[HttpsURLConnection]
-     connection.setSSLSocketFactory(sslFactory)
+      val connection = url.openConnection()
+
+      if (remoteUrl.toLowerCase.startsWith("https")) {
+        val httpsConnection: HttpsURLConnection = connection.asInstanceOf[HttpsURLConnection]
+        httpsConnection.setSSLSocketFactory(ctx.getSocketFactory)
+      } else {
+        logger.warn(s"Connection to schema registry at $remoteUrl is not secure and trust/key stores are not used." +
+          s" Fix by setting config value of $SchemaRegistryUrlConfigKey to begin with 'https'.")
+      }
+
 
       val mimeType = SchemaController.avscContentType // only AVSC is expected to come from the schema registry
       val fileStream = Source.fromInputStream(connection.getInputStream)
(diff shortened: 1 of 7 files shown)


menas/src/main/scala/za/co/absa/enceladus/menas/services/SchemaRegistryService.scala
```diff
@@ -61,6 +64,48 @@ class SchemaRegistryService @Autowired()() {
...
 
+  /**
+   * JKS based TrustManagerFactory option based on [[SecureConfig.Keys.javaxNetSslTrustStore]] and
+   * [[SecureConfig.Keys.javaxNetSslTrustStorePassword]] presence in [[config]]
+   */
+  lazy val trustManagerFactory: Option[TrustManagerFactory] = {
+    if (SecureConfig.hasTrustStoreProperties(config)) {
+      val trustStore = KeyStore.getInstance(defaultStoreType)
+
+      val tsInputStream = new FileInputStream(config.getString(SecureConfig.Keys.javaxNetSslTrustStore))
```
Zejnilovic, line 75: Is this safe? What if it does not exist?

menas/src/main/scala/za/co/absa/enceladus/menas/services/SchemaRegistryService.scala
```diff
@@ -61,6 +64,48 @@ class SchemaRegistryService @Autowired()() {
...
+
+  /**
+   * JKS based KeyManagerFactory option based on [[SecureConfig.Keys.javaxNetSslKeyStore]] and
+   * [[SecureConfig.Keys.javaxNetSslKeyStorePassword]] presence in [[config]]
+   */
+  lazy val keyManagerFactory: Option[KeyManagerFactory] = {
+    if (SecureConfig.hasKeyStoreProperties(config)) {
+      val ks = KeyStore.getInstance(defaultStoreType)
+
+      val ksInputStream = new FileInputStream(config.getString(SecureConfig.Keys.javaxNetSslKeyStore))
```
Zejnilovic, line 96: Is this safe? What if it does not exist?

menas/src/main/scala/za/co/absa/enceladus/menas/services/SchemaRegistryService.scala
```diff
@@ -69,8 +114,34 @@ class SchemaRegistryService @Autowired()() {
    */
   def loadSchemaByUrl(remoteUrl: String): SchemaResponse = {
     try {
+      if (keyManagerFactory.isEmpty) {
+        logger.warn("keyManagerFactory is not, secure schema registry integration may not function properly. " +
+          s"Set ${SecureConfig.Keys.javaxNetSslKeyStore} and ${SecureConfig.Keys.javaxNetSslKeyStorePassword} to fix the issue.")
+      }
+      if (trustManagerFactory.isEmpty) {
+        logger.warn("trustManagerFactory is not, secure schema registry integration may not function properly. " +
```
Zejnilovic, line 122: is not defined? 
Just `is not` seems lacking a word

menas/src/main/scala/za/co/absa/enceladus/menas/services/SchemaRegistryService.scala
```diff
@@ -69,8 +114,34 @@ class SchemaRegistryService @Autowired()() {
...
+          s"Set ${SecureConfig.Keys.javaxNetSslTrustStore} and ${SecureConfig.Keys.javaxNetSslTrustStorePassword} to fix the issue.")
+      }
+
+      val ctx = SSLContext.getInstance(defaultSslContextProtocol)
+      ctx.init(
+        keyManagerFactory.map(_.getKeyManagers).orNull,
+        trustManagerFactory.map(_.getTrustManagers).orNull,
+        null // scalastyle:ignore null - Java API
```
Zejnilovic, line 130: This could be a piece of syntactic sugar to not use null.

```suggestion
        None.orNull
```

menas/src/main/scala/za/co/absa/enceladus/menas/services/SchemaRegistryService.scala
dk1844: `SecureConfig.hasKeyStoreProperties(config)` should guard you for this very case.

menas/src/main/scala/za/co/absa/enceladus/menas/services/SchemaRegistryService.scala
dk1844: `SecureConfig.hasTrustStoreProperties(config)` should guard you for this very case.

menas/src/main/scala/za/co/absa/enceladus/menas/services/SchemaRegistryService.scala
dk1844: Honestly, I don't know. `null` here is much more obvious and _not wrong_, because the API directly specifies it, I am a bit siding to keep it as is. How strong is your opinion on this?

menas/src/main/scala/za/co/absa/enceladus/menas/services/SchemaRegistryService.scala
dk1844: yes, definitely, thank you!

menas/src/main/scala/za/co/absa/enceladus/menas/services/SchemaRegistryService.scala
Zejnilovic: No strong feelings. I just think it looks better. Functions the same. If you do not like it, you don't like it.

menas/src/main/scala/za/co/absa/enceladus/menas/services/SchemaRegistryService.scala
Zejnilovic: Sorry, I mean what is the load fails somehow. Does it leave an open stream?

dk1844: Keen observation, I have put the closing in `finally`.

Zejnilovic (approved):
Menas runs. I don't have the env to test if it runs on our infrastructure.

menas/src/main/scala/za/co/absa/enceladus/menas/services/SchemaRegistryService.scala
```diff
@@ -69,8 +120,34 @@ class SchemaRegistryService @Autowired()() {
    */
   def loadSchemaByUrl(remoteUrl: String): SchemaResponse = {
     try {
+      if (keyManagerFactory.isEmpty) {
```
benedeki, line 123: Can it work without `keyManagerFactory` or `trustManagerFactory`?
If yes, I would add a possibility to suppress the warnings.

menas/src/main/scala/za/co/absa/enceladus/menas/services/SchemaRegistryService.scala
```diff
@@ -69,8 +120,34 @@ class SchemaRegistryService @Autowired()() {
...
+      )
+
       val url = new URL(remoteUrl)
       val connection = url.openConnection()
+
+      if (remoteUrl.toLowerCase.startsWith("https")) {
+        val httpsConnection: HttpsURLConnection = connection.asInstanceOf[HttpsURLConnection]
+        httpsConnection.setSSLSocketFactory(ctx.getSocketFactory)
+      } else {
+        logger.warn(s"Connection to schema registry at $remoteUrl is not secure and trust/key stores are not used." +
```
benedeki, line 146: Ditto on adding a suppression option for the warning.

menas/src/main/scala/za/co/absa/enceladus/menas/services/SchemaRegistryService.scala
```diff
@@ -69,8 +120,34 @@ class SchemaRegistryService @Autowired()() {
    */
   def loadSchemaByUrl(remoteUrl: String): SchemaResponse = {
     try {
+      if (keyManagerFactory.isEmpty) {
+        logger.warn("keyManagerFactory is not defined, secure schema registry integration may not function properly. " +
```
benedeki, line 124: It's a logging within a web app backend. Will be hard to spot for the app owner. 🤔

benedeki (commented):
Some questions. Some remarks. Not necessary need to change, but would like to consider the implications.

menas/src/main/scala/za/co/absa/enceladus/menas/services/SchemaRegistryService.scala
```diff
@@ -61,6 +64,54 @@ class SchemaRegistryService @Autowired()() {
     }
   }
 
+  /**
+   * JKS based TrustManagerFactory option based on [[SecureConfig.Keys.javaxNetSslTrustStore]] and
+   * [[SecureConfig.Keys.javaxNetSslTrustStorePassword]] presence in [[config]]
+   */
+  lazy val trustManagerFactory: Option[TrustManagerFactory] = {
+    if (SecureConfig.hasTrustStoreProperties(config)) {
```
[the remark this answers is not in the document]
dk1844: I have thought about it during development, too, and I have (as you are hinting) concluded I could not do it in a way that would be efficient and readable (hence: not worth it in my eyes).

menas/src/main/scala/za/co/absa/enceladus/menas/services/SchemaRegistryService.scala
```diff
@@ -61,6 +64,54 @@ class SchemaRegistryService @Autowired()() {
...
+      val trustStore = KeyStore.getInstance(defaultStoreType)
+
+      val tsInputStream = new FileInputStream(config.getString(SecureConfig.Keys.javaxNetSslTrustStore))
+      try {
+        trustStore.load(tsInputStream, config.getString(SecureConfig.Keys.javaxNetSslTrustStorePassword).toCharArray)
```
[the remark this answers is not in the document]
dk1844: TrustStore password is only used once. For KeyStore password, see below.

menas/src/main/scala/za/co/absa/enceladus/menas/services/SchemaRegistryService.scala
dk1844: It can work with both being missing (not secured SR case). Theoretically, I can imagine only one being needed but makes little sense to me.

menas/src/main/scala/za/co/absa/enceladus/menas/services/SchemaRegistryService.scala
dk1844: Well, here it only applies to 'https', so I believe the warning is quite correct. Don't you think? But still, I can probably let it drive with the same flag as above.

menas/src/main/scala/za/co/absa/enceladus/menas/services/SchemaRegistryService.scala
```diff
@@ -15,30 +15,33 @@
...
   private val config = ConfigFactory.load()
 
-  if (SecureConfig.hasKeyStoreProperties(config) && SecureConfig.hasTrustStoreProperties(config)) {
-    SecureConfig.setKeyStoreProperties(config)
-    SecureConfig.setTrustStoreProperties(config)
-  }
+  private val defaultStoreType = "JKS"
+  private val defaultSslContextProtocol = "TLS"
 
   lazy val schemaRegistryBaseUrl: Option[String] = {
```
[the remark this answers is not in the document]
dk1844: My logic was that it did not have to be populated if not used, but true, we use it anyway. Removed `lazy`.

menas/src/main/scala/za/co/absa/enceladus/menas/services/SchemaRegistryService.scala
```diff
@@ -15,30 +15,33 @@
...
 
+  private val logger = LoggerFactory.getLogger(this.getClass)
   private val config = ConfigFactory.load()
 
-  if (SecureConfig.hasKeyStoreProperties(config) && SecureConfig.hasTrustStoreProperties(config)) {
-    SecureConfig.setKeyStoreProperties(config)
-    SecureConfig.setTrustStoreProperties(config)
-  }
+  private val defaultStoreType = "JKS"
+  private val defaultSslContextProtocol = "TLS"
```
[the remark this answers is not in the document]
dk1844: True. Done.

menas/src/main/scala/za/co/absa/enceladus/menas/services/SchemaRegistryService.scala
```diff
@@ -61,6 +64,54 @@ class SchemaRegistryService @Autowired()() {
     }
   }
 
+  /**
+   * JKS based TrustManagerFactory option based on [[SecureConfig.Keys.javaxNetSslTrustStore]] and
+   * [[SecureConfig.Keys.javaxNetSslTrustStorePassword]] presence in [[config]]
+   */
+  lazy val trustManagerFactory: Option[TrustManagerFactory] = {
```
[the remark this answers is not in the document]
dk1844: True, done.

menas/src/main/scala/za/co/absa/enceladus/menas/services/SchemaRegistryService.scala
```diff
@@ -61,6 +64,54 @@ class SchemaRegistryService @Autowired()() {
...
+   * JKS based KeyManagerFactory option based on [[SecureConfig.Keys.javaxNetSslKeyStore]] and
+   * [[SecureConfig.Keys.javaxNetSslKeyStorePassword]] presence in [[config]]
+   */
+  lazy val keyManagerFactory: Option[KeyManagerFactory] = {
+    if (SecureConfig.hasKeyStoreProperties(config)) {
+      val ks = KeyStore.getInstance(defaultStoreType)
+
+      val ksInputStream = new FileInputStream(config.getString(SecureConfig.Keys.javaxNetSslKeyStore))
+      try {
+        ks.load(ksInputStream, config.getString(SecureConfig.Keys.javaxNetSslKeyStorePassword).toCharArray)
```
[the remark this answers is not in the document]
dk1844: Yes, true, thanks, done.

menas/src/main/scala/za/co/absa/enceladus/menas/services/SchemaRegistryService.scala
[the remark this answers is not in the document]
dk1844: Note to self: `hasTrustStoreProperties` and `hasKeyStoreProperties` could offer functional alternative to return `Option[]`...

menas/src/main/scala/za/co/absa/enceladus/menas/services/SchemaRegistryService.scala
dk1844: Note from discussions: allow javax.ssl env properties be the default and specific config settings to override them?

menas/src/main/scala/za/co/absa/enceladus/menas/services/SchemaRegistryService.scala
dk1844: Note: If cannot retrieve the schema, try to let the user know what went wrong (http return code?)

AdrianOlosutean (dismissed):
Code reviewed

menas/src/main/scala/za/co/absa/enceladus/menas/services/SchemaRegistryService.scala
dk1844: @benedeki - haha, having been thinking about this and having tested it, it actually works like this out-of-the-box. If one does not specify key/trust store, null is passed as the factory and in such a case, the "default security providers" are used 😉

menas/src/main/scala/za/co/absa/enceladus/menas/services/SchemaRegistryService.scala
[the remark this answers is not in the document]
dk1844: Redone as suggested + unit-test-covered. Of course, is better readable and much more scala-like. 😄

menas/src/main/scala/za/co/absa/enceladus/menas/services/SchemaRegistryService.scala
dk1844: After some digging, it turns out that the http return code is only available as part of the _cause string_ and only in some of the use-cases. So exposing the cause string seems like the best option for the user, it actually dramatically improves the level of understanding what the problem might be:

keyStoreNotFound
KeyStorePwdError
ssr403
ssrHostException

menas/src/main/scala/za/co/absa/enceladus/menas/services/SchemaRegistryService.scala
benedeki: Great!

menas/src/main/scala/za/co/absa/enceladus/menas/services/SchemaRegistryService.scala
benedeki: Wondering it it won't disclose some sensitive information or not too "programmer's language" for UI.
But let's leave it like this for now 👍

utils/src/test/scala/za/co/absa/enceladus/utils/config/SecureConfigSuite.scala
```diff
@@ -0,0 +1,47 @@
...
+import za.co.absa.enceladus.utils.config.SecureConfig.StoreDef
+
+class SecureConfigSuite extends FlatSpec with Matchers {
+
+  val emptyConfig = ConfigFactory.empty()
+  val keyStoreNoPassConfig = emptyConfig.withAnyRefValue("javax.net.ssl.keyStore", "/path/to/keystore")
+  val keyStoreConfig = keyStoreNoPassConfig.withAnyRefValue("javax.net.ssl.keyStorePassword", "ksPwd1")
+
+  val trustStoreNoPassConfig = emptyConfig.withAnyRefValue("javax.net.ssl.trustStore", "/path/to/trustStore")
+  val trustStoreConfig = trustStoreNoPassConfig.withAnyRefValue("javax.net.ssl.trustStorePassword", "tsPwd1")
```
benedeki, line 30: `private` or type declaration, please. First better IMHO

benedeki (changes requested):

utils/src/test/scala/za/co/absa/enceladus/utils/config/SecureConfigSuite.scala
dk1844: Sure. Done.

menas/src/main/scala/za/co/absa/enceladus/menas/services/SchemaRegistryService.scala
dk1844: To the extent I tested it and thought about it, it looked reasonable.

benedeki (approved):
LGTM 👍

AdrianOlosutean (approved):
Code reviewed

Zejnilovic (approved):
Code reviewed. Run not required. As discussed, this is mostly an integration issue that is hard to almost impossible to be tested on our machines. 
It being run by @dk1844 and @hamc17 should be enough.

PR #1545, opened 2020-10-05
Upgrade to Hyperdrive v4.0.0

kevinwallimann:
Hyperdrive v4.0.0 introduced multiple transformers (https://github.com/AbsaOSS/hyperdrive/issues/82). With 4.0.0, a transformer component only sees its subset of the configuration, i.e. `HyperConformance` doesn't need to prepend the prefix `transformer.hyperconformance` to its properties anymore.

It's up to the user of hyperdrive to make sure the prefix is unique in the global configuration.

### DIFF
--- a/spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/HyperConformance.scala
+++ b/spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/HyperConformance.scala
@@ -88,22 +88,21 @@ class HyperConformance (implicit cmd: ConformanceConfig,
 /**
  * This is the definition of Dynamic Conformance as a component of Hyperdrive.
  *
- * In order to use it in hyperdrive the component needs to be configured in 'ingestion.properties' as follows:
+ * See https://github.com/AbsaOSS/hyperdrive#configuration for instructions how the component needs to be configured in Hyperdrive.
+ * Example values are given below:
  * {{{
- * component.transformer.id.1=[hyperconformance]
- * component.transformer.class.[hyperconformance]=za.co.absa.enceladus.conformance.HyperConformance
- * transformer.[hyperconformance].menas.rest.uri=http://localhost:8080
- * transformer.[hyperconformance].dataset.name=example
- * transformer.[hyperconformance].dataset.version=1
- * transformer.[hyperconformance].report.date=2020-01-29
- * transformer.[hyperconformance].report.version=1
- * transformer.[hyperconformance].event.timestamp.column=EV_TIME
+ * menas.rest.uri=http://localhost:8080
+ * dataset.name=example
+ * dataset.version=1
+ * report.date=2020-01-29
+ * report.version=1
+ * event.timestamp.column=EV_TIME
  *
  * # Either plain credentials
- * transformer.[hyperconformance].menas.credentials.file=/path/menas.credentials
+ * menas.credentials.file=/path/menas.credentials
  *
  * # Or a keytab
- * transformer.[hyperconformance].menas.auth.keytab=/path/to/keytab
+ * menas.auth.keytab=/path/to/keytab
  * }}}
  */
 object HyperConformance extends StreamTransformerFactory with HyperConformanceAttributes {
--- a/spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/HyperConformance.scala
+++ b/spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/HyperConformance.scala
@@ -90,18 +90,20 @@ class HyperConformance (implicit cmd: ConformanceConfig,
  *
  * In order to use it in hyperdrive the component needs to be configured in 'ingestion.properties' as follows:
  * {{{
- * transformer.hyperconformance.menas.rest.uri=http://localhost:8080
- * transformer.hyperconformance.dataset.name=example
- * transformer.hyperconformance.dataset.version=1
- * transformer.hyperconformance.report.date=2020-01-29
- * transformer.hyperconformance.report.version=1
- * transformer.hyperconformance.event.timestamp.column=EV_TIME
+ * component.transformer.id.1=[hyperconformance]
+ * component.transformer.class.[hyperconformance]=za.co.absa.enceladus.conformance.HyperConformance
+ * transformer.[hyperconformance].menas.rest.uri=http://localhost:8080
+ * transformer.[hyperconformance].dataset.name=example
+ * transformer.[hyperconformance].dataset.version=1
+ * transformer.[hyperconformance].report.date=2020-01-29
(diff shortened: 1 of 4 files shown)


spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/HyperConformance.scala
```diff
@@ -90,18 +90,20 @@ class HyperConformance (implicit cmd: ConformanceConfig,
  *
  * In order to use it in hyperdrive the component needs to be configured in 'ingestion.properties' as follows:
  * {{{
- * transformer.hyperconformance.menas.rest.uri=http://localhost:8080
- * transformer.hyperconformance.dataset.name=example
- * transformer.hyperconformance.dataset.version=1
- * transformer.hyperconformance.report.date=2020-01-29
- * transformer.hyperconformance.report.version=1
- * transformer.hyperconformance.event.timestamp.column=EV_TIME
+ * component.transformer.id.1=[hyperconformance]
```
benedeki, line 93: What does the enclosing in brackets do/mean here? Optional?

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/HyperConformance.scala
```diff
@@ -90,18 +90,20 @@ class HyperConformance (implicit cmd: ConformanceConfig,
...
+ * component.transformer.class.[hyperconformance]=za.co.absa.enceladus.conformance.HyperConformance
+ * transformer.[hyperconformance].menas.rest.uri=http://localhost:8080
```
benedeki, line 95: What about the two dots surrounding the []?

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/HyperConformanceAttributes.scala
```diff
@@ -18,18 +18,17 @@ package za.co.absa.enceladus.conformance
...
-  val menasUriKey = s"$keysPrefix.menas.rest.uri"
-  val menasCredentialsFileKey = s"$keysPrefix.menas.credentials.file"
-  val menasAuthKeytabKey = s"$keysPrefix.menas.auth.keytab"
-  val datasetNameKey = s"$keysPrefix.dataset.name"
-  val datasetVersionKey = s"$keysPrefix.dataset.version"
-  val reportDateKey = s"$keysPrefix.report.date"
-  val reportVersionKey = s"$keysPrefix.report.version"
-  val eventTimestampColumnKey = s"$keysPrefix.event.timestamp.column"
-  val eventTimestampPatternKey = s"$keysPrefix.event.timestamp.pattern"
+  val menasUriKey = "menas.rest.uri"
```
benedeki, line 23: While not sure about the `[hyperconformance]` being optional, it seems the `transformer` prefix is in any case still in place?

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/HyperConformance.scala
kevinwallimann: It doesn't mean anything per se. Here, whatever you define in `component.transformer.id.1` is used in the key of other properties. In this sense, the brackets show that this part of the key is dynamic and user-defined. But you could also not use brackets, here they are just characters that stand out. I didn't think it could be interpreted as optional.

So, you could also write
```
component.transformer.id.1=asdf
component.transformer.class.asdf=...
transformer.asdf.means.rest.uri=...
```

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/HyperConformance.scala
kevinwallimann: The dots are syntactically relevant for the configuration (to select subsets of the config), while the brackets aren't

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/HyperConformanceAttributes.scala
kevinwallimann: I see your confusion. `[hyperconformance]` is not optional. Hyperdrive needs the prefix, e.g. `transformer.[hyperconformance].` (or `transformer.asdf` or whatever the user defines) and passes only the properties with that prefix. The prefix is cut off, because it's dynamic.
From the transformer component's point of view, there is no prefix.

dk1844 (approved):

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/HyperConformanceAttributes.scala
benedeki: Explained live.

benedeki (approved):
- [x] code reviewed
- [ ] pulled
- [ ] built
- [ ] run

benedeki:
Please upon merging, with the linked Issue (#1547) do the following:
* assign it to current release
* add a comment starting with _Release Notes:\n_ text followed by an explanation what is new thanks to the issue (a sentence or two).
* close the issue

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/HyperConformance.scala
```diff
@@ -88,20 +88,21 @@ class HyperConformance (implicit cmd: ConformanceConfig,
...
+ * See https://github.com/AbsaOSS/hyperdrive#configuration for instructions how the component needs to be configured in Hyperdrive.
+ * Example values are given below:
  * {{{
- * transformer.hyperconformance.menas.rest.uri=http://localhost:8080
- * transformer.hyperconformance.dataset.name=example
- * transformer.hyperconformance.dataset.version=1
- * transformer.hyperconformance.report.date=2020-01-29
- * transformer.hyperconformance.report.version=1
- * transformer.hyperconformance.event.timestamp.column=EV_TIME
+ * menas.rest.uri=http://localhost:8080
```
Zejnilovic, line 94: `menas.rest.uri` is already being used as a property key. Are we not scared of theses merging and overwriting themselves?

kevinwallimann: The properties listed here have to be passed to Hyperdrive, not Enceladus. In the Hyperdrive config, there will not be a conflict.

More detail about Hyperdrive config: In Hyperdrive, the user has to prefix the properties with a prefix that he chooses. Hyperdrive then passes only the properties with that prefix to `HyperConformance`. Because the prefix is not static, it will be stripped off, before passing to the HyperConformance Transformer

kevinwallimann:
Closed because Jenkins build cannot be triggered

PR #1549, opened 2020-10-08
1548: Helper script run_standardization_conformance.sh doesn't work

### LINKED ISSUE #1548: Helper script run_standardization_conformance.sh doesn't work
## Describe the bug
Running the combined **Standardization And Conformance** job using the helper script is not possible, because of error in the shell script.

## To Reproduce
1. Call `run_standardization_conformance.sh`, no parameters are even needed
2. Nothing happens

## Expected behaviour
Spark job is started

benedeki:
* run_standardization_conformance.sh
* CLASS variable definition removed from `enceladus_env.*` as it doesn't really belong there
* added support for --help

Closes #1548


scripts/bash/run_standardization_conformance.sh
```diff
@@ -17,7 +17,7 @@ SRC_DIR=`dirname "$0"`
 
 source ${SRC_DIR}/enceladus_env.sh
 
-export CLASS=${STD_CONF_CLASS}
+export CLASS="za.co.absa.enceladus.standardization_conformance.StandardizationAndConformanceJob"
```
dk1844, line 20: I see the move from the `enceladus_env.template.sh` and I support it. However, how could this simple move have helped with the non-working helper script?

scripts/bash/run_enceladus.sh
```diff
@@ -440,6 +447,9 @@ add_to_cmd_line "--experimental-mapping-rule" ${EXPERIMENTAL_MAPPING_RULE}
 add_to_cmd_line "--catalyst-workaround" ${CATALYST_WORKAROUND}
 add_to_cmd_line "--autoclean-std-folder" ${AUTOCLEAN_STD_FOLDER}
 add_to_cmd_line "--persist-storage-level" ${PERSIST_STORAGE_LEVEL}
+if [ "$HELP_CALL" == "1" ]; then
+    CMD_LINE="$CMD_LINE --help"
```
dk1844, line 451: It took me a while to understand that `--help` operates on a different level than one would conventionally expect. I believe that the expectation for `--help` argument is that:

1. it will prevent the program from running normally (transitively kind of true)
2. it will give helpful information on the program/command **directly** called (not really true)

Though seeming logically correct, my opinion is that the **argument should be named differently** (e.g. `--pass-help`), while `--help` would be kept reserved if we ever want to offer to provide first-hand help information for the script itself.

scripts/cmd/run_enceladus.cmd
```diff
@@ -382,11 +383,18 @@ IF "%1"=="--set-dra" (
     SHIFT
     GOTO CmdParse
 )
+IF "%1"=="--help" (
```
dk1844, line 386: same as expressed above

Moreover, Windows-style help is more customarily `/help` or `/?` (I think), but this gradually ceased to be true with many ported programs anyway.

dk1844 (approved):
(just read through)

scripts/bash/run_standardization_conformance.sh
Zejnilovic: Interesting. I am actually against pulling variable strings out of env file. This might need changes down the road, maybe even for debug reasons, maybe for some other reasons. As such I would not like this to be in the static part of scripts.

scripts/bash/run_enceladus.sh
benedeki: All the parameters are passed to the Spark job eventually, well except the `-`dry-run`.
Although the `--pass-help` would be something not intuitive, requiring extra documentation which kind of contradicts the parameter's purpose 😉 
Regarding 2, - right now the help in _SparkJobs.jar_ is broken. It's being addressed (actually already fixed) in @AdrianOlosutean PR #1511

scripts/cmd/run_enceladus.cmd
benedeki: All the other parameters are in `--parameter` form and this is not help on the helper script, but the _SparkJobs.jar` execution.

scripts/bash/run_standardization_conformance.sh
Zejnilovic: > @Zejnilovic The class is not environment specific. It's constant for the job called. This is why I believe it's more then justified to be moved out.

It is not environment-specific, but it feels like it is run specific. Not static. So if you would like to debug an experimental class like HyperConformance, you would need to change more then one place.

benedeki: It's exactly on one place (for each of job types, that is).
And if one wants to do such things, I am not sure  helper script is what one should use. At least not the "main stock" helper scripts. And if fiddling with them in this manner, it imho needs expert knowledge.
I perhaps see what you mean, but then we should create a `encaladus_const.sh` file. IMHO it's better to have this strictly defined somewhere, then to depend on something that is specific for env.

benedeki:
WTF. How did it got merged? 😮

benedeki:
> 
> 
> WTF. How did it got merged? 😮

OK, I by mistake pushed to develop branch. Will try to roll back...

benedeki:
OK, rolled back `develop` and created a new PR - #1550 . Sorry for interrupting the conversation.

scripts/bash/run_enceladus.sh
dk1844: But `--help` parameter is usually somewhat special as it prevents running and processing other params (there can be exceptions). 

I do not, however, strongly object to this, logically it makes sense. I just felt the need to call out the disparity between the function and possible user's expectation. After I have pointed this out and you still think this is the best form, feel free to merge it as is. Oh snap, you already have 😆

PR #1550, opened 2020-10-09
1548: Helper script run_standardization_conformance.sh doesn't work II

### LINKED ISSUE #1548: Helper script run_standardization_conformance.sh doesn't work

benedeki:
* run_standardization_conformance.sh
* CLASS variable definition removed from `enceladus_env.*` as it doesn't really belong there
* added support for --help

This is a redo of #1549 PR - which got merged by accident and then rolled back.

Closes #1548


dk1844 (dismissed):

scripts/bash/run_enceladus.sh
```diff
@@ -440,6 +447,9 @@ add_to_cmd_line "--experimental-mapping-rule" ${EXPERIMENTAL_MAPPING_RULE}
 add_to_cmd_line "--catalyst-workaround" ${CATALYST_WORKAROUND}
 add_to_cmd_line "--autoclean-std-folder" ${AUTOCLEAN_STD_FOLDER}
 add_to_cmd_line "--persist-storage-level" ${PERSIST_STORAGE_LEVEL}
+if [ "$HELP_CALL" == "1" ]; then
```
Zejnilovic, line 450: I don't like this personally, but I can understand there is no real compelling reason to do it otherwise as it copies what jars do. But here are my findings:

If I run something with an issue in the preceding arguments, there will be errors and then help, which is weird on its own. But also, not all errors are printed.

If I run `...--dataset-version alfa --raw-format parquet --trimValues true --help...` I get an error for `dataset-version` and help, but I don't get that raw-format parquet and trimValues do not work together.

**What I get**
```
Error: Option --dataset-version expects a number but was given 'alfa'
Enceladus v2.15.0-SNAPSHOT
Usage: Standardization Job [options]
...
```
**What I expect (option a)**
just help. Nothing else. Ignore arguments

**What I expect (option b)**
```
Error: Option --dataset-version expects a number but was given 'alfa'
Error: The --trimValues option is supported only for FixedWidth format
Enceladus v2.15.0-SNAPSHOT
Usage: Standardization Job [options]
...
```

I am aware the jar behaves the same. But if we are providing **helper** scripts I would probably have them be more "precise" (cannot come up with the word I am looking for).

scripts/bash/enceladus_const.sh
```diff
@@ -0,0 +1,20 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Static configuration
+
+STD_CLASS="za.co.absa.enceladus.standardization.StandardizationJob"
```
Zejnilovic, line 18: I really like this solution 👍

Zejnilovic:
Can we add `set -e` to the script? This will make the script abort if any one line of it fails. 

I have for example just experienced a tmp file creation failure and it still ran.

scripts/bash/run_enceladus.sh
```diff
@@ -284,36 +289,38 @@ set -- "${POSITIONAL[@]}" # restore positional parameters
...
-# Validation
-VALID="1"
-validate() {
-    if [[ -z "$2" ]]; then
-        echo "Missing mandatory option $1"
-        VALID="0"
+if [ "$HELP_CALL" == "0" ]; then
+    # Validation (only if not help called)
+    VALID="1"
+    validate() {
```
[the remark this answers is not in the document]
benedeki: Otherwise the helper script doesn't allow to execute the `spark-submit` making the `-help` of limited use.
This way `run_standardization.sh --help` will display the help.

scripts/bash/run_enceladus.sh
```diff
@@ -15,6 +15,8 @@
 
 # Command line for the script itself
 
+SET -e
```
Zejnilovic, line 18:
```suggestion
set -e
```

scripts/bash/run_enceladus.sh
```diff
@@ -478,19 +481,20 @@ if [[ -z "$DRY_RUN" ]]; then
     echo "$CMD_LINE" >> "$TMP_PATH_NAME"
     echo "The log will be saved to $TMP_PATH_NAME"
     # Run the job and return exit status of the last failed command in the subshell pipeline (Issue #893)
+    SET +e
```
Zejnilovic, line 484:
```suggestion
    set +e
```

Zejnilovic (commented):
SET (capital does not exist)

Zejnilovic (dismissed):

dk1844 (approved):

Zejnilovic (approved):

dk1844 (approved):

AdrianOlosutean (approved):
Ran the changes in the environments

Issue #1538, opened 2020-10-01
CommonJobExecution.runPostProcessing performance improvement proposal


Zejnilovic:
## Background
`CommonJobExecution.runPostProcessing` reads a parquet ([link](https://github.com/AbsaOSS/enceladus/blob/develop/spark-jobs/src/main/scala/za/co/absa/enceladus/common/CommonJobExecution.scala#L113)) that is an output of std or conf. This parquet (actually `Dataframe`) can be supplied by `process"Type"Result` (conformance or standardization). Weirdly `processStandardizationResult` returns a `DataFrame`, but conformance doesn't.

## Feature
Align the usage of `process"Type"Result` method and don't read again the written parquet in `runPostProcessing`

benedeki:
The situation is actually more complicated. It relates to the (in-)stability of computation within a distributed cluster.
It should definitely be addressed though - to make it consistent here at least.
Let's discuss it with @yruslan too and consider solutions for S3 (and its costs)

dk1844:
Release notes:
processStandardizationResult & processConformanceResult force postProcessing to reread data written to disk to avoid possible data instability

dk1844:
Do we close this issue and the future ones will just be linked? @benedeki @Zejnilovic

Zejnilovic:
I think you can. So we have "nice" release notes. Next issue might be broader and be done during our "audit" time.

benedeki:
Yup, let's close it. 👍

PR #1551, opened 2020-10-09
#1538 postProcessing reread output - normalization

dk1844:
This change directly addresses #1538 - now both `process*Result` methods return `Unit` and thus make any further post-processing routines to reload data from FS as originally intended. (i.e. prohibit using "returned" dataframe to be used)

The rereading is deliberate to ensure that the same data that is being post-processed are already written to the output location and the postProcessing could not be affected by any kind of plan-rerun that could result in newly computed data (this is especially a concern if the output is unstable).

### DIFF
--- a/spark-jobs/src/main/scala/za/co/absa/enceladus/common/CommonJobExecution.scala
+++ b/spark-jobs/src/main/scala/za/co/absa/enceladus/common/CommonJobExecution.scala
@@ -110,6 +110,9 @@ trait CommonJobExecution {
     }
   }
 
+  /**
+   * Post processing rereads the data from a path on FS (based on `sourcePhase`)
+   */
   protected def runPostProcessing[T](sourcePhase: SourcePhase, preparationResult: PreparationResult, jobCmdConfig: JobConfigParser[T])
                                     (implicit spark: SparkSession, fileSystemVersionUtils: FileSystemVersionUtils): Unit = {
     val outputPath = sourcePhase match {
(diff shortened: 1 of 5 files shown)


benedeki (commented):
Looks good. 
But when I was looking for a way how to "release" the memory the previously used DataFrames might hold, i stumble upon this article: https://sparkbyexamples.com/spark/spark-dataframe-cache-and-persist-explained/
Wouldn't that be the solution for our case? General solution. 🤔

dk1844:
> But when I was looking for a way how to "release" the memory the previously used DataFrames might hold, I stumbled upon this article: https://sparkbyexamples.com/spark/spark-dataframe-cache-and-persist-explained/
> Wouldn't that be the solution for our case? General solution. 🤔

Could you please _connect the dots_ for me here? Do you mean the unpersisting?

benedeki:
> 
> 
> > But when I was looking for a way how to "release" the memory the previously used DataFrames might hold, I stumbled upon this article: https://sparkbyexamples.com/spark/spark-dataframe-cache-and-persist-explained/
> > Wouldn't that be the solution for our case? General solution. 🤔
> 
> Could you please _connect the dots_ for me here? Do you mean the unpersisting?

Well, how I understood it, using `cache()` can make the computation stable. E.g. ensure it's not recomputed, and that even if the node fails (at least under some settings).

benedeki:
I think the release notes are too technical (developer level). I would write something like:
Switched to re-read data from disk between processing phases to ensure data stability (no change in volatile data).

dk1844:
> Well, how I understood it, using `cache()` can make the computation stable. E.g. ensure it's not recomputed, and that even if the node fails (at least under some settings).

I don't think it solves the situation completely, but we can definitely consider that an improvement of the chances and deem it good enough. I would be content with this approach. As the linked article states:

> When you persist a dataset, each node stores it’s partitioned data in memory and reuses them in other actions on that dataset. And Spark’s persisted data on nodes are fault-tolerant meaning if any partition of a Dataset is lost, it will automatically be recomputed using the original transformations that created it.

Zejnilovic (approved):
Code reviewed

benedeki (approved):

PR #1558, opened 2020-10-15
Github Actions PoC

Zejnilovic:
Made a small PoC for GitHub actions
This only runs the tests, integration tests and scala-style
It runs it in parallel
Caches maven dependencies


.github/workflows/unit_test.yml
```diff
@@ -0,0 +1,58 @@
...
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Set up JDK 1.8
+        uses: actions/setup-java@v1
+        with:
+          java-version: 1.8
+      - uses: actions/cache@v1
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-${{ hashFiles('**/pom.xml') }}
```
benedeki, line 20: What dos this line doing?

pom.xml
```diff
@@ -365,6 +366,7 @@
                         </goals>
                         <configuration>
                             <suffixes>(?&lt;!IntegrationSuite)</suffixes>
+                            <skipTests>${skip.unit.tests}</skipTests>
```
benedeki, line 369: To understand, why was this one needed?

.github/workflows/unit_test.yml
```diff
@@ -0,0 +1,58 @@
+name: Run-Tests
```
benedeki, line 1: The file name is `unit_test.yaml`. But this does three types of tests.
Maybe let's rename this to something better describing this complexity and `unit_test.yaml` have the `unit-test` job?

.github/workflows/unit_test.yml
```diff
@@ -0,0 +1,58 @@
...
+          java-version: 1.8
+      - uses: actions/cache@v1
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            ${{ runner.os }}-
+      - name: Run the scalastyle maven plugin
+        run: mvn -B scalastyle:check
+  unit-test:
```
benedeki, line 25: Tiny: Shouldn't be `unit-tests` better to show, it's a battery of tests?

.github/workflows/unit_test.yml
```diff
@@ -0,0 +1,58 @@
...
+          java-version: 1.8
+      - uses: actions/cache@v1
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            ${{ runner.os }}-
+      - name: Run the Unit tests
+        run: mvn -B test
+  integration-test:
```
benedeki, line 42: Ditto on plural.

.github/workflows/unit_test.yml
Zejnilovic: caches the dependencies, this is the key it caches them by.

https://docs.github.com/en/free-pro-team@latest/actions/guides/caching-dependencies-to-speed-up-workflows#input-parameters-for-the-cache-action

pom.xml
Zejnilovic: So I can run parallel integration tests and unit tests

.github/workflows/unit_test.yml
Zejnilovic: the file name is bad. It was when my first PoC had 4 files. Ended with one and forgot to rename it. I would rename the file to `pr_tests.yml`. How about that?

pom.xml
benedeki: Sorry for being slow, how does this enable that? Can you override the pom property in the yaml file of the action?

.github/workflows/unit_test.yml
```diff
@@ -0,0 +1,58 @@
...
+        with:
+          java-version: 1.8
+      - uses: actions/cache@v1
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            ${{ runner.os }}-
+      - name: Run the scalastyle maven plugin
+        run: mvn -B scalastyle:check
```
benedeki, line 24: We have a ton of warnings there. Won't they cause issues?

pom.xml
benedeki: Or is `true` the default?

Zejnilovic: false is the default. False needs to be set using `-Dskip.unit.tests=true`. That is used in integration tests.

.github/workflows/unit_test.yml
Zejnilovic: Warnings are ok. I think. It is failing on dependencies

pom.xml
benedeki: Understood now. The property is overridden in the job in the file above.

dk1844 (approved):

Zejnilovic:
Doesn't make sense to do until we solve the integration tests and even the external dependencies.

Issue #763, opened 2019-08-05
Add Schema after conformance to the Dataset view


Zejnilovic:
## Background
We currently possess transitive schema when creating conformance rules but we do not show this scheme anywhere.

## Feature
Add a tab `Schema after conformance` (the name is WIP) or add a switch to the current schema to see how the schema will look at the end of all conformance rules.

## Bonus points
Be able to export the final schema - is there a use case of this? Will someone need it?

Zejnilovic:
Possible duplicate of #763

dk1844:
Release notes:
Menas UI now displays projected after-conformance schema for datasets.

dk1844:
Released notes:
Menas UI now displays projected after-conformance schema for datasets.

PR #1561, opened 2020-10-16
#763 conformed schema in Menas UI

dk1844:
This feature allows the user to see what the conformed schema is about to look like (i.e. current schema + conformance rules applied). The result is not saved anywhere, it is in UI only.
There is an icon indication for the to-be-added fields, this also works in depth.

I haven't invented the conf rule application logic, only adjusted the front-end to view it (added one parameter to the logic to distinguish original vs added schema fields. Some other necessary adjustments had to be made to reuse the existing treatable fragment (original clashed on ids when reused).

This is what it looks like:
conformedRules-preview

### DIFF
--- a/menas/ui/components/SchemaTable.js
+++ b/menas/ui/components/SchemaTable.js
@@ -20,9 +20,9 @@ class SchemaTable {
     this._parentId = fragmentId;
 
     this._schemaTable = oController.byId(sap.ui.core.Fragment.createId(this._parentId, "schemaFieldsTreeTable"));
-    oController.byId(sap.ui.core.Fragment.createId(this._parentId, "metadataButton")).attachPress(this.metadataPress, this); // worked with "schemaFragment--metadataButton"
+    oController.byId(sap.ui.core.Fragment.createId(this._parentId, "metadataButton")).attachPress(this.metadataPress, this);
 
-    this._oPopoverTemplate = new sap.m.List({})
+    this._oPopoverTemplate = new sap.m.List({});
     this._oPopover = new sap.m.Popover({
       title: "Metadata",
       content: [this._oPopoverTemplate],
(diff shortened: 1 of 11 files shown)


Zejnilovic:
> added one parameter to the logic to distinguish original vs added schema fields

I love this. Is there anything for deleted/dropped columns?

dk1844:
Release notes suggestion:
Menas UI now displays projected after-conformance schema for datasets.

dk1844:
> I love this. Is there anything for deleted/dropped columns?

Thanks. No, only columns that are expected to exist after conformance are shown. Since dropped columns would not be part of the resulting scheme, they are not shown. (Albeit I understand that one may want to see this somewhere)

Zejnilovic:
I know this was not introduced by this PR, so it is questionable if this PR should address this, but when we have SingleColumn CR then structs don't resolve.

I have a struct `ManipulationBox`, not even sure what is in it. But when I apply SingleColumn CR it produces struct but doesn't show what is in the struct.

### Rule definition
image
### Schema after Conformance
image
### Resolved schema when creating Rules
image

dk1844:
> I have a struct `ManipulationBox`, not even sure what is in it. But when I apply SingleColumn CR it produces struct but doesn't show what is in the struct.

@Zejnilovic, can you even apply SingleColumn CR to a non-leaf entry (struct/array)? What should be the result? I have looked into the github-pages doc to gain understanding of this rule logic, didn't understand it much. I have checked out the test suite for it then ([SingleColumnRuleSuite.scala](https://github.com/AbsaOSS/enceladus/blob/develop/spark-jobs/src/test/scala/za/co/absa/enceladus/conformance/interpreter/rules/SingleColumnRuleSuite.scala)) and there it seems that the CR is always applied to the leaf-nodes of the hierarchy.

So I am a little confused if this should even work and if so, how?

AdrianOlosutean (approved):
Ran it, looks good

Issue #1565, opened 2020-10-16
Secure schema registry setting names are inconsistent 


benedeki:
## Describe the bug
In the Menas' settings  - `application.conf`/`application.conf.template` - there are
`# menas.schemaRegistryBaseUrl=https://localhost:8081`
and
`#menas.schemaRegistry.warnUnsecured=true`

And such also used within the application. The two should adhere to the same pattern.

## Suggested solution
Change either `menas.schemaRegistry.warnUnsecured` to `menas.schemaRegistryWarnUnsecured`, or `menas.schemaRegistryBaseUrl` to `menas.schemaRegistry.baseUrl`
The later seems as a preferred solution.

benedeki:
Release notes:
Schema registry URL config (in Menas) key has been changed from `menas.schemaRegistryBaseUrl` to `menas.schemaRegistry.baseUrl` to conform to other `menas.*` config keys.

PR #1566, opened 2020-10-16
menas config unification

dk1844:
Menas config key change:
 - `menas.schemaRegistryBaseUrl` => `menas.schemaRegistry.baseUrl`

(the underlying reason is explained in the issue #1565 - to conform to other key(s) in the `menas.*` config space)

### DIFF
--- a/menas/src/main/resources/application.properties.template
+++ b/menas/src/main/resources/application.properties.template
@@ -147,7 +147,7 @@ menas.oozie.splineMongoURL=mongodb://localhost:27017
 #----------- Schema Registry
 # URL of schema registry [optional]:
 # when unset, the ability to load schema by subject name from the schema repository will not be present
-# menas.schemaRegistryBaseUrl=https://localhost:8081
+#menas.schemaRegistry.baseUrl=https://localhost:8081
 
 # When schema registry is used and unsecured, there will be log warnings unless switched to false: (default=true)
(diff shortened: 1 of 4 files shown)


dk1844:
Release notes sugesstion:
Schema registry URL config (in Menas) key has been changed from `menas.schemaRegistryBaseUrl` to `menas.schemaRegistry.baseUrl` to conform to other `menas.*` config keys.

Zejnilovic:
Is this not a breaking change?

benedeki:
> 
> 
> Is this not a breaking change?

My point of view is that it's not. Because the secure schema registry haven't actually worked yet.  The fix wis part of the 2.15 release. This is why I wanted to have it done right away, to include it in 2.15, and @dk1844 offered to create the PR.

benedeki (approved):
I am approving it, finding it OK. But please wait for @Zejnilovic opinion as well.

dk1844:
> Is this not a breaking change?

That is what I thought as well at first and that was the reason why I didn't unify the config keys before. But @benedeki's explanation makes sense in practice.

Zejnilovic (approved):
All good if we shove it in 2.15

benedeki:
Agreed with @dk1844 I would merge the PR.

dk1844:
Thank you, @benedeki, for the night shift 🧛‍♂️ 😆

PR #1583, opened 2020-10-22
#1582: Missing version for maven-jar-plugin

### LINKED ISSUE #1582: Missing version for maven-jar-plugin
## Describe the bug
Compilation throws warning like:
```
Some problems were encountered while building the effective model for za.co.absa.enceladus:spark-jobs:jar:2.17.0-SNAPSHOT
'build.plugins.plugin.version' for org.apache.maven.plugins:maven-jar-plugin is missing. @ line 253, column 21
```

## To Reproduce
1. Download the source
2. Execute `mvn clean install`
3. See error

## Expected behaviour
No warnings

benedeki:
* added version to Menas and SparkJobs poms
* removed maven-jar-plugin from Utils pom, not needed there

Closes #1582


Zejnilovic (approved):
Built, run, reviewed.

`PerformanceMetricTools` still writes the correct version to _INFO, but I was actually scared it would not since it is an `object` in utils and I thought it will be using that `org.apache.maven.plugins` and version from utils. Any explanation?

benedeki:
> 
> 
> Built, run, reviewed.
> 
> `PerformanceMetricTools` still writes the correct version to _INFO, but I was actually scared it would not since it is an `object` in utils and I thought it will be using that `org.apache.maven.plugins` and version from utils. Any explanation?

My belief was, as it's a runtime information, that it's important to have the info in the final package. No impact if it's in the other pre-compiled embedded packages.

PR #1585, opened 2020-10-26
1554 Tomcat with TLS in Docker container

### LINKED ISSUE #1554: Secure communication between ELB and Fargate
## Background
To secure the communication in the public cloud the services need to use TLS communication.

## Feature
Use TLS for connection between ELB and our Fargate. At start self-signed certificate is sufficient. But eventually it should be an Absa-signed one, leveraging the _CertMe_ service.

## Proposed Solution
For the moment it seems _Envoy_ might be the way to go.

AdrianOlosutean:
The current configuration expects the following environment variables to be passed at runtime: `PRIVATE_KEY`, `CERTIFICATE`, and `CA_CHAIN`. The `start_menas.sh` script will create the private.pem, certificate.pem, cachain.pem files which will be taken by Tomcat and then starts the server.  
This configuration uses HTTP2, which would supposedly make the application responses faster.
In case of need, I can also provide the Envoy configuration in this PR
Closes #1554


menas/src/main/resources/server.xml
```diff
@@ -0,0 +1,167 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
hamc17, line 1: This whole file can be condensed, remove any unneeded lines, I would remove anything extra other than the licence

menas/src/main/resources/start_menas.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+echo "$PRIVATE_KEY" >> conf/private.pem
```
hamc17, line 2: Change "$variable" to "${variable}"
Just a better convention, no functional difference here but is helpful in cases where debugging may be required

hamc17 (commented):
Lgtm

menas/Dockerfile
```diff
@@ -9,34 +9,21 @@
...
     copyright="2020 ABSA Group Limited" \
     license="Apache License, version 2.0" \
     name="Menas"
 
-ENV TOMCAT_MAJOR=9 \
-    TOMCAT_VERSION=9.0.37 \
-    TOMCAT_HOME=/opt/tomcat \
-    CATALINA_HOME=/opt/tomcat
+ADD src/main/resources/start_menas.sh start_menas.sh
+RUN chmod +x start_menas.sh && rm -rf webapps/*
```
Zejnilovic, line 21: each command on a new line separated with `\` please

menas/Dockerfile
```diff
@@ -9,34 +9,21 @@
 #       menas
 # You can skip the environment properties if you set up your own properties before build
 # FROM openjdk:8-jdk-alpine
```
Zejnilovic, line 11: This seems like my leftovers. Please delete.

Zejnilovic (commented):
How do I run this locally?

menas/src/main/resources/start_menas.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+echo "${PRIVATE_KEY}" >> conf/private.pem
```
benedeki, line 2: Maybe this is a common trick in docker world, but I find it a neat trick 🏅

benedeki (commented):
Still missing some new-lines in some files' end.

menas/src/main/resources/envoy/start_envoy.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+mkdir -p /etc/ssl
+
+echo -e "$PRIVATE_KEY" >> /etc/ssl/envoy.key
+echo -e "$CERTIFICATE" >> /etc/ssl/envoy.crt
+echo -e "$CA_CHAIN" >> /etc/ssl/envoy.crt
```
Zejnilovic, line 7: As per @hamc17 's previous suggestion

```suggestion
echo -e "${PRIVATE_KEY}" >> /etc/ssl/envoy.key
echo -e "${CERTIFICATE}" >> /etc/ssl/envoy.crt
echo -e "${CA_CHAIN}" >> /etc/ssl/envoy.crt
```

Zejnilovic (approved):
Looks good, I have that one nit there.

lokm01 (approved):
LGTM

Zejnilovic (approved):
Apologies, I remembered I approved, but with the change proposal.

PR #1586, opened 2020-11-02
Feature/1556 file access PoC using Hadoop FS API

### LINKED ISSUE #1556: S3 file access PoC using Hadoop FS API
## Background
An interesting point has been raised by Tony in regard of the S3 PoC file access, currently written in terms of AWS SDK for S3.
> Any reason for going through the pain of lower level AWS Java SDK libraries instead of using the Hadoop FileSystem API methods of emrfs?
If we’re directly using the AWS Java SDK libraries then we only ever have eventual consistency, there is no possibility to use any consistency features such as s3guard (for s3a) or emrfs consistent view.

## Feature
A PoC attempt should be made to use Hadoop FS API to gain access to consistency features.

dk1844:
File access PoC using Hadoop FS API

 - SDK S3 approach has been removed completely. In theory, there could be 3 different FileSystems (hdfs/s3 - based on the string path supplied for _raw_, _std_, and _publish_), these FSs are held in the `FileSystems` case class. Therefore, implicit filesystem passing is now less frequent, because we should be passing the correct FS to work with in different cases (raw vs std vs publish).
 - The Filesystems are cached and reused (`FileSystemUtils.FileSystemExt.toFsUtils`).

Test-ran on EMR (EMRFS-backed with having all input/output/intermediates on S3). No changes in output from the previous atum-3.0.0-based version that leveraged sdk-s3 approach.

~Blocked by: https://github.com/AbsaOSS/atum/pull/44~

Closes #1556


spark-jobs/src/main/scala/za/co/absa/enceladus/common/CommonJobExecution.scala
```diff
@@ -82,18 +81,19 @@ trait CommonJobExecution {
...
                              (implicit dao: MenasDAO,
                               cmd: JobConfigParser[T],
-                              fsUtils: DistributedFsUtils,
                               spark: SparkSession): PreparationResult = {
     val confReader: ConfigReader = new ConfigReader(conf)
     confReader.logEffectiveConfigProps(Constants.ConfigKeysToRedact)
-
     dao.authenticate()
+
+    implicit val hadoopConf = spark.sparkContext.hadoopConfiguration
```
AdrianOlosutean, line 89: I'm thinking if this would make more sense to be a field

AdrianOlosutean (approved):
Code reviwed. Looks good to me

utils/src/main/scala/za/co/absa/enceladus/utils/fs/FileSystemUtils.scala
```diff
@@ -0,0 +1,69 @@
...
+        val s3BucketUri: String = s"s3://${s3Location.bucketName}" // s3://<bucket>
+        val s3uri: URI = new URI(s3BucketUri)
+        FileSystem.get(s3uri, hadoopConf)
+
+      case None =>
+        FileSystem.get(hadoopConf) // HDFS
+    }
+  }
+
+  private[fs] val fsUtilsCache = TrieMap[FileSystem, HadoopFsUtils]()
```
benedeki, line 51: No need for the `[fs]` access modifier.

utils/src/main/scala/za/co/absa/enceladus/utils/fs/FileSystemUtils.scala
```diff
@@ -0,0 +1,69 @@
...
+
+import scala.collection.concurrent.TrieMap
+
+object FileSystemUtils {
+
+  val log: Logger = LoggerFactory.getLogger(this.getClass)
+
+  /**
+   * Will yeild a [[FileSystem]] for path. If path prefix suggest S3, S3 FS is returned, HDFS otherwise.
+   * @param path
```
benedeki, line 34: Missing descriptions for parameters - I wonder if it's better to have them empty or not at all.

spark-jobs/src/test/scala/za/co/absa/enceladus/conformance/interpreter/ArrayConformanceSuite.scala
```diff
@@ -17,20 +17,18 @@ package za.co.absa.enceladus.conformance.interpreter
...
 import za.co.absa.enceladus.conformance.datasource.DataSource
-import za.co.absa.enceladus.dao.MenasDAO
 import za.co.absa.enceladus.conformance.samples._
-import za.co.absa.enceladus.utils.fs.HdfsUtils
-import za.co.absa.enceladus.utils.testUtils.SparkTestBase
+import za.co.absa.enceladus.dao.MenasDAO
+import za.co.absa.enceladus.utils.testUtils.{HadoopFsTestBase, SparkTestBase}
 
-class ArrayConformanceSuite extends AnyFunSuite with SparkTestBase with BeforeAndAfterAll {
+class ArrayConformanceSuite extends AnyFunSuite with SparkTestBase with BeforeAndAfterAll with HadoopFsTestBase {
```
benedeki, line 28: What about having `EnceladusTestBase` as a combination of `SparkTestBase` and `HadoopFsTestBase` as they are combined so often?

spark-jobs/src/main/scala/za/co/absa/enceladus/common/config/FileSystems.scala
```diff
@@ -0,0 +1,36 @@
...
+
+import org.apache.hadoop.conf.Configuration
+import org.apache.hadoop.fs.FileSystem
+import org.slf4j.{Logger, LoggerFactory}
+import za.co.absa.enceladus.utils.fs.FileSystemUtils
+
+case class FileSystems(rawFs: FileSystem, publishFs: FileSystem, standardizationFs: FileSystem)
+
+object FileSystems {
+  val log: Logger = LoggerFactory.getLogger(this.getClass)
```
benedeki, line 26: I don't think this is used.

benedeki (commented):
Nice!

utils/src/main/scala/za/co/absa/enceladus/utils/fs/FileSystemUtils.scala
```diff
@@ -0,0 +1,69 @@
...
+
+  private[fs] val fsUtilsCache = TrieMap[FileSystem, HadoopFsUtils]()
+
+  implicit class FileSystemExt(fs: FileSystem) {
+    /**
+     * Given the FileSystem object `fs` that this method is called on,
+     * the appropriate HadoopFsUtils is either newly created or returned form cache.
+     * @return
+     */
+    def toFsUtils: HadoopFsUtils = {
```
Zejnilovic, line 59: Optionall idea for naming
```suggestion
    def usingHdfsUtils: HadoopFsUtils = {
```

spark-jobs/src/main/scala/za/co/absa/enceladus/common/CommonJobExecution.scala
```diff
@@ -52,7 +51,7 @@ trait CommonJobExecution {
   protected case class PreparationResult(dataset: Dataset,
                                          reportVersion: Int,
                                          pathCfg: PathConfig,
-                                         s3Config: S3Config,
+                                         fileSystems: FileSystems,
```
Zejnilovic, line 54: Thought: I feel like `pathCfg` and `fileSystems` could be one object.

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/ConformanceExecution.scala
```diff
@@ -46,32 +43,33 @@ import za.co.absa.enceladus.utils.schema.SchemaUtils
 import scala.util.control.NonFatal
 import scala.util.{Failure, Success, Try}
 
-trait ConformanceExecution extends CommonJobExecution with S3DefaultCredentialsProvider {
+trait ConformanceExecution extends CommonJobExecution{
```
Zejnilovic, line 46:
```suggestion
trait ConformanceExecution extends CommonJobExecution {
```

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/StandardizationExecution.scala
```diff
@@ -108,16 +102,17 @@ trait StandardizationExecution extends CommonJobExecution with S3DefaultCredenti
...
+    validateIfPathAlreadyExists(pathConfig.standardizationPath)(fileSystems.standardizationFs.toFsUtils)
   }
 
   protected def readStandardizationInputData[T](schema: StructType,
                                                 cmd: StandardizationConfigParser[T],
                                                 path: String,
                                                 dataset: Dataset)
                                                (implicit spark: SparkSession,
-                                                fsUtils: DistributedFsUtils,
+                                                rawFs: FileSystem,
```
Zejnilovic, line 115: Not sure if this should be kept as an implicit

utils/src/main/scala/za/co/absa/enceladus/utils/testUtils/HadoopFsTestBase.scala
```diff
@@ -13,12 +13,22 @@
...
+package za.co.absa.enceladus.utils.testUtils
 
-trait S3DefaultCredentialsProvider {
+import org.apache.hadoop.fs.FileSystem
+import org.apache.spark.sql.SparkSession
+import za.co.absa.enceladus.utils.fs.FileSystemUtils.FileSystemExt
 
-  implicit val defaultCredentialsProvider = DefaultCredentialsProvider.create()
 
+trait HadoopFsTestBase extends HasSparkSession {
```
Zejnilovic, line 24: Could this make it easier for us?

```suggestion
trait HadoopFsTestBase extends SparkTestBase {
```

utils/src/main/scala/za/co/absa/enceladus/utils/testUtils/HadoopFsTestBase.scala
```diff
@@ -13,12 +13,22 @@
...
+
+  implicit val fs = FileSystem.get(spark.sparkContext.hadoopConfiguration)
+  implicit val fsUtils = fs.toFsUtils
+}
+
+trait HasSparkSession {
```
Zejnilovic, line 30: If you do not apply my `extends SparkTestBase` comment, I would still think of moving this to a separate file as it is used in the `SparkTestBase` the same way it is used in `HadoopFsTestBase`

utils/src/main/scala/za/co/absa/enceladus/utils/fs/FileSystemUtils.scala
benedeki: If keeping the original, it's a verb, so should be a function with `()`

utils/src/main/scala/za/co/absa/enceladus/utils/fs/HadoopFsUtils.scala
```diff
@@ -31,10 +28,10 @@ import scala.util.Try
  * A set of functions to help with the date partitioning and version control
  */
 
-class HdfsUtils(conf: Configuration) extends DistributedFsUtils {
+class HadoopFsUtils()(implicit fs: FileSystem) extends DistributedFsUtils {
```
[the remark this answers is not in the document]
dk1844: I was trying to insinuate to the user/reader that this is not bound to the "standard" HDFS - the usually used distributed FS used with Spark, but backed by Hadoop FS API - which can turn out to be HDFS, S3, on anything else.

I know that the stack layering can be confusing, that is roughly what I was trying to demonstrate.

What is your option on this when you know my reasoning? Would you suggest keeping it as is, even though it may be a bit misleading?

utils/src/main/scala/za/co/absa/enceladus/utils/fs/FileSystemUtils.scala
dk1844: @Zejnilovic why? 
@benedeki where is the verb in `toFsUtils`?

Zejnilovic: Calling `"FileSystem".toFsUtils` sounds a bit weird to me. Just felt this sounds better. I do not have any hard reasons.

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/ConformanceExecution.scala
dk1844: Thanks, suggestion applied.

spark-jobs/src/main/scala/za/co/absa/enceladus/common/config/FileSystems.scala
dk1844: True, removed.

utils/src/main/scala/za/co/absa/enceladus/utils/fs/FileSystemUtils.scala
dk1844: True. Fixed.

spark-jobs/src/main/scala/za/co/absa/enceladus/common/CommonJobExecution.scala
dk1844: field of what? Do you mean a non-implicit parameter of the method `prepareJob[T]()`?

utils/src/main/scala/za/co/absa/enceladus/utils/fs/FileSystemUtils.scala
dk1844: Thanks, desc added.

spark-jobs/src/test/scala/za/co/absa/enceladus/conformance/interpreter/ArrayConformanceSuite.scala
dk1844: I have checked the numbers, `with SparkTestBase` is used about 64 times, and in conjunction of `with HadoopFsTestBase` it is used 16 times. This means, that I could only cover a minority of `SparkTestBase` encounters. I could change those 16 instances of `SparkTestBase with HadoopFsTestBase` for `EnceladusTestBase`.

I actually find this way much more explanatory. With this, you could directly see what is needed (spark, Hadoop FS), for other Enceladus-related suites, there may be other necessary traits, ... Or the other point - I don't think that all Enceladus test suites always need Spark + Hadoop FS.

So I am siding to keeping it as-is. But if you see where I am wrong, go ahead and point it out to me, please.

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/StandardizationExecution.scala
dk1844: Specifically here or in general for FileSystem? I thought about this as well. In cases where only a single FS is used, I think it could. 

With multiple fs, it makes no sense with the current types (I thought about introducing something like `case class InputFs(fs: FileSystem)` and `outputFs` in a similar fashion, then you could have these as implicits and create implicits as needed (because `stdFs` is sometimes an `inputFs` and sometimes an `outputFs`), but I have abandoned this idea because it seemed that it is not worth doing for the sake of 'using implicits', for those cases just explicitly passing a specific FS seems more appropriate.

I will, nonetheless, review the implicit FS usage, your point makes sense.

dk1844: Here, the implicit usage seems justified (there no filesystems around to get confused with). Or are you suggesting to use explicit filesystems in Enceladus everywhere?

utils/src/main/scala/za/co/absa/enceladus/utils/testUtils/HadoopFsTestBase.scala
dk1844: Well, even though I don't think this is the best approach, I tried to do it. Thinking, well, ok, would be fine for tests, I guess. Fewer traits, less mess. But then you stumble upon cases like `za.co.absa.enceladus.examples.CustomRuleSample1` and you realize that some tests might have justified reasons why to provide their own spark session while still needing to test the HadoopFs features, too. In those cases, it would not be correct to base HadoopFsTestBase on SparkTestBase directly, because that already brings its own spark session that might differ from what's needed there. 

I know you could override the spark-session in those cases, too, but that seems quite against the logic of the unit tests, they should IMHO be created on the simples grounds possible and to have anything unrelated dragged in. :(

utils/src/main/scala/za/co/absa/enceladus/utils/testUtils/HadoopFsTestBase.scala
dk1844: Sure, gladly. Done.

utils/src/test/scala/za/co/absa/enceladus/utils/fs/HdfsUtilsSpec.scala
```diff
@@ -20,13 +20,13 @@ import java.io.FileNotFoundException
...
-import za.co.absa.enceladus.utils.testUtils.SparkTestBase
+import za.co.absa.enceladus.utils.testUtils.{HadoopFsTestBase, SparkTestBase}
 
 /**
   * Unit tests for File system utils
   */
-class HdfsUtilsSpec extends AnyWordSpec with Matchers with SparkTestBase {
-  val hdfsUtils = new HdfsUtils(spark.sparkContext.hadoopConfiguration)
+class HdfsUtilsSpec extends AnyWordSpec with Matchers with SparkTestBase with HadoopFsTestBase {
+  val hdfsUtils = fsUtils
```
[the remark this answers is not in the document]
dk1844: True. Fixed.

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/StandardizationExecution.scala
Zejnilovic: I put my comment on a wrong line. My main concern was `PerformanceMetricTools.addJobInfoToAtumMetadata`

dk1844: Oh, you were spot on. I have redone the API of that with relation the FileSystems+PathConfig merge. The implicit is now gone.

spark-jobs/src/main/scala/za/co/absa/enceladus/common/CommonJobExecution.scala
dk1844: Yes, I found a way, I believe it is a tad better now.

utils/src/main/scala/za/co/absa/enceladus/utils/fs/FileSystemUtils.scala
dk1844: Redone.
Now, an instance of `HadoopFsUtil` is obtained via `HadoopFsUtil.getOrCreate(fs: FileSystem)`. This is cache controlled in a similar manner as `fs.toFsUtils` was. Hope you find it more suitable.

The clear upsides are:
 - the method name already suggest cache controlling mechanism being used
-  the implicit class wrapper is not needed
- ... and the infamous and widely-unpopular `toFsUtils` is gone.

utils/src/main/scala/za/co/absa/enceladus/utils/fs/FileSystemUtils.scala
dk1844: Redone, so not relevant anymore.

spark-jobs/src/main/scala/za/co/absa/enceladus/standardization/StandardizationExecution.scala
```diff
@@ -222,9 +219,10 @@ trait StandardizationExecution extends CommonJobExecution {
...
   //scalastyle:off parameter.number
 
-  private def ensureSplittable(df: DataFrame, path: String, schema: StructType)
-                              (implicit spark: SparkSession, fsUtils: DistributedFsUtils): DataFrame = {
-    if (fsUtils.isNonSplittable(path)) {
+  private def ensureSplittable(df: DataFrame, input: PathWithFs, schema: StructType)
+                              (implicit spark: SparkSession): DataFrame = {
+    implicit val fsUtils = input.fileSystem.toFsUtils
+    if (fsUtils.isNonSplittable(input.path)) {
       convertToSplittable(df, schema)
```
Zejnilovic, line 224: If you do not mind touching `StandardizationExecution.convertToSplittable`, I would also make `fsUtils` not implicit. Your new arch makes it obsolete and I favour your new approach.

dk1844: OK, done.

Zejnilovic (approved):
Code Reviewed

AdrianOlosutean (approved):

PR #1594, opened 2020-11-04
1592 Helper scripts don't ignore unknown options

### LINKED ISSUE #1592: Helper scripts swallow any additional options and ignore them
## Describe the bug
Helper scripts swallow any additional options and ignore them

## Expected behaviour
- Print a warning when there is an unrecognized option passed to the scripts
- **Optional**: Have a switch for warning or non-zero exit in case there is an unrecognized option passed

Zejnilovic:
Closes #1592 

RN: Helper scripts now have a setting to either print a warning or do a non-zero exit in case unknown parameters are passed to it. This config variable `EXIT_ON_UNRECOGNIZED_OPTIONS` can be found in `enceladus_env.sh.template`. Can be set to `"true"` or `"false"`.


benedeki (approved):
Looks good. Will test.

benedeki (approved):
- [x] code reviewed
- [ ] pulled
- [ ] built
- [x] run

Works well.
I would consider putting each entry starting with `--` and new line. But not crucial.

benedeki:
Oh, one note. Windows equivalent again 😉

Zejnilovic:
> I would consider putting each entry starting with `--` and new line. But not crucial.

I have thought of this. But the issue is that problematic stuff can be:
- `--datasset-nam alfa` > is a tuple
- `--dry-ran` > is single
- `--dataset-name some name` > should be all three

How do we choose? I picked the lazy way of printing everything unknow out together as I am accustomed in other tools. Or am I having a bad memory?

benedeki:
Well if only entries with `--` would be prefixed with `\n` the grouping would work nicely, I believe. But this is really minor.

Zejnilovic:
> Well if only entries with `--` would be prefixed with `\n` the grouping would work nicely, I believe. But this is really minor.

would not work for the third, in my opinion, the most common mistake. It would show only `name` and it might show it coupled with a different key.

benedeki:
> 
> 
> > Well if only entries with `--` would be prefixed with `\n` the grouping would work nicely, I believe. But this is really minor.
> 
> would not work for the third, in my opinion, the most common mistake. It would show only `name` and it might show it coupled with a different key.

Nope. `name` doesn't have the `--` prefix so there would be no `\n` before it. Therefore `--dataset-name some name` all on one line.

Zejnilovic:
> > > Well if only entries with `--` would be prefixed with `\n` the grouping would work nicely, I believe. But this is really minor.
> > 
> > 
> > would not work for the third, in my opinion, the most common mistake. It would show only `name` and it might show it coupled with a different key.
> 
> Nope. `name` doesn't have the `--` prefix so there would be no `\n` before it. Therefore `--dataset-name some name` all on one line.

But `--dataset-name some` is valid as far as we know. It would not get printed out. Only `name` would be printed out.

benedeki:
> 
> 
> > > > Well if only entries with `--` would be prefixed with `\n` the grouping would work nicely, I believe. But this is really minor.
> > > 
> > > 
> > > would not work for the third, in my opinion, the most common mistake. It would show only `name` and it might show it coupled with a different key.
> > 
> > 
> > Nope. `name` doesn't have the `--` prefix so there would be no `\n` before it. Therefore `--dataset-name some name` all on one line.
> 
> But `--dataset-name some` is valid as far as we know. It would not get printed out. Only `name` would be printed out.

Agreed to keep on one line. I will do the Windows scripts.

scripts/bash/run_enceladus.sh
```diff
@@ -285,12 +285,26 @@ case $key in
...
 done
-set -- "${POSITIONAL[@]}" # restore positional parameters
+
+too_many_options_print() {
+  echo "$1: Found unrecognized options passed to the script. Parameters are:"
+  echo "    $2"
+  echo ""
+}
+
+if [[ -n "${OTHER_PARAMETERS[*]}" ]]; then
```
benedeki, line 300: I am pretty convinced that this should be executed only if  `[ "$HELP_CALL" == "0" ];` So should be placed on current line 330.

benedeki (changes requested):

scripts/bash/run_enceladus.sh
Zejnilovic: When we try and run Git `--help` for example. If there is something specified, it tries to help you with the stuff specified. It even opens man pages for you if you run `git branch --help`. I am not sure we want to offer any services like those. And when you run `git alfa --help` it throws an error that `alfa` is unknown. So, I think this is fine.

And to add to the argument. Our Help executes spark context and runs the spark-jobs. I do not wanna run it more times than needed.

benedeki: I don't know imagine somebody runs a comad with `--version` instead of `--dataset-version`, and adding `--help`...
OK, probably a convoluted counter-examples. Let's keep it like you created it 😉

benedeki (approved):
As I retracted the previous request change I am re-approving.
But also removing the `test run` label, as I added the _Windows_ flavour of the Helper scripts. @dk1844 please test/run it. (I did the _Bash_ ones.)

dk1844:
Tested on Windows, seems to work as expected: 👍 
## Having `SET EXIT_ON_UNRECOGNIZED_OPTIONS=false`
and running `run_standardization.cmd --dataset-name Foo --dataset-version 5 --report-date 2018-07-16 --report-version 1 --raw-format csv --dry-run --bogus` keeps error level at 0:
```
echo %ERRORLEVEL%
0
```
## `Having SET EXIT_ON_UNRECOGNIZED_OPTIONS=true` and
running `run_standardization.cmd --dataset-name Foo --dataset-version 5 --report-date 2018-07-16 --report-version 1 --raw-format csv --dry-run --bogus` sets error level to 127:
```
echo %ERRORLEVEL%
127
```

Non-bogus dry run always results in `ERRORLEVEL` 0 on Windows, no matter the `EXIT_ON_UNRECOGNIZED_OPTIONS` value.

dk1844 (approved):

PR #1596, opened 2020-11-09
622 api export import docs

Zejnilovic:
Documentation for the Import Export

I expect we change the documentation in the audit phase but we need something to go together with the 2.17.

### DIFF
--- a/_docs/2.0.0/usage/menas-api.md
+++ b/_docs/2.0.0/usage/menas-api.md
@@ -24,9 +24,13 @@ This endpoint exports a single entity from Menas. All exports are versionless, e
 - name is the name of the entity as seen in Menas
 - version is the version of the entity as seen in Menas. Version is optional and in that case the latest version will be used
 
-#### Example Import Response for Schema
+**Resposes:**
+| Code | Description |
+|------|-------------|
+| 200  | Successful export with exported entity in the response payload |
+| 404  | Requested entity was not found |
 
-Returns code is 200.
+#### Example Import Response for Schema
 
 **Top level keys**
 
@@ -108,10 +112,16 @@ This endpoint imports a single entity. All imports are versionless. If the impor
 
 Versions of connected entities need to be specified properly. Export of a Dataset carries a Schema and maybe some Mapping tables as connected entities. These have versions and these versions need to exist on Import.
 
-`POST {menas_url}//api/{entity}/importItem`
+`POST {menas_url}/api/{entity}/importItem`
 
 - entity is `schema`, `dataset` or `mappingTable`
 - expects a JSON payload
+
+**Resposes:**
+| Code | Description |
+|------|-------------|
+| 201  | Successful imported with created/updated entity in the response payload |
+| 400  | Bad request. Import payload had validation errors. Errors returned in response json payload |
 
 #### JSON payload
 
@@ -119,7 +129,7 @@ JSON payload is the same as the JSON response from the [export](#example-import-
 
 #### Example Export Response for Schema
 
-On success, it is the same as JSON payload of update or create API, depending if the entity name already existed or not. Response code is 201.
+On success, it is the same as JSON payload of update or create API, depending if the entity name already existed or not.
 
 On failure, you will get a list of errors produced by the validation like bellow
 
--- a/_docs/2.0.0/usage/menas-api.md
+++ b/_docs/2.0.0/usage/menas-api.md
@@ -16,7 +16,7 @@ categories:
 
 ### Export Entity
 
-This feature exports a single entity from Menas. All exports are version less. Except for connected entities. They are kept as is, but need to be validated before import, that they match.
+This feature exports a single entity from Menas. All exports are versionless, except for connected entities. They are kept as is, but need to be validated before import, that they match.
 
 `GET /menas/api/{entity}/exportItem/{name}/{version}`
 
@@ -102,7 +102,7 @@ Example:
 
 ### Entity Import
 
-This feature imports a single entity. All imports are version less. If import does not find an entity with the same name it will create a new one and start the version from 1. If import finds a version it will update the previous version.
+This feature imports a single entity. All imports are versionless. If the import does not find an entity with the same name it will create a new one and start the version from 1. If the import finds an existing version, it will update the previous version.
 
 Versions of connected entities need to be specified properly. Export of a Dataset carries a Schema and maybe a Mapping table as connected entities. These have versions and these versions need to exist on Import.
 
@@ -119,7 +119,7 @@ JSON payload is the same as the JSON response from the [export](#example-import-
 
 On success, it is the same as JSON payload of update or create API, depending if the entity name already existed or not.
 
-On failure, you will get a list off error produced by the validation like bellow
+On failure, you will get a list of errors produced by the validation like bellow
 
 ```json
 {
--- a/_docs/2.0.0/usage/menas-api.md
+++ b/_docs/2.0.0/usage/menas-api.md
@@ -1,0 +1,140 @@
+---
+layout: docs
+title: Usage - Menas API
+version: '2.0.0'
+categories:
+    - '2.0.0'
+    - usage
+---
+
+- [Export Entity](#export-entity)
+  - [Example Import Response for Schema](#example-import-response-for-schema)
+- [Entity Import](#entity-import)
+  - [JSON payload](#json-payload)
+  - [Example Export Response for Schema](#example-export-response-for-schema)
+
+
+### Export Entity
+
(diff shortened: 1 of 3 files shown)


_docs/2.0.0/usage/menas-api.md
```diff
@@ -0,0 +1,140 @@
...
+- [Export Entity](#export-entity)
+  - [Example Import Response for Schema](#example-import-response-for-schema)
+- [Entity Import](#entity-import)
+  - [JSON payload](#json-payload)
+  - [Example Export Response for Schema](#example-export-response-for-schema)
+
+
+### Export Entity
+
+This feature exports a single entity from Menas. All exports are version less. Except for connected entities. They are kept as is, but need to be validated before import, that they match.
```
AdrianOlosutean, line 19:
```suggestion
This feature exports a single entity from Menas. All exports are versionless, except for connected entities. They are kept as is, but need to be validated before import, that they match.
```

_docs/2.0.0/usage/menas-api.md
```diff
@@ -0,0 +1,140 @@
...
+
+#### JSON payload
+
+JSON payload is the same as the JSON response from the [export](#example-import-response-for-schema).
+
+#### Example Export Response for Schema
+
+On success, it is the same as JSON payload of update or create API, depending if the entity name already existed or not.
+
+On failure, you will get a list off error produced by the validation like bellow
```
AdrianOlosutean, line 122:
```suggestion
On failure, you will get a list of errors produced by the validation like bellow
```

_docs/2.0.0/usage/menas-api.md
```diff
@@ -0,0 +1,140 @@
...
+        "absolutePath": "Byte"
+      }
+    ]
+  }
+}
+```
+
+### Entity Import
+
+This feature imports a single entity. All imports are version less. If import does not find an entity with the same name it will create a new one and start the version from 1. If import finds a version it will update the previous version.
```
AdrianOlosutean, line 105:
```suggestion
This feature imports a single entity. All imports are versionless. If the import does not find an entity with the same name it will create a new one and start the version from 1. If the import finds an existing version, it will update the previous version.
```

_docs/2.0.0/usage/menas-api.md
zakera786: I would say ' They are kept as is however need to be validated before import to ensure they match '

AdrianOlosutean (approved):

benedeki (commented):
I believe `_docs/usage/menas-api.md` has to exist as well for redirect to work.

_docs/2.0.0/usage/menas-api.md
```diff
@@ -0,0 +1,140 @@
...
+- [Export Entity](#export-entity)
+  - [Example Import Response for Schema](#example-import-response-for-schema)
+- [Entity Import](#entity-import)
+  - [JSON payload](#json-payload)
+  - [Example Export Response for Schema](#example-export-response-for-schema)
+
+
+### Export Entity
+
+This feature exports a single entity from Menas. All exports are versionless, except for connected entities. They are kept as is, but need to be validated before import, that they match.
```
zakera786, line 19: They are kept as is however would need to be validated before import to ensure that they match.

zakera786: to ensure they match ( remove that)

AdrianOlosutean (dismissed):

_docs/2.0.0/usage/menas-api.md
```diff
@@ -0,0 +1,140 @@
...
+- [Entity Import](#entity-import)
+  - [JSON payload](#json-payload)
+  - [Example Export Response for Schema](#example-export-response-for-schema)
+
+
+### Export Entity
+
+This feature exports a single entity from Menas. All exports are versionless, except for connected entities. They are kept as is, but need to be validated before import.
+
+`GET /menas/api/{entity}/exportItem/{name}/{version}`
```
dk1844, line 21: How about the fact that schema attachments can also be exported via `/menas/api/schema/export/{name}/{version}`?

Also, maybe this would only complicate things, so I can see why you took this shortcut, just want to know if it was deliberate or not: Even though with our deployments, Menas and its API is usually deployed under `/menas`, technically, it can be operated under a different path-prefix, so maybe one could at least mention it somewhere that this mental shortcut is used in the API docs everywhere.

dk1844 (dismissed):

_docs/2.0.0/usage/menas-api.md
Zejnilovic: - `/menas` is a mistake it was meant to be something like `{menas_url}/api/...`. Will fix
- I know of export, but it is used for something else and this export does not have any import counterpart. It is not a full entity export but only schema export

benedeki (commented):
Shouldn't possible HTML Return codes be listed too?

_docs/2.0.0/usage/menas-api.md
```diff
@@ -0,0 +1,142 @@
...
+
+`GET {menas_url}/api/{entity}/exportItem/{name}/{version}`
+
+- entity is `schema`, `dataset` or `mappingTable`
+- name is the name of the entity as seen in Menas
+- version is the version of the entity as seen in Menas. Version is optional and in that case the latest version will be used
+
+#### Example Import Response for Schema
+
+Returns code is 200.
```
benedeki, line 29: This does not belong to the examples chapter, I bet. But few lines higher, I guess.

Also no 404 status or maybe other?

_docs/2.0.0/usage/menas-api.md
```diff
@@ -0,0 +1,142 @@
...
+### Export Entity
+
+This endpoint exports a single entity from Menas. All exports are versionless, except for connected entities. They are kept as is, but need to be validated before import.
+
+`GET {menas_url}/api/{entity}/exportItem/{name}/{version}`
+
+- entity is `schema`, `dataset` or `mappingTable`
+- name is the name of the entity as seen in Menas
+- version is the version of the entity as seen in Menas. Version is optional and in that case the latest version will be used
+
```
benedeki, line 26:
```suggestion

**Resposes:**
| Code | Description |
|------|-------------|
| 200  | Successful export with exported entity in the response payload |
| 404  | Requested entity was not found |

```

_docs/2.0.0/usage/menas-api.md
```diff
@@ -0,0 +1,142 @@
...
+`POST {menas_url}//api/{entity}/importItem`
+
+- entity is `schema`, `dataset` or `mappingTable`
+- expects a JSON payload
+
+#### JSON payload
+
+JSON payload is the same as the JSON response from the [export](#example-import-response-for-schema).
+
```
benedeki, line 119: | Code | Description |
|------|-------------|
| 201  | Successful import of the entity |
| ???  | Error, with payload describing the error |

_docs/2.0.0/usage/menas-api.md
```diff
@@ -0,0 +1,142 @@
...
+#### Example Export Response for Schema
+
+On success, it is the same as JSON payload of update or create API, depending if the entity name already existed or not. Response code is 201.
```
benedeki, line 122: Ditto with position of the code.

_docs/2.0.0/usage/menas-api.md
```diff
@@ -0,0 +1,142 @@
...
+}
+```
+
+### Entity Import
+
+This endpoint imports a single entity. All imports are versionless. If the import does not find an entity with the same name it will create a new one and start the version from 1. If the import finds an existing version, it will update the previous version.
+
+Versions of connected entities need to be specified properly. Export of a Dataset carries a Schema and maybe some Mapping tables as connected entities. These have versions and these versions need to exist on Import.
+
+`POST {menas_url}//api/{entity}/importItem`
```
dk1844, line 111:
```suggestion
`POST {menas_url}/api/{entity}/importItem`
```

_docs/2.0.0/usage/menas-api.md
dk1844: If there truly is a Location header (that 201 Created should be accompanied with), maybe there could be a mention of that, too?

dk1844 (dismissed):

AdrianOlosutean (dismissed):

_docs/2.0.0/usage/menas-api.md
Zejnilovic: @dk1844 I do not know what is location header.

benedeki (approved):

AdrianOlosutean (approved):

PR #1635, opened 2021-01-14
1634: Little project fixes

### LINKED ISSUE #1634: Little project fixes
## Background
Some changes that surfaced during higher Spline version integration. These things resurface often when keeping up the branch with the `Develop`, while they are not _Spline_ related

## Feature
Include the little changes  - improvements - in the code-base, to be earlier in the `Develop` branch then the new _Spline_.

benedeki:
* Maven Shade Plugin upgraded to version 3.2.0
* Added httpclient.version and bower.chart.js.version variables to `pom.xml`
* Menas: Same origin headers
* `ConfigReaderSuite` moved to correct directory

Closes #1634

### DIFF
--- a/menas/pom.xml
+++ b/menas/pom.xml
@@ -47,7 +47,7 @@
         <dependency>
             <groupId>org.apache.httpcomponents</groupId>
             <artifactId>httpclient</artifactId>
-            <version>4.4.1</version>
+            <version>${httpclient.version}</version>
         </dependency>
         <!-- Spark -->
         <dependency>
@@ -251,7 +251,7 @@
         <dependency>
             <groupId>org.webjars.bower</groupId>
             <artifactId>chart.js</artifactId>
-            <version>2.7.3</version>
+            <version>${bower.chart.js.version}</version>
         </dependency>
         <!-- Data formats -->
         <dependency>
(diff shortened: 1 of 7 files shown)


menas/pom.xml
```diff
@@ -251,7 +251,7 @@
         <dependency>
             <groupId>org.webjars.bower</groupId>
             <artifactId>chart.js</artifactId>
-            <version>2.7.3</version>
+            <version>${bower.chart.js.version}</version>
```
AdrianOlosutean, line 254: I don't think this is needed, because it will be replaced in the separation PR

menas/src/main/scala/za/co/absa/enceladus/menas/WebSecurityConfig.scala
```diff
@@ -72,6 +72,9 @@ class WebSecurityConfig @Autowired()(beanFactory: BeanFactory,
         .and()
           .addFilterBefore(kerberosFilter, classOf[UsernamePasswordAuthenticationFilter])
           .addFilterAfter(jwtAuthFilter, classOf[SpnegoAuthenticationProcessingFilter])
+        .headers()
```
AdrianOlosutean, line 75: I think this will also be removed in the separation PR

menas/pom.xml
benedeki: I see, that it will be probably the case.
But just to be sure, I would keep this one now, and removed when not needed.

menas/src/main/scala/za/co/absa/enceladus/menas/WebSecurityConfig.scala
benedeki: I can see that. And as it's nor crucial now, I will remove it.

Zejnilovic (approved):

dk1844 (approved):
(just read through)

AdrianOlosutean (approved):
Looks good

PR #1636, opened 2021-01-15
1634: Little project fixes - optional

benedeki:
* Alphabetical ordering of `pom.xml` properties

Open to opinions if this is useful or not.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -131,72 +131,71 @@
         <encoding>UTF-8</encoding>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <!--plugin versions-->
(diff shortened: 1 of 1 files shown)


AdrianOlosutean (approved):
I would say it's useful, maybe some separating comments with the start and end of kind of dependencies would make it more readable.

benedeki:
> 
> 
> I would say it's useful, maybe some separating comments with the start and end of kind of dependencies would make it more readable.

The dependencies are done this way. I didn't think it's necessary for versions. IMHO in their case it decreases the orientation rather than increases.

Zejnilovic (approved):
I like it

PR #1637, opened 2021-01-18
2.17 release notes update

Zejnilovic:


benedeki (approved):

Zejnilovic:
Release Notes:
Property Definition Updates can be enabled using a switch `menas.feature.property.definitions.allowUpdate` in properties. Thus also a new property `menas.feature.property.definitions.allowUpdate` that can be true or false.

PR #1652, opened 2021-01-27
Atum updated to v3.3.0

dk1844:
Atum version increased to 3.3.0, notable changes related to that:
 - since Atum v3.3.0 is modularized, it is sufficient to use `atum-model` in `migrations`, `utils` and `menas`.
 - `ControlUtils`'s serialization functionality has been moved to `atum-model` as `SerializationUtils` (thus, the usage now does not require full atum, just the lightweight model is sufficient
 - minor updates

## Naive Test run
 - I have test-ran this version on EMR (std + conf), _INFO file appeared and correctly contains information about being created by Atum 3.3.0

### Release note suggestion:
Atum has been updated to v3.3.0 (modularized version, hence smaller footprint in some submodules)


Zejnilovic (approved):

AdrianOlosutean (approved):

menas/pom.xml
```diff
@@ -265,17 +265,6 @@
             <artifactId>guava</artifactId>
             <version>${guava.version}</version>
         </dependency>
-        <dependency>
```
benedeki, line 268: Why to remove the dependency altogether and not just the exclusion? The Atum is used within Menas, as the files bellow show. This way the dependency is transitive via _Utils_

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/ConformanceExecution.scala
```diff
@@ -22,7 +22,7 @@ import org.apache.hadoop.fs.FileSystem
 import org.apache.spark.sql.functions.{lit, to_date}
 import org.apache.spark.sql.{DataFrame, SparkSession}
 import za.co.absa.atum.AtumImplicits
```
benedeki, line 24: Not needed. Just need to remove `AtumImplicits.SparkSessionWrapper`  on line 117 and 122

benedeki (changes requested):
Please change the PR to be against `develop`?

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/ConformanceExecution.scala
dk1844: Thanks for noticing, done.

menas/pom.xml
dk1844: Thanks for the explanation, it is indeed cleaner as suggested. Done.

benedeki (dismissed):
- [x] code reviewed
- [ ] pulled
- [ ] built
- [ ] run

spark-jobs/src/test/scala/za/co/absa/enceladus/standardization/interpreter/stages/SchemaCheckerSuite.scala
```diff
@@ -10,6 +10,5 @@ class SchemaCheckerSuite extends AnyFunSuite with SparkTestBase {
     val sourceFile = FileReader.readFileAsString("src/test/resources/data/bug.json")
     val schema = DataType.fromJson(sourceFile).asInstanceOf[StructType]
     val output = SchemaChecker.validateSchemaAndLog(schema)
```
benedeki, line 12:
```suggestion
    val output = SchemaChecker.validateSchemaAndLog(schema)
    val expected = (
        List(
          "Validation error for column 'Conformed_TXN_TIMESTAMP', pattern 'yyyy-MM-ddTHH:mm:ss.SSSX': Illegal pattern character 'T'"
        ),
        List()
      )
    assert(output == expected)    
```

dk1844: Of course; applying suggestion.

benedeki (approved):
Thanks for the good boy scout attitude regarding the `println`s 👍

Zejnilovic (approved):
I might be weird but I really like how it came to be

Looks good

- [x] code reviewed
- [x] pulled
- [x] built
- [x] run

Issue #1653, opened 2021-01-27
HDFS browser doesn't work with EMR HDFS


lokm01:
## Describe the bug
When running menas and pointing it to EMR based HDFS, the service throws the following exception:

`
java.lang.IllegalStateException: Illegal access: this web application instance has been stopped already. Could not load [commons-logging.properties]. The following stack trace is thrown for debugging purposes as well as to attempt to terminate the thread which caused the illegal access.
	at org.apache.catalina.loader.WebappClassLoaderBase.checkStateForResourceLoading(WebappClassLoaderBase.java:1385)
	at org.apache.catalina.loader.WebappClassLoaderBase.findResources(WebappClassLoaderBase.java:985)
	at org.apache.catalina.loader.WebappClassLoaderBase.getResources(WebappClassLoaderBase.java:1086)
	at org.apache.htrace.commons.logging.LogFactory$4.run(LogFactory.java:1388)
	at java.security.AccessController.doPrivileged(Native Method)
	at org.apache.htrace.commons.logging.LogFactory.getResources(LogFactory.java:1407)
	at org.apache.htrace.commons.logging.LogFactory.getConfigurationFile(LogFactory.java:1469)
	at org.apache.htrace.commons.logging.LogFactory.getFactory(LogFactory.java:457)
	at org.apache.htrace.commons.logging.LogFactory.getLog(LogFactory.java:685)
	at org.apache.htrace.Tracer.<clinit>(Tracer.java:33)
	at org.apache.htrace.Trace.isTracing(Trace.java:154)
	at org.apache.htrace.Trace.startSpan(Trace.java:84)
	at org.apache.htrace.Trace.startSpan(Trace.java:70)
	at org.apache.hadoop.hdfs.DFSClient.getPathTraceScope(DFSClient.java:3560)
	at org.apache.hadoop.hdfs.DFSClient.getFileInfo(DFSClient.java:2115)
	at org.apache.hadoop.hdfs.DistributedFileSystem$22.doCall(DistributedFileSystem.java:1305)
	at org.apache.hadoop.hdfs.DistributedFileSystem$22.doCall(DistributedFileSystem.java:1301)
	at org.apache.hadoop.fs.FileSystemLinkResolver.resolve(FileSystemLinkResolver.java:81)
	at org.apache.hadoop.hdfs.DistributedFileSystem.getFileStatus(DistributedFileSystem.java:1301)
[...]

PR #1654, opened 2021-01-27
#1653 HDFS browser doesn't work with EMR HDFS

lokm01:
Adds htrace dependency to menas. Without this, HDFS browser throws an exception when trying to list folders on EMR HDFS.


lokm01:
Closing this, I think it'd environment related.

Issue #1571, opened 2020-10-19
UI Changes for Dataset properties - pilot property: Data Classification


benedeki:
## Background
Dataset properties is a map attached to a Dataset.
The keys of the map are defined on application setup level. They are common to all Datasets of the deployment.
The values then are entered by users and are specific per each Dataset.

## Feature
Implement the UI changes to display properties and the control for selection from the list for pilot _Data Classification_ property.
Including the API calls for dealing with the properties.

Zejnilovic:
Next step: Will add analysis and needs

Zejnilovic:
We have a new Properties API. One is for setting the Property Definitions and one is to add properties to Datasets. This issue is about adding the properties do Dataset. Setting properties on Dataset should come with the Create/Edit dialogue window. But we need to keep in mind that properties will be in later versions added to all versionedModel entities (schema, mapping table)

All property definitions can be found at `BASE_URL/api/properties/datasets`. This returns an array of PropertyDefinitions. The latest version of each. Here we are interested in `$[*].propertyType,essentiality`. That jsonPath would show off the property type and if it is mandatory or not. 

Currently, types are a string or enum. For each of these types, we would have a `display` and `add` fragment. These would be located (presumably) in properties folder `properties` in `/menas/ui/components` and have a similar setup as the ConformanceRules that dynamically load appropriate fragment. 

In Dataset Create/Edit fragments there would be a (presumably) `CustomListItem` that would then invoke appropriate add fragments. to be edited. In Dataset Details appropriate detail fragments.

Behaviour on essentiality:
 - mandatory - needs to be filled in
 - optional - a pop-up warning if the person is sure
 - optional - leave it be

Attributes of String Property type:
 - suggested value (might be empty)

Attributes of Enum Property type
- allowed values
- suggested value (has to be one of the allowed ones)

lokm01:
Release Notes:
[Menas] Implemented the UI changes for dataset properties

aaishaosman:
SECURITY CLASSIFICATION per dataset – selection from dropdown Public, Internal, Confidential, Secret

aaishaosman:
For future…i iwll still define these
CONTENT TYPE per dataset (customer data, transactional etc)
RETENTION per dataset – still figuring out what we need here to help eventually drive data lifecycle management on AWS
GENERAL – per datasets, open text tagging concept to assist with searching etc

PR #1657, opened 2021-02-01
#1571 UI Changes for Dataset properties

lokm01:
This implements the first draft of UI changes for dataset properties.


menas/ui/components/dataset/addDataset.fragment.xml
```diff
@@ -15,7 +15,7 @@
 
 <core:FragmentDefinition xmlns="sap.m" xmlns:core="sap.ui.core" xmlns:form="sap.ui.layout.form"
                          xmlns:hdfs="components.hdfs">
-    <Dialog id="addDatasetDialog" title="{entity>/title} Dataset" stretchOnPhone="true" class="minW40rem">
+    <Dialog id="addDatasetDialog" title="{entity>/title} Dataset" stretchOnPhone="true" class="minW55rem maxW75p w100p">
```
Zejnilovic, line 18: Not sure if from here. But the dialogue window is across the whole window now.

image

menas/ui/components/dataset/addDataset.fragment.xml
```diff
@@ -29,20 +29,37 @@
...
+                          <InputListItem class="inputListItemSmallLabel" label="{entity>name}" tooltip="{entity>description}">
+                            <HBox alignItems="Center" justifyContent="End">
+                                <Input type="Text" width="20rem" visible="{= ${entity>propertyType/_t} === 'StringPropertyType'}" 
+                                    value="{entity>value}" valueState="{entity>validation}" valueStateText="{entity>validationText}"></Input>
+                                <Select width="20rem" items="{path: 'entity>propertyType/allowedValues', templateShareable: false}" 
+                                    selectedKey="{entity>value}" valueState="{entity>validation}" valueStateText="{entity>validationText}"
+                                    visible="{= ${entity>propertyType/_t} === 'EnumPropertyType'}">
+                                    <core:Item key="{entity>value}" text="{entity>value}"></core:Item>
+                                </Select>
+                                <ObjectStatus text="{entity>essentiality/_t}" class="essentialityStatus"
```
Zejnilovic, line 57: Is it possible to have the icon on the right?
image

Zejnilovic (commented):
Looks good

- [ ] code reviewed
- [x] pulled
- [x] built
- [x] run

menas/ui/components/dataset/addDataset.fragment.xml
lokm01: Yeah should stretch 75% width max. Much easier to read in my opinion.

menas/ui/components/dataset/addDataset.fragment.xml
lokm01: Yeah it is, it didn't look very good :) can try style it up a bit more

menas/ui/components/dataset/addDataset.fragment.xml
lokm01: Happy to restyle if you have suggestions :)

menas/ui/service/EntityValidationService.js
```diff
@@ -71,4 +71,26 @@ var EntityValidationService = new function () {
...
 
+  this.hasValidProperties = function(aProperties) {
+    let isOk = true;
+
+    aProperties.map((oProp) => {
+      //check essentiality - if Mandatory property has no or empty value, we fail
+      if(oProp.essentiality._t === "Mandatory" && !oProp.value) {
+        isOk = false;
+        oProp.validation = "Error";
+        oProp.validationText = `Provide valid ${oProp.name} - ${oProp.description}`;
```
dk1844, line 82: This might look a bit weird if no description was given (`propName1 - `), so maybe the string could be different in that case.

menas/ui/service/EntityValidationService.js
```diff
@@ -71,4 +71,26 @@ var EntityValidationService = new function () {
     return isOk;
   };
 
+  this.hasValidProperties = function(aProperties) {
+    let isOk = true;
+
+    aProperties.map((oProp) => {
+      //check essentiality - if Mandatory property has no or empty value, we fail
+      if(oProp.essentiality._t === "Mandatory" && !oProp.value) {
```
dk1844, line 79: I think it would be worth showing some `validationText` informing about a `Recommended` property having no value, too. I believe this is not addressed.

dk1844 (approved):
(just read the code) LGTM.

menas/ui/service/EntityValidationService.js
lokm01: Yes I know... I need to check how to trigger this validation on focus out - currently it happens on the form submission

menas/ui/components/dataset/addDataset.fragment.xml
lokm01: Alright so I restyled this, made the default width smaller and also optimized for smaller screen. I think this looks better than before and is more usable imho

menas/ui/service/EntityValidationService.js
lokm01: Yeah changed that slightly

menas/ui/service/EntityValidationService.js
lokm01: Alright so this is implemented in the last commit, don't hate me for the timeout.

I tried to do this more declaratively, but I'd have to dig deeper into how these binding templates are initialized. I think for now we should try this and if it's causing issues, we'll fix in a separate issue.

dk1844 (dismissed):

lokm01:
Created #1668 which suggests an clean approach to fixing that timeout hack. However let's leave this as a separate issue for now and based on how this works we can prioritize it properly.

dk1844 (dismissed):

AdrianOlosutean (dismissed):
Looks good

Zejnilovic (approved):
It works. I like it. I have ran it.

Question: Why not have the properties with the basic info? Seems like a bigger use-case then most from the tabs.

lokm01:
> It works. I like it. I have ran it.
> 
> Question: Why not have the properties with the basic info? Seems like a bigger use-case then most from the tabs.

Yeah that was the original plan :) rather went for a quick solution. But I agree, let me see if I can restyle very quickly, otherwise we put it into a separate ticket.

lokm01:
> It works. I like it. I have ran it.
> 
> Question: Why not have the properties with the basic info? Seems like a bigger use-case then most from the tabs.

Alright check now. Had to change the form element used in that tab, but I think its better.

menas/ui/components/dataset/datasetInfo.fragment.xml
```diff
@@ -14,41 +14,80 @@
...
+            </f:fields>
+          </f:FormElement>
+          <f:FormElement label="Created By">
+            <f:fields>
+              <Text id="currentDatasetUserCreated" text="{dataset>/userCreated}" />
+            </f:fields>
+          </f:FormElement>
+        </f:formElements>
+      </f:FormContainer>
+      <f:FormContainer formElements="{/currentDataset/propertiesDisplay}">
```
Zejnilovic, line 84: What are the chances to have them in the same order as when editing dataset?

lokm01: Will check, this will probably create those margins, but we should be able to force-style otherwise.

image

lokm01: Ok, restyled those margins to be less dramatic, tried without them but I think they actually make it easier to read
image

Zejnilovic: While the change looks good to me. What I talked about is sorting the list of propertiesDisplay the same way when you edit dataset

#### Detail
image

#### Dataset Edit
image

lokm01: Will have a look

lokm01: Ok added a sorter. This will only work on a best-effort basis. In cases where the dataset loads faster than the property definitions, a default order will be used.

dk1844 (approved):

dk1844 (approved):

Zejnilovic (approved):
Looks good

- [ ] code reviewed
- [x] pulled
- [x] built
- [x] run

AdrianOlosutean (approved):
Tested locally, Looks good

PR #1660, opened 2021-02-02
1659 Fix Property Definition Update

### LINKED ISSUE #1659: Import of PropertyDefinition does not update Essentiality
## Describe the bug
When import and presumably update is used on the property definition it does not update essentiality

## To Reproduce
1. have a property definition
2. update the definition with different essentiality
3. it does not update

## Expected behaviour
I expect it to update

Zejnilovic:
Fixes #1659 

RN: Update and import API of Property Def now properly updates essentiality, putIntoInfoFile and propertyType

### DIFF
--- a/menas/src/main/scala/za/co/absa/enceladus/menas/services/PropertyDefinitionService.scala
+++ b/menas/src/main/scala/za/co/absa/enceladus/menas/services/PropertyDefinitionService.scala
@@ -34,7 +34,11 @@ class PropertyDefinitionService @Autowired()(propertyDefMongoRepository: Propert
 
   override def update(username: String, propertyDef: PropertyDefinition): Future[Option[PropertyDefinition]] = {
     super.update(username, propertyDef.name, propertyDef.version) { latest =>
-      latest.setDescription(propertyDef.description)
+      latest
+        .setPropertyType(propertyDef.propertyType)
+        .setPutIntoInfoFile(propertyDef.putIntoInfoFile)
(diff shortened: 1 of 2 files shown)


dk1844 (approved):

menas/src/main/scala/za/co/absa/enceladus/menas/services/PropertyDefinitionService.scala
```diff
@@ -34,7 +34,11 @@ class PropertyDefinitionService @Autowired()(propertyDefMongoRepository: Propert
 
   override def update(username: String, propertyDef: PropertyDefinition): Future[Option[PropertyDefinition]] = {
     super.update(username, propertyDef.name, propertyDef.version) { latest =>
-      latest.setDescription(propertyDef.description)
+      latest
```
benedeki, line 37: Why using the setters here, creating 4 new instances, instead of using the copy constructor directly, once, with the four properties?

Zejnilovic: Keeping up with what is already established in dataset, schema and mapping table services

benedeki: Not blocking for sure. Just considered it little ineffective. 😉

AdrianOlosutean (approved):
LGTM

lokm01 (approved):
Reviewed the code, looks good.

PR #1667, opened 2021-02-04
#1666: `overrideMappingTableOwnFilter` in `MappingConformanceRule` makes its old serialization incompatible

### LINKED ISSUE #1666: `overrideMappingTableOwnFilter` in `MappingConformanceRule` makes its old serialization incompatible
## Describe the bug
The new property `MappingConformanceRule.overrideMappingTableOwnFilter` in  defined as not optional. That makes it old serialization incompatible with the new definition - property value os considered `null`.

## To Reproduce
1. Have a pre 2.18 serialziation of  `MappingConformanceRule`
2. Load it
3. Try to see it
4. See error (attached picture)
image

## Expected behaviour
No error

benedeki:
* `overrideMappingTableOwnFilter` made optional
* added UT testing the old serialization
* fixed a few compilation warnings
Fixes #1666 

Release notes:
`overrideMappingTableOwnFilter` was made optional in `MappingConformanceRule` de-/serialization so old serializations of conformance rules without the property are still readable.

### DIFF
--- a/data-model/src/main/scala/za/co/absa/enceladus/model/conformanceRule/package.scala
+++ b/data-model/src/main/scala/za/co/absa/enceladus/model/conformanceRule/package.scala
@@ -18,7 +18,7 @@ package za.co.absa.enceladus.model
 import com.fasterxml.jackson.annotation.JsonSubTypes.Type
 import com.fasterxml.jackson.annotation.{JsonSubTypes, JsonTypeInfo}
 import za.co.absa.enceladus.model.dataFrameFilter.DataFrameFilter
-import za.co.absa.enceladus.model.conformanceRule.MappingConformanceRule.DEFAULT_OVERRIDE_MAPPING_TABLE_OWN_FILTER
+import za.co.absa.enceladus.model.conformanceRule.MappingConformanceRule.DefaultOverrideMappingTableOwnFilter
 
 package object conformanceRule {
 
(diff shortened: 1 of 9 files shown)


data-model/src/main/scala/za/co/absa/enceladus/model/conformanceRule/package.scala
```diff
@@ -153,7 +157,8 @@ package object conformanceRule {
 
   object MappingConformanceRule {
     // attributeMappings property has key's with dot's that mongo doesn't accept; this symbol is used to replace the dots
-    val DOT_REPLACEMENT_SYMBOL: Char = '^'
+    final val DOT_REPLACEMENT_SYMBOL: Char = '^'
+    final val DEFAULT_OVERRIDE_MAPPING_TABLE_OWN_FILTER: Boolean = false
```
dk1844, line 161: I don't think this is the proper letter casing in Scala for constants. I believe `CapitalizedCamelCase` should be used

dk1844 (approved):
Otherwise LGTM.

data-model/src/main/scala/za/co/absa/enceladus/model/conformanceRule/package.scala
benedeki: I agree. But I adhered to the existing constant. On the other hand, why not to change both of them... 🤔

AdrianOlosutean: I also had the same first impression: [here](https://docs.scala-lang.org/style/naming-conventions.html)

dk1844 (approved):

Zejnilovic (approved):
Code Reviewed

Zejnilovic (approved):
Built and Run. Menas no longer has the issue it had with old datasets

AdrianOlosutean (approved):
Looks good

HuvarVer (approved):
Build,
run,
tested MappingConformanceRule

PR #1672, opened 2021-02-08
Hdfs browser hiding

### LINKED ISSUE #1627: Add hide functionality to the HDFS (file) browser component
## Background
There might be cases when HDFS (file) browser might not be able to connect to HDFS or S3, but Menas should still have full functionality as a metadata store of sorts

## Feature
Add hide functionality to the HDFS (file) browser so if it is unable to read files, let the user just specify them by hand in the text box above.

dk1844:
HDFS Browser can now be hidden
 - by a click of a button
 - automatically on dialog open when one of the paths cannot be displayed in the HDFS Browser

Background:
 - the HDFS listing AJAX call was originally part of the HDFS browser itself; the call routine has been extracted into a HdfsService, directly used by the HDFS Browser. 
 - jQuery & ES6 Promise wrappers have been added for the HDFS listing call to enable promise composition to be used in dialogs (dataset dialog) where multiple HDFS checks are needed without custom chaining
 - The dialog assembly logic has been changed for the service loading to be done first and the dialog to be shown later, otherwise the dialog elements may flicker to react to the listing changes.

### Dataset dialog
simple-path-dataset
hdfs-browser-path-dataset
### Mapping table dialog
simple-path-mapping-table

## Release notes suggestion:
 - Entering HDFS/S3 paths can now be done either using the existing HDFS Browser or via a simple input field. The mode is chosen automatically on dialog open and can be switched manually back and forth.

Closes #1627.

### DIFF
--- a/menas/ui/components/dataset/addDataset.fragment.xml
+++ b/menas/ui/components/dataset/addDataset.fragment.xml
@@ -32,8 +32,7 @@
                             <!-- HDFS Browser and its label -->
                         	<Input type="Text" id="selectedRawHDFSPathLabel" visible="{entity>/hdfsBrowserEnabled}" />
                             <hdfs:HDFSBrowser id="newDatasetRawHDFSBrowser" height="300px" width="100%"
-                                          busyControl="addDatasetDialog"
-                                          pathInput ="selectedRawHDFSPathLabel" restURI="{entity>/hdfsRestUri}"
+                                          busyControl="addDatasetDialog" pathInput ="selectedRawHDFSPathLabel"
                                           HDFSPath="{entity>/hdfsPath}" visible="{entity>/hdfsBrowserEnabled}" />
                             <!-- Simple field as an alternative -->
                             <Input id="newDatasetRawSimplePath" type="Text" placeholder="Dataset raw path" width="100%"
@@ -45,8 +44,7 @@
                             <!-- HDFS Browser and its label -->
                             <Input type="Text" id="selectedPublishHDFSPathLabel" visible="{entity>/hdfsBrowserEnabled}" />
                             <hdfs:HDFSBrowser id="newDatasetPublishHDFSBrowser" height="300px" width="100%"
-                                          busyControl="addDatasetDialog"
-                                          pathInput="selectedPublishHDFSPathLabel" restURI="{entity>/hdfsRestUri}"
+                                          busyControl="addDatasetDialog" pathInput="selectedPublishHDFSPathLabel"
                                           HDFSPath="{entity>/hdfsPublishPath}" visible="{entity>/hdfsBrowserEnabled}" />
                             <!-- Simple field as an alternative -->
                             <Input id="newDatasetPublishSimplePath" type="Text" placeholder="Dataset publish path" width="100%"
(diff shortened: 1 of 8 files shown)


menas/ui/components/dataset/addDataset.fragment.xml
```diff
@@ -29,19 +29,43 @@
                         <core:Fragment type="XML" fragmentName="components.schema.selector.schemaSelector"/>
                         <Label text="HDFS raw data folder path"/>
```
lokm01, line 30: Can we rename this to something more generic? Alternatively change the text with the switch as well?

menas/ui/components/dataset/addDataset.fragment.xml
```diff
@@ -29,19 +29,43 @@
...
                         <VBox>
-                        	<Input type="Text" id="selectedRawHDFSPathLabel"/>
+                            <HBox>
+                                <!-- HDFS Browser input -->
+                                <Input type="Text" id="selectedRawHDFSPathLabel" visible="{entity>/hdfsBrowserEnabled}">
+                                    <layoutData>
+                                        <FlexItemData growFactor="5"/>
```
lokm01, line 36: Interesting, what does this do? :)

I never learned how to use these 👍

lokm01 (commented):
Reviewed the code, overall I like it.

I think we need to rename those labels too to make it a bit more intuitive.

Zejnilovic (commented):
- [ ] code reviewed
- [x] pulled
- [x] built
- [x] run

Question: 
Could we improve when we call /api/hdfs/list ? 
It is called four times if both field and list are shown
It is called twice even though it is not evaluating anything in the hidden mode.

menas/ui/components/dataset/addDataset.fragment.xml
dk1844: I wouldn't say I have learned how to use it, but here it allows input to grow beyond what it normally would. Without the `growFactor`, it would look like this:
no-grow-factor
(both elements in the `HBox` are only willing to grow to half of the width if allowed to grow at all)
Setting a higher `growFactor` has the effect of allowing the `Input` to grow beyond its fair-share space and thus, the `Input `takes all the free available space.

menas/ui/components/dataset/addDataset.fragment.xml
dk1844: True, makes sense. Renamed.

menas/ui/components/dataset/addDataset.fragment.xml
lokm01: Nice!

dk1844:
Hi @Zejnilovic, 

> Question:
> Could we improve when we call /api/hdfs/list ?
> It is called four times if both field and list are shown
> It is called twice even though it is not evaluating anything in the hidden mode.

I assume you are talking about DatasetDialog and not MappingTableDialog?

For DatasetDialog, there two calls are necessary because there are two paths that need checking.
For MappingTableDialog, there is one call necessary since there is only one path to check.

Now, this is happening before the dialog is shown and based on the outcome (all paths checks return 200 OK -> show HDFS Brower; simple mode otherwise). If the simple mode is chosen, the HDFS Browser(s) is/are hidden and no checks other checks are run.

So far so good, I would say.

If HDFS Browser mode is enabled, it drives its own service call which I have not changed (except for the service call extraction to make it reusable). This is, I assume, the issue you are referring to - that there are too many calls. I am correct in the understanding of the problem statement?

Zejnilovic (approved):

PR #1674, opened 2021-02-08
1637 Add simple switch for prop definitions updates

Zejnilovic:
So I tried a few approaches. Flips library, togglz library, refresh scope, cloud starter, profiles, custom annotations/Aspect and maybe more. Not sure. Got lost. Anyway, all too big or complicated. Might be hard to remove.

Also, found out that changing values of app props on the flight is Spring anti-pattern. 

**Solution** is a "simple" property that is set at the beginning of service start, has a default of false and only enables and disables property definitions updates. Will also be quick to revert without leaving any ghost annotations or configs. 

Nagging: Would be quite easier if we already had that stand-alone Configuration object using `@ConfigurationProperties`. 

Fixes #1637 

Relase Notes: Property Definition Updates can be enabled using a switch `menas.feature.property.definitions.update` in properties. Thus also a new property `menas.feature.property.definitions.update` that can be true or false.

### DIFF
--- a/menas/src/main/resources/application.properties.template
+++ b/menas/src/main/resources/application.properties.template
@@ -157,3 +157,6 @@ menas.oozie.splineMongoURL=mongodb://localhost:27017
 #javax.net.ssl.trustStorePassword=somePassword
 #javax.net.ssl.keyStore=/path/to/keystore.jks
 #javax.net.ssl.keyStorePassword=somePassword
+
+# Temporary fix for locking of property definition updates #1673
+# menas.feature.property.definitions.allowUpdate=false
--- a/menas/src/main/scala/za/co/absa/enceladus/menas/controllers/PropertyDefinitionController.scala
+++ b/menas/src/main/scala/za/co/absa/enceladus/menas/controllers/PropertyDefinitionController.scala
@@ -16,14 +16,18 @@
 package za.co.absa.enceladus.menas.controllers
 
 import java.net.URI
+import java.util.Optional
 import java.util.concurrent.CompletableFuture
 
-import org.springframework.beans.factory.annotation.Autowired
+import com.mongodb.client.result.UpdateResult
+import org.springframework.beans.factory.annotation.{Autowired, Value}
 import org.springframework.http.{HttpStatus, ResponseEntity}
 import org.springframework.security.core.annotation.AuthenticationPrincipal
 import org.springframework.security.core.userdetails.UserDetails
 import org.springframework.web.bind.annotation._
+import za.co.absa.enceladus.menas.exceptions.EndpointDisabled
 import za.co.absa.enceladus.menas.services.PropertyDefinitionService
+import za.co.absa.enceladus.model.ExportableObject
 import za.co.absa.enceladus.model.properties.PropertyDefinition
 
 import scala.concurrent.ExecutionContext.Implicits.global
@@ -39,6 +43,9 @@ class PropertyDefinitionController @Autowired()(propertyDefService: PropertyDefi
 
   import za.co.absa.enceladus.menas.utils.implicits._
 
+  @Value("${menas.feature.property.definitions.allowUpdate:false}")
+  private val enabled: Boolean = false
+
   @GetMapping(Array(""))
   def getAllDatasetProperties(): CompletableFuture[Seq[PropertyDefinition]] = {
     logger.info("retrieving all dataset properties in full")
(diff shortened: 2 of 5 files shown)


lokm01:
Checked the code, LGTM

Also I think this approach is fine for now.

lokm01 (dismissed):

menas/src/main/resources/application.properties.template
```diff
@@ -157,3 +157,6 @@ menas.oozie.splineMongoURL=mongodb://localhost:27017
 #javax.net.ssl.trustStorePassword=somePassword
 #javax.net.ssl.keyStore=/path/to/keystore.jks
 #javax.net.ssl.keyStorePassword=somePassword
+
+# Temporary fix for locking of property definition updates #1673
+menas.feature.property.definitions.update=true
```
AdrianOlosutean, line 162: In the description it's `has a default of false and only enables and disables property definitions updates`, which I would agree with

Zejnilovic:
```suggestion
# menas.feature.property.definitions.update=false
```

lokm01 (dismissed):

menas/src/main/resources/application.properties.template
```diff
@@ -157,3 +157,6 @@ menas.oozie.splineMongoURL=mongodb://localhost:27017
 #javax.net.ssl.trustStorePassword=somePassword
 #javax.net.ssl.keyStore=/path/to/keystore.jks
 #javax.net.ssl.keyStorePassword=somePassword
+
+# Temporary fix for locking of property definition updates #1673
+# menas.feature.property.definitions.update=false
```
benedeki, line 162: **Minor"**
```suggestion
# menas.feature.property.definitions.allowUpdate=false
```

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/PropertyDefinitionController.scala
```diff
@@ -75,4 +86,33 @@ class PropertyDefinitionController @Autowired()(propertyDefService: PropertyDefi
     super.getVersionDetail(propertyName, version)
   }
 
+  @PostMapping(Array("/importItem"))
```
benedeki, line 89: New endpoint?

Zejnilovic: Parent endpoints. Not new, they just check for the switch and then call super

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/RestExceptionHandler.scala
```diff
@@ -56,6 +56,11 @@ class RestExceptionHandler {
     ResponseEntity.notFound().build[Any]()
   }
 
+  @ExceptionHandler(value = Array(classOf[EndpointDisabled]))
+  def handleEndpointDisabled(exception: EndpointDisabled): ResponseEntity[Any] = {
+    ResponseEntity.status(HttpStatus.I_AM_A_TEAPOT).build[Any]() // Could change for LOCKED but I like this more
```
Zejnilovic, line 61: I am quite surprised no one comment on this.

menas/src/main/resources/application.properties.template
Zejnilovic: The life of this prop is gonna be 1 month, but sure.

benedeki: You are right. And it's really minor. So unless you change something else too I would say don't bother.

Zejnilovic: I did it already 😄

benedeki: Ah, already changed meanwhile. 🙇 (Sorry for the unneeded nitpicking)

benedeki (approved):
- [x] code reviewed
- [ ] pulled
- [ ] built
- [ ] run

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/RestExceptionHandler.scala
benedeki: The comment in code is there, that explains it 😉

benedeki:
Btw, don't you want to merge this into The [Hotfix 2.18.1 branch](https://github.com/AbsaOSS/enceladus/tree/hotfix/2.18.1)?

Zejnilovic:
Changed base to hotfix 2.18.1. Please re-approve

benedeki (approved):

PR #1680, opened 2021-02-10
#1679 Landing page information takes a minute to load with 1.3M runs

### LINKED ISSUE #1679: Landing page information takes a minute to load with 1.3M runs
## Describe the bug
The landing page statistics take a shocking 1 minute to load when there's a lot of entities and namely runs. From UX perspective this is quite unacceptable.

This query should be cached or regularly materialized - the landing page statistics - graphs & tiles should load almost immediately.

lokm01:
Fixes #1679 

This PR proposes how such expensive queries could be addressed before we invest into the massive data model rebuild or before we got requirements with more advanced caching needs (where products like redis could be used)

Here Spring will run the recalculation of the landing page statistics upon the start up and subsequently in 15 minute intervals. These statistics are stored in a mongo collection and the API endpoint simply reads the latest record.

### DIFF
--- a/menas/src/main/scala/za/co/absa/enceladus/menas/controllers/LandingPageController.scala
+++ b/menas/src/main/scala/za/co/absa/enceladus/menas/controllers/LandingPageController.scala
@@ -19,7 +19,6 @@ import java.util.concurrent.CompletableFuture
 
 import scala.concurrent.Future
 
-import org.slf4j.LoggerFactory
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.scheduling.annotation.Async
 import org.springframework.scheduling.annotation.Scheduled
--- a/menas/src/main/scala/za/co/absa/enceladus/menas/controllers/LandingPageController.scala
+++ b/menas/src/main/scala/za/co/absa/enceladus/menas/controllers/LandingPageController.scala
@@ -46,7 +46,7 @@ class LandingPageController @Autowired() (datasetRepository: DatasetMongoReposit
   import za.co.absa.enceladus.menas.utils.implicits._
 
   @GetMapping(path = Array("/info"))
-  def retreiveLandingPageInfo(): CompletableFuture[LandingPageInformation] = {
+  def retrieveLandingPageInfo(): CompletableFuture[LandingPageInformation] = {
     landingPageRepository.get()
   }
 
@@ -65,7 +65,7 @@ class LandingPageController @Autowired() (datasetRepository: DatasetMongoReposit
   def scheduledLandingPageStatsRecalc(): CompletableFuture[_] = {
     logger.info("Running scheduled landing page statistics recalculation")
     for {
-      newStats <- landingPageInfo
+      newStats <- landingPageInfo()
       res <- landingPageRepository.updateStatistics(newStats)
     } yield res 
   }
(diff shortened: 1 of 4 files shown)


menas/src/main/scala/za/co/absa/enceladus/menas/controllers/LandingPageController.scala
```diff
@@ -48,4 +59,14 @@ class LandingPageController @Autowired() (datasetRepository: DatasetMongoReposit
       todaysStats <- runsService.getTodaysRunsStatistics()
     } yield LandingPageInformation(dsCount, mtCount, schemaCount, runCount, todaysStats)
   }
+
+  @Async
+  @Scheduled(fixedDelay = 900000)
+  def scheduledTest(): CompletableFuture[_] = {
```
dk1844, line 65:
```suggestion
  @Scheduled(initialDelay = 1000, fixedDelay = 900000)
  def scheduledLandingPageStatsRecalc(): CompletableFuture[_] = {
```
I hope I understand the `@Scheduled` annotation correctly, but it seems to me that we are risking that in the first 15 minutes there is no landing page info cached.

dk1844 (approved):

dk1844 (approved):
(Just read the code)

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/LandingPageController.scala
lokm01: Good point, I also wondered! :) Thanks

dk1844 (dismissed):

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/LandingPageController.scala
```diff
@@ -48,4 +59,14 @@ class LandingPageController @Autowired() (datasetRepository: DatasetMongoReposit
       todaysStats <- runsService.getTodaysRunsStatistics()
     } yield LandingPageInformation(dsCount, mtCount, schemaCount, runCount, todaysStats)
   }
+
+  @Async
+  @Scheduled(initialDelay = 1000, fixedDelay = 900000)
```
benedeki, line 64: Also, is 15 minutes refresh a good frequency?

benedeki (commented):
- [x] code reviewed
- [x] pulled
- [x] built
- [x] run

The comments are about code styling & typos.
Generally it's more a workaround then proper solution IMHO. But seems to work as intended.

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/LandingPageController.scala
lokm01: Good question, these annotations also expect constants so I couldn't find a quick way to make them configurable

benedeki: As written elsewhere, I hope this is a quickfix. So let's not dwell on it for too long. 
On second thought, 15 minutes or 5 doesn't make much difference :-)

lokm01: Hmm maybe 5 mins is more reasonable from user perspective.

benedeki (approved):

PR #1699, opened 2021-02-25
Sending errors to a secured kafka

### LINKED ISSUE #1352: Sending errors to Kafka should work with Secure Kafka cluster
## Background
Since our jobs are supposed to be running on corp environments we expect secure Kafka can be used there,

## Feature
Add support for secure Kafka (Kerberos) and Schema registry (PKI) for the plugin that sends all standardization and conformance errors to Kafka.

Jobs are already configured with keystore and truststore paths, as well as jaas.conf. The driver already initializes secure Kafka. Probably need to make sure secure Kafka executors initialize secure Kafka as well. And need to add 
```
kafka.security.protocol="SASL_SSL"
kafka.sasl.mechanism="GSSAPI"
```
if configured in `application.conf`

dk1844:
As the issue specifies, the following settings are needed for SSL secured Kafka topic to be produced messages to. Those can be applied either by updating the `spark-jobs`' `application.conf` or via the `--conf "spark.driver.extraJavaOptions=-Dconfig.key.here=config.value.here` method.

```
# Enabling the error-sending plugins (info plugins analogy)
standardization.plugin.postprocessor.1=za.co.absa.enceladus.plugins.builtin.errorsender.mq.kafka.KafkaErrorSenderPlugin
conformance.plugin.postprocessor.1=za.co.absa.enceladus.plugins.builtin.errorsender.mq.kafka.KafkaErrorSenderPlugin

# Have your clientId and your error topic name defined (info plugins analogy, too)
kafka.error.client.id="my-error-client-id1"
kafka.error.topic.name="example.error.publishing.topic"

# Kafka URL settings
kafka.schema.registry.url:"https://my1017.schema.registry.example.com:8081"
kafka.bootstrap.servers="my1013.bootstrap.server.example.com:9094,my1014.bootstrap.server.example.com:9094,my1015.bootstrap.server.example.com:9094"

# security settings for SSL
kafka.security.protocol="SASL_SSL"
kafka.sasl.mechanism="GSSAPI"

# SSL configuration specifics
javax.net.ssl.trustStore=/path/to/my-truststore.jks
javax.net.ssl.trustStorePassword=trustStoreExamplePassword1
javax.net.ssl.keyStore=/path/to/my-keystore.jks
javax.net.ssl.keyStorePassword=keyStoreExamplePassword1
java.security.auth.login.config="/path/to/jaas.config"
# JAAS config should contain your `KafkaClient { ... }` settings - Kerberos-based or other
```

Small changes in the error msgs producing to apply the security settings for the _structured streaming_ approach as well. 

## Tested on in-company secured Kafka cluster/topics
error-publish-ssl-test

### Release notes suggestion
Error messages publishing now works with SSL-secured Kafka topics. Refer to the [`spark-jobs`' `reference.conf`](https://github.com/AbsaOSS/enceladus/blob/develop/spark-jobs/src/main/resources/reference.conf) on how to set it up.

Closes #1352

### DIFF
--- a/plugins-builtin/src/main/scala/za/co/absa/enceladus/plugins/builtin/errorsender/mq/KafkaErrorSenderPluginImpl.scala
+++ b/plugins-builtin/src/main/scala/za/co/absa/enceladus/plugins/builtin/errorsender/mq/KafkaErrorSenderPluginImpl.scala
@@ -18,9 +18,9 @@ package za.co.absa.enceladus.plugins.builtin.errorsender.mq
 import org.apache.log4j.LogManager
 import org.apache.spark.sql.functions.{col, explode, lit, size, struct, typedLit}
 import org.apache.spark.sql.types.DataTypes
-import org.apache.spark.sql.{Column, DataFrame, Encoder, Encoders}
+import org.apache.spark.sql.{Column, DataFrame, DataFrameWriter, Encoder, Encoders}
 import za.co.absa.enceladus.plugins.api.postprocessor.PostProcessor
-import za.co.absa.enceladus.plugins.builtin.common.mq.kafka.KafkaConnectionParams
+import za.co.absa.enceladus.plugins.builtin.common.mq.kafka.{KafkaConnectionParams, KafkaSecurityParams}
 import za.co.absa.enceladus.plugins.builtin.errorsender.DceError
 import za.co.absa.enceladus.plugins.builtin.errorsender.mq.KafkaErrorSenderPluginImpl.SingleErrorStardardized
 import za.co.absa.enceladus.utils.schema.SchemaUtils
@@ -145,6 +145,7 @@ case class KafkaErrorSenderPluginImpl(connectionParams: KafkaConnectionParams,
       .option("kafka.bootstrap.servers", connectionParams.bootstrapServers)
       .option("topic", connectionParams.topicName)
       .option("kafka.client.id", connectionParams.clientId)
+      .withOptionalKafkaSecurityParams(connectionParams.security)
       .option("path", "notReallyUsedButAtumExpectsItToBePresent") // TODO Atum issue #32
       .save()
   }
@@ -193,6 +194,15 @@ object KafkaErrorSenderPluginImpl {
   def errorCodesForSource(sourceId: SourcePhase): Seq[String] = sourceId match {
     case SourcePhase.Standardization => ErrorCodes.standardizationErrorCodes
     case SourcePhase.Conformance => ErrorCodes.conformanceErrorCodes
+  }
+
+  implicit class DataFrameWriterOptionExt[T](dataFrameWriter: DataFrameWriter[T]) {
+    def withOptionalKafkaSecurityParams(optSecParams: Option[KafkaSecurityParams]): DataFrameWriter[T] = {
+      optSecParams match {
+        case None => dataFrameWriter
+        case Some(secParams) => dataFrameWriter.options(secParams.toMap)
+      }
+    }
   }
 
 }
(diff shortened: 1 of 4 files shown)


plugins-builtin/src/main/scala/za/co/absa/enceladus/plugins/builtin/errorsender/mq/KafkaErrorSenderPluginImpl.scala
```diff
@@ -92,13 +92,19 @@ case class KafkaErrorSenderPluginImpl(connectionParams: KafkaConnectionParams,
 
     val allowedErrorCodes = KafkaErrorSenderPluginImpl.errorCodesForSource(params.sourceId)
 
+    val reportDateCol: Column = if (dataFrame.columns.contains(ColumnNames.reportDate)) {
```
Zejnilovic, line 95: So is there an option for no ReportDate? When does that happen? Does the information still have the same informational value?

Zejnilovic (commented):
I am not gonna pretend I understand all of it. But I like it.

plugins-builtin/src/main/scala/za/co/absa/enceladus/plugins/builtin/errorsender/mq/KafkaErrorSenderPluginImpl.scala
dk1844: Well, there are datasets where `ReportDate` is not present - e.g. Cobol ones, but in general, this is possible - the presence of this field is not forced in Enceladus in any way. Here, I only fixed the obvious error in that case.

Unless someone deliberately disables/mangles the `enceladus_record_id` generation, that alone is sufficient to uniquely identify the error origin. `ReportDate` is just a helpful field.

Zejnilovic: The report Date that we add is, I think, Data Quality requirement. But this is not a place for that discussion. 👍

Zejnilovic (approved):
Code Reviewed

spark-jobs/src/main/resources/reference.conf
```diff
@@ -89,12 +89,20 @@ timezone="UTC"
...
 
-#kafka.error.client.id="DqErrorSender"
+#kafka.error.client.id="errorInfo"
 #kafka.error.topic.name="dq.errors"
 
 # Optional security settings
 #kafka.security.protocol="SASL_SSL"
 #kafka.sasl.mechanism="GSSAPI"
 
+# SSL configuration specifics
```
AdrianOlosutean, line 99: @dvagapov was thinking to use this when trying to establish a Menas secured Mongo connection. Would it be posible to reuse some of these?

plugins-builtin/src/main/scala/za/co/absa/enceladus/plugins/builtin/errorsender/mq/KafkaErrorSenderPluginImpl.scala
```diff
@@ -189,4 +196,13 @@ object KafkaErrorSenderPluginImpl {
     case SourcePhase.Conformance => ErrorCodes.conformanceErrorCodes
   }
 
+  implicit class DataFrameWriterOptionExt[T](dataFrameWriter: DataFrameWriter[T]) {
```
AdrianOlosutean, line 199: Nice, I like having implicit classses

AdrianOlosutean (approved):
Code reviewed

spark-jobs/src/main/resources/reference.conf
dvagapov: I already played with TLS for Kafka. This is a config for Kafka-client.
Mongo client doesn't have that kind of files https://docs.mongodb.com/manual/tutorial/configure-ssl-clients/ only args.

benedeki (approved):

Issue #1707, opened 2021-03-02
Make Enceladus compilation Warning free and keep it that way


benedeki:
## Background
A clean code should have no compilation warnings. It's a sign of potential and/or future issues. Enceladus is mature enough to not have any.

## Feature
Get rid of the deprecation warning in `ActiveDirectoryLdapAuthoritiesPopulator`, the last compilation warning. Set the project to fail on future copmilation  warnings.

benedeki:
Release notes:
Project Enceladus is now built without any warnings. (No deprecated code calls, no advanced language features). Project is set to fail  on compilation in case of any such warnings.

PR #1708, opened 2021-03-02
#1707: Make Enceladus compilation Warning free and keep it that way

benedeki:
* Deprecated call replaced in `ActiveDirectoryLdapAuthoritiesPopulator`
* Whole project set to fail on compilation warning
* `scala-maven-plugin` upgraded to version 4.4.0

Release notes:
Project Enceladus is now built without any warnings. (No deprecated code calls, no advanced language features). Project is set to fail  on compilation in case of any such warnings.

### DIFF
--- a/dao/pom.xml
+++ b/dao/pom.xml
@@ -79,6 +79,13 @@
                         </goals>
                     </execution>
                 </executions>
+                <configuration>
+                    <args>
+                        <arg>-Xfatal-warnings</arg>
+                        <arg>-unchecked</arg>
+                        <arg>-deprecation</arg>
+                    </args>
+                </configuration>
             </plugin>
         </plugins>
     </build>
--- a/data-model/pom.xml
+++ b/data-model/pom.xml
@@ -78,6 +78,13 @@
                         </goals>
                     </execution>
                 </executions>
+                <configuration>
+                    <args>
+                        <arg>-Xfatal-warnings</arg>
+                        <arg>-unchecked</arg>
+                        <arg>-deprecation</arg>
+                    </args>
+                </configuration>
             </plugin>
         </plugins>
     </build>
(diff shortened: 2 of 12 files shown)


dao/pom.xml
```diff
@@ -79,6 +79,13 @@
                         </goals>
                     </execution>
                 </executions>
+                <configuration>
+                    <args>
+                        <arg>-Xfatal-warnings</arg>
+                        <arg>-unchecked</arg>
+                        <arg>-deprecation</arg>
```
Zejnilovic, line 86: Can we get a quick link or explanation?

benedeki: https://docs.scala-lang.org/overviews/compiler-options/index.html

`-Xfatal-warnings` makes the compilation fail on warnings
`-unchecked` - adds location and describes in detail warnings where generated code depends on assumptions
`-deprecation` - adds location and describes in detail the deprecation warning

Actually should have included `-feature` more than `-unchecked` as it is the one we had experienced together with `-deprecation`. I will add it.
For the record, including the switches does not change if the warning is generated, but a difference between:
_There were 2 deprecation warnings_
vs.
_There's a deprecation warning in file XYZ using class OPQ on line 123 and 301._

Zejnilovic (approved):
Code Reviewed

benedeki:
Btw, for testing I recommend reverting the changed class to previous version - and the build will fail.

AdrianOlosutean (approved):
Code reviewed. Had some maven central connectivity problems, so I could not test locally yet

dk1844 (approved):
It looks very useful, I am just wondering if this could be defined on parent and shared or use a comparable approach to avoid defining the plugin and its configuration on each submodule.

PR #1714, opened 2021-03-05
license copyright update

dk1844:
Copyright fix, to ~`master`~ `develop`.


LICENSE.md
```diff
@@ -187,7 +187,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright 2017 Barclays Africa Group Limited
+   Copyright 2017 ABSA Group Limited
```
Zejnilovic, line 190: It should be `Copyright [yyyy] [name of copyright owner]`

benedeki:
I got these files by running:
```
grep -rIL \
 --exclude-dir=target --exclude-dir=.idea \
 --include \*.scala --include \*.java \
 --include \*.html --include \*.js --include \*.css --include \*.xml --include \*.js \
 --include \*.conf --include \*.properties --include \*.template \
 --include \*.sh --include \*.bat --include \*.cmd \
 --include \*.md --include \*.yaml --include \*.yml \
 --include \*.editorconfig --include \*.gitattributes --include \*.gitignore \
 "Licensed under the Apache License, Version 2\.0"
```
Couple of questions arose...

benedeki:
What about files:
* `spark-jobs/src/main/resources/META-INF/services/za.co.absa.hyperdrive.ingestor.api.transformer.StreamTransformerFactoryProvider` Not even sure what this file is about. Maybe @kevinwallimann would help?
* `scripts/cmd/tee.bat`
* `.github/ISSUE_TEMPLATE/*`

benedeki:
Do we put the license into the `.md` files?
And if so, do we put it there as it is now - as kind of a header or as a comment?
In some `.md` files it had been already there as comment. So I went with that style in my commit. On the other hand the comment style is in my opinion much more suitable for github.io/ghpages branch

benedeki:
Do we want to use the `apache-rat-plugin`? If so, then in every build, in `integration` profile or some dedicated one?

kevinwallimann:
> What about files:
> 
> * `spark-jobs/src/main/resources/META-INF/services/za.co.absa.hyperdrive.ingestor.api.transformer.StreamTransformerFactoryProvider` Not even sure what this file is about. Maybe @kevinwallimann would help?
> * `scripts/cmd/tee.bat`
> * `.github/ISSUE_TEMPLATE/*`

`spark-jobs/src/main/resources/META-INF/services/za.co.absa.hyperdrive.ingestor.api.transformer.StreamTransformerFactoryProvider` is a provider configuration file. This is part of the Java SPI (Service Provider Interface) mechanism to load implementations of an interface. Any line starting with `#` is a comment. See e.g. here: https://github.com/AbsaOSS/hyperdrive/blob/develop/ingestor-default/src/main/resources/META-INF/services/za.co.absa.hyperdrive.ingestor.api.transformer.StreamTransformerFactoryProvider

LICENSE.md
dk1844: Oh, verbatim, not replacing the values?! Sorry, I did not get that.

dk1844:
@benedeki, thank you for the commit.

> Why is this marked as hotfix?
because it merges to `master` (circumventing `develop`)

> Do we put the license into the `.md` files?
> And if so, do we put it there as it is now - as kind of a header or as a comment?
> In some `.md` files it had been already there as comment. So I went with that style in my commit. On the other hand the comment style is in my opinion much more suitable for github.io/ghpages branch
I think what you did makes sense.

> What about files:
> 
> * `spark-jobs/src/main/resources/META-INF/services/za.co.absa.hyperdrive.ingestor.api.transformer.StreamTransformerFactoryProvider` Not even sure what this file is about. Maybe @kevinwallimann would help?
> * `scripts/cmd/tee.bat`
> * `.github/ISSUE_TEMPLATE/*`
I can see the header in the first, added it to the second. For the third, the github templates, I would not do it, because it would interfere with their function.

scripts/cmd/tee.bat
```diff
@@ -1,5 +1,18 @@
 @ECHO OFF
 
+:: Copyright 2018 ABSA Group Limited
```
benedeki, line 3: I didn't want to do that right away, before discussing with you guys.
This script is from Internet, not written by us, nor for Absa...

dk1844: Oh, I did not know that... so what is the original license there? can we even use it?

benedeki: The contribution is at the bottom. I kept the file as it was. Doesn't have any license mentioned on the site, the other doesn't even exist anymore. I think it's a pretty old script.

benedeki:
> Do we want to use the `apache-rat-plugin`? If so, then in every build, in `integration` profile or some dedicated one?
> It seems that we already have it in Enceladus, added by @yruslan 🙇 ! it has its own profile `license-check`, which makes sense to me - it can be a separate CI check

I didn't see it 🤦 . Sorry.

Zejnilovic (approved):
+1 for linting the md file.

AdrianOlosutean (dismissed):

benedeki (approved):
Good in my opinion. But I did a part of it, so not sure, my approval is "good enough".

Zejnilovic (approved):

Zejnilovic:
Would it be "needlessly increasing size of a PR" to add the GitHub action to run the rat plugin as in Atum?

benedeki:
> 
> 
> Would it be "needlessly increasing size of a PR" to add the GitHub action to run the rat plugin as in Atum?

I don't think so. Let's do it.

Issue #1706, opened 2021-03-01
Spark 3.0.x


benedeki:
## Background
We need to move on higher version of Spark

## Goal
Making Enceladus working with Spark 3.0

## Expected Task List [Optional]
A list of expected issues that will be needed to achieve this Epic
1. Figure out dependencies are ready
2. Cross-compilation for Scala 2.11 and 2.12
3. How it will work on prem? Spark 3 with Hadoop 2

**NB! Don't forget to assign the Epic into a Milestone**

PR #1731, opened 2021-03-15
Feature/1706 spark3

dk1844:
note on branching: we will have a `develop-spark3` branch, but will wait with its creation until develop contains `develop-ver-3.0`

Spark 3/Scala 2.12 branch
- buildable, but there are some unsolved issues, that require attention -- see the epic https://github.com/AbsaOSS/enceladus/issues/1706

The current state could become a basis for the rest of the development. (spline agent, hyperdrive, DateTime patterns, udf usage, ...)

Some details:
 - significant version updates: Scala 2.12, Spark 3.0.2, Hadoop 3.2.1, Guaba 30.1-jar
 - Atum 3.4.0 used, with `eventually` `_INFO` file existence checking
    - Jenkinsfile build defines patience config scaling 6x
 - is mergeable from `develop` for now, so license checking and falling warnings are included
 - some of the "deliberately incorrect" data parsing has been updated to the newer spark behavior (seems better in all touched cases).

Has not been run so far, probably at least the UDF should get fixed first?


utils/src/test/scala/org/apache/hadoop/io/nativeio/NativeIO.java
```diff
@@ -0,0 +1,989 @@
...
+     *
+     * @param path input path
+     * @param desiredAccess ACCESS_READ, ACCESS_WRITE or ACCESS_EXECUTE
+     * @return true if access is allowed
+     * @throws IOException I/O exception on error
+     */
+    public static boolean access(String path, AccessRight desiredAccess)
+      throws IOException {
+      //return access0(path, desiredAccess.accessRight());
+      return true; // workaround to make NativeIo work on Windows 10 https://www.programmersought.com/article/33926846559/
```
dk1844, line 647: This, I am really ashamed of doing, but I don't know how to otherwise make it work for Windows in general without making assumptions on the host system (like forcing users to install Spark3+Hadoop 3.2, setting up paths, and `winutils`. That would be ok to force for running Enceladus, but just for building it with tests it seems too harsh.

The "good" (in sense it could be worse) things are that this file only exists in `test` (`main` exec code is unaffected) and only once (in `utils`) and it gets reused (in `spark-jobs` by relying on the `util`'s test-jar).

utils/src/test/scala/org/apache/hadoop/io/nativeio/NativeIO.java
```diff
@@ -0,0 +1,989 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.hadoop.io.nativeio;
+
+// the file is copied from hadoop3.2.1, with just method org.apache.hadoop.io.nativeio.NativeIO.Windows.access changed
+// locally to make the tests run on Windows 10
+
```
AdrianOlosutean, line 20: So is this file optional for other operating systems?

utils/src/main/scala/za/co/absa/enceladus/utils/udf/UDFLibrary.scala
```diff
@@ -27,60 +27,76 @@ import scala.collection.mutable
...
+  UDFLibrary.registerUdf() // re-registration would be unnecessary
+}
 
-  spark.udf.register(stdCastErr, { (errCol: String, rawValue: String) =>
-    ErrorMessage.stdCastErr(errCol, rawValue)
-  })
+object UDFLibrary {
 
-  spark.udf.register(stdNullErr, { errCol: String => ErrorMessage.stdNullErr(errCol) })
+  private var isRegistered = false
```
AdrianOlosutean, line 35: I know it's a small use case, but could there be a way to avoid using this var?

AdrianOlosutean (approved):
Code reviewed

utils/src/test/scala/org/apache/hadoop/io/nativeio/NativeIO.java
dk1844: Yes

utils/src/main/scala/za/co/absa/enceladus/utils/udf/UDFLibrary.scala
dk1844: I thought about leveraging `lazy` or some AtomicReference, but their logic is similar. It seems that for Spark3 the UDF registration works differently and thus, Spark3 is complaining about reregistering these UDFs. But there are still some leftovers. I will add a point to address it in the UDF task, it is related anyway.

AdrianOlosutean (approved):

PR #1739, opened 2021-03-22
1732 Spline 0.6 integration

### LINKED ISSUE #1732: Integrated Spline 0.6
## Background
Spline 0.6 is nearing release. Let's upgrade to this version and release with it.

## Feature
Upgrade dependencies in #1175 to 0.6.

- [ ] SparkJobs - include Spline Spark Agent, possible configurations update
- [ ] Menas - UI integration

AdrianOlosutean:
Closes #1732


benedeki:
Just curious why you included the Swagger in this PR? 😉

AdrianOlosutean:
> Just curious why you included the Swagger in this PR? 😉

I considered it's quite useful to add it after separating Menas. If you think it's better, I will put the changes in a separate PR

menas/src/main/resources/application.properties.template
```diff
@@ -130,8 +126,11 @@ menas.oozie.mavenSparkJobsJarLocation=/za/co/absa/enceladus/spark-jobs/@project.
 #Menas URL for submitted std and conf jobs
 menas.oozie.menasApiURL=http://menasHostname:8080/menas/api
 
-#Mongo address for spline for the submitted jobs
-menas.oozie.splineMongoURL=mongodb://localhost:27017
+#The URL address of Spline for the submitted jobs
```
benedeki, line 129: I would get rid of it _Spline_ reference from Oozie code completely. Let's take if from `application.conf`.

menas/src/main/scala/za/co/absa/enceladus/menas/repositories/OozieRepository.scala
```diff
@@ -96,8 +95,11 @@ class OozieRepository @Autowired() (oozieClientRes: Either[OozieConfigurationExc
   @Value("${menas.oozie.menasApiURL:}")
   val menasApiURL: String = ""
 
-  @Value("${menas.oozie.splineMongoURL:}")
-  val splineMongoURL: String = ""
+  @Value("${menas.oozie.lineageWriteApiUrl:}")
```
benedeki, line 98: As mentioned above, I would get rid of Oozie code to anyhow know about _Spline_ or lineage.

scripts/bash/enceladus_env.template.sh
```diff
@@ -16,10 +16,9 @@
 # Environment configuration
 STD_HDFS_PATH="/bigdata/std/std-{0}-{1}-{2}-{3}"
 
-# MongoDB connection configuration for Spline
+# Configuration for Spline
 # Important! Special characters should be escaped using triple backslashes (\\\)
-SPLINE_MONGODB_URL="mongodb://localhost:27017"
-SPLINE_MONGODB_NAME="spline"
+SPLINE_PRODUCER_URL="http://localhost:8080/spline/producer"
```
benedeki, line 21: If done as suggested above, should not be needed.

scripts/bash/run_enceladus.sh
```diff
@@ -447,7 +447,7 @@ else
 fi
 
 JVM_CONF="spark.driver.extraJavaOptions=-Dstandardized.hdfs.path=$STD_HDFS_PATH \
--Dspline.mongodb.url=$SPLINE_MONGODB_URL -Dspline.mongodb.name=$SPLINE_MONGODB_NAME -Dhdp.version=$HDP_VERSION \
+-Dspline.producer.url=$SPLINE_PRODUCER_URL -Dhdp.version=$HDP_VERSION \
```
benedeki, line 450: Part of "no Spline configuration on command line" clearing.

benedeki (commented):
> 
> 
> > Just curious why you included the Swagger in this PR? 😉
> 
> I considered it's quite useful to add it after separating Menas. If you think it's better, I will put the changes in a separate PR

It is definitely useful, no doubt. But I see no crossover with the Spline integration. So having it in one PR - particularly when there already was a PR (maybe not on separated Menas) - seems contra productive, just increases the size of the PR.
Furthermore Spline is dragging behind a little. Swagger could be already merged.... Perhpas it would be worth to separate into its own PR. 🤔

pom.xml
```diff
@@ -141,7 +141,7 @@
         <maven.rat.plugin.version>0.12</maven.rat.plugin.version>
         <maven.scala.version>4.4.0</maven.scala.version>
         <maven.scoverage.version>1.3.0</maven.scoverage.version>
-        <maven.shade.version>3.2.1</maven.shade.version>
+        <maven.shade.version>3.1.1</maven.shade.version>
```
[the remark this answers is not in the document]
AdrianOlosutean: No particular reason, probably a merge leftover

menas-web/ui/service/RestDAO.js
```diff
@@ -232,4 +232,5 @@ class ConfigRestClient {
   static getEnvironmentName() {
     return RestClient.get(`/configuration/environment`)
   }
+
```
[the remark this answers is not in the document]
AdrianOlosutean: Nope

menas/src/main/scala/za/co/absa/enceladus/menas/HDFSConfig.scala
```diff
@@ -28,7 +28,7 @@ import org.springframework.context.annotation.{Bean, Configuration}
 class HDFSConfig @Autowired() (spark: SparkSession) {
   private val logger = LoggerFactory.getLogger(this.getClass)
 
-  @Value("${menas.hadoop.auth.method}")
+  @Value("${menas.hadoop.auth.method:}")
```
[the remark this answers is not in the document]
AdrianOlosutean: It used to fail before because of this, but we solved it another way. Anyway, it seems to be the only one without a colon here

menas/src/main/resources/application.properties.template
AdrianOlosutean: I also thought of this, but wouldn't it break Oozie integration?

menas/src/main/scala/za/co/absa/enceladus/menas/HDFSConfig.scala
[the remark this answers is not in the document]
dk1844: Let's not forget what the colon means - in general, it changes the meaning of the property `${"my.property.key"}` (without default value - so failing to find it in the config should cause a failure) to `${"my.property.key:defaultValueForThatKey"}` (where failing to find the key in config just fallbacks to the default - and the _nothing_ after the colon means default empty string, but having default provided nonetheless)

Most probably, you all knew all this, just stating it so we just "don't add" colons because it is present elsewhere, but we consider the true meaning - and here that it should be _whether it is reasonable to provide default empty string when the key is not found in properties_.

dk1844 (approved):
LGTM (read the code and test-ran on EKS/Fargate with @AdrianOlosutean )

dk1844 (approved):
LGTM (read the code and test-ran on EKS/Fargate with @AdrianOlosutean )

PR #1744, opened 2021-04-09
1710 Spark 3 UDF Fix

### LINKED ISSUE #1710: Spark3 equivalent of UserDefinedFunction usage
Current (Spark 2) Enceladus directly uses `UserDefinedFunction` case class API. This structure has been converted into a sealed trait in Spark 3, so different means of UDF callings must be used instead.

Also, Spark 3 behaves seems to behave differently in the case of named UDF's registration. There has been a lot of re-registration warnings, some of those were "solved" in `UDFLibrary`, some still remain unsolved. A comprehensive approach to the problem could be part of this PR. Related: https://issues.apache.org/jira/browse/SPARK-34804

AdrianOlosutean:
Closes #1710


dk1844 (commented):
It is great that you got it working for Scala 3, @AdrianOlosutean!

It would probably be nicer if we could use the Scala UDF API, instead of the Java one, but better than nothing.

I believe a part of the issue is also the problem of multiple named UDF re-registration, which I believe has not been addressed in this PR.

utils/src/main/scala/za/co/absa/enceladus/utils/broadcast/BroadcastUtils.scala
```diff
@@ -102,338 +97,120 @@ object BroadcastUtils {
...
-     param6: Any, param7: Any, param8: Any, param9: Any, param10: Any) => {
-      val mt = mappingTable.value.map
-      if (mt.contains(Seq(param1, param2, param3, param4, param5, param6, param7, param8, param9, param10))) {
-        null
-      } else {
-        ErrorMessage.confMappingErr(outputColumn, Seq(safeToString(param1), safeToString(param2), safeToString(param3),
-          safeToString(param4), safeToString(param5), safeToString(param6), safeToString(param7), safeToString(param8),
-          safeToString(param9), safeToString(param10)), mappings)
-      }
+    val errorMessageType = ScalaReflection.schemaFor[ErrorMessage].dataType
```
AdrianOlosutean, line 177: Here  I use Java UDFs, but there is an alternative. Using this alternative, however, doesn't allow reusing a function like `applyError` and only works if it's inline, otherwise will fail with something like "Cannot have return type any". 

The alternative does not need a return type parameter and looks like this:
`case 1 => udf((p1: Any) => if (mappingTable.value.contains(Seq(p1))) {
        null
      } else {
        val strings: Seq[String] = Seq(p1).map(a => safeToString(a))
        ErrorMessage.confMappingErr(outputColumns.mkString(","), strings, mappings)
      })`

@dk1844 , which one would you consider preferable?

dk1844: It seems to me that _doesn't allow reusing a function like `applyError` and only works if it's inline_ is too much of a limitation, so I would probably keep it as is.

dk1844 (dismissed):
LGTM, also checked out and built ✔️

utils/src/main/scala/za/co/absa/enceladus/utils/broadcast/BroadcastUtils.scala
```diff
@@ -125,70 +104,68 @@ object BroadcastUtils {
...
-      case 10 => (p1: Any, p2: Any, p3: Any, p4: Any, p5: Any, p6: Any, p7: Any, p8: Any, p9: Any, p10: Any) =>
-        mappingTable.value.getRowWithDefault(Seq(p1, p2, p3, p4, p5, p6, p7, p8, p9, p10), defaultValue)
-      case n => throw new IllegalArgumentException(s"Mapping UDFs with $n arguments are not supported. Should be between 1 and 10.")
-    }
-  }
-
-  private def getMultipleMappingLambda(mappingTable: Broadcast[LocalMappingTable],
-                                       defaultValues: Map[String, Any], numberOfArguments: Int): AnyRef = {
-    val defaultRow = if (defaultValues.isEmpty) {
+  private def getMappingLambda(mappingTable: Broadcast[LocalMappingTable],
```
benedeki, line 110: I would refactor this considerably:
```
private def getLambda(numberOfArguments: Int, fnc: Seq[Any] => Any): UserDefinedFunction = {
...
}

private def getLambdaFnc(, defaultValue: Any)
```

In `getMappingUdfForSingleOutput`:
```
   ...
    val defaultValue = defaultValueExprMap.get(mappingTable.value.outputColumns.values.head).orNull
    val fnc: Seq[Any] => Any  = (seq: Seq[Any] ) => mappingTable.value.getRowWithDefault(seq, defaultValue)
    getLambda(numberOfArguments, fnc)   
```
or alternatively:
```
   ...
    val defaultValue = defaultValueExprMap.get(mappingTable.value.outputColumns.values.head).orNull
    getLambda(numberOfArguments, mappingTable.value.getRowWithDefault(_, defaultValue))   
```

Similarly in `getMappingUdfForMultipleOutputs` and `getErrorUdf`

utils/src/main/scala/za/co/absa/enceladus/utils/broadcast/BroadcastUtils.scala
```diff
@@ -125,70 +104,68 @@ object BroadcastUtils {
...
-        mappingTable.value.getRowWithDefault(Seq(p1, p2, p3, p4, p5, p6), defaultRow)
-      case 7 => (p1: Any, p2: Any, p3: Any, p4: Any, p5: Any, p6: Any, p7: Any) =>
-        mappingTable.value.getRowWithDefault(Seq(p1, p2, p3, p4, p5, p6, p7), defaultRow)
-      case 8 => (p1: Any, p2: Any, p3: Any, p4: Any, p5: Any, p6: Any, p7: Any, p8: Any) =>
-        mappingTable.value.getRowWithDefault(Seq(p1, p2, p3, p4, p5, p6, p7, p8), defaultRow)
-      case 9 => (p1: Any, p2: Any, p3: Any, p4: Any, p5: Any, p6: Any, p7: Any, p8: Any, p9: Any) =>
-        mappingTable.value.getRowWithDefault(Seq(p1, p2, p3, p4, p5, p6, p7, p8, p9), defaultRow)
-      case 10 => (p1: Any, p2: Any, p3: Any, p4: Any, p5: Any, p6: Any, p7: Any, p8: Any, p9: Any, p10: Any) =>
-        mappingTable.value.getRowWithDefault(Seq(p1, p2, p3, p4, p5, p6, p7, p8, p9, p10), defaultRow)
+      case 1 => udf(new UDF1[Any, Any] {
```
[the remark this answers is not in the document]
AdrianOlosutean: User Defined Aggregate Functions are not relevant here. 
`udf((p1: Any) => mappingTable.value.getRowWithDefault(Seq(p1), defaultValue))` will fail with `Schema for type Any is not supported`, since Spark 3.x changed the behavior, so the only option left is to use Java UDFs. I can explain more in detail if you want why this is the only option left

utils/src/main/scala/za/co/absa/enceladus/utils/broadcast/BroadcastUtils.scala
AdrianOlosutean: Good idea, I will also need to add the valueType parameter

utils/src/main/scala/za/co/absa/enceladus/utils/broadcast/BroadcastUtils.scala
```diff
@@ -64,30 +64,46 @@ object BroadcastUtils {
...
+    val defaultValues = defaultValueExpressions.mapValues(getValueOfSparkExpression)
+    val defaultRow = if (defaultValues.isEmpty) {
+      null
+    } else {
+      Row.fromSeq(mappingTable.value.outputColumns.values.toSeq.map(field => defaultValues.getOrElse(field, null)))
+    }
 
-    val lambda = getMappingLambda(mappingTable, defaultValueMap, numberOfArguments, isSingleOutput = false)
+    val fnc: Seq[Any] => Any = (seq: Seq[Any] ) => mappingTable.value.getRowWithDefault(seq, defaultRow)
+    val structFields: Seq[StructField] = mappingTable.value.outputColumns.keys
```
benedeki, line 101: This is one I wanted to suggest for refactoring during the original multicolumne result back then around Easter (but didn't happen for reasons 😉 ).
Anyway, I think it should be a property of `LocalMappingTable`, the returned `DataType` of the lookup. IMHO it belongs there conceptually, and might make the overall code tighter - construct and define the `returnedType: DataType` (name is just a suggestion) during `LocalMappingTable` creation.

AdrianOlosutean: But how to do with the ErrorMessage value type?

benedeki: Ah, of course. After all `LocalMappingTable` is not its (`getLambda`) parameter anymore anyway.

utils/src/main/scala/za/co/absa/enceladus/utils/broadcast/LocalMappingTable.scala
```diff
@@ -83,7 +86,8 @@ object LocalMappingTable {
       (keys, if (values.length == 1) values.head else new GenericRowWithSchema(values, rowSchema))
     }).toMap
 
-    LocalMappingTable(mappingTable, outputColumns, keyTypes, valueTypes)
+    val valueType: DataType = if (outputColumns.size == 1) valueTypes.head else rowSchema
+    LocalMappingTable(mappingTable, outputColumns, keyTypes, valueType)
```
benedeki, line 90: I would restructure this piece of code - line 76-89. 
I don't like that the forking between one-value and multiple-value mapping table is done on more places. Also the code is not that obvious IMO. And the _select_ on the dataFrame might be worth to avoid. 
Suggesting this code instead (put the subfunction and the top of the method):
```
    def oneValue(seq: Seq[Any]): Any = seq.head
    def moreValues(seq: Seq[Any], schema: StructType): Any = new GenericRowWithSchema(seq.toArray, schema)
    val (valueType, valueFnc) = if (outputColumns.size == 1) {
      (valueTypes.head, oneValue _)
    } else {
      (rowSchema, moreValues(_, rowSchema))
    }
    val mappingTable = mappingTableDf.collect().map(row => {
      val keys = keyFields.map(name => row.getAs[Any](name))
      val values = targetAttributes.map(name => row.getAs[Any](name))
      (keys, valueFnc(values))
    }).toMap
```

spark-jobs/src/test/scala/za/co/absa/enceladus/conformance/interpreter/fixtures/MultipleMappingFixture.scala
```diff
@@ -90,7 +90,8 @@ trait MultipleMappingFixture extends BeforeAndAfterAll with SparkTestBase {
     val value = "../examples/data/e2e_tests/data/hdfs/std/mappingConformanceRuleData/2020/03/23/v1/_INFO"
     val infoFileContents: String = fromFile(value, "UTF-8").mkString
 
-    FileUtils.writeStringToFile(new File(s"${getRawDataPath(mappingDS.hdfsPath)}/_INFO"), infoFileContents, Charset.defaultCharset())
+    FileUtils.writeStringToFile(new File(s"${getRawDataPath(mappingDS.hdfsPath)}/_INFO"), infoFileContents,
```
Zejnilovic, line 93: 1. Why append true?
2. `Charset.defaultCharset` is being used in `NestedStructsFixture.scala` and `TradeConformance.scala`. I would like us to use the same across the board.

utils/src/main/scala/za/co/absa/enceladus/utils/broadcast/LocalMappingTable.scala
AdrianOlosutean: Using `getAs[Any](name)` looks good, but does not really work for accessing nested fields, unless a kind of recursive function is used, which I don't believe is suitable here.

spark-jobs/src/test/scala/za/co/absa/enceladus/conformance/interpreter/fixtures/MultipleMappingFixture.scala
AdrianOlosutean: I remember it didn't use to work without append true with the newer versions

benedeki: If the boolean param is needed, add naming to it please.
Also if touching it, there's a hint for `fromFile` to not be closed. Usually `FileReader` provides a safe way.

utils/src/main/scala/za/co/absa/enceladus/utils/broadcast/LocalMappingTable.scala
benedeki: Darn, you are right. 😞 
Well it could be done with:
```
    def getRowValue(row: Row, fieldNameSegments: Seq[String]): Any = {
      fieldNameSegments match {
        case Nil => null
        case head :: Nil => row.getAs[Any](head)
        case head :: tail => getRowValue(row.getAs[Row](head), tail)
      }
    }
```
And have a function in `def SchemaUtils.fieldNameToSegments(fieldName: String): Seq[String]`

**OR**

Keep it as it is, despite not nice. `select` is not an _action_ but _transformation_ I checked, so from that point of view it's OK. Then just one point: 
`val targetColumns: Seq[Column]` is declared (line 76), but `keyFields.map(c => col(c))` is used directly (line 78). Let's do it one way or another, the mix is confusing.

spark-jobs/src/test/scala/za/co/absa/enceladus/conformance/interpreter/fixtures/MultipleMappingFixture.scala
```diff
@@ -88,9 +88,12 @@ trait MultipleMappingFixture extends BeforeAndAfterAll with SparkTestBase {
       .mode(SaveMode.Overwrite)
       .parquet(getRawDataPath(mappingDS.hdfsPath))
     val value = "../examples/data/e2e_tests/data/hdfs/std/mappingConformanceRuleData/2020/03/23/v1/_INFO"
-    val infoFileContents: String = fromFile(value, "UTF-8").mkString
+    val fileContent = fromFile(value, "UTF-8")
+    val infoFileContents: String = fileContent.mkString
+    fileContent.close()
```
benedeki, line 93: Why not using the `FileReader` object?

spark-jobs/src/test/scala/za/co/absa/enceladus/conformance/interpreter/fixtures/MultipleMappingFixture.scala
```diff
@@ -88,9 +88,12 @@ trait MultipleMappingFixture extends BeforeAndAfterAll with SparkTestBase {
       .mode(SaveMode.Overwrite)
       .parquet(getRawDataPath(mappingDS.hdfsPath))
     val value = "../examples/data/e2e_tests/data/hdfs/std/mappingConformanceRuleData/2020/03/23/v1/_INFO"
-    val infoFileContents: String = fromFile(value, "UTF-8").mkString
+    val fileContent = fromFile(value, "UTF-8")
```
benedeki, line 91: Just a few line below (originally one), it has been suggested to not use the `"UTF-8"` literal string but instead `Charset.defaultCharset`. But please do use the `FileReader` and if the encoding is needed, enhance the function there.

spark-jobs/src/test/scala/za/co/absa/enceladus/conformance/interpreter/fixtures/MultipleMappingFixture.scala
benedeki: I tried, `append = true` is not needed. Actually thinking about it, I have a suspicion it would cause problems on test reruns.
And **do name boolean parameters** when sending literals there.

utils/src/main/scala/za/co/absa/enceladus/utils/broadcast/LocalMappingTable.scala
```diff
@@ -66,24 +66,35 @@ object LocalMappingTable {
...
+    val structFields: Seq[StructField] = outputColumns.keys
+      .map(SchemaUtils.stripPathFromFieldName).toSeq
+      .zip(valueTypes)
       .map { case (name: String, fieldType: DataType) => StructField(name, fieldType) }
     val rowSchema = StructType(structFields)
 
     val targetColumns: Seq[Column] = targetAttributes.map(targetAttribute => col(targetAttribute))
-
-    val mappingColumns = targetColumns ++ keyFields.map(c => col(c))
+    val keyColumns = keyFields.map(c => col(c))
```
benedeki, line 77: At the risk of being seen as nitpicky 🤓. Future reader will wonder why `targetColumns` needs type declaration while `keyColumns` not. I believe it's good to keep consistency, at least in so adjacent code. (one or another, don't care which 😉 )

spark-jobs/src/test/scala/za/co/absa/enceladus/conformance/interpreter/fixtures/MultipleMappingFixture.scala
AdrianOlosutean: It seems like the recommended way of reading files in Scala is `Source.fromFile`

spark-jobs/src/test/scala/za/co/absa/enceladus/conformance/interpreter/fixtures/MultipleMappingFixture.scala
AdrianOlosutean: This file is not currently used anywhere, because its use is in `HyperConformanceMappingIntegrationSuite`. I would propose to move this discussion to #1745

benedeki (approved):

PR #1749, opened 2021-04-13
Stricter enforcement of DRA in helper scripts

Zejnilovic:


scripts/bash/enceladus_env.template.sh
```diff
@@ -34,12 +34,14 @@ STD_DEFAULT_DRIVER_MEMORY=""
 STD_DEFAULT_DRIVER_CORES=""
 STD_DEFAULT_EXECUTOR_MEMORY=""
 STD_DEFAULT_EXECUTOR_CORES=""
+STD_DEFAULT_DRA_EXECUTOR_MEMORY=""
+STD_DEFAULT_DRA_EXECUTOR_CORES=""
```
lokm01, line 38: executor cores default should always be 1

scripts/bash/enceladus_env.template.sh
```diff
@@ -53,13 +55,14 @@ CONF_DEFAULT_DRIVER_MEMORY=""
 CONF_DEFAULT_DRIVER_CORES=""
 CONF_DEFAULT_EXECUTOR_MEMORY=""
 CONF_DEFAULT_EXECUTOR_CORES=""
-CONF_DEFAULT_EXECUTOR_CORES=""
+CONF_DEFAULT_DRA_EXECUTOR_MEMORY=""
+CONF_DEFAULT_DRA_EXECUTOR_CORES=""
```
lokm01, line 59: Same as above

scripts/bash/enceladus_env.template.sh
```diff
@@ -34,12 +34,14 @@ STD_DEFAULT_DRIVER_MEMORY=""
 STD_DEFAULT_DRIVER_CORES=""
 STD_DEFAULT_EXECUTOR_MEMORY=""
 STD_DEFAULT_EXECUTOR_CORES=""
+STD_DEFAULT_DRA_EXECUTOR_MEMORY=""
```
lokm01, line 37: How about this is set based on the executor-memory (if passed in)?

lokm01 (commented):
I'd tweak the defaults if I understand this correctly, otherwise lgtm

scripts/bash/enceladus_env.template.sh
Zejnilovic: This can be done on the Enviroment Level

scripts/bash/enceladus_env.template.sh
Zejnilovic: This can be done on the Environment Level

scripts/bash/enceladus_env.template.sh
Zejnilovic: So I intentionally have them separated the normal executor memory and the DRA executor memory, since when you sent executor memory during the DRA it behaves like the "initial number of executor" and will never scale below this number. Even if there is no reason to have X executors, it will keep them.

scripts/bash/run_enceladus.sh
```diff
@@ -386,33 +401,37 @@ SPARK_CONF="--conf spark.logConf=true"
...
+  if [ -n "$NUM_EXECUTORS" ]; then
+      echo "WARNING: num-executors should NOT be set when using Dynamic Resource Allocation and will be ignored."
+      echo "Set dra-num-executors if you know what you are doing or disable dra"
+  fi
+
+  echo "Using values from --dra-executor-memory and --dra-executor-cores. --executor-memory and --executor-cores ignored"
+
+  add_spark_conf_cmd "spark.dynamicAllocation.enabled" "true"
+  add_spark_conf_cmd "spark.shuffle.service.enabled" "true"
+  add_spark_conf_cmd "spark.sql.adaptive.enabled" "true"
```
dk1844, line 422: I am wondering if all there should be enabled together under one switch. The `shuffle.service` probably should (but are we always certain that the service is set up - as the documentation states?).

The `sql.adaptive` can helpful, but does it really relate with `dynamicAllocation` so that is switched on/off together?

dk1844 (approved):
LGTM (just read the code)

scripts/bash/run_enceladus.sh
Zejnilovic: So I did not change this. This is how it was the first set up and is currently working.
Maybe @DzMakatun can tell you more about it

scripts/bash/enceladus_env.template.sh
benedeki: I think 1 is a sensible default for the template.

scripts/bash/run_enceladus.sh
```diff
@@ -386,33 +401,37 @@ SPARK_CONF="--conf spark.logConf=true"
...
-    if [ -n "$DRA_ALLOCATION_RATIO" ]; then
-        add_spark_conf_cmd "spark.dynamicAllocation.executorAllocationRatio" "${DRA_ALLOCATION_RATIO}"
-    fi
-    if [ -n "$ADAPTIVE_TARGET_POSTSHUFFLE_INPUT_SIZE" ]; then
-        add_spark_conf_cmd "spark.sql.adaptive.shuffle.targetPostShuffleInputSize" "${ADAPTIVE_TARGET_POSTSHUFFLE_INPUT_SIZE}"
-    fi
+  echo "Dynamic Resource Allocation enabled"
+
+  if [ -n "$NUM_EXECUTORS" ]; then
+      echo "WARNING: num-executors should NOT be set when using Dynamic Resource Allocation and will be ignored."
```
benedeki, line 414: I think nobody will notice this warning. Don't we want to stop the script maybe?

scripts/bash/run_enceladus.sh
```diff
@@ -386,33 +401,37 @@ SPARK_CONF="--conf spark.logConf=true"
...
+      echo "Set dra-num-executors if you know what you are doing or disable dra"
+  fi
+
+  echo "Using values from --dra-executor-memory and --dra-executor-cores. --executor-memory and --executor-cores ignored"
```
benedeki, line 418: I would write this only if any of the later two are defined.

scripts/bash/run_enceladus.sh
benedeki: Good enough for now IMHO. It's a herper script after all. If one wants to run it with all the tweaks, they can do that via spark-submit.

scripts/bash/run_enceladus.sh
```diff
@@ -100,6 +107,14 @@ case $key in
...
     shift 2 # past argument and value
     ;;
+  --dra-executor-cores)
+    DRA_EXECUTOR_CORES="$2"
+    shift 2 # past argument and value
+    ;;
+  --dra-executor-memory)
+    DRA_EXECUTOR_MEMORY="$2"
+    shift 2 # past argument and value
+    ;;
```
benedeki, line 117: **Major:** There is no way how to disable DRA other then changing the value in `enceladus_env.sh`. Didn't we want to have a switch `--disable-dra`?

scripts/bash/enceladus_env.template.sh
```diff
@@ -34,12 +34,14 @@ STD_DEFAULT_DRIVER_MEMORY=""
...
+STD_DEFAULT_DRA_EXECUTOR_CORES=""
 # setting num executors disables DRA
 STD_DEFAULT_NUM_EXECUTORS=""
 
 # Dynamic Resource Allocation
 # also enables external shuffle service and adaptive execution for consistent setup
-STD_DEFAULT_DRA_ENABLED=false
+STD_DEFAULT_DRA_ENABLED=true
```
benedeki, line 44: **To consider:** If we will have a switch `--disable-dra` and no `--enable-dra` that this doesn't have much sense IMO.
On the other hand, I think it would be good to have both switches and then this default makes sense.

(Obviously same applies to `CONF_DEFAULT_DRA_ENABLED`)

benedeki (commented):
Considering this is a possible serious paradagim shift (DRA on by default instead of off), don't we want to keep this for Enceladus 3?
Obviously we would deploy it on prem asap.

scripts/bash/run_enceladus.sh
Zejnilovic: They can have Default values from the env script, thus being defined always.

scripts/bash/run_enceladus.sh
Zejnilovic: You can use `--enable-dra false`

benedeki: Was not aware of this switch already being there. 🙈

scripts/bash/enceladus_env.template.sh
benedeki: Having switch `--set-dra` makes this comment moot.

scripts/bash/run_enceladus.sh
benedeki: Which is OK. But if not defined just noise in the output.

scripts/bash/run_enceladus.sh
benedeki: Discussed and clarified.

scripts/bash/enceladus_env.template.sh
Zejnilovic: I am not arguing about the number. But I do think that this is not the place to put it. Either we come up with numbers for most of these in the template file or we leave them be, as we did so far and it is set on the servers.
I cannot see why to have a default for `dra_executor_cores` here but not for the 10 of the others.

scripts/bash/run_enceladus.sh
Zejnilovic: add an if

dk1844 (approved):

benedeki (approved):
- [x] code reviewed
- [ ] pulled
- [ ] built
- [ ] run

scripts/bash/enceladus_env.template.sh
DzMakatun: --num-executors is the 'initial number of executor' in DRA. executor-memory is pretty much the same w and w/o DRA =memory per executor.

DzMakatun: Also, if we override some of the params for DRA, I would recommend to override all of them together. Otherwise, if someone wants '4 executors x 8 cores + 24 GB' and we override only part of it, it may result with something like 'DRA with 1 core and 24 GB per executor'.

scripts/bash/run_enceladus.sh
DzMakatun: **Having adaptive with DRA is a must.** Here is why: if adaptive is disabled (and the code does not partition data explicitly), any shuffle will generate 200 tasks by default, regardless the data size, even for 100 KB of data. As a result DRA will ask for (200 * executorAllocationRatio) executors and  max out to maxExecutors.

AdrianOlosutean (approved):
Code reviewed

Zejnilovic:
Put the label there myslef as we ran on production today and it worked.

PR #1752, opened 2021-04-15
Feature/merging develop ver.30

dk1844:
This is essentially merging `develop` into `develop-ver3.0`, but since it was non-trivial, this PR was created.


benedeki:
Non-trivial 😞 
Then I would wait the develop to be updated to the expected (hopefully) version - e.g. merge of #1740

menas-web/Dockerfile
```diff
@@ -11,7 +11,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM nginx:1-alpine
+ARG DOCKER_BASE_ARTIFACT=nginx
+ARG DOCKER_BASE_TAG=1-alpine
+
+FROM ${DOCKER_BASE_ARTIFACT}:${DOCKER_BASE_TAG}
```
Zejnilovic, line 17: I know this is a new feature that people are excited about, but I think it was mainly created for multi staged builds.

What does it give us here? What are the pros of this approach? I honestly see only possible problems here + the extra (undeclared) properties in Maven.

menas/src/main/resources/docker.properties.template
```diff
@@ -34,11 +34,18 @@ spring.resources.chain.strategy.fixed.version=@project.version@
...
+menas.auth.jwt.lifespan.hours=8
+
 # Timeout (in seconds) for MongoDB migration queries
 migrations.mongo.query.timeout.seconds=300
 
 menas.spark.master=local[1]
 
+# The template for Lineage (Spline) display. If empty Lineage button is not shown.
+# Avoiding the http(s) protocol will cause that the same protocol will be used as for Menas itself, avoiding issues
+spline.urlTemplate=//localhost:8080/spline/dataset/lineage/_search?path=%s&application_id=%s
```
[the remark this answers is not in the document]
AdrianOlosutean: I think this is not even needed anymore, since it will be removed from spline integration

dk1844:
#1740 merged into `develop`, updated this PR/branch by merging `develop` into it.

dao/src/main/scala/META-INF/MANIFEST.MF
```diff
@@ -1,3 +1,3 @@
-Manifest-Version: 1.0
-Class-Path: 
-
+Manifest-Version: 1.0
```
AdrianOlosutean, line 1: Let's remove this file

AdrianOlosutean (dismissed):

dao/src/main/scala/META-INF/MANIFEST.MF
dk1844: This is just a CRLF->LF switch, a possible removal can be done separately.

menas-web/Dockerfile
Zejnilovic: I would put a comment in there above the ARGVs as to why it is like that.

menas/src/main/resources/docker.properties.template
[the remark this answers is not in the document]
dk1844: @dvagapov removed it.

Zejnilovic (approved):
I only went over Docker stuff

AdrianOlosutean (approved):

PR #1779, opened 2021-05-03
Feature/1722 binary tests

dk1844:
_Test improvements:_

There were many tests whose results were defined in a show-like manner (effectively reading the result of `dataframe.show()`). This works, but is disadvantageous for a couple of reasons:
 - nuances like correct datatypes (strings vs same-valued toString representation) are hard to cover (this also applies to differentiating "null" (string) vs null value serialized in the same fashion)
 - for future use, there are problems with forward-compatibility, because for spark 3.0 vs spark 3.1, there are some changes in the show-representations (`[content, of, struct-object], [content, of array]` -> `{content, of, struct-object}, [content, of array]`).

Replacing the show-based expected data with in-code defined aims to alleviate both of the situations outlined above.

_Use hermes to compare dataframes instead of mrpower's dataset compared when id generation is fixed? ->_  separate issue: https://github.com/AbsaOSS/enceladus/issues/1782

### Test-merge
I have also test merged this PR with the currently in progress spark 3.1 #1755 and it is confirmed to be working - effectively substituting a good portion of the 3.1-related changes. (However, there are be some manual merging efforts).


Zejnilovic:
> TODO: use hermes to compare dataframes instead of mrpower's dataset compared?

From what I read, I would like to note that mrpower's dataset compared uses `collect` and then `sameElements` methods to compare the data and thus need to have the data always read in the same order. Same for the larger comparison, but just different methods are used.

Also, I do not see him pinpointing the difference. Rather printing everything out.

I have not compared the speeds of each of them. There, his approach might be faster.

Don' forget I am biased...

dk1844:
> > TODO: use hermes to compare dataframes instead of mrpower's dataset compared?
> 
> From what I read, I would like to note that mrpower's dataset compared uses `collect` and then `sameElements` methods to compare the data and thus need to have the data always read in the same order. Same for the larger comparison, but just different methods are used.

That's entirely possible.

> 
> Also, I do not see him pinpointing the difference. Rather printing everything out.

The difference is shown in color (gray - same, red - different data)

> 
> I have not compared the speeds of each of them. There, his approach might be faster.

For the size of the test DF's, it should not matter much.

> Don' forget I am biased...

That's entirely possible.

spark-jobs/src/test/scala/za/co/absa/enceladus/standardization/StandardizationParquetSuite.scala
```diff
@@ -363,41 +327,42 @@ class StandardizationParquetSuite extends FixtureAnyFunSuite with SparkTestBase
...
     val schema = StructType(seq)
     // stableHashId will always yield the same ids
-    val destDF = StandardizationInterpreter.standardize(sourceDF, schema, cmd.rawFormat, recordIdGenerationStrategy = IdType.StableHashId)
+    val actualDf = StandardizationInterpreter.standardize(sourceDF, schema, cmd.rawFormat, recordIdGenerationStrategy = IdType.StableHashId)
+
+    val expectedData = Seq(
+      Row(1L, Array("A", "B"), Row(false), Array(), 1950798873),
+      Row(2L, Array("C"), Row(true), Array(), -988631025)
+    )
+    val expectedDF = expectedData.toDfWithSchema(actualDf.schema) // checking the just data, not the schema here
```
benedeki, line 336:
```suggestion
    val expectedDF = expectedData.toDfWithSchema(actualDf.schema) // checking just the data, not the schema here
```
I think it's copy pasted on multiple places.

utils/src/main/scala/za/co/absa/enceladus/utils/testUtils/DataFrameTestUtils.scala
```diff
@@ -0,0 +1,28 @@
...
+
+package za.co.absa.enceladus.utils.testUtils
+
+import org.apache.spark.sql.types.StructType
+import org.apache.spark.sql.{DataFrame, Row, SparkSession}
+
+object DataFrameTestUtils {
+
+  implicit class RowSeqToDf(val data: Seq[Row]) extends AnyVal {
+    def toDfWithSchema(schema: StructType)(implicit spark: SparkSession): DataFrame = {
```
benedeki, line 24: Nice 👍

dk1844: Thanks.

spark-jobs/src/test/scala/za/co/absa/enceladus/standardization/StandardizationParquetSuite.scala
dk1844: Yes it is. Thanks for finding this, I was already blind to it. Fixed.

dk1844:
I have assigned the label _no testing needed_ as this PR only concerns tests that get run on Jenkins anyway.

Zejnilovic (approved):

AdrianOlosutean (approved):

PR #1781, opened 2021-05-05
#1769 Rename Menas to rest-api

### LINKED ISSUE #1769: Rename Menas to rest-api

AdrianOlosutean:
Brace yourselves, the first part of renaming
Closes #1769


AdrianOlosutean:
I'm not sure how exactly to handle the separation + renaming in README. Any proposals?

rest-api/src/main/resources/banner.txt
```diff
@@ -0,0 +1,7 @@
+  ____  _____ ____ _____       _    ____ ___
+ |  _ \| ____/ ___|_   _|     / \  |  _ \_ _|
+ | |_) |  _| \___ \ | |_____ / _ \ | |_) | |
+ |  _ <| |___ ___) || |_____/ ___ \|  __/| |
+ |_| \_\_____|____/ |_|    /_/   \_\_|  |___|
+
+version ${application.version}
```
dk1844, line 7: Very cool 😎

rest-api/Dockerfile
```diff
@@ -11,14 +11,14 @@
...
 
-# To build, first build menas and be in enceladus/menas folder. Then run:
-# $> docker build -t menas .
+# To build, first build rest-api and be in enceladus/rest-api folder. Then run:
+# $> docker build -t rest-api .
 # Run image using:
 # $> docker run \
 #       -e MENAS_MONGO_CONNECTION_STRING=mongodb://host.docker.internal:27017 \
-#       -e MENAS_MONGO_CONNECTION_DATABASE=menas \
+#       -e MENAS_MONGO_CONNECTION_DATABASE=rest_api \
```
dk1844, line 19: Are you sure about this? Since the `application.properties` has not changed, I assume that the database name is kept (`menas`), so the db name should stay here.

Or do we plan to rename the database, too? (I probably wouldn't unless it was necessary, because then there might need to be a migration script etc.)

dk1844 (dismissed):
I see nothing suspicious -> LGTM (just read the code)

rest-api/Dockerfile
AdrianOlosutean: I forgot this one. I will revert this change

rest-api/src/main/resources/banner.txt
Zejnilovic: While I agree, I would do this
```suggestion
  ____  _____ ____ _____       _    ____ ___
 |  _ \| ____/ ___|_   _|     / \  |  _ \_ _|
 | |_) |  _| \___ \ | |_____ / _ \ | |_) | |
 |  _ <| |___ ___) || |_____/ ___ \|  __/| |
 |_| \_\_____|____/ |_|    /_/   \_\_|  |___|

Enceladus version ${application.version}
```

dk1844:
> I'm not sure how exactly to handle the separation + renaming in README. Any proposals?

What do you have in mind specifically?
I would do about the following:
 - in README, 
   - in the  `### Menas` section,  I'd update that it is only UI front-end and a general-purpose service API exists behind it.
   - in the `## How to run` / `#### Menas requirements` section, I'd state that it's `menas-web.war` now and reword that some requirements are tied to the service REST API instead of to Menas directly.
   - (not aware of any other adjustments needed there, but anything is possible)
 
Then, consider what else must change in the GHP description, what actually public API endpoints to describe (if the time is right?) and publish (in a sense of _making them public_) them that way. But GHP changes are another story (i.e. PR) anyway.

rest-api/src/test/scala/za/co/absa/enceladus/rest_api/package.scala
```diff
@@ -15,7 +15,7 @@
 
 package za.co.absa.enceladus
 
-package object menas {
+package object restapi {
```
AdrianOlosutean, line 18: I could not name this package object in tests to rest_api. Hopefully it's not too confusing, given that the module is rest-api and the package is rest_api

dk1844: I consider this acceptable.

README.md
```diff
@@ -132,7 +144,7 @@ password=changeme
 --row-tag <tag>
 ```
 * Here `row-tag` is a specific option for `raw-format` of type `XML`. For more options for different types please see our WIKI.
-* In case Menas is configured for in-memory authentication (e.g. in dev environments), replace `--menas-auth-keytab` with `--menas-credentials-file`
+* In case REST API is configured for in-memory authentication (e.g. in dev environments), replace `--menas-auth-keytab` with `--menas-credentials-file`
```
dk1844, line 147: I wonder if `--menas-auth-keytab` or  `--menas-credentials-file` shouldn't also relate to the _rest-api_ rather than _menas_.

PR #1783, opened 2021-05-12
#417 Spark-XML hack removed

### LINKED ISSUE #417: Remove extra code when spark-xml handles empty arrays
Remove extra code in `StandardizationInterpreter` when spark-xml handles empty arrays

dk1844:
1.  first a unit test has been added to assert the current functionality 
2.  SparkXML-hack removed
3.  the test holds

###RN suggestion
_XML input standardization: SparkXML-hack for arrays removed, no change in function should be noticed._

Closes #417


AdrianOlosutean (approved):
Looks good.

dk1844:
After discussion with @benedeki and @Zejnilovic, I have assigned label _no test needed_ as the added test case and its way of originating should suffice as evidence of correct behavior, guarding from regression.

PR #1788, opened 2021-05-18
#1787: MenasDAO.getDataset backwards compatibility fix

### LINKED ISSUE #1787: MenasDAO.getDataset backwards compatibility fix
## Describe the bug
`MenasDAO#getDataset` method has parameter of `ValidationLevel` which is mandatory and has no default. This is not backwards compatible.

Plus, even if people set `noValidation`, the `noValidation` param is introduced in the URL. In this case, there should be no param in the URL.

## Expected behaviour
`MenasDAO#getDataset` should have a default `ValidationLevel` set to `noValidation` and if `noValidation` is chosen, do not add it to the URL.

benedeki:
* defined `DefaultValidationLevel`
* default value for `validationLevel` in `getDataset`
* if value for validationLevel is equal to DefaultValidationLevel, validation level is not added to query string

Release notes:
MenasDAO.getDataset made backward compatible.

Closes #1787

### DIFF
--- a/utils/src/main/scala/za/co/absa/enceladus/utils/validation/ValidationLevel.scala
+++ b/utils/src/main/scala/za/co/absa/enceladus/utils/validation/ValidationLevel.scala
@@ -21,7 +21,7 @@ object ValidationLevel extends Enumeration {
   final val NoValidation, ForRun, Strictest = Value
 
   object Constants { // Not to mix with enumeration items
-    final val DefaultValidationLevel:ValidationLevel  = NoValidation
+    final val DefaultValidationLevel: ValidationLevel = NoValidation
 
     final val DefaultValidationLevelName = "NoValidation" // This is used in an annotation - for that the type MUST NOT be specified
   }
--- a/utils/src/main/scala/za/co/absa/enceladus/utils/validation/ValidationLevel.scala
+++ b/utils/src/main/scala/za/co/absa/enceladus/utils/validation/ValidationLevel.scala
@@ -20,5 +20,10 @@ object ValidationLevel extends Enumeration {
 
   final val NoValidation, ForRun, Strictest = Value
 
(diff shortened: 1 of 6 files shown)


Zejnilovic (approved):
Code reviewed

AdrianOlosutean (approved):
Looks good(code review)

utils/src/main/scala/za/co/absa/enceladus/utils/validation/ValidationLevel.scala
```diff
@@ -20,5 +20,10 @@ object ValidationLevel extends Enumeration {
 
   final val NoValidation, ForRun, Strictest = Value
 
-  final val NoValidationName = "NoValidation" // This is used in an annotation - for that the type MUST NOT be specified
+  object Constants { // Not to mix with enumeration items
+    final val DefaultValidationLevel:ValidationLevel  = NoValidation
```
dk1844, line 24:
```suggestion
    final val DefaultValidationLevel: ValidationLevel = NoValidation
```

Also, I keep wondering if the existence of `DefaultValidationLevel` is necessary, but sure, why not.

dk1844 (approved):
LGTM.

utils/src/main/scala/za/co/absa/enceladus/utils/validation/ValidationLevel.scala
benedeki: I think it does. It prevents accidental removal/change without realizing the consequences - at least I believe it does.

Zejnilovic (approved):
Ran a spark-jobs standardization and custom project

AdrianOlosutean (approved):

dk1844 (approved):
LGTM (just read the code)

PR #1789, opened 2021-05-19
Suggested value not auto apply

### LINKED ISSUE #1784: Properties suggested value is applied automatically and it's not optional
## Describe the bug
Part of property definition is a suggested value. For enumeration it actually have to be an item from the list.
When editing a dataset and the property has not be defined the suggested value is used and accepted. And the suggested value is not optional.

Also empty string is accepted as a valid property value, which is strange.

## To Reproduce
1. Have some properties defined, with a non-trivial suggested value, set as _Mandatory_
2. In _Menas_ create new dataset. Save the dataset without changing the suggested value
3. It's accepted
4. This might not be desirable, at least not in all cases

## Expected behavior
Suggested values are not applied automatically, at least not always.

## Suggested solutions
* Make suggested value optional
  * Empty string should serve well
  * Maybe conversion to `Option[String]` might not cause problems either
* Make empty string to be seen as undefined value for property
* **_Bonus:_** use suggested value in Menas UI controls as real suggestions, not a value directly

dk1844:
This PR addresses problems outlined in #1784, namely fixes the following:

## Backend
 - `suggestedValue` for propertyTypes has been made optional - simply put, some properties may not have a defined suggested value
... //todo

Closes #1784


menas/ui/generic/functions.js
```diff
@@ -53,5 +53,15 @@ var Functions = new function () {
...
     })
+  };
+
+  /**
+   * Checks if the property is enum-like by making sure that non-empty inerable of `allowedValues` exists
+   * @param propertyType
+   * @returns {boolean}
+   */
+  this.isEnum = function(propertyType) {
+    return propertyType.hasOwnProperty("allowedValues") && propertyType.allowedValues.length > 0;
```
benedeki, line 64: What if it's an enum without allowed values (`allowedValues.length == 0`)?

menas/ui/service/EntityDialog.js
```diff
@@ -97,7 +97,32 @@ class DatasetDialog extends EntityDialog {
...
+      let allowedMap = oProp.propertyType.allowedValues.map(val => {
+        if (val == oProp.propertyType.suggestedValue) {
+          return {value: val, text: `${val} (suggested value)`}
+        } else {
+          return {value: val, text: val}
+        }
+      });
 
+      if (oProp.essentiality._t !== "Mandatory") {
+        allowedMap = [{value: "", text: ""}, ...allowedMap] // (ES6 prepending) - ability to undefine the property
```
benedeki, line 117: ES6?

menas/ui/service/EntityDialog.js
```diff
@@ -97,7 +97,32 @@ class DatasetDialog extends EntityDialog {
 
  }
 
+  /**
+   * Will created oProps's allowedValues mapped into displayable sequence of objects, e.g.
```
benedeki, line 101: Seems as a incorrect grammar in the comment _Will created..._

menas/ui/service/EntityService.js
```diff
@@ -325,7 +325,22 @@ class DatasetService extends EntityService {
     }
   }
 
+  // if _properties with empty string exist, they will be replaced with undefs in place (backend expecting an empty option)
+  replaceBlankProperties(dataset) {
```
benedeki, line 329: I would do something similar on backend side as well. After all, the API can be called by anyone, any "enemy", not just this "friendly" UI.

menas/ui/service/EntityService.js
```diff
@@ -325,7 +325,22 @@ class DatasetService extends EntityService {
     }
   }
 
+  // if _properties with empty string exist, they will be replaced with undefs in place (backend expecting an empty option)
```
benedeki, line 328:
```suggestion
  // if properties with empty string exist, they will be replaced with undefs in place (backend expecting an empty option)
```

benedeki:
RN typo:
Dataset property values are **now** suggested...

menas/ui/service/EntityDialog.js
dk1844: EcmaScript 6

benedeki:
Haven't tested it yet, but based on the screenshots looks great. Excellent work 🥇

menas/ui/service/EntityService.js
dk1844: No. Both `properties` and `_properties` are present in the object. The former holds the initial information, while the latter hold the submitted data. So `_properties` is, indeed, the desired filed to be adjusted.

menas/ui/service/EntityDialog.js
dk1844: True, result of re-editing, will fix, thanks.

menas/ui/service/EntityService.js
dk1844: Could be. 
I wonder, however, if are general in the feature, are you absolutely certain we will not need empty string value, e.g. for enum as one value, so we can afford to do this?

menas/ui/service/EntityService.js
benedeki: Missed that, sorry.

menas/ui/generic/functions.js
benedeki: I like both suggestions 👍

menas/ui/service/EntityDialog.js
benedeki: Always learning new stuff. Thank you.

menas/ui/generic/functions.js
dk1844: Done. 
 - `isEnum` -> `hasValidAllowedValues` ✔️ 
 - enum property type(s) the allowedValues: does not allow "" to be present + must have 1+ items ✔️

menas/ui/service/EntityService.js
dk1844: Done, an empty string for a dataset's value is converted to an _unset value_. Applies for creating & editing a dataset and for updating dataset's properties (test added to cover this, too.)

dk1844:
About a bug shared with @benedeki using other channels - properties defs not being present in UI on the first load - I believe I have now fixed it by refreshing the properties model after login. 

The root case seemed to lie in the fact that the properties defs are loaded on SPA first load, but if the user is not logged in, they fail to load, on refresh, it worked due to the existing session (that expired after a while).

Zejnilovic (approved):
- [X] code reviewed
- [X] pulled
- [X] built
- [X] run (Menas and REST API tests)

data-model/src/main/scala/za/co/absa/enceladus/model/properties/propertyType/package.scala
```diff
@@ -37,7 +37,11 @@ package object propertyType {
     override def isValueConforming(value: String): Try[Unit] = Success(Unit)
   }
 
-  case class EnumPropertyType(allowedValues: Set[String], suggestedValue: Option[String]) extends PropertyType {
+  case class EnumPropertyType(allowedValues: Seq[String], suggestedValue: Option[String]) extends PropertyType {
+    require(allowedValues.nonEmpty, "At least one allowed value must be defined for EnumPropertyType.")
```
benedeki, line 41: Like these wordings. And was not aware/forgot about the `require` usage. 👍

data-model/src/main/scala/za/co/absa/enceladus/model/properties/propertyType/package.scala
```diff
@@ -37,7 +37,11 @@ package object propertyType {
     override def isValueConforming(value: String): Try[Unit] = Success(Unit)
   }
 
-  case class EnumPropertyType(allowedValues: Set[String], suggestedValue: Option[String]) extends PropertyType {
+  case class EnumPropertyType(allowedValues: Seq[String], suggestedValue: Option[String]) extends PropertyType {
```
benedeki, line 40: Could we use `LinkedHashSet`?

data-model/src/main/scala/za/co/absa/enceladus/model/properties/propertyType/package.scala
dk1844: Thanks

data-model/src/main/scala/za/co/absa/enceladus/model/properties/propertyType/package.scala
dk1844: Solved by require on `size=toSet.size`, thanks for the hint.

menas/ui/service/EntityService.js
```diff
@@ -325,7 +325,22 @@ class DatasetService extends EntityService {
...
+    const updatedProperties = dataset._properties.map((oProp) => {
+      if (oProp.value === "") {
+        const changedProp = jQuery.extend({}, oProp);
+        changedProp.value = undefined;
+        return changedProp;
+      } else {
+        return oProp; // unchanged
+      }
+    });
+    dataset._properties = updatedProperties; // changed in place
```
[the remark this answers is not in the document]
dk1844: I'd naturally prefer fn-way, too, but I figured that for JS/ES the imperative way of working is sort of fitting :)

OK, redone, looks more readable.

benedeki (approved):
- [x] code reviewed
- [x] pulled
- [X] built
- [X] run

PR #1796, opened 2021-05-27
Hotfix: suggestedValueNotApplied for 2.21.1

dk1844:
#1789 backport to 2.21.1

## Test-run
My own naive UI testing worked:
 - edited an existing dataset to contain properties, again editing them - unsetting some of the properties. Correctly show in UI and expected properties maps also existed in the Mongo DB
 - created a new DS with some non-mandatory properties not set, mongo DB collection contained the expected map.


benedeki (approved):
- [x] code reviewed
- [x] pulled
- [X] built
- [X] run

dao/src/main/scala/META-INF/MANIFEST.MF
```diff
@@ -1,3 +1,3 @@
-Manifest-Version: 1.0
-Class-Path: 
-
+Manifest-Version: 1.0
```
Zejnilovic, line 1: what are these?

data-model/src/main/scala/za/co/absa/enceladus/model/properties/propertyType/package.scala
```diff
@@ -28,16 +28,22 @@ package object propertyType {
...
     override def isValueConforming(value: String): Try[Unit] = Success(Unit)
   }
 
-  case class EnumPropertyType(allowedValues: Set[String], suggestedValue: String) extends PropertyType {
+  case class EnumPropertyType(allowedValues: Seq[String], suggestedValue: Option[String]) extends PropertyType {
+    require(allowedValues.nonEmpty, "At least one allowed value must be defined for EnumPropertyType.")
+    require(allowedValues.forall(_.nonEmpty), "Empty string is disallowed as an allowedValue for EnumPropertyType." +
+      "Consider using the property in non-mandatory context instead.")
+    require(allowedValues.size == allowedValues.toSet.size,
+      s"Allowed values for EnumPropertyType should be unique, but $allowedValues was given.")
```
Zejnilovic, line 45: Question not stopping from releasing and would go into a new issue: 

How about printing only the duplicates?
Quick code option:
```scala
allowedValues
  .foldLeft(Map.empty[String, Int].withDefaultValue(0))((acc, v) => acc.updated(v, acc(v) + 1))
  .filter((m) => m._2 > 1)
  .keys
```

dao/src/main/scala/META-INF/MANIFEST.MF
benedeki: They were removed in later PRs iirc

data-model/src/main/scala/za/co/absa/enceladus/model/properties/propertyType/package.scala
benedeki: 👍 Let's do it with merge from Develop to Develop3

dao/src/main/scala/META-INF/MANIFEST.MF
dk1844: Yeah, here those are just CRLF-LF changes

Zejnilovic: Then, does anybody know why it is ignored by `.editorconfig` which should "force" `lf` endings?

PR #1797, opened 2021-05-28
Hotfix: suggestedValueNotApplied for 2.20.1 #1796

dk1844:
#1789 backport to 2.20.1

## Test-run
My own naive UI testing worked:
 - edited an existing dataset to contain properties, again editing them - unsetting some of the properties. Correctly show in UI and expected properties maps also existed in the Mongo DB
 - created a new DS with some non-mandatory properties not set, mongo DB collection contained the expected map.


benedeki (approved):
- [x] code reviewed
- [x] pulled
- [X] built
- [X] run

PR #1799, opened 2021-06-02
#1790: Replace the discontinued Bintray as source of packages

### LINKED ISSUE #1790: Replace the discontinued Bintray as source of packages
## Background
Bintray.com has been discontinued as a service. It needs to be replaced.

## Feature
Bintray.com was used as a source for OpenUI library package. It needs to be replaced.

## Proposed Solution
Solution Ideas 
1. Consider using GitHub https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-apache-maven-registry

benedeki:
* OpenUI5 webjar is now downloaded from webjar.org instead of bintray.com
* cRonstrue is now included within the Menas module directly instead of as a webjar
* bintray.com as repository removed from the project

Closes #1790 

Release notes:
Removed dependencies on discontinued bintray.com repository.


menas/pom.xml
```diff
@@ -391,6 +386,16 @@
...
+            <plugin>
+                <groupId>org.apache.rat</groupId>
+                <artifactId>apache-rat-plugin</artifactId>
+                <version>${maven.rat.plugin.version}</version>
+                <configuration>
+                    <excludes>
+                        <exclude>ui/3rdParty/cRonstrue/cronstrue.min.js</exclude>
+                    </excludes>
+                </configuration>
+            </plugin>
```
dk1844, line 398: Why is it that it was not satisfactory to put the _exlude_ path in the list of _exlucde_ paths we have in the root `pom.xml` already?

(I saw a discussion on WA about this, but it was not clear)

dk1844 (approved):
LGTM (just read the code)

menas/pom.xml
benedeki: Because the check is done in the parent and then with each module (with path relative to the module root). And this new file belongs to_Menas_ module. So the configuration had to be done within the affected module.
(I guess the relative patch could have been in the parent `pom.xml` as well. But that would be: first confusing, second potentially exclude unwanted files in other modules.)

dk1844: Well, I am saying that because it seems that some of the exclusions we have in the root config are also sometimes actually subpackage specific, e.g. `**/node_modules/**`, `**/_site/**`, `**/ui/dist/**`. So perhaps we could rethink those, too.

benedeki: Oh, them having wildcards, I didn't realize. Are these all, or should I check more?

benedeki: Moved `**/node_modules/**`, other is not present.

dk1844: The ones I listed were the only sort of non-general cases. So IMHO only those were actually in question.

dk1844 (approved):
Still LGTM (just read the code)

benedeki:
> 
> 
> Still LGTM (just read the code)

Could you test it too please. Should be easy IMHO 😉

dk1844:
Tested:
 - builds for me
   - in the log it shows that OpenUI is download from a public repo:
  `Downloaded from central: https://repo.maven.apache.org/maven2/org/webjars/openui5/1.67.1/openui5-1.67.1.jar `
 - runs @ local tomcat 
   - `http://<menas root>/3rdParty/cRonstrue/cronstrue.min.js` is loaded and being called (breakpointed there)
   - navigated through, viewed a dataset, updated a dataset, created a new one.

AdrianOlosutean (approved):
Looks good

PR #1802, opened 2021-06-03
#1790: Replace the discontinued Bintray as source of packages - 2.20.1

benedeki:
* OpenUI5 webjar is now downloaded from webjar.org instead of bintray.com
* cRonstrue is now included within the Menas module directly instead of as a webjar
* bintray.com as repository removed from the project
* moved `**/node_modules/**` rat plugin exclude from parent to Menas module
* added missing licenses

Same as #1787, but into 2.20


dk1844 (approved):
LGTM (just read the code)

Zejnilovic (approved):
code checked

PR #1803, opened 2021-06-04
cronstrue not being loaded without existing session

dk1844:
Allowing the path for cRonstrue to be loaded without authorization seems to be needed to get correct Menas behavior without having to refresh the page after login.

### Test-ran
 - naively test ran on my machine: even on login page, the library is now loaded immediately:
 cronstrue-loaded-1st-from-3rdParty2
 
Also, having an existing custom dataset with defined schedule and having oozie enabled in config (so that the Scheduling tab is reachable), now the dataset data and its scheduling details are also correctly shown in the UI (this was not the case for the first run after login before this fix)

hints: of course, we should remerge/cherrypick this to other branches (`hotfix/2.{20,21}.1`, `develop`, ...)


benedeki (approved):

Zejnilovic (approved):

benedeki:
Merging to make the release

PR #1806, opened 2021-06-07
#1809 NativeIO removal for windows build

### LINKED ISSUE #1809: Replace custom NativeIO on with proper setup
## Background
Custom is currently part of `org/apache/hadoop/io/nativeio/NativeIO.java` `utils` module for tests and shared to `spark-jobs` as a test-jar. With proper platform setup, this should not be necessary.

## Proposed Solution
1. The proper solution should in providing `winutils.exe` and `hadoop.dll` in `PATH`. That way, the NativeIO-dependent tests should work correctly.

dk1844:
Custom `NativeIO` is no longer needed for windows build, this is solved by having necessary windows binaries (`winutils.exe` and `hadoop.dll`) on `PATH`.
 - no need to build and share the `utils` test-jar, too, anymore
 - README updated to provide explanatory info and a binaries source 

Closes #1809

## Test run
Test-ran on win10 box machine after removal


utils/pom.xml
```diff
@@ -137,18 +137,6 @@
                     </execution>
                 </executions>
             </plugin>
-            <plugin>
-                <groupId>org.apache.maven.plugins</groupId>
```
benedeki, line 141: Why removal of this?
I admit, I am not sure what it does. But it was not added with the file in question.

README.md
```diff
@@ -83,6 +90,8 @@ Ensure the properties there fit your environment.
 
 The coverage reports are written in each module's `target` directory and aggregated in the root `target` directory.
 
+
```
benedeki, line 93: On purpose or accidental?

utils/pom.xml
dk1844: The point was the following:
The custom `NativeIO` was needed in two places in tests - in `utils` and in `spark-jobs`. Instead of adding the same file twice in tests, you'd want to define a dependency from one module to the other. Except normally, the tests are not packaged, only `main ` is - so to reuse the class, you either move it from `test/` to `main/` if it makes sense or you define a test-jar to be built and depend on the test-jar. This was the latter option.

You can see that in `spark-jobs/pom.xml:125`, the util test-jar dependency is also removed.

README.md
dk1844: Accidental, thanks, fixed.

utils/pom.xml
benedeki: Thank you for you explanation.

benedeki (approved):
- [x] code reviewed
- [x] pulled
- [x] built
- [x] run

PR #1811, opened 2021-06-07
rat-plugin config for menas in a profile (same as root pom.xml)

dk1844:
For `menas`, the existing licence check ignore items (particular for `menas`) are moved to the `license-check` profile to align the with the `root.pom` (and actually not to prevent otherwise successful build)

 - the `<exclude>ui/dist/**</exclude>` entry is added for convenience when license-checking on a built codebase.

### DIFF
--- a/menas/pom.xml
+++ b/menas/pom.xml
@@ -386,17 +386,6 @@
                     </execution>
                 </executions>
             </plugin>
-            <plugin>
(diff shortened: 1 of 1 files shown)


dk1844:
I have assigned "no test needed" as it essentially gets tested in GH actions anyway.

benedeki (approved):

Issue #1816, opened 2021-06-14
UDF registration does not allow registration for mulitple SparkSessions


benedeki:
## Describe the bug
The way, how UDF registration is done in Enceladus prevents registering the function in multiple Spark sessions.
It is unlikely to have more, but hard to explain if so.

## To Reproduce
1. Create two (or more) Spark sessions
2. Try to register UD Functions in `UDFLibrary` for both sessions

## Expected behaviour
Possible to register UDF for both sessions

PR #1817, opened 2021-06-14
#1816: UDF registration does not allow registration for mulitple Spar…

benedeki:
…kSessions

* Registration is governed by the SparkSession's hash not a simple boolean

### DIFF
--- a/utils/src/main/scala/za/co/absa/enceladus/utils/udf/UDFLibrary.scala
+++ b/utils/src/main/scala/za/co/absa/enceladus/utils/udf/UDFLibrary.scala
@@ -34,65 +34,74 @@ object UDFLibrary {
 
   private var isRegistered = false
 
-  private def registerUdf()(implicit spark: SparkSession) {
+  private var registry: Set[Int] = Set.empty
+
+  private def shouldRegister(forSpark: SparkSession): Boolean = {
+    var result = false
     this.synchronized {
-      if (!isRegistered) {
-
-        spark.udf.register(stdCastErr, { (errCol: String, rawValue: String) =>
-          ErrorMessage.stdCastErr(errCol, rawValue)
-        })
-
-        spark.udf.register(stdNullErr, { errCol: String => ErrorMessage.stdNullErr(errCol) })
-
-        spark.udf.register(stdSchemaErr, { errRow: String => ErrorMessage.stdSchemaError(errRow) })
-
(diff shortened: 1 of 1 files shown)


utils/src/main/scala/za/co/absa/enceladus/utils/udf/UDFLibrary.scala
```diff
@@ -34,65 +34,74 @@ object UDFLibrary {
 
   private var isRegistered = false
 
-  private def registerUdf()(implicit spark: SparkSession) {
+  private var registry: Set[Int] = Set.empty
+
+  private def shouldRegister(forSpark: SparkSession): Boolean = {
+    var result = false
```
AdrianOlosutean, line 40: Could this variable be removed so we just use the registry variable?

benedeki: I don't know how. The only option would be to place the whole registration into the `synchronized` block. I prefer using a `var` to a long `synchronized` section.

AdrianOlosutean (approved):
Looks ok(no testing done)

benedeki:
Right now a general solution for "run a function once per object" doesn't seem possible.

utils/src/main/scala/za/co/absa/enceladus/utils/udf/UDFLibrary.scala
```diff
@@ -34,65 +34,74 @@ object UDFLibrary {
 
   private var isRegistered = false
 
-  private def registerUdf()(implicit spark: SparkSession) {
+  private var registry: Set[Int] = Set.empty
```
dk1844, line 37: Wouldn't it be easier to use a existing synchronized set/map. E.g. `java.util.concurrent.ConcurrentHashMap[K, Unit].putIfAbsent`?

dk1844 (approved):
LGTM (just read the code)

utils/src/main/scala/za/co/absa/enceladus/utils/udf/UDFLibrary.scala
benedeki: Because I didn't know the class. 👼

benedeki:
Changed it per @dk1844 suggestion. And noticed some warnings about UDF registrations under the same name. Did some refactoring to get rid of the warnings.
More refactoring would be still needed though - see #1886

PR #1822, opened 2021-06-20
1820 Fix auditable message for Dataset Properties

### LINKED ISSUE #1820: No changes detected in Dataset when property value is changed
## Describe the bug
When a property is defined or changed in a Dataset, the version is correctly increased. But the audit log show "no changes".

## To Reproduce
1. Have a property defined
2. Open an existing dataset definition
3. Add a value to the property
4. Save the changes
5. See the version increased
6. Go to Audit tab of the Dataset
7. See _No changes_

## Expected behaviour
The changes are properly displayed in the audit log.

## Screenshots
image

Zejnilovic:
**How it looks now:**

image

**Release notes**: 
Now Audit Trail in Menas shows added, removed and changed properties

Fixes #1820

### DIFF
--- a/data-model/src/main/scala/za/co/absa/enceladus/model/menas/audit/Auditable.scala
+++ b/data-model/src/main/scala/za/co/absa/enceladus/model/menas/audit/Auditable.scala
@@ -79,6 +79,74 @@ trait Auditable[T <: Product] { self: T =>
     }).getOrElse(Seq())
   }
 
+
+
+  private[model] def getOptionalMapFieldsAudit(newValue: T, fieldName: AuditFieldName)(implicit ct: ClassTag[T]): Seq[AuditTrailChange] = {
+    def tupleToSomeString(v: (String, _)): Option[String] = Some(s"""key: "${v._1}" value: "${v._2}"""")
+
+    val index = getFieldIndex(fieldName.declaredField)
+
+    index.map({i =>
+      val maybeNewMap = newValue.productElement(i).asInstanceOf[Option[Map[String, _]]]
+      val maybeOldMap = self.productElement(i).asInstanceOf[Option[Map[String, _]]]
+
+      (maybeNewMap, maybeOldMap) match {
+        case (Some(newMap), Some(oldMap)) =>
+          val added = newMap.toSet.diff(oldMap.toSet).toMap
+          val removed = oldMap.toSet.diff(newMap.toSet).toMap
+          val changedKey = added.keySet.intersect(removed.keySet)
+          val onlyAdded = added -- changedKey
+          val onlyRemoved = removed -- changedKey
+
+          val addedList = onlyAdded.map(t =>
+            AuditTrailChange(field = fieldName.declaredField,
+              oldValue = None,
+              newValue = tupleToSomeString(t),
+              message = s"${fieldName.humanReadableField} added."
+            )
+          )
+          val removedList = onlyRemoved.map(t =>
+            AuditTrailChange(
+              field = fieldName.declaredField,
+              oldValue = tupleToSomeString(t),
+              newValue = None,
+              message = s"${fieldName.humanReadableField} removed."
+            )
+          )
+          val changedList = changedKey.map(k =>
+            AuditTrailChange(
+              field = fieldName.declaredField,
+              oldValue = tupleToSomeString(k, removed(k)),
+              newValue = tupleToSomeString(k, added(k)),
+              message = s"${fieldName.humanReadableField} changed."
+            )
+          )
+
+          removedList ++ addedList ++ changedList
+        case (None, Some(oldMap)) =>
+          oldMap.map((v: (String, Any)) =>
+            AuditTrailChange(
+              field = fieldName.declaredField,
+              oldValue = tupleToSomeString(v),
+              newValue = None,
+              message = s"${fieldName.humanReadableField} removed."
+            )
+          )
+        case (Some(newMap), None) =>
+          newMap.map((v: (String, Any)) =>
+            AuditTrailChange(
+              field = fieldName.declaredField,
+              oldValue = None,
+              newValue = tupleToSomeString(v),
+              message = s"${fieldName.humanReadableField} added."
+            )
+          )
+        case (_, _) => Set.empty
+      }
+
+    }).getOrElse(Set.empty).toSeq
+  }
+
   /**
    * Utility function to unwrap defined option values (if Some) or keep the original value otherwise.
    */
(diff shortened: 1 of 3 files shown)


data-model/src/main/scala/za/co/absa/enceladus/model/menas/audit/Auditable.scala
```diff
@@ -79,6 +79,74 @@ trait Auditable[T <: Product] { self: T =>
     }).getOrElse(Seq())
   }
 
+
+
+  private[model] def getOptionalMapFieldsAudit(newValue: T, fieldName: AuditFieldName)(implicit ct: ClassTag[T]): Seq[AuditTrailChange] = {
+    def tupleToSomeString(v: (String, _)): Option[String] = Some(s"""key: "${v._1}" value: "${v._2}"""")
```
benedeki, line 85: **Minor:** The function name is not very informative IMHO. What about `makeAuditTrailChangeValue`? Or anything 😄

data-model/src/main/scala/za/co/absa/enceladus/model/menas/audit/Auditable.scala
```diff
@@ -79,6 +79,74 @@ trait Auditable[T <: Product] { self: T =>
...
+
+    val index = getFieldIndex(fieldName.declaredField)
+
+    index.map({i =>
+      val maybeNewMap = newValue.productElement(i).asInstanceOf[Option[Map[String, _]]]
+      val maybeOldMap = self.productElement(i).asInstanceOf[Option[Map[String, _]]]
+
+      (maybeNewMap, maybeOldMap) match {
```
benedeki, line 93: **Suggestion:** This function is rather long. I would put the cases into sub-functions.
* Shorter
* `oldMap.toSet` and `newMap.toSet` wouldn't need to be done multiple times
* The second and the third `case` is code duplication for parts of the first `case`

benedeki (changes requested):
Please change the PR target from `develop`.
The other comments are not requests  to change, just suggestions open to discussion.

data-model/src/main/scala/za/co/absa/enceladus/model/menas/audit/Auditable.scala
Zejnilovic: Followed convention of `getPrimitiveFieldsAudit`, `getSeqFieldsAudit`. Would you change these as well?

data-model/src/test/scala/za/co/absa/enceladus/model/menas/audit/AuditableTest.scala
```diff
@@ -75,4 +84,21 @@ class AuditableTest extends AnyFunSuite {
...
+
+  test("Testing changed getOptionalMapFieldsAudit") {
+    val newObj2 = obj2.copy(properties = Some(Map("key1" -> "value previous")))
+    val result = newObj2.getOptionalMapFieldsAudit(obj3, AuditFieldName("properties", "Property"))
+    assertResult(Seq(
+      AuditTrailChange(field = "properties", newValue = Some("""key: "key1" value: "value1""""), oldValue = Some("""key: "key1" value: "value previous""""), message = "Property changed.")
+    ))(result)
+  }
+
+
```
dk1844, line 103: I believe the _removed_ test case should also be present.

```suggestion
  test("Testing removed getOptionalMapFieldsAudit") {
    val result = obj3.getOptionalMapFieldsAudit(obj2, AuditFieldName("properties", "Property"))
    assertResult(Seq(
      AuditTrailChange(field = "properties", newValue = None, oldValue = Some("""key: "key1" value: "value1""""), message = "Property removed.")
    ))(result)
  }
```

dk1844: Depending on the rest of the decision, you may want to include the cases for both `properties = Some(Map.empty)|None` for the _added_ and _removed_ test cases, to make sure that you correctly cover the trail creation for these cases, too.

Whether to spell it out and forego some DRYness or use the infamous `Seq(None, Some(Map.empty)).forEach{ props => test(s"almost the same test differentiated by $prop")}` is up to you.

For my brief'n'naive™ testing I have just spelled it out (steal away if it seems useful):
```scala
  test("Testing added getOptionalMapFieldsAudit - explicitly None properties") {
    val result = obj2.copy(properties = None).getOptionalMapFieldsAudit(obj3, AuditFieldName("properties", "Property"))
    assertResult(Seq(
      AuditTrailChange(field = "properties", newValue = Some("""key: "key1" value: "value1""""), oldValue = None, message = "Property added.")
    ))(result)
  }

  test("Testing removed getOptionalMapFieldsAudit  - explicitly None properties") {
    val result = obj3.getOptionalMapFieldsAudit(obj2.copy(properties = None), AuditFieldName("properties", "Property"))
    assertResult(Seq(
      AuditTrailChange(field = "properties", newValue = None, oldValue = Some("""key: "key1" value: "value1""""), message = "Property removed.")
    ))(result)
  }
```

data-model/src/main/scala/za/co/absa/enceladus/model/menas/audit/Auditable.scala
dk1844: Agreed on being a bit long.

But I have the feeling that maybe only the first case processing is actually needed (because what is the auditing difference between having `None` vs `Some(Map.empty)`), so this may be a way to shorten it. I hope I am not missing something but would do:

```scala
index.map({ i =>
      // because from AuditTrail perspective None vs Some(Map.empty) yields the same AuditTrail result:
      val newMap = newValue.productElement(i).asInstanceOf[Option[Map[String, _]]].getOrElse(Map.empty)
      val oldMap = self.productElement(i).asInstanceOf[Option[Map[String, _]]].getOrElse(Map.empty)

      val added = ... // same as before
      val removed = ... // same as before
     ...

      val addedList = ... // same as before
      val removedList = ... // same as before
      val changedList = ... // same as before

      removedList ++ addedList ++ changedList
    }).getOrElse(Set.empty).toSeq

```

I believe that after this, the result might already be simple enough. But some sub-functions, as @benedeki suggested, will still be possible, of course. Your call.

dk1844 (commented):
- [x] code reviewed
- [x] pulled
- [x] built
- [ ] run
(just ran the `AuditableTest` suite).

Apart from the suggestions, LGTM.

data-model/src/test/scala/za/co/absa/enceladus/model/menas/audit/AuditableTest.scala
Zejnilovic: A great catch. Did not notice properties were Empty by default.

data-model/src/main/scala/za/co/absa/enceladus/model/menas/audit/Auditable.scala
benedeki: Sorry not being clear. No problem with `getOptionalMapFieldsAudit`.
I was talking about `tupleToSomeString`.

benedeki (approved):
- [x] code reviewed
- [x] pulled
- [x] built
- [x] run

dk1844 (approved):
LGTM (just read the code)

PR #1823, opened 2021-06-21
1819 Fix properties warnings not being logged in spark-jobs

Zejnilovic:
Added another level to validation. It is warnings. Defaults to an empty map, so it should be fine.

image

```json
{
  "propertiesValidation": {
    "errors": {
      "firstProp123": ["Dataset property 'firstProp123' is mandatory, but does not exist!"]
    },
    "warnings": {
      "firstProp2": ["Property 'firstProp2' is recommended to be present, but was not found!"],
      "firstProp3": ["Property 'firstProp3' is recommended to be present, but was not found!"]
    }
  }
}
```

### DIFF
--- a/data-model/src/main/scala/za/co/absa/enceladus/model/Validation.scala
+++ b/data-model/src/main/scala/za/co/absa/enceladus/model/Validation.scala
@@ -55,13 +55,15 @@ case class Validation (errors: Map[String, List[String]] = Map(), warnings: Map[
   }
 
   def merge(validation: Validation): Validation = {
-    val mergedErrMaps = validation.errors.foldLeft(errors) { case (acc, (key, list)) =>
-      acc + (key -> (acc.getOrElse(key, List.empty[String]) ++ list))
+    def mergeKeyedMaps(first: Map[String, List[String]],
+                       second: Map[String, List[String]]): Map[String, List[String]] = {
+      first.foldLeft(second) { case (acc, (key, list)) =>
+        acc + (key -> (acc.getOrElse(key, List.empty[String]) ++ list))
+      }
     }
 
-    val mergedWarnMaps = validation.warnings.foldLeft(warnings) { case (acc, (key, list)) =>
-      acc + (key -> (acc.getOrElse(key, List.empty[String]) ++ list))
-    }
+    val mergedErrMaps = mergeKeyedMaps(validation.errors, errors)
+    val mergedWarnMaps = mergeKeyedMaps(validation.warnings, warnings)
 
     Validation(mergedErrMaps, mergedWarnMaps)
   }
--- a/data-model/src/main/scala/za/co/absa/enceladus/model/Validation.scala
+++ b/data-model/src/main/scala/za/co/absa/enceladus/model/Validation.scala
@@ -38,7 +38,7 @@ case class Validation (errors: ValidationRecord = Map(), warnings: ValidationRec
   def isEmpty: Boolean = errors.isEmpty && warnings.isEmpty
 
   @JsonIgnore
-  def nonEmpty: Boolean = errors.nonEmpty && warnings.nonEmpty
+  def nonEmpty: Boolean = errors.nonEmpty || warnings.nonEmpty
 
   def withError(key: String, error: String): Validation = {
     this.copy(errors = errors + (key -> (error :: errors.getOrElse(key, Nil))))
--- a/data-model/src/main/scala/za/co/absa/enceladus/model/Validation.scala
+++ b/data-model/src/main/scala/za/co/absa/enceladus/model/Validation.scala
@@ -16,18 +16,20 @@
 package za.co.absa.enceladus.model
 
 import com.fasterxml.jackson.annotation.JsonIgnore
+import za.co.absa.enceladus.model.Validation.ValidationRecord
 
 object Validation {
+
+  type ValidationRecord = Map[String, List[String]]
 
   val NotSpecified = "not specified"
 
   val empty: Validation = Validation()
 
   def merge(a: Validation, b: Validation) : Validation = a.merge(b)
-
 }
 
-case class Validation (errors: Map[String, List[String]] = Map(), warnings: Map[String, List[String]] = Map()) {
+case class Validation (errors: ValidationRecord = Map(), warnings: ValidationRecord = Map()) {
 
   @JsonIgnore
   def isValid: Boolean = errors.isEmpty
@@ -55,8 +57,8 @@ case class Validation (errors: Map[String, List[String]] = Map(), warnings: Map[
   }
 
   def merge(validation: Validation): Validation = {
-    def mergeKeyedMaps(first: Map[String, List[String]],
-                       second: Map[String, List[String]]): Map[String, List[String]] = {
+    def mergeKeyedMaps(first: ValidationRecord,
+                       second: ValidationRecord): ValidationRecord = {
       first.foldLeft(second) { case (acc, (key, list)) =>
         acc + (key -> (acc.getOrElse(key, List.empty[String]) ++ list))
       }
--- a/data-model/src/main/scala/za/co/absa/enceladus/model/Validation.scala
+++ b/data-model/src/main/scala/za/co/absa/enceladus/model/Validation.scala
@@ -27,24 +27,42 @@ object Validation {
 
 }
 
-case class Validation (errors: Map[String, List[String]] = Map()) {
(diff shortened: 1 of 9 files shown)


menas/src/main/scala/za/co/absa/enceladus/menas/services/DatasetService.scala
```diff
@@ -155,12 +155,12 @@ class DatasetService @Autowired()(datasetMongoRepository: DatasetMongoRepository
         propDef.essentiality match {
           case Mandatory(true) if forRun =>
             log.warn(s"Property '${propDef.name}' is required to be present, but was not found! (Run enabled)")
-            acc
+            acc.withWarning(propDef.name, s"Property '${propDef.name}' is recommended to be present, but was not found!")
```
dk1844, line 158: the `acc.withWarning` seems correct (turning required property issue into a warning when allowing to run), but the message seems a bit off (it is actually _required_ to be present) and the fact that it is allowed to run nonetheless should be probably worded slightly differently?

data-model/src/main/scala/za/co/absa/enceladus/model/Validation.scala
```diff
@@ -27,24 +27,42 @@ object Validation {
...
+    val mergedErrMaps = validation.errors.foldLeft(errors) { case (acc, (key, list)) =>
+      acc + (key -> (acc.getOrElse(key, List.empty[String]) ++ list))
+    }
+
+    val mergedWarnMaps = validation.warnings.foldLeft(warnings) { case (acc, (key, list)) =>
       acc + (key -> (acc.getOrElse(key, List.empty[String]) ++ list))
     }
 
-    Validation(mergedMaps)
+    Validation(mergedErrMaps, mergedWarnMaps)
```
dk1844, line 66: These could share a common generalized aux method and reuse it, e.g.:
```scala
  def merge(validation: Validation): Validation = {
    def mergeKeyedMaps(first: Map[String, List[String]], second: Map[String, List[String]]): Map[String, List[String]] = {
      first.foldLeft(second) { case (acc, (key, list)) =>
        acc + (key -> (acc.getOrElse(key, List.empty[String]) ++ list))
      }
    }

    val mergedErrMaps = mergeKeyedMaps(validation.errors, errors)
    val mergedWarnMaps = mergeKeyedMaps(validation.warnings, warnings)

    Validation(mergedErrMaps, mergedWarnMaps)
  }
```

While it is a bit DRIer, it is just a light suggestion.

data-model/src/main/scala/za/co/absa/enceladus/model/Validation.scala
```diff
@@ -27,24 +27,42 @@ object Validation {
 
 }
 
-case class Validation (errors: Map[String, List[String]] = Map()) {
+case class Validation (errors: Map[String, List[String]] = Map(), warnings: Map[String, List[String]] = Map()) {
```
dk1844, line 30: I keep wondering if a different structure like the following would be better or worse:

```scala
  sealed trait IssueType
  object IssueTypes {
    case object Error extends IssueType
    case object Warning extends IssueType
  }
  
  case class Issue(tpe: IssueType, messages: List[String])

  case class Validation(issues: Map[String, Issue] = Map()) {...}
```

Some things could get simplified (like merging validations, 1-1 key-issueType mapping existence), some would get a tad harder (`def isValid = issues.values.exists(_.tpe == IssueTypes.Error)`).

Just thinking out loud, feel free to decide either way.

dk1844 (approved):
Just some ideas... LGTM.

data-model/src/main/scala/za/co/absa/enceladus/model/Validation.scala
Zejnilovic: After an offline discussion we came to a conclusion that we will keep the case class as is.

menas/src/main/scala/za/co/absa/enceladus/menas/services/DatasetService.scala
Zejnilovic: Added a hopefully better wording.

dk1844 (dismissed):
LGTM

AdrianOlosutean (dismissed):
Code reviewed

data-model/src/main/scala/za/co/absa/enceladus/model/Validation.scala
```diff
@@ -27,24 +27,44 @@ object Validation {
 
 }
 
-case class Validation (errors: Map[String, List[String]] = Map()) {
```
benedeki, line 30: **Minor suggestion:** `Map[String, List[String]]` could have be defined as a `type` in the companion object. Would shorten, simplify and perhaps make it easier to understand the code.

(Placed the comment on original code, as GitHub doesn't allow me to place it to the changed one for unknown reason.)

menas/src/main/scala/za/co/absa/enceladus/menas/services/DatasetService.scala
```diff
@@ -155,12 +155,14 @@ class DatasetService @Autowired()(datasetMongoRepository: DatasetMongoRepository
         propDef.essentiality match {
           case Mandatory(true) if forRun =>
             log.warn(s"Property '${propDef.name}' is required to be present, but was not found! (Run enabled)")
```
benedeki, line 157: This warning should be removed now, with the implementation of `.withWarning`.

menas/src/main/scala/za/co/absa/enceladus/menas/services/DatasetService.scala
```diff
@@ -155,12 +155,14 @@ class DatasetService @Autowired()(datasetMongoRepository: DatasetMongoRepository
...
-            acc
+            acc.withWarning(propDef.name,
+              s"""Property '${propDef.name}' is required to be present, but was not
+                 | found! This warning will turn into error after the transition period""".stripMargin)
           case Mandatory(_) =>
             acc.withError(propDef.name, s"Dataset property '${propDef.name}' is mandatory, but does not exist!")
           case Recommended =>
             log.warn(s"Property '${propDef.name}' is recommended to be present, but was not found!")
```
benedeki, line 164: Same as line 157, the warning is an undesired side-effect here. Please remove.

benedeki (changes requested):
Great implementation, thank you (I hoped for something like this). Just please remove the two warnings into log please.

benedeki (commented):
Wanted to approve, but fails on `DatasetServiceTest`  during my build...
(Will investigate more later)

Zejnilovic:
> Wanted to approve, but fails on `DatasetServiceTest` during my build...
> (Will investigate more later)

@benedeki, somebody forgot to update tests after updating the validation messages... 🤷‍♂️

benedeki (approved):
> @benedeki, somebody forgot to update tests after updating the validation messages... 🤷‍♂️
Sorry was too tired at that moment to figure it out.

- [x] code reviewed
- [x] pulled
- [x] built
- [ ] run

spark-jobs/src/main/scala/za/co/absa/enceladus/common/CommonJobExecution.scala
```diff
@@ -92,6 +92,9 @@ trait CommonJobExecution extends ProjectMetadata {
         throw new IllegalStateException("Dataset validation failed, errors found in fields:\n" +
           validation.errors.map { case (field, errMsg) => s" - '$field': $errMsg" }.mkString("\n")
         )
+      case Some(validation) if validation.nonEmpty =>
+        log.warn("Dataset validation had some warnings:\n" +
+          validation.warnings.map { case (field, warnMsg) => s" - '$field': $warnMsg" }.mkString("\n"))
```
benedeki, line 97: After fixing, this turns out not very nice:
```
21/07/11 18:57:41 WARN StandardizationAndConformanceJob$: Dataset validation had some warnings:
 - 'Color': List(Property 'Color' is recommended to be present, but was not found!)
 - 'SNOW_CORRELATION_ID': List(Property 'SNOW_CORRELATION_ID' is recommended to be present, but was not found!)
```
```suggestion
          validation.warnings.map { case (field, warnMsg) => {
            val header = s" - '$field': "
            s"$header${warnMsg.mkString(s"\n$header")}" }.mkString("\n")}
        )
```

benedeki (changes requested):

dk1844 (approved):
The code still LGTM. (it was already pretty good before). Just read the code this time.

But it seems to be failing on Jenkins, I see.

benedeki (approved):
- [x] code reviewed
- [x] pulled
- [x] built
- [x] run

PR #1824, opened 2021-06-21
strip blank properties fix

dk1844:
This PR addresses 2 problems: 
1. when updating a dataset (using _Edit_), new ds property values are in `_properties` (while the old values are in `properties`); while for conformance rules update, the values are only in `properties`, as they are not being changed. (so, `_properties` is not defined)
2. when changing CRs, the ds properties are not kept (they get stripped) - update/delete/move (therefore simple `if(!dataset._properties) {...}` does not suffice)

The code now works as follows: only `_properties` are being blank-checked-and-fixed (moved to `DatasetService.cleanupEntity`), while for CR updates, the `properties` are used.

Note to self: we are so missing an UI testing framework being employed :(

## Testing done:
 - on localhost:
 1. create a dataset with ds properties -> checked in UI and DB for correct result
 1. change ds properties -> observed the blanks being removed
 1. create a conformance rule (a CastingConformanceRule) -> observe CR being added to a new version while the ds properties being still present on DS
 1. delete the conformance rule -> observe CR removed and ds properties being still present on DS

## RN suggestion:
Fixed: Dataset properties checking broke some Conformance Rules updates.

### DIFF
--- a/menas/ui/service/EntityService.js
+++ b/menas/ui/service/EntityService.js
@@ -307,7 +307,9 @@ class DatasetService extends EntityService {
 
     if (oEntity._properties && oEntity._properties.map) {
       oEntity._properties.map((oProp) => {
-        oProps[oProp.name] = oProp.value
+        if (oProp.value !== "") { // strips blank properties
+          oProps[oProp.name] = oProp.value
+        }
       });
     }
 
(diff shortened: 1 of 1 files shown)


Zejnilovic (approved):
Tested on both compliant and non-compliant dataset:
- edit Dataset
- edit conformance rules
- add conformance rules
- move conformance rules
- delete conformance rules 
- review of audit trail

Zejnilovic:
**1. Question:**
Property with white space is not blank?

image

AdrianOlosutean (approved):
Looks good

dk1844:
> **1. Question:**
> Property with white space is not blank?
> 
> image
> 
> I know it behaves as it used to.

No, only empty-string property values are considered blank and converted to `None`. Values with length > 0 are allowed, I was not aware of any requirements to disallow it. Do you think we should reconsider, @Zejnilovic?

Zejnilovic:
> > **1. Question:**
> > Property with white space is not blank?
> > image
> > I know it behaves as it used to.
> 
> No, only empty-string property values are considered blank and converted to `None`. Values with length > 0 are allowed, I was not aware of any requirements to disallow it. Do you think we should reconsider, @Zejnilovic?

I think it is ok right now as is. But I would make some improvements from the UX point of view. Maybe even allow regex validations for string properties. I will put my thoughts down and create an issue later.

PR #1828, opened 2021-06-23
1807 Add schema registry security params

### LINKED ISSUE #1807: Use authenticated user when sending _INFO to Kafka
## Background
In case Kafka requires user authentication to write into a topic, this needs to be supported by the Kafka plugin

## Feature
It's to be discussed and decided if the user executing the `SparkJobs` would be the one authenticating against Kafka, or there would be one user set-up to do the writing. Or the former being a fallback of the later.

Zejnilovic:
New configs for SparkJobs using kafak plugins
- kafka.schema.registry.basic.auth.credentials.source
- kafka.schema.registry.basic.auth.user.info

Release Notes:
Added new properties for KafkaPlugins to be able to connect to Secured SchemaRegistry

Fixes #1807

### DIFF
--- a/plugins-builtin/src/main/scala/za/co/absa/enceladus/plugins/builtin/common/mq/kafka/KafkaSecurityParams.scala
+++ b/plugins-builtin/src/main/scala/za/co/absa/enceladus/plugins/builtin/common/mq/kafka/KafkaSecurityParams.scala
@@ -16,7 +16,7 @@
 package za.co.absa.enceladus.plugins.builtin.common.mq.kafka
 
 import com.typesafe.config.Config
-import za.co.absa.enceladus.plugins.builtin.common.mq.kafka.KafkaSecurityParams.{SaslMechanismKey, SecurityProtocolKey}
+import za.co.absa.enceladus.plugins.builtin.common.mq.kafka.KafkaSecurityParams._
 import za.co.absa.enceladus.utils.config.ConfigUtils.ConfigImplicits
 
 /**
@@ -27,8 +27,12 @@ case class KafkaSecurityParams(
                                 saslMechanism: Option[String]
                               ) {
   def toMap: Map[String, String] = {
-    Map(SecurityProtocolKey -> securityProtocol) ++
-      saslMechanism.fold(Map.empty[String, String])(saslValue => Map(SaslMechanismKey -> saslValue))
+    Map(
+      SecurityProtocolKey -> Some(securityProtocol),
+      SaslMechanismKey -> saslMechanism
+    ).collect {
+      case (key, Some(value)) => key -> value
+    }
   }
 }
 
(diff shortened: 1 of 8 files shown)


plugins-builtin/src/main/scala/za/co/absa/enceladus/plugins/builtin/common/mq/kafka/KafkaSecurityParams.scala
```diff
@@ -27,8 +27,12 @@ case class KafkaSecurityParams(
...
                               ) {
   def toMap: Map[String, String] = {
-    Map(SecurityProtocolKey -> securityProtocol) ++
-      saslMechanism.fold(Map.empty[String, String])(saslValue => Map(SaslMechanismKey -> saslValue))
+    Map(
+      SecurityProtocolKey -> Some(securityProtocol),
+      SaslMechanismKey -> saslMechanism
+    ).collect {
+      case (key, Some(value)) => key -> value
+    }
```
Zejnilovic, line 35: Did this in the first iteration when I added 3 more params. I would just leave it be if no one has any comments.

AdrianOlosutean: There is also this option to use:
`for ((k, Some(v)) <- map) yield k -> v`
But I would not say it's that much better

AdrianOlosutean (approved):
Looks good so far, Im wondering where to test it

plugins-builtin/src/main/scala/za/co/absa/enceladus/plugins/builtin/common/mq/kafka/SchemaRegistrySecurityParams.scala
```diff
@@ -0,0 +1,34 @@
...
+
+case class SchemaRegistrySecurityParams(credentialsSource: String,
+                                        userInfo: Option[String])
+
+object SchemaRegistrySecurityParams {
+  val BasicAuthCredentialsSourceKey = "kafka.schema.registry.basic.auth.credentials.source"
+  val BasicAuthUserInfoKey = "kafka.schema.registry.basic.auth.user.info"
+
+  def fromConfig(conf: Config): Option[SchemaRegistrySecurityParams] = {
+    conf.getOptionString(BasicAuthCredentialsSourceKey).map { source =>
```
benedeki, line 29: This is neat! 😎

plugins-builtin/src/main/scala/za/co/absa/enceladus/plugins/builtin/common/mq/kafka/KafkaSecurityParams.scala
benedeki: I think it's OK , how it is.

benedeki (approved):
- [x] code reviewed
- [x] pulled
- [x] built
- [ ] run

dk1844:
Test-ran on Chewbacca, the info plugin actually wrote data to a Kafka topic while using configuration such as
```properties
kafka.schema.registry.basic.auth.credentials.source=USER_INFO
kafka.schema.registry.basic.auth.user.info="SVC-user-accounT:secretPasswordExampleHere#2+a"
```

Issue #1835, opened 2021-06-30
Support multiple date formats in Helper scripts


benedeki:
## Background
In Helper scripts date is parsed for a log file naming.

## Feature
On Windows the `date` OS function result is locale dependent. Make the Helper scripts easy to configure for different  date formats

benedeki:
Release notes:
Windows _Helper scripts_ are now parse the date independently of locale (for the log file name purpose).

PR #1837, opened 2021-06-30
#1835: Support multiple date formats in Helper scripts

benedeki:
* added configuration for date parsing type
* .gitattributes changed to hopefully better handle CRLF in native Windows files
* LF -> CRLF in *.cmd and *.bat files

I needed to implement the partial changes for my testing for other issues. So I thought to make it more general and include in code-base.

_Release notes:_
Configuration of the date parsing (locale dependent) to correctly name the log file.

### DIFF
--- a/scripts/cmd/_run_enceladus.cmd
+++ b/scripts/cmd/_run_enceladus.cmd
@@ -464,6 +464,11 @@ CALL :validate --dataset-version,%DATASET_VERSION%
 CALL :validate --report-date,%REPORT_DATE%
 CALL :validate_either --menas-credentials-file,%MENAS_CREDENTIALS_FILE%,--menas-auth-keytab,%MENAS_AUTH_KEYTAB%
 
+IF NOT DEFINED DATE_PARSING_TYPE (
+    ECHO DATE_PARSING_TYPE not defined in `_enceladus_env.cmd`
+    SET VALID=="0"
+)
+
 :: For now this check is disabled in Windows version of the script
 :: IF NOT "%MASTER%"=="yarn" (
 ::   ECHO "Master '%MASTER%' is not allowed. The only allowed master is 'yarn'."
@@ -602,6 +607,8 @@ IF "%ASYNCHRONOUS_MODE%"=="true" (
 GOTO :eof
 
 :client_run
+:end
+
 CALL :temp_log_file TMP_PATH_NAME
 
 :: Initializing Kerberos ticket
@@ -638,7 +645,7 @@ IF EXIST %ERROR_SIGNAL_FILE% (
 :: Report the result and log location
 ECHO ""
 ECHO Job %RESULT%. Refer to logs at %TMP_PATH_NAME% | tee -a %TMP_PATH_NAME%
-EXIT /B %EXIT_STATUS%
+
 
 :: Functions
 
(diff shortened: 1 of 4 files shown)


dk1844 (dismissed):
LGTM (just read the code)

scripts/cmd/_run_enceladus.cmd
```diff
@@ -658,13 +665,25 @@ EXIT /B 0
...
 
 :temp_log_file
-    SET yyyy=%date:~-4%
-    SET MM=%date:~3,2%
-    SET DD=%date:~0,2%
-    ::alternate date parsing
-    ::SET YYYY=%date:~10,4%
-    ::SET MM=%date:~4,2%
-    ::SET DD=%date:~7,2%
+    IF %DATE_PARSING_TYPE%==parser1 (
```
dk1844, line 668: In the most usual cases, this will probably work fine.
In general, the users can change the formats, so even the position-based parsing may not be sufficient (e.g. for the situation of (non)-padding with 0, short/long year, etc.

I suggest taking inspiration at https://ss64.com/nt/syntax-gettime.html

benedeki: Nice. I will try to implement this in a free moment. This is no high priority. I pushed the change only because I needed a different parsing, so made it switchable, albeit not the smartest. 😉

dk1844: Yes, I agree that the current solution is probably good enough for the use case. 

It just made sense to provide some information in the problematics in the broader sense, we may very well decide to leave it as-is, it would just be a bit more informed decision 😄

benedeki: Should be universal now 🤞 
Also added configuration to prevent editor conversion from CRLF to LF

scripts/cmd/_run_enceladus.cmd
```diff
@@ -665,29 +659,20 @@ EXIT /B 0
...
-        :: 2021-06-25
-        SET yyyy=%date:~0,4%
-        SET MM=%date:~5,2%
-        SET DD=%date:~8,2%
+    ::Date&Time parsing
+    FOR /F "skip=1 tokens=1-6" %%A IN ('WMIC Path Win32_LocalTime Get Day^,Hour^,Minute^,Month^,Second^,Year /Format:table') DO (
+        if "%%B" NEQ "" (
+            SET /A FDATE=%%F*10000+%%D*100+%%A
+            :: prefixing 1000000 to ensure the hours is two digit (starting with 0 before 10)
+            SET /A FTIME=1000000+%%B*10000+%%C*100+%%E
```
dk1844, line 667: This is a very interesting approach, but why not. I would have naturally attempted to created a _number-padding_ method, but this works well, too.

scripts/cmd/_run_enceladus.cmd
```diff
@@ -665,29 +659,20 @@ EXIT /B 0
...
-        SET DD=%date:~7,2%
-    )
-    IF %DATE_PARSING_TYPE%==parser3 (
-        :: 2021-06-25
-        SET yyyy=%date:~0,4%
-        SET MM=%date:~5,2%
-        SET DD=%date:~8,2%
+    ::Date&Time parsing
+    FOR /F "skip=1 tokens=1-6" %%A IN ('WMIC Path Win32_LocalTime Get Day^,Hour^,Minute^,Month^,Second^,Year /Format:table') DO (
+        if "%%B" NEQ "" (
```
dk1844, line 664: Again, naturally, I would have looked for cmd alternative for shell `myCommand | head -n2 | tail -n1`, but this seems to be simply available only in powershell 🙄

dk1844 (approved):
- [x] code reviewed
- [x] pulled
- [x] built
- [x] run and disected

LGTM. 

Personal remark: Oh, I so don't want to be writing stuff for CMD (Powershell seems to fill the blanks, at least. And the GOTO command 🙄 )

scripts/cmd/_run_enceladus.cmd
benedeki: I found it the easiest. Batch scripts are designed to manipulate strings, as I see it 😉

scripts/cmd/_run_enceladus.cmd
benedeki: I admit, I found this on the Internet, not my invention at all.

benedeki:
> 
> Personal remark: Oh, I so don't want to be writing stuff for CMD (Powershell seems to fill the blanks, at least. And the GOTO command 🙄 )

Funny I never learned PowerShell, and I find batch script surprisingly powerful. 👼 
And the `GOTO` - seemed as the clearest solution, could have done it with reverting the `if`, perhaps with some `{...}` 🤷

benedeki:
> 
> 
>     * [x]  code reviewed
>     * [x]  pulled
>     * [x]  built
>     * [x]  run and disected

Based on this I dare to add `PR: Testes` label. 😉

benedeki:
Release notes used:
Windows _Helper scripts_ are now parse the date independently of locale (for the log file name purpose).

PR #1838, opened 2021-06-30
#1818 errCol added in Conformance is not Nullable

### LINKED ISSUE #1818: Investigate why errCol added in Conformance is not Nullable
## Describe the bug
With the introduction of _Spark 3_ suddenly if `errCol` is added in _Conformance_ phase it became `Nullable=false`. This could mean an issue for Hyperdrive for example.
But it seems the column does stay nullable in _Standardization_ phase.

## To Reproduce
1. See `MappingGroupExplodeSuite` vs `StandardizationInterpreter_ArraySuite`

## Expected behaviour
`MappingInterpreterSuite.normalizeErrColNullability` should not be needed probably.

dk1844:
# The Problem 
The problem is at length described in https://github.com/AbsaOSS/enceladus/issues/1818, but to sum up:
When running conformance under Spark 3.1, the `errCol` is not nullable; while it is nullable under Spark 2.4

# The Cause
[utils/src/main/scala/za/co/absa/enceladus/utils/explode/ExplodeTools.scala](https://github.com/AbsaOSS/enceladus/blob/cbc06ff3b329b6f5eb169b499a65b4d36f567ce4/utils/src/main/scala/za/co/absa/enceladus/utils/explode/ExplodeTools.scala#L184) calls 
```scala
        decDf.orderBy(orderByRecordCol, orderByInsideArray)
          .groupBy(groupByColumns: _*)
          .agg(collect_list(deconstructedField).as(tmpColName),
            array_distinct(flatten(collect_list(col(errorCol)))).as(errorCol))
```
On Spark 3, the `errCol` in `decDf` is nullable, while in the result of this operation, `errCol` is not nullable anymore. The underlying **root cause seems to be function `collect_list`**. **When operating on a nullable column, under Spark 2.4 the result remains nullable while under Spark 3.1 is changed to not-nullable**.

## Evidence
The following code demonstrates the difference when ran on Spark 2.4 and Spark 3.1:

```scala
// hint: use ":paste" command to paste all at-once in spark-shell

import org.apache.spark.sql.Row
import org.apache.spark.sql.types.{StringType, StructField, StructType}

val arrayStructData = Seq(
  Row("James", "Java"),
  Row("James", "Python"),
  Row("Michael", "Java"),
  Row("Washington", null)
)

val arrayStructSchema = StructType(Seq(
  StructField("name", StringType, nullable = true),
  StructField("language", StringType, nullable = true)
))

val df1 = spark.createDataFrame(spark.sparkContext.parallelize(arrayStructData), arrayStructSchema)
df1.show
df1.printSchema

val df2 = df1.groupBy("name").agg(collect_list("language").as("langs"))

df2.show
df2.printSchema
```

All output is the same except for the df2's schema, which differs: `langs: array (nullable = true)` for Spark 2 vs `langs: array (nullable = false)` for Spark3.1

### Spark 2 output
```
+----------+--------+
|      name|language|
+----------+--------+
|     James|    Java|
|     James|  Python|
|   Michael|    Java|
|Washington|    null|
+----------+--------+

root
 |-- name: string (nullable = true)
 |-- language: string (nullable = true)

+----------+--------------+
|      name|         langs|
+----------+--------------+
|     James|[Java, Python]|
|Washington|            []|
|   Michael|        [Java]|
+----------+--------------+

root
 |-- name: string (nullable = true)
 |-- langs: array (nullable = true)
 |    |-- element: string (containsNull = true)
``` 
### Spark 3 output
```
+----------+--------+
|      name|language|
+----------+--------+
|     James|    Java|
|     James|  Python|
|   Michael|    Java|
|Washington|    null|
+----------+--------+

root
 |-- name: string (nullable = true)
 |-- language: string (nullable = true)

+----------+--------------+
|      name|         langs|
+----------+--------------+
|     James|[Java, Python]|
|Washington|            []|
|   Michael|        [Java]|
+----------+--------------+

root
 |-- name: string (nullable = true)
 |-- langs: array (nullable = false)
 |    |-- element: string (containsNull = false)
``` 

## Conclusion
The change in Spark 3 is logically correct because the result of `array_collect` is guaranteed to be non-null, hence the observed nullability change.
Albeit a reasonable logic, this change can cause migration issues where the resulting schema change is significant.

# Fixing the issue
  - errCol nullablity (on the dataframe level) in conformance and standardization is now driven by config value `enceladus.errCol.nullable` (default: true = original Eneceladus 2.x behavior) 
    - but **beware, that parquet, when read by spark, will contain all columns nullable** (see the [doc](https://spark.apache.org/docs/3.1.2/sql-data-sources-parquet.html)) 
  - tests updated: normalizeErrColNullability removed
 - added DataFrameImplicits.setNullableStateOfColumn + unitTests

Closes #1818.

### DIFF
--- a/spark-jobs/src/main/resources/reference.conf
+++ b/spark-jobs/src/main/resources/reference.conf
@@ -28,7 +28,7 @@ menas.rest.uri="http://localhost:8080,host2:9000/menas;https://localhost:8080,ho
 # column will not be added at all. Allowed values: "uuid", "stableHashId", "none"
 enceladus.recordId.generation.strategy="uuid"
 
-# errCol's schema output to be (non)nullable in datagrame - both for Standardization and Conformance, regardless of errCol schema input
+# errCol's schema output to be (non)nullable in dataframe - both for Standardization and Conformance, regardless of errCol schema input
 # Warning: this works on dataframe level, when writing to/reading from parquet, columns are by default nullable, see:
 # https://spark.apache.org/docs/3.1.2/sql-data-sources-parquet.html
 enceladus.errCol.nullable=true
--- a/spark-jobs/src/main/resources/reference.conf
+++ b/spark-jobs/src/main/resources/reference.conf
@@ -31,7 +31,7 @@ enceladus.recordId.generation.strategy="uuid"
 # errCol's schema output to be (non)nullable in datagrame - both for Standardization and Conformance, regardless of errCol schema input
 # Warning: this works on dataframe level, when writing to/reading from parquet, columns are by default nullable, see:
 # https://spark.apache.org/docs/3.1.2/sql-data-sources-parquet.html
-enceladus.errCol.nullable=false
+enceladus.errCol.nullable=true
 
 standardized.hdfs.path="/tmp/conformance-output/standardized-{0}-{1}-{2}-{3}"
 
--- a/spark-jobs/src/main/resources/reference.conf
+++ b/spark-jobs/src/main/resources/reference.conf
@@ -28,6 +28,9 @@ menas.rest.uri="http://localhost:8080,host2:9000/menas;https://localhost:8080,ho
 # column will not be added at all. Allowed values: "uuid", "stableHashId", "none"
 enceladus.recordId.generation.strategy="uuid"
 
+# errCol's schema in output to be (non)nullable - both for Standardization and Conformance, regardless of errCol schema input
+enceladus.errCol.nullable=false
+
 standardized.hdfs.path="/tmp/conformance-output/standardized-{0}-{1}-{2}-{3}"
 
 # Pattern to look for mapping table for the specified date
--- a/spark-jobs/src/main/resources/reference.conf
+++ b/spark-jobs/src/main/resources/reference.conf
@@ -28,7 +28,9 @@ menas.rest.uri="http://localhost:8080,host2:9000/menas;https://localhost:8080,ho
 # column will not be added at all. Allowed values: "uuid", "stableHashId", "none"
 enceladus.recordId.generation.strategy="uuid"
 
-# errCol's schema in output to be (non)nullable - both for Standardization and Conformance, regardless of errCol schema input
+# errCol's schema output to be (non)nullable in datagrame - both for Standardization and Conformance, regardless of errCol schema input
+# Warning: this works on dataframe level, when writing to/reading from parquet, columns are by default nullable, see:
+# https://spark.apache.org/docs/3.1.2/sql-data-sources-parquet.html
 enceladus.errCol.nullable=false
 
 standardized.hdfs.path="/tmp/conformance-output/standardized-{0}-{1}-{2}-{3}"
--- a/spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/ConformanceExecution.scala
+++ b/spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/ConformanceExecution.scala
@@ -30,7 +30,7 @@ import za.co.absa.enceladus.common.plugin.menas.MenasPlugin
 import za.co.absa.enceladus.common.{CommonJobExecution, Constants, RecordIdGeneration}
 import za.co.absa.enceladus.conformance.config.{ConformanceConfig, ConformanceConfigParser}
 import za.co.absa.enceladus.conformance.interpreter.rules.ValidationException
-import za.co.absa.enceladus.conformance.interpreter.{FeatureSwitches, DynamicInterpreter}
+import za.co.absa.enceladus.conformance.interpreter.{DynamicInterpreter, FeatureSwitches}
 import za.co.absa.enceladus.dao.MenasDAO
 import za.co.absa.enceladus.dao.auth.MenasCredentials
 import za.co.absa.enceladus.model.Dataset
--- a/spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/ConformanceExecution.scala
+++ b/spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/ConformanceExecution.scala
@@ -30,7 +30,7 @@ import za.co.absa.enceladus.common.plugin.menas.MenasPlugin
 import za.co.absa.enceladus.common.{CommonJobExecution, Constants, RecordIdGeneration}
 import za.co.absa.enceladus.conformance.config.{ConformanceConfig, ConformanceConfigParser}
 import za.co.absa.enceladus.conformance.interpreter.rules.ValidationException
-import za.co.absa.enceladus.conformance.interpreter.{DynamicInterpreter, FeatureSwitches}
+import za.co.absa.enceladus.conformance.interpreter.{FeatureSwitches, DynamicInterpreter}
 import za.co.absa.enceladus.dao.MenasDAO
 import za.co.absa.enceladus.dao.auth.MenasCredentials
 import za.co.absa.enceladus.model.Dataset
--- a/spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/ConformancePropertiesProvider.scala
+++ b/spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/ConformancePropertiesProvider.scala
@@ -21,10 +21,11 @@ import za.co.absa.enceladus.conformance.config.ConformanceConfigParser
 import za.co.absa.enceladus.utils.config.ConfigUtils.ConfigImplicits
 import za.co.absa.enceladus.conformance.interpreter.{FeatureSwitches, ThreeStateSwitch}
 import ConformancePropertiesProvider._
+import za.co.absa.enceladus.common.ErrorColNormalization
 
 /**
-  * Reads conformance properties from the configuration file
-  */
+ * Reads conformance properties from the configuration file
+ */
 class ConformancePropertiesProvider {
   private val enableCF: Boolean = true
   private val log: Logger = LoggerFactory.getLogger(this.getClass)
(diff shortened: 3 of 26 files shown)


benedeki (commented):
So do we do the option to enable/disable this? 😉

dk1844:
> So do we do the option to enable/disable this? 😉

I have tried to do it first by making the change I have done dependable on a config property, but I have realized that this change only applies under specific conditions (arrays types with explosion/explosion revertion is used, etc), so the config property then only drives the behavior in those cases. Other times, the errCol nullability is not affected.

So the options are:
 1. keep the change to where on Explosion revert (ExplodeTools:187) the errCol is always made nullable
 2. make the nullability change configurable in the case above (ExplodeTools:187)
 3. introduce a configuration that would set the `errCol` nullability in general (regardless of the explosion reversion being used or not)

Option 1 has an expected result - the `errCol` is going to be nullable after Conformance (no matter the type of CRs)
Option 3 also has an expected - the `errCol` is going to be (non)-nullable (depending on the property settings) (no matter the type of CRs)
Option 2 is in my perspective misleading because if you set it to `true`, you will get the result of option 1; but setting it to `false` will have a confusing effect - only for certain types of CRs being used, the nullability of the `errCol` will change, otherwise, it won't.

So I am hesitant to push the change to "simply make original change config-driven", because it would result in hard-to-understand behavior (Option 2).

Zejnilovic (dismissed):
code reviewed, played with Unit Tests
I would say no integration tests needed as a pure function was updated and it is for spark-3

utils/src/main/scala/za/co/absa/enceladus/utils/implicits/DataFrameImplicits.scala
```diff
@@ -68,6 +69,24 @@ object DataFrameImplicits {
       SparkUtils.withColumnIfDoesNotExist(df, colName, col)
     }
 
+    /**
+     * Set nullable property of column.
+     *
+     * @param columnName is the column name to change
+     * @param nullable   boolean flag to set the nullability of the column `columnName` to
+     */
+    def setNullableStateOfColumn(columnName: String, nullable: Boolean): DataFrame = {
```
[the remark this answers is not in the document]
AdrianOlosutean: I would suggest it to be called `withNullableColumnState`, in order to be differentiated from the Java style setters

utils/src/main/scala/za/co/absa/enceladus/utils/implicits/DataFrameImplicits.scala
```diff
@@ -68,6 +69,24 @@ object DataFrameImplicits {
...
+      // Courtesy of https://stackoverflow.com/a/33195510
+
+      // modify [[StructField] with name `columnName`
+      val newSchema = StructType(df.schema.map {
+        case StructField(c, t, _, m) if c.equals(columnName) => StructField(c, t, nullable = nullable, m)
+        case y: StructField => y
+      })
+      // apply new schema
+      df.sqlContext.createDataFrame(df.rdd, newSchema)
```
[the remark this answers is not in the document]
dk1844: Done.

utils/src/main/scala/za/co/absa/enceladus/utils/implicits/DataFrameImplicits.scala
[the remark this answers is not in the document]
dk1844: Done.

dk1844:
> ...
> 
> So the options are:
> 
> 1. keep the change to where on Explosion revert (ExplodeTools:187) the errCol is always made nullable
> 2. make the nullability change configurable in the case above (ExplodeTools:187)
> 3. introduce a configuration that would set the `errCol` nullability in general (regardless of the explosion reversion being used or not)
> 

So, as discussed, I have implemented option 3. The config property to control the behavior is `enceladus.errCol.nullable`, by default set to `false`. It affects both Standardization and Conformance and is oblivious to the original nullability of the `errCol`, should it already exist (e.g. when running Conformance, there may be already some Standardization errors present).

AdrianOlosutean (approved):
LGTM

dk1844:
... Aaand it is probably all for naught. And the reason is (drum-roll, please):

_Parquet is a columnar format that is supported by many other data processing systems. Spark SQL provides support for both reading and writing Parquet files that automatically preserves the schema of the original data. **When reading Parquet files, all columns are automatically converted to be nullable for compatibility reasons.**_ src: https://spark.apache.org/docs/3.1.2/sql-data-sources-parquet.html

I have tested the implementation and when reading back in spark, the errCol, indeed, always appears _nullable_.
I am planning to investigate further if at least the parquet actually written with the intended nullability, but it seems that the point of the external switch is becoming quite useless if spark-parquet is behaving this way.

Unless something extraordinary comes up, I think that retracting the codebase back to where the `errCol` is made always nullable as a result of conformance on DF-level is the most compatible and most sensible option now.

spark-jobs/src/main/resources/reference.conf
```diff
@@ -28,6 +28,11 @@ menas.rest.uri="http://localhost:8080,host2:9000/menas;https://localhost:8080,ho
 # column will not be added at all. Allowed values: "uuid", "stableHashId", "none"
 enceladus.recordId.generation.strategy="uuid"
 
+# errCol's schema output to be (non)nullable in datagrame - both for Standardization and Conformance, regardless of errCol schema input
```
AdrianOlosutean, line 31: Little typo
```suggestion
# errCol's schema output to be (non)nullable in dataframe - both for Standardization and Conformance, regardless of errCol schema input
```

spark-jobs/src/main/resources/reference.conf
```diff
@@ -28,6 +28,11 @@ menas.rest.uri="http://localhost:8080,host2:9000/menas;https://localhost:8080,ho
...
+# Warning: this works on dataframe level, when writing to/reading from parquet, columns are by default nullable, see:
```
AdrianOlosutean, line 32: Good to have this comment

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/ConformanceExecution.scala
```diff
@@ -30,7 +30,7 @@ import za.co.absa.enceladus.common.plugin.menas.MenasPlugin
 import za.co.absa.enceladus.common.{CommonJobExecution, Constants, RecordIdGeneration}
 import za.co.absa.enceladus.conformance.config.{ConformanceConfig, ConformanceConfigParser}
 import za.co.absa.enceladus.conformance.interpreter.rules.ValidationException
-import za.co.absa.enceladus.conformance.interpreter.{DynamicInterpreter, FeatureSwitches}
+import za.co.absa.enceladus.conformance.interpreter.{FeatureSwitches, DynamicInterpreter}
```
AdrianOlosutean, line 33: Maybe unnecessary?

AdrianOlosutean (commented):
Just small things

spark-jobs/src/main/resources/reference.conf
dk1844: Thank you.

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/ConformanceExecution.scala
dk1844: Reverted as it was, thanks for noticing.

spark-jobs/src/main/resources/reference.conf
dk1844: Fixed, thanks for noticing.

AdrianOlosutean (approved):
Nothing else to say

utils/src/main/scala/za/co/absa/enceladus/utils/explode/ExplodeTools.scala
```diff
@@ -178,10 +179,13 @@ object ExplodeTools {
       case Some(errorCol) =>
         // Implode taking into account the error column
         // Errors should be collected, flattened and made distinct
+        // decDf.schema: "errCol: array (nullable = true)", while the result would contain "errCol: array (nullable = false)"
+        // bc. collect_list. See issue #1818
         decDf.orderBy(orderByRecordCol, orderByInsideArray)
           .groupBy(groupByColumns: _*)
           .agg(collect_list(deconstructedField).as(tmpColName),
             array_distinct(flatten(collect_list(col(errorCol)))).as(errorCol))
+          .withNullableColumnState(errorCol, true) // explicitly nullable: expected schema outcome, see issue #1818
```
benedeki, line 188: Shouldn't this be based on nullability configuration?

utils/src/main/scala/za/co/absa/enceladus/utils/implicits/DataFrameImplicits.scala
```diff
@@ -68,6 +69,29 @@ object DataFrameImplicits {
...
+     * Set nullable property of column.
+     *
+     * @param columnName is the column name to change
+     * @param nullable   boolean flag to set the nullability of the column `columnName` to
+     */
+    def withNullableColumnState(columnName: String, nullable: Boolean): DataFrame = {
+      // Courtesy of https://stackoverflow.com/a/33195510
+
+      // modify [[StructField] with name `columnName` if its nullability differs
+      val newSchema = StructType(df.schema.map {
```
benedeki, line 82: **Minor:** I think a `foldLeft` would be able to signal back if change has happened avoiding the need for comparison right after.

utils/src/main/scala/za/co/absa/enceladus/utils/implicits/DataFrameImplicits.scala
```diff
@@ -68,6 +69,29 @@ object DataFrameImplicits {
...
+        case StructField(c, t, n, m) if c.equals(columnName) && n != nullable =>
+          StructField(c, t, nullable = nullable, m)
+        case y: StructField => y
+      })
+
+      if (newSchema == df.schema) {
+        df // no change
+      } else {
+        df.sqlContext.createDataFrame(df.rdd, newSchema) // apply new schema
```
benedeki, line 91: A very paranoid thought: Could it never happen that the `errCol` is `null` , but the settings says it should not be nullable?

benedeki (approved):
Beside the minor  comments, great work.
I also appreciate the small code improvements around. 👍

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/ConformancePropertiesProvider.scala
```diff
@@ -21,10 +21,11 @@ import za.co.absa.enceladus.conformance.config.ConformanceConfigParser
 import za.co.absa.enceladus.utils.config.ConfigUtils.ConfigImplicits
 import za.co.absa.enceladus.conformance.interpreter.{FeatureSwitches, ThreeStateSwitch}
 import ConformancePropertiesProvider._
+import za.co.absa.enceladus.common.ErrorColNormalization
 
 /**
-  * Reads conformance properties from the configuration file
-  */
+ * Reads conformance properties from the configuration file
```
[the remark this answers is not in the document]
dk1844: I wonder if this is something that can be instilled in the scalastyle-config 🤔

utils/src/main/scala/za/co/absa/enceladus/utils/explode/ExplodeTools.scala
dk1844: On this level, it is deliberate to maintain consistency. (Spark 2 it was always nullable here, for Spark 3 we make it the same). Because setting it here by the configuration value (I have tried) has the effect of only for certain types of CRs being used (dubbed as _Option 2_ above)

Then, on the level of Standardization and Conformance as a whole, the nullability configuration is applied on that level, specifically:
```
DynamicInterpreter.interpret:74
StandardizationInterpreter.standardize:73
```

So we have internals that behave uniformly and externals that behave according to the configuration. By applying the config here, in ExplodeToools:184, too,  we would have forgo internal uniformity.

utils/src/main/scala/za/co/absa/enceladus/utils/implicits/DataFrameImplicits.scala
dk1844: I don't think that it is currently possible.

utils/src/main/scala/za/co/absa/enceladus/utils/explode/ExplodeTools.scala
benedeki: Thanks for the explanation. 👍

utils/src/main/scala/za/co/absa/enceladus/utils/implicits/DataFrameImplicits.scala
benedeki: OK, cool. I trust you 😉

PR #1839, opened 2021-07-01
#1808SparkVersionGuard: For Spark 3, min version = 3.0.0

### LINKED ISSUE #1808: Adjust Spark version guard
## Background
Moving to Spark 3. Not having backward compatibility now or yet.

## Feature
Adjust the version guard, so the run is executed and supported version of Spark

dk1844:
SparkVersionGuard: For Spark 3, the min version has been set to 3.0.0

Closes #1808.

### DIFF
--- a/spark-jobs/src/main/scala/za/co/absa/enceladus/common/SparkCompatibility.scala
+++ b/spark-jobs/src/main/scala/za/co/absa/enceladus/common/SparkCompatibility.scala
@@ -26,6 +26,6 @@ import za.co.absa.commons.version.Version._
 import za.co.absa.commons.version.impl.SemVer20Impl.SemanticVersion
 
 object SparkCompatibility {
-  val minSparkVersionIncluded: SemanticVersion = semver"2.4.2"
+  val minSparkVersionIncluded: SemanticVersion = semver"3.1.2"
   val maxSparkVersionExcluded: Option[SemanticVersion] = Some(semver"3.2.0")
 }


Zejnilovic (dismissed):

AdrianOlosutean (approved):

spark-jobs/src/main/scala/za/co/absa/enceladus/common/SparkCompatibility.scala
```diff
@@ -26,6 +26,6 @@ import za.co.absa.commons.version.Version._
 import za.co.absa.commons.version.impl.SemVer20Impl.SemanticVersion
 
 object SparkCompatibility {
-  val minSparkVersionIncluded: SemanticVersion = semver"2.4.2"
+  val minSparkVersionIncluded: SemanticVersion = semver"3.0.0"
```
benedeki, line 29: Considering the known problems with lower versions, what about setting this to 3.1.2?

dk1844: It does not make sense to me for this version, since it is spark3.0-based. we can do this in the spar3.1 version or if we merge the spark3.1 into `develop-spark-3`. Does this make sense to you?

benedeki: But that (#1755) has been already merged. 😉

dk1844: Oh, I didn't notice that. In that case, you are completely correct and I will make the adjustment as suggested 👍

dk1844: Done.

benedeki (approved):
- [x] code reviewed
- [ ] pulled
- [ ] built
- [ ] run

Issue #1836, opened 2021-06-30
Standardization ignores `timezone` metadata for timestamp type in case the source is already timestamp


benedeki:
## Describe the bug
When the input of Standardization is strongly type (e.g. Parquet) the _metadata_ of `timezone` for timestamp type is ignored in the case the input is already of timestamp type.
This causes extra issues in case when `defaultTimestampTimeZone` (other as UTC) is defined in application configuration, because that *is* take into account.

## To Reproduce
1. Configure `defaultTimestampTimeZone` in `application.conf` to a non-UTC value
2. Have a source data in strongly type format (parquet) containing a field of type timestamp
2. Create a schema and dataset for the source data, add metadata `timestamp` with value `UTC` to the timestamp field
2. Run the standardization of the dataset
3. Check the data
4. The timestamp field got shifted by the difference  between `defaultTimestampTimeZone` and UTC

## Expected behaviour
`timezone` metadata are taken into account. The timestamp from the **To reproduce** paragraph is not shifted.

benedeki:
Release notes:
Time zone in metadata is taken into account even if the source data is already of timestamp type.

PR #1840, opened 2021-07-01
#1836: Standardization ignores timezone metadata for timestamp type in case the source is already timestamp (2.21)

benedeki:
* Timezone in metadata is taken into account even if the source is already timestamp or date

### DIFF
--- a/spark-jobs/src/test/scala/za/co/absa/enceladus/standardization/StandardizationParquetSuite.scala
+++ b/spark-jobs/src/test/scala/za/co/absa/enceladus/standardization/StandardizationParquetSuite.scala
@@ -443,18 +443,23 @@ class StandardizationParquetSuite extends FixtureAnyFunSuite with SparkTestBase 
       " --report-version 1 --menas-auth-keytab src/test/resources/user.keytab.example " +
       "--raw-format parquet").split(" ")
 
+    /* This might seem confusing for a quick observer. The reason why this is the correct result:
+       the source data has two timestamps 12:00:00AM and 23:00:00PM *without* time zone.
+       The metadata then signal the timestamp are to be considered in CET time zone. The data are ingested with that
+       time zone and adjusted to system time zone - UTC. Therefore they are seemingly shifted by one hour. */
     val expected =
-      """+-------------------+------+
-        ||ts                 |errCol|
-        |+-------------------+------+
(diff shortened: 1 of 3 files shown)


spark-jobs/src/test/scala/za/co/absa/enceladus/standardization/StandardizationParquetSuite.scala
```diff
@@ -436,6 +437,32 @@ class StandardizationParquetSuite extends FixtureAnyFunSuite with SparkTestBase
...
+    val args = (s"--dataset-name $datasetName --dataset-version $datasetVersion --report-date 2019-07-23" +
+      " --report-version 1 --menas-auth-keytab src/test/resources/user.keytab.example " +
+      "--raw-format parquet").split(" ")
+
+    val expected =
+      """+-------------------+------+
+        ||ts                 |errCol|
+        |+-------------------+------+
+        ||1969-12-31 23:00:00|[]    |
+        ||1969-12-31 22:00:00|[]    |
```
dk1844, line 451: I'm sorry, maybe I am missing the point here, but I will try:
It seems counter-intuitive that dataset with `(UTC midnight and CET midnight)` can end up as `(11pm CET and 10pm CET)`. 

If this case is truly correct, a bit of explanatory comment could help.

dk1844 (commented):
The main code makes sense to me.

- [x] code reviewed
- [x] pulled
- [x] built
- ran just tests

spark-jobs/src/test/scala/za/co/absa/enceladus/standardization/StandardizationParquetSuite.scala
benedeki: I see, I will add a comment. But is correct. Let me try to explain:
The timestamps are not `(UTC midnight and CET midnight)`, that's just their source. They are timestamp without timezone, one saying 12:00:00AM the other 23:00:00PM. And the _metadata_ suggests the timestamp is in CET for the conversion. Then displayed in UTC...

dk1844: Oh, I see. Yes, commenting in this fashion would help, then.

dk1844 (approved):
👍 LGTM now (just read the code)

PR #1841, opened 2021-07-01
#1836: Standardization ignores timezone metadata for timestamp type in case the source is already timestamp

benedeki:
* Timezone in metadata is taken into account even if the source is already timestamp or date

Release note:
Time zone in metadata is taken into account even if the source data is already of timestamp type.


dk1844 (approved):
- [x] code reviewed
- [x] pulled
- [x] built
- [x] run

LGTM. Thank, @benedeki, for adding the additional explanatory comments on the usecase and for support with testing.

Issue #1842, opened 2021-07-07
Properties name cut-off


benedeki:
## Describe the bug
A dataset property name longer than to fit into the give space is used - which is not very long - the rest of the name is cut off marked with an ellipsis. This doesn't help comprehension at all.

## Expected behaviour
Use more space for the label containing the property name. Consider autoresizing the window to fit the whole name,

## Screenshots
![image](https://user-images.githubusercontent.com/14905969/124733082-22bbb900-df14-11eb-9452-c3995582458e.png)

benedeki:
**Idea/suggestion:**
As resizing the parent window based on the length of the label might be complicated.
What about creating a group box with a caption _Dataset properties:_
Inside would then be properties using the whole width of the window. Furthermore if this extra space is not enough, in case the label is longer, let's put it on its own line, while the edit box and icon bellow it.
And even if that is not enough - IMHO clearly misused name - the whole name could be put into the hint when hovering over the label.

dk1844:
Release notes
Dataset properties editing UX has been enhanced - now, there is more room for the property name, a useful tooltip is provided and the dialog is resizable (Small by default).

PR #1845, opened 2021-07-08
1842 Dataset Dialog - properties editing update

dk1844:
As outlined in #1842, a longer dataset property name may have previously been cut off and the user was unable to reach the whole string.

## Solution
The dialog has undergone several changes to address the problem at hand, namely:
 - width of the value field has been slightly decreased (the values from the selection will be fully visible when opened if smaller)
 - default width of the dialog has been set to 55em (sufficient in testing), but mainly, the dialog has been made stretchable -- so if the user wants to enlarge the dialog to have more space, they can
 - more sensible tooltip has been provided - revealing the full property name string and optionally the description if available -- thus, the user can read a very long property name this way, too, if ellipsized

(I have also attempted to diminish the blank space at the label's side, but I was unsuccessful in this regard 
- with the current `ResponsiveLayout`, it can be done using `layoutData.ResponsiveFlowLayoutData.weight, but then the labels are not aligned. 
- with `ResponsiveGridLayout`, it can be done as well quite nicely with `labelSpanM` settings, but the solution on the other hand warns to clash on sync/async calls in the console. Esthetically, this solution would look the best if no issues were involved.

## Showcase
Showing a resizable dialog with enhanced hints on the DS properties:
ds-props-resized-hints-resizable
Smaller width version (when resized) is also fully usable:
ds-props-smaller-display

## Release notes suggestion
Dataset properties editing UX has been enhanced - now, there is more room for the property name, a useful tooltip is provided and the dialog is resizable.

### DIFF
--- a/menas/ui/components/dataset/addDataset.fragment.xml
+++ b/menas/ui/components/dataset/addDataset.fragment.xml
@@ -28,7 +28,7 @@
                         <Input id="newDatasetDescription" type="Text" placeholder="Description" width="100%"
                                value="{entity>/description}"/>
                         <Label text="Dataset Properties"/>
-                        <List width="100%" items="{path: 'entity>/_properties', sorter: {path: 'order'}}" class="sapUiMediumMarginBottom">
+                        <List width="100%" items="{path: 'entity>/_properties', sorter: {path: 'order'}}" class="sapUiMediumMarginBottom dsProps">
                             <InputListItem class="inputListItemSmallLabel" label="{entity>name}"
                                            tooltip="{= ${entity>description} ? ${entity>name} + ': ' + ${entity>description} : ${entity>name} }">
                                 <HBox alignItems="Center" justifyContent="End" renderType="Bare">
--- a/menas/ui/components/dataset/addDataset.fragment.xml
+++ b/menas/ui/components/dataset/addDataset.fragment.xml
@@ -16,7 +16,7 @@
 <core:FragmentDefinition xmlns="sap.m" xmlns:core="sap.ui.core" xmlns:form="sap.ui.layout.form"
                          xmlns:hdfs="components.hdfs">
     <Dialog id="addDatasetDialog" title="{entity>/title} Dataset" stretchOnPhone="true" resizable="true"
-            draggable="true" contentWidth="55em">
+            draggable="true" contentWidth="573px">
         <content>
             <VBox class="sapUiSmallMargin">
                 <form:SimpleForm adjustLabelSpan="false" editable="true">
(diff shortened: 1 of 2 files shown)


benedeki (changes requested):
Please direct the PR towards `support/2.21` branch.
And eventually another PR into `hotfix/2.22`.

benedeki:
There's definitely more space. 👍 
But I would create a box around the properties to distinguish them. And make all the edits same width and align with the other edits of the dataset (see the first two of the of the picture).
Having the labels aligned to left instead of right like of the "major" edits might stay, again as a distinguishing sign. 🤔 
image

dk1844:
> There's definitely more space. 👍
> But I would create a box around the properties to distinguish them. And make all the edits same width and align with the other edits of the dataset (see the first two of the of the picture).
...

I have added a border around the properties list, usable for Small and Medium dialog sizing. The edits are aligned on my machine, so I am surprised that it does not work for you. They should be, could you please recheck with the latest commit?

ds-border-suggestion-wider
ds-border-suggestion

benedeki (dismissed):
It needed a couple of (forced) refreshes and reloads - it now shows the edits aligned.
Thanks for the frame. I like it this way now. But please have it approved by somebody closer to users. 😉

Zejnilovic:
Not critical but, the whole of the row is a popup for the name
Jul-13-2021 13-14-10

dk1844:
> Not critical but, the whole of the row is a popup for the name
> Jul-13-2021 13-14-10

Yes, I know. However, I don't really know how to fix this, because you do not have code-access to the label object on the list. The code goes like this:
```xml
<List>
    <InputListItem tooltip="<tooltip for the row>">
        <HBox>
            <Input type="Text" /> <!-- string input, or: -->
            <Select /> <!-- choice input -->
            <Icon tooltip="<tooltip for the essentiality icon>" />
        </HBox>
    </InputListItem>
</List>
```

So I fear there is nowhere to specify that the row tooltip should actually be only assigned to the row label. The `InputListItem` takes this functionality from `Element`, so there is no necessary specialization to achieve this.

So unless we want to resort to some amazing stuff like [css tooltips](https://stackoverflow.com/questions/7117073/add-a-tooltip-to-a-div), I do not see a straightforward way to do it.

dk1844:
Based on feedback (@zakera786, @aaishaosman), the Small size of the Dataset editing dialog has been made the default view (still resizable if the user needed):
DsEditsmallSizeDefault

benedeki:
> 
> 
> Based on feedback (@zakera786, @aaishaosman), the Small size of the Dataset editing dialog has been made the default view (still resizable if the user needed):
> DsEditsmallSizeDefault

I guess, this justifies the PR: Tested label.

benedeki (approved):

Issue #1844, opened 2021-07-07
When saving a dataset with undefined recommended property show a warning dialog


benedeki:
## Background
_Recommended_ property when not defined upon save doesn't hint that it should be filled, unless there's a problem preventing saving the Dataset.

## Feature
Show a warning similar to HDFS path not existing when a _Recommneded_ property is not filled in

benedeki:
Release notes:
When saving a dataset with undefined recommended property a warning dialog is shown.

PR #1846, opened 2021-07-10
#1844: When saving a dataset with undefined recommended property show a warning dialog (2.21)

benedeki:
* Show dialogs only on submit not in case of validation on the fly
* Show property warning only if all Mandatory has passed validation

Release notes:
When saving a dataset with undefined recommended property a warning dialog is shown


Zejnilovic (approved):
- [x] code reviewed
- [x] pulled
- [x] built
- [x] run

Tested:
- all filled
- one recommended not filled
- two recommended not filled
- mandatory not filled
- optional not filled

previous steps combined with HDFS path not existing or being empty

PR #1851, opened 2021-07-15
Feature/1705 show properties missing on dataset main page

### LINKED ISSUE #1705: Show property values and missing ones on dataset's main page
## Background
Staring version v2.18.0 is possible to set dataset properties. And v2.20.0 UI supports the UI supports proprties as well.
But the properties are only shown when editing the dataset, not on the overview page.

## Feature
Show the properties defined for the dataset on the main page, not only when the dataset is edited.
Also shown defined but missing properties.

## Example 
Somewhere here. Or as a new button.
image

AdrianOlosutean:
Closes #1705 

Showing all properties and if absent, "Missing" will be written in yellow if it's Recommended, blue if Optional or red if Mandatory

Screenshot 2021-07-15 at 18 27 10

### DIFF
--- a/menas/ui/components/dataset/datasetDetail.controller.js
+++ b/menas/ui/components/dataset/datasetDetail.controller.js
@@ -374,7 +374,7 @@ sap.ui.define([
         currentDataset.propertiesDisplay = [];
 
         for(let property of aPropTemplates) {
-          let propValue = 'Missing';
+          let propValue = null;
           if(currentDataset.properties && currentDataset.properties[property.name]) {
             propValue = currentDataset.properties[property.name];
           }
--- a/menas/ui/components/dataset/datasetDetail.controller.js
+++ b/menas/ui/components/dataset/datasetDetail.controller.js
@@ -375,7 +375,7 @@ sap.ui.define([
 
         for(let property of aPropTemplates) {
           let propValue = 'Missing';
-          if(currentDataset.properties) {
+          if(currentDataset.properties && currentDataset.properties[property.name]) {
             propValue = currentDataset.properties[property.name];
           }
           currentDataset.propertiesDisplay.push({
--- a/menas/ui/components/dataset/datasetDetail.controller.js
+++ b/menas/ui/components/dataset/datasetDetail.controller.js
@@ -370,15 +370,19 @@ sap.ui.define([
 
       //transform properties into a bindable format
       const aPropTemplates = this._model.getProperty(`/properties`) || [];
-      if(currentDataset && currentDataset.properties) {
+      if(currentDataset) {
         currentDataset.propertiesDisplay = [];
-        for(let sName in currentDataset.properties) {
-          const oPropTemplMatch = aPropTemplates.find((oTempl) => { return oTempl.name === sName });
-          const iOrder = oPropTemplMatch ? oPropTemplMatch.order : Infinity;
+
+        for(let property of aPropTemplates) {
+          let propValue = 'Missing';
+          if(currentDataset.properties) {
+            propValue = currentDataset.properties[property.name];
+          }
           currentDataset.propertiesDisplay.push({
-            name: sName,
-            value: currentDataset.properties[sName],
-            order: iOrder
+            name: property.name,
+            value: propValue,
+            order: property.order,
+            essentiality: property.essentiality._t
           });
         }
       }
--- a/menas/ui/components/dataset/datasetInfo.fragment.xml
+++ b/menas/ui/components/dataset/datasetInfo.fragment.xml
@@ -39,9 +39,9 @@
       </f:FormContainer>
       <f:FormContainer formElements="{path: '/currentDataset/propertiesDisplay', sorter: {path:'order'}}">
         <f:FormElement label="{name}">
-            <ObjectStatus text="{parts: [{path: 'value'}],
-                                formatter: '.formatter.italicMissingProp'}"
-                          state="{= ${value} !== 'Missing' ? undefined : ${essentiality} === 'Optional' ? 'Information' : ${essentiality} === 'Recommended' ? 'Warning' : ${essentiality} === 'Mandatory' ? 'Error' : undefined }"/>
+            <ObjectStatus text="{parts: [{path: 'value'}], formatter: 'Formatters.italicMissingProp'}"
+                          state="{= ${value} !== 'Missing' ? undefined : ${essentiality} === 'Optional' ? 'Information' : ${essentiality} === 'Recommended' ? 'Warning' : ${essentiality} === 'Mandatory' ? 'Error' : undefined }">
+            </ObjectStatus>
         </f:FormElement>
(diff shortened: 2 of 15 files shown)


menas/ui/components/dataset/datasetDetail.controller.js
```diff
@@ -370,15 +370,19 @@ sap.ui.define([
...
+            propValue = currentDataset.properties[property.name];
+          }
           currentDataset.propertiesDisplay.push({
-            name: sName,
-            value: currentDataset.properties[sName],
-            order: iOrder
+            name: property.name,
+            value: propValue,
+            order: property.order,
+            essentiality: property.essentiality._t
```
dk1844, line 385: I would maybe personally prefer a more functional approach, for Javascript, it's fine, I guess. It seems correct.

menas/ui/components/dataset/datasetInfo.fragment.xml
```diff
@@ -39,7 +39,8 @@
       </f:FormContainer>
       <f:FormContainer formElements="{path: '/currentDataset/propertiesDisplay', sorter: {path:'order'}}">
         <f:FormElement label="{name}">
-          <Text text="{value}" />
+            <ObjectStatus text="{value}"
+                          state="{= ${value} !== 'Missing' ? undefined : ${essentiality} === 'Optional' ? 'Information' : ${essentiality} === 'Recommended' ? 'Warning' : ${essentiality} === 'Mandatory' ? 'Error' : undefined }"/>
```
dk1844, line 43: 👍 for using SAP UI's built-in states, that is smart 😎

dk1844 (dismissed):
- [x] code reviewed
- [x] pulled
- [x] built (also with `-Pintegration`)
- [x] run

LGTM, shows as expected

missing-properties

menas/ui/css/style.css
```diff
@@ -127,3 +127,15 @@ h5[id$=datasetDetailView--scheduleTimingTitle] {
...
+  font-style: italic;
+}
+
+.sapMObjStatusWarning{
+  font-style: italic;
+}
+
+.sapMObjStatusInformation{
+  font-style: italic;
+}
```
dk1844, line 141: This could be simplified to
```css
.sapMObjStatusError, .sapMObjStatusWarning, .sapMObjStatusInformation{
  font-style: italic;
}
```

But even so, I don't think that it is ideal, because you could be inadvertently modifying style globally - it can have an adverse effect somewhere else by changing the style of these universal classes (which I believe you don't intend to do). 

Maybe it would be more fitting to find another way to introduce the change to the properties in dataset view only, of the top of my head, the options might be:
 - if you can, wrap the missing labels with a custom class (is this even possible with `FormElement` or `ObjectStatus`?) and define the style there, e.g. 
```css
.missingObjectStatus {
  font-style: italic;
}```

 - define a wrapper class higher in the form-container structure and limit the effect, as in
```css
.myDsView .sapMObjStatusError, .myDsView .sapMObjStatusWarning, .myDsView .sapMObjStatusInformation{
  font-style: italic;
}
```
  - if you had trouble doing this (too), perhaps using one of the generated ids (`__navigation0---datasetDetailView--info--Grid`?,  `__navigation0---datasetDetailView--info`?) could work as well, in the worst case scenario

AdrianOlosutean: I would try the second option

menas/ui/css/style.css
```diff
@@ -127,3 +127,7 @@ h5[id$=datasetDetailView--scheduleTimingTitle] {
   padding: 8px 2px;
   margin-top: 4px;
 }
+
+.italic{
```
benedeki, line 131: I would not name the class _how it is_, but _what it is_. E.g. `missingPropertyValue`

dk1844 (dismissed):
- [x] code reviewed
- [x] pulled
- [x] built
- [x] tested locally 

LGTM:
missing-ds-property-italics

dk1844:
Btw. @AdrianOlosutean, were there problems taking a declarative approach for the class?

AdrianOlosutean:
If you mean the second approach you suggested, yes. I was not able to make it work by applying the class higher up in xml, it just didn't appear

> Btw. @AdrianOlosutean, were there problems taking a declarative approach for the class?

menas/ui/css/style.css
AdrianOlosutean: There is already the `italicMissingProp` formatter, so I would consider leaving that class `italic` in case it would be reused for other purposes

benedeki: That's the thing. I am not sure it should be reused as it is. Such a class is not different from assigning a style directly. Not very modular.

menas/ui/css/style.css
```diff
@@ -127,3 +127,7 @@ h5[id$=datasetDetailView--scheduleTimingTitle] {
...
+  font-style: italic;
+}
```
benedeki, line 133: And I have a side reason for above 😉 

```suggestion
}

div.sapMObjStatusInformation.missingPropertyValue span.sapMObjStatusText{
  color: silver;
}

To get "rid of" the blue.
```

benedeki (commented):
What is weird, this works for me only in Chrome, no other browser.
Not even Opera, which is Chromium based for sure. Other tried browsers Waterfox (Firefox clone) and Edge.
Properties are there but the word _Missing_ in any styling not. Asked other to double-check.

benedeki:
> 
> 
> What is weird, this works for me only in Chrome, no other browser.
> Not even Opera, which is Chromium based for sure. Other tried browsers Waterfox (Firefox clone) and Edge.
> Properties are there but the word _Missing_ in any styling not. Asked other to double-check.

Opera: 77.0.4054.254 (freshest at the moment)
Waterfox: 

> 
> 
> What is weird, this works for me only in Chrome, no other browser.
> Not even Opera, which is Chromium based for sure. Other tried browsers Waterfox (Firefox clone) and Edge.
> Properties are there but the word _Missing_ in any styling not. Asked other to double-check.

Opera: 77.0.4054.254 (freshest at the moment)
Waterfox: G3.2.4.1 based on Gecko 8.12.0.1 (freshest at the moment)
Edge: 91.0.864.71 (freshest at the moment)

benedeki:
Adding `PR: Tested` back. Saw it working on other browsers too, so probably a local issue (suspected so somewhat)

menas/ui/generic/formatters.js
```diff
@@ -130,4 +130,11 @@ var Formatters = new function() {
     }
   };
 
+  this.italicMissingProp = function (sStatus) {
+    if (sStatus === 'Missing') {
```
benedeki, line 134: This is actually marks the value `Missing` as _Missing_. So it's broken actually,

benedeki (changes requested):

benedeki:
Tested in all four mentioned browsers. Now works perfectly.

PR #1852, opened 2021-07-16
1842 Dataset Dialog - properties editing update

dk1844:
_This is the same as https://github.com/AbsaOSS/enceladus/pull/1845 for Enceladus develop_

- as stated in https://github.com/AbsaOSS/enceladus/pull/1845, Dataset edit dialog has been updated:
  - layout tweaked, resizable (default: Small), enhanced tooltips, properties border 

DsEditsmallSizeDefault2withToolTip

## Test-run
I have test-ran it on local Menas and noticed no change from the original in #1845.

Relates to #1842.


dk1844:
Not needed, because 2.22.2 will be merged to master/develop with all it contains in the end.

PR #1859, opened 2021-07-22
gitflow maven plugin update 1.16.0

dk1844:
The build plugin `com.amashchenko.maven.plugin:gitflow-maven-plugin` has been updated to latest version as-of today, i.e `1.16.0`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -136,6 +136,7 @@
         <maven.sources.version>3.0.1</maven.sources.version>
         <maven.surefire.version>2.18.1</maven.surefire.version>
         <maven.war.version>2.2</maven.war.version>
+        <gitflow.maven.plugin.version>1.16.0</gitflow.maven.plugin.version>
         <!--dependency versions-->
         <abris.version>3.1.1</abris.version>
         <atum.version>3.3.0</atum.version>
@@ -416,7 +417,7 @@
             <plugin>
                 <groupId>com.amashchenko.maven.plugin</groupId>
                 <artifactId>gitflow-maven-plugin</artifactId>
-                <version>1.12.0</version>
+                <version>${gitflow.maven.plugin.version}</version>
                 <configuration>
                     <installProject>true</installProject>
                     <verbose>true</verbose>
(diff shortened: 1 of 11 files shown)


Zejnilovic (dismissed):

pom.xml
```diff
@@ -136,6 +136,7 @@
         <maven.sources.version>3.0.1</maven.sources.version>
         <maven.surefire.version>2.18.1</maven.surefire.version>
         <maven.war.version>2.2</maven.war.version>
+        <gitflow.maven.plugin.version>1.16.0</gitflow.maven.plugin.version>
```
benedeki, line 139:
```suggestion
        <maven.gitflow,plugin.version>1.16.0</maven.gitflow.plugin.version>
```
And they are actually alphabetically ordered. 

Bonus point if you change `<dockerfile.maven.version>` too. 👼

pom.xml
```diff
@@ -416,7 +417,7 @@
             <plugin>
                 <groupId>com.amashchenko.maven.plugin</groupId>
                 <artifactId>gitflow-maven-plugin</artifactId>
-                <version>1.12.0</version>
+                <version>${gitflow.maven.plugin.version}</version>
```
benedeki, line 420:
```suggestion
                <version>${maven.gitflow.plugin.version}</version>
```

benedeki (changes requested):

benedeki:
Just a support plugin upgrade, no need to test extra.

pom.xml
benedeki: It's actually correct by the plugin name.

benedeki (approved):

AdrianOlosutean (dismissed):

pom.xml
dk1844: If I am offer an alternative, I have committed a change in the following fashion:
The version property names 
 - now follow the logic of `plugin-artifact-name` -> `plugin.artifact.name.version`
 - are now alphabetically ordered
 - and the unused properties are removed (`maven.compiler.source`, `maven.compiler.target`)

How about that, @benedeki?

benedeki (approved):
👍

Issue #1887, opened 2021-08-19
`defaultTimestampTimeZone` can be source type specific


benedeki:
## Background
When a _timestamp_ is read without having time zone specification - either in data or in metadata - the system setting of `defaultTimestampTimeZone` is used. For various source types this setting might cause

## Feature
One global setting of `defaultTimestampTimeZone` can cause confusion and unwanted timestamp shifts for certain. For example _parquet_ timestamps are generally normalized to UTC. This can be addressed by metadata configuration of the schema, but it's tedious.
Implement `defaultTimestampTimeZone`  specific per source type, which would take precedence to the global `defaultTimestampTimeZone`. The global would be used as a fallback, similarly as system time zone is a fallback in case of the global setting missing.

## Additional
Update the documentation with this configuration option.

## Proposed Solution
Solution Ideas 
1. Upon startup, have `Defaults` implicit be based on source type
2. First check for `defaultTimestampTimeZone.`[source_type], if not found use the existing logic.

benedeki:
Release notes:
Default time zone used during Standardization can now be overridden/configured for each raw format type. The configuration name is then `defaultTimestampTimeZone.[format]` and takes precedence over the global setting of `enceladus.defaultTimestampTimeZone.default`. (`defaultTimestampTimeZone` was made obsolete, but still supported as a fallback)

PR #1916, opened 2021-09-15
#1887 defaultTimestampTimeZone can be source type specific

benedeki:
* rename of the configuration prefix from `enceladus.` to `standardization.`

Just a little change in the new configuration name. Believe it doesn't need retesting, that the UTs are enough.


AdrianOlosutean (approved):

dk1844 (approved):
LGTM (just read the code)

Let's just reflect the change in RNs - https://github.com/AbsaOSS/enceladus/issues/1887#issuecomment-920018540

benedeki:
> 
> Let's just reflect the change in RNs - [#1887 (comment)](https://github.com/AbsaOSS/enceladus/issues/1887#issuecomment-920018540)

Thanks for the reminder 👍 . Done.

PR #1924, opened 2021-09-20
#1893 autoclean on s3

### LINKED ISSUE #1893: `conformance.autoclean.standardized.hdfs.folder=true` does work properly on S3
## Describe the bug
When using S3 as the intermediate storage between _Standardization_ and _Conformance_ then the autocleaning doesn't work perfectly leaving some files behind.

## To Reproduce
1. Set `standardized.hdfs.path` to a an S3 path
2. Set `conformance.autoclean.standardized.hdfs.folder=true`
3. Run _Standardization_ followed by _Conformance_
4. The subfolder specific to the job in the folder specified by `standardized.hdfs.path` have some leftover files

## Expected behavior
All files are deleted.

AdrianOlosutean:
Closes #1893

### DIFF
--- a/_docs/2.0.0/usage/config.md
+++ b/_docs/2.0.0/usage/config.md
@@ -30,5 +30,12 @@ spark-submit --conf "spark.driver.extraJavaOptions= -Dkey1=value1 -Dkey2=value2"
 
 <!-- specific sections on Standardization & Conformance options may follow in the future -->
 
+Note, that
+  - `conformance.autoclean.standardized.hdfs.folder` when set to true and the job is writing to S3, there will be a leftover empty file like `conformance-output_$folder$` after the autoclean.
+  This, however, will not negatively impact the functionality of other jobs and is due to the EMR comitter. The recommendation is that the same `standardized.hdfs.path` value to be set for
+   all jobs performing autoclean so that there won't be more of these files. 
+  If this is not the case, a bucket lifecycle policy can be set so that all files ending in `$folder$` be deleted regularly
+  
+
 [readme]: https://github.com/AbsaOSS/enceladus/blob/master/README.md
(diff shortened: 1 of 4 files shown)


_docs/2.0.0/usage/config.md
```diff
@@ -30,5 +30,12 @@ spark-submit --conf "spark.driver.extraJavaOptions= -Dkey1=value1 -Dkey2=value2"
 
 <!-- specific sections on Standardization & Conformance options may follow in the future -->
 
+Note, that
+  - `conformance.autoclean.standardized.hdfs.folder` when set to true and the job is writing to S3, there will be a leftover empty file like `conformance-output_$folder$` after the autoclean.
```
benedeki, line 34: I would like to have this note linked from the `conformance.autoclean.standardized.hdfs.folder`, similarly as you can see in https://absaoss.github.io/enceladus/2021/02/12/release-2.18.1.html

_docs/2.0.0/usage/config.md
```diff
@@ -30,5 +30,12 @@ spark-submit --conf "spark.driver.extraJavaOptions= -Dkey1=value1 -Dkey2=value2"
...
+  This, however, will not negatively impact the functionality of other jobs and is due to the EMR comitter. The recommendation is that the same `standardized.hdfs.path` value to be set for
```
benedeki, line 35:
```suggestion
  This, however, will not negatively impact the functionality of other jobs even when using the same path and is due to the EMR comitter.
```
Here I would end the _Note_, the rest is S3/AWS thing not Enceladus.

_docs/2.0.0/usage/config.md
benedeki:
```suggestion
  - `conformance.autoclean.standardized.hdfs.folder` when set to true and the job is writing to S3, there could be a leftover empty file like `conformance-output_$folder$` after the autoclean.
```
The file is put there only in case the folder would be all empty. If there are other folders, it won't be created.

PR #1929, opened 2021-09-22
Feature/1863 mapping table filtering

dk1844:
_In progress._

 - The mapping table UI contains means for viewing and editing MT filter in a visual way.
   - validation is included, it is now based on filter tree model extension


dk1844:
WIP = column type to be derived from MT schema; columns available to be derived from MT schema, too. Otherwise testable

PR #1933, opened 2021-09-24
1898 Conformed columns for enceladus_info_version

### LINKED ISSUE #1898: Info Version from conformed columns
## Background
Currently the Info Version  can only be taken from columns prior to conformance, which is a bit restrictive

## Feature
Add the ability to specify conformed columns as columns for `enceladus_info_version`

AdrianOlosutean:
Closes #1898 
I decided to only include `enceladus_info_version` in InfoVersionFactory since the logic is simple enough to allow conformed columns, but it's much more complex for InfoDateFactory since the type needs to be known. 

If my proposal is accepted, I will create another issue for InfoDateFactory to accept conformed columns

I kept `ConformedSchema.create` as a start


data-model/src/main/scala/za/co/absa/enceladus/model/ConformedSchema.scala
```diff
@@ -0,0 +1,46 @@
...
+
+case class ConformedSchema(schema: List[StructField], dataset: Dataset) {
+  def hasField(field: String): Boolean = {
+    if (schema.exists(_.name == field)) true else {
+      val ss = dataset.conformance.find {
+        case MappingConformanceRule(_, _, _, _, _, _, outputColumn, additionalColumns, _, _, _) =>
+          outputColumn == field || additionalColumns.getOrElse(Map()).contains(field)
+        case SingleColumnConformanceRule(_, _, outputColumn, _, inputColumnAlias) =>
+          outputColumn == field || field == outputColumn + "." + inputColumnAlias
+        case DropConformanceRule(_, _, _) => false
```
kevinwallimann, line 29: This line is redundant, right? Found it a bit confusing

kevinwallimann:
Tested OK
- Existing use case with hard-coded report version (`report.version=1`)
- Existing use case with report version from column (`report.version.column=height`)
- Report version from conformed column (CoalesceConformanceRule)

- On Laptop, Spark 2.4.3, Hyperdrive 4.4.1

data-model/src/main/scala/za/co/absa/enceladus/model/ConformedSchema.scala
AdrianOlosutean: The point is that DropConformanceRule will be ignored initally at this point. Then if the field is found in a regular rule, a drop one will be looked for later on in order to establish if the final schema has the column or not

kevinwallimann: Ah I see, I didn't realize that `DropConformanceRule` is a subtype of `ConformanceRule`. Makes sense now

kevinwallimann (approved):

PR #1946, opened 2021-10-13
1945 Mapping Rule on s3

### LINKED ISSUE #1945: Make mapping rule work with S3
## Background
A clear and concise description of where the limitation lies.

## Feature
A description of the requested feature.

## Example [Optional]
A simple example if applicable.

Solution Ideas:
1. 
2. 
3.

AdrianOlosutean:
Closes #1945


spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/interpreter/DynamicInterpreter.scala
```diff
@@ -309,7 +308,10 @@ case class DynamicInterpreter(implicit inputFs: FileSystem) {
     val mappingTableDef = ictx.dao.getMappingTable(rule.mappingTable, rule.mappingTableVersion)
     val mappingTablePath = PartitioningUtils.getPartitionedPathName(mappingTableDef.hdfsPath,
       ictx.progArgs.reportDate)
-    val mappingTableSize = HadoopFsUtils.getOrCreate(inputFs).getDirectorySizeNoHidden(mappingTablePath)
+    //accommodate different fs for the mapping table or different bucket
```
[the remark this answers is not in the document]
benedeki: Yes, it can go. 👍

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/interpreter/DynamicInterpreter.scala
```diff
@@ -309,7 +308,10 @@ case class DynamicInterpreter(implicit inputFs: FileSystem) {
...
+    val mappingTableFs = PathWithFs.fromPath(mappingTablePath)(ictx.spark.sparkContext.hadoopConfiguration)
+
+    val mappingTableSize = HadoopFsUtils.getOrCreate(mappingTableFs.fileSystem).getDirectorySizeNoHidden(mappingTableFs.path)
```
benedeki, line 314: I would consider making this a function in the `HadoopFSUtils` object, only taking in the path as a `String` and `implcit HadoopConfiguration`.

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/interpreter/DynamicInterpreter.scala
```diff
@@ -309,7 +308,10 @@ case class DynamicInterpreter(implicit inputFs: FileSystem) {
...
     (mappingTableSize / (1024 * 1024)).toInt
```
benedeki, line 315:
```suggestion
    val mb = (mappingTableSize / (1024 * 1024)).toInt
    log.debug(s"$mappingTablePath size: ${mb}MB")
    mb
```

benedeki (dismissed):
- [x] code reviewed
- [x] pulled
- [x] built
- [x] run
Tested on (local)HDFS, there it works, so no regression. Needs test on S3 too.

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/interpreter/DynamicInterpreter.scala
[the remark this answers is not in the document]
AdrianOlosutean: Probably right now it cannot be removed since it's used by Atum

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/interpreter/DynamicInterpreter.scala
AdrianOlosutean: I think something like this would indeed make sense on a larger refactoring, but this way we would need to pass the path 2 times: once when instantiating the object and then when calling `getDirectorySizeNoHidden`. Is it good enough for now?

benedeki: Not sure, I understand the double passing of path. But sure enough good enough now (as the approval suggests)

benedeki:
Tested on S3 (thanks @dk1844)

benedeki (approved):

PR #1950, opened 2021-10-18
1919 Enhancement for data migration

### LINKED ISSUE #1919: Add new features to Enceladus needed for data migration
## Background
We will be migrating Enceladus (Menas) data between two location, while both will be used at the same time. This might require extra capabilities of the project to support this migration.

## Feature
Add identified features required or useful for successfully exercising and concluding the migration. Features are yet to be identified and described.

## Proposed Features
1. Locking a dataset/any Menas entity, including data model change and API
2. ???

AdrianOlosutean:
Closes #1919


dk1844 (dismissed):
LGTM, just read the code.

(only suggesting the two simple test cases below)

data-model/src/main/scala/za/co/absa/enceladus/model/Dataset.scala
```diff
@@ -43,6 +43,8 @@ case class Dataset(name: String,
                    disabled: Boolean = false,
                    dateDisabled: Option[ZonedDateTime] = None,
                    userDisabled: Option[String] = None,
+
+                   modifiable: Boolean = true,
```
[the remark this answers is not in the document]
AdrianOlosutean: Disabled is Boolean, why not this one?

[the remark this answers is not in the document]
Zejnilovic: Well, it is a new field. Will the DB be compatible? Will there need to be a migration?

AdrianOlosutean:
I will need to figure out how to test this

Zejnilovic (commented):
Is Locked in the update method or export/import?

menas/src/main/scala/za/co/absa/enceladus/menas/services/VersionedModelService.scala
```diff
@@ -224,8 +224,8 @@ abstract class VersionedModelService[C <: VersionedModel with Product with Audit
         Future.failed(NotFoundException(s"Version $itemVersion of $itemName not found"))
       } else if (versionToUpdate.get.version != itemVersion) {
         Future.failed(ValidationException(Validation().withError("version", s"Version $itemVersion of $itemName is not the latest version, therefore cannot be edited")))
-      } else if (versionToUpdate.get.locked) {
-        Future.failed(new NotAllowedException("Entity is locked"))
+      } else if (versionToUpdate.get.locked.getOrElse(false)) {
+        Future.failed(LockedEntityException(s"Entity $itemName is locked"))
```
dk1844, line 228: Nicely done to bring your own, semantically explanatory, exception 👍

data-model/src/main/scala/za/co/absa/enceladus/model/MappingTable.scala
```diff
@@ -43,7 +43,7 @@ case class MappingTable(name: String,
                         dateDisabled: Option[ZonedDateTime] = None,
                         userDisabled: Option[String] = None,
 
-                        locked: Boolean = false,
+                        locked: Option[Boolean] = None,
```
dk1844, line 46: Using the option wrapper makes sense to differentiate the unset situation by `None`. However, what I find less ideal is the fact, that now you have to default-to-false on empty and you are now doing it in many places.

Some places call `locked.getOrElse(false)`, other places `forall(_.locked.contains(true)` etc. - so the defaulting to `false` is spread around (and god-forbid you ever had to change it and forgot one place).

I think it would be better to do the defaulting once and for all - e.g. by providing:
```scala
class Dataset {
  def isLocked: Boolean = locked.getOrElse(false)
}
```

That way, all other calls would never have to default themselves.

dk1844 (dismissed):
Aside from the locked-value-defaulting, LGTM (just read the code)

data-model/src/main/scala/za/co/absa/enceladus/model/MappingTable.scala
AdrianOlosutean: This does make sense, however if I add that method to the `VersionedModel`, it looks like serializing and deserializing doesn't work the same anymore and on every update the value is added as false, which I'm not sure if it's what we want.

dk1844: Perhaps the method could be excepted from being serialized (some annotation perhaps?)

dk1844: So, I have tried what I suggested and, truly, integration tests fail on SerDe-related errors. The nature of the errors, however, suggests that the method `isLocked` is used as a getter for the SerDe processing and that's why the tests fail.

I have also tried the same, just with the method bearing a different name (e.g. `lockedWithDefault`) and the problem disappeared. So, I believe the idea can be used, just rename the method :D

AdrianOlosutean: Tried `lockedWithDefault`

dk1844 (approved):
LGTM (just read the code this time)

data-model/src/main/scala/za/co/absa/enceladus/model/MappingTable.scala
benedeki: I must admit, I am not really fond of the function name `lockedWithDefault`. I don't have much better idea though.
`lockedDefined`?
Or a way I remember from Delphi, where the field has an _F_ prefix and getter/setter was without it. What do you think about this style:
```scala
    val fLocked: Option[Boolean]
    def Locked: Boolean = fLocked.getOrElse(false)
```

data-model/src/main/scala/za/co/absa/enceladus/model/versionedModel/VersionedModel.scala
```diff
@@ -35,6 +35,8 @@ trait VersionedModel extends Exportable {
   val dateDisabled: Option[ZonedDateTime]
   val userDisabled: Option[String]
 
+  val locked: Option[Boolean]
+
```
benedeki, line 39: **IMHO important:** I would add `userLocked` and `dateLocked` too.

benedeki (commented):
It would be nice to have a visual indication of locking as well. But of course not in this PR. 😉

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/VersionedModelController.scala
```diff
@@ -158,4 +158,17 @@ abstract class VersionedModelController[C <: VersionedModel with Product with Au
...
+  @PutMapping(Array("/lock/{name}"))
+  @ResponseStatus(HttpStatus.OK)
+  def lock(@PathVariable name: String): CompletableFuture[UpdateResult] = {
+    versionedModelService.setLock(name, isLocked = true)
+  }
+
+  @PutMapping(Array("/unlock/{name}"))
+  @ResponseStatus(HttpStatus.OK)
+  def unlock(@PathVariable name: String): CompletableFuture[UpdateResult] = {
+    versionedModelService.setLock(name, isLocked = false)
```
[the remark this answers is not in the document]
dk1844: This is a very good point. Given the fact how we use versioning of entities, it probably should increase the version. 🙇

data-model/src/main/scala/za/co/absa/enceladus/model/versionedModel/VersionedModel.scala
dk1844: Why, could you elaborate, please?

data-model/src/main/scala/za/co/absa/enceladus/model/MappingTable.scala
benedeki: Yeah, the f-prefix has been rejected on the CQC meeting, so it's of the table. 😉 
And I can live with `lockedWithDefault`, so up to @AdrianOlosutean if he keeps it or replaces it 😉

data-model/src/main/scala/za/co/absa/enceladus/model/versionedModel/VersionedModel.scala
benedeki: Locking does not change the version. Would be good to know who and when did it.

menas/src/main/scala/za/co/absa/enceladus/menas/repositories/VersionedMongoRepository.scala
```diff
@@ -132,9 +132,17 @@ abstract class VersionedMongoRepository[C <: VersionedModel](mongoDb: MongoDatab
...
+    } else {
+      (null, null)
+    }
+
     collection.updateMany(getNameFilter(name), combine(
-      set("locked", isLocked))).toFuture()
+      set("locked", isLocked),
+      set("dateLocked", additionalFields._1),
+      set("userLocked", additionalFields._2))).toFuture()
   }
```
dk1844, line 146:
```suggestion
 def setLockState(name: String, isLocked: Boolean, username: String, datetime: ZonedDateTime = ZonedDateTime.now()): Future[UpdateResult] = {
    val (dateLocked, userLocked) = if (isLocked) {
      (datetime, username)
    } else {
      (null, null)
    }

    collection.updateMany(getNameFilter(name), combine(
      set("locked", isLocked),
      set("dateLocked", dateLocked),
      set("userLocked", userLocked))).toFuture()
  }
```

How about making the now-datetime overridable from param (easier testing) and directly naming the tuple items?

dk1844 (dismissed):
LGTM. (just read the code)

AdrianOlosutean:
Question: Should locking/unlocking be done only by admins?

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/VersionedModelController.scala
```diff
@@ -158,4 +158,17 @@ abstract class VersionedModelController[C <: VersionedModel with Product with Au
     versionedModelService.disableVersion(name, v)
   }
 
+  @PutMapping(Array("/lock/{name}"))
+  @ResponseStatus(HttpStatus.OK)
+  def lock(@PathVariable name: String): CompletableFuture[UpdateResult] = {
```
benedeki, line 163: Please make it limited to admin roles only. As the property definitions are in `PropertyDefinitionController`
```scala
  @PreAuthorize("@authConstants.hasAdminRole(authentication)")
```

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/VersionedModelController.scala
```diff
@@ -158,4 +158,17 @@ abstract class VersionedModelController[C <: VersionedModel with Product with Au
...
+    versionedModelService.setLock(name, isLocked = true)
+  }
+
+  @PutMapping(Array("/unlock/{name}"))
+  @ResponseStatus(HttpStatus.OK)
+  def unlock(@PathVariable name: String): CompletableFuture[UpdateResult] = {
```
benedeki, line 169: Same as above:
```scala
  @PreAuthorize("@authConstants.hasAdminRole(authentication)")
```

Bonus! I would consider adding this to the `delete` too.

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/VersionedModelController.scala
```diff
@@ -158,4 +158,17 @@ abstract class VersionedModelController[C <: VersionedModel with Product with Au
     versionedModelService.disableVersion(name, v)
   }
 
+  @PutMapping(Array("/lock/{name}"))
+  @ResponseStatus(HttpStatus.OK)
+  def lock(@PathVariable name: String): CompletableFuture[UpdateResult] = {
+    versionedModelService.setLock(name, isLocked = true)
```
benedeki, line 164: Why the `setLock` method has to figure out the authenticated user itself and it's not coming from API itself (`@AuthenticationPrincipal principal: UserDetails`), like in most other places (well except `delete`)?

menas/src/main/scala/za/co/absa/enceladus/menas/repositories/VersionedMongoRepository.scala
```diff
@@ -131,6 +131,18 @@ abstract class VersionedMongoRepository[C <: VersionedModel](mongoDb: MongoDatab
...
       set("userDisabled", username))).toFuture()
   }
+ def setLockState(name: String, isLocked: Boolean, username: String, datetime: ZonedDateTime = ZonedDateTime.now()): Future[UpdateResult] = {
+    val (dateLocked, userLocked) = if (isLocked) {
+      (datetime, username)
+    } else {
+      (null, null)
+    }
+
+    collection.updateMany(getNameFilter(name), combine(
```
benedeki, line 141: I think it updates all versions of the entity. But on second thought, it's probably not a problem. 🤔

AdrianOlosutean: Yes, I remember this was the purpose

menas/src/main/scala/za/co/absa/enceladus/menas/repositories/VersionedMongoRepository.scala
```diff
@@ -24,7 +24,7 @@ import org.mongodb.scala.bson.conversions.Bson
 import org.mongodb.scala.model.Aggregates._
 import org.mongodb.scala.model.Filters._
 import org.mongodb.scala.model.Projections._
-import org.mongodb.scala.model.Updates._
+import org.mongodb.scala.model.Updates.{set, _}
```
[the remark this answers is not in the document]
AdrianOlosutean: Yes

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/VersionedModelController.scala
AdrianOlosutean: My reason for asking this was also because disable now is not limited to admin. Should it be the case as well?

benedeki: Admin came much later and was not considered/forgotten for delete. And see the bonus note bellow 😉

benedeki:
> Question: Should locking/unlocking be done only by admins?

Sorry, missed the question. And based on the comments, the answer is obviously "yes". 😄

menas/src/main/scala/za/co/absa/enceladus/menas/controllers/VersionedModelController.scala
```diff
@@ -159,15 +161,19 @@ abstract class VersionedModelController[C <: VersionedModel with Product with Au
   }
 
   @PutMapping(Array("/lock/{name}"))
+  @PreAuthorize("@authConstants.hasAdminRole(authentication)")
   @ResponseStatus(HttpStatus.OK)
-  def lock(@PathVariable name: String): CompletableFuture[UpdateResult] = {
-    versionedModelService.setLock(name, isLocked = true)
+  def lock(@PathVariable name: String,
```
benedeki, line 166: In all other API cases (except one) the @AuthenticationPrincipal is the first param. I would keep it consistent.

benedeki (dismissed):

menas/src/main/scala/za/co/absa/enceladus/menas/repositories/VersionedMongoRepository.scala
```diff
@@ -132,6 +132,20 @@ abstract class VersionedMongoRepository[C <: VersionedModel](mongoDb: MongoDatab
       set("userDisabled", username))).toFuture()
   }
 
+  def setLockState(name: String, isLocked: Boolean, username: String,
```
Zejnilovic, line 135: Is this what we want? Like it is probably okay but not really helpful 
```json
{
	"matchedCount": 2,
	"modifiedCount": 2,
	"upsertedId": null,
	"modifiedCountAvailable": true
}
```

Zejnilovic, line 135: Not a fan of methods with boolean params, when it could be nicely 2 method `lock` and `unlock`, but I am not sure of our stance of that as a group.

Zejnilovic (approved):
Code reviewed

PR #1956, opened 2021-10-22
1955 run_enceladus can now run --help

Zejnilovic:


benedeki:
Agreed to not support running `run_enceladus.sh` directly.

PR #1962, opened 2021-10-26
1959 Add local and S3 keytabs to --files in spark-submit

### LINKED ISSUE #1959: Helper scripts evaluate where from is keytab and add it to files
## Background
Current helper scripts just add keytab to the spark-submit. But with cloud and S3 it becomes a bit more complicated

## Feature
Helper scripts evaluate if a script starts with S3:// or is on local FS and adds the path to the `--files`.

```shell

add_keytab_to_files() {
    MENAS_AUTH_KEYTAB_NAME = echo "${MENAS_AUTH_KEYTAB}" | grep -o '[^/]*$'
    FILES="${FILES},${MENAS_AUTH_KEYTAB}#${MENAS_AUTH_KEYTAB_NAME}"
    MENAS_AUTH_KEYTAB = "${MENAS_AUTH_KEYTAB_NAME}"
}

if [[ "${MENAS_AUTH_KEYTAB}" =~ "^(s|S)3://.*" ]]; then
  echo "Using Keytab from S3"
  add_keytab_to_files
elif [[ -f "${MENAS_AUTH_KEYTAB}" ]]; then
  echo "Using Keytab from local FS"
  add_keytab_to_files
else 
  echo "Using Keytab from HDFS"
fi
```

Zejnilovic:
Closes #1959

RN: Helper scripts now evaluate if a keytab is on S3 (prepended path with `s3://`), on local FS (tested using -F) or on HDFS (no test, defaults to this). If the keytab is on S3 or local FS, then it is added to the `--files` in spark-submit so it is distributed to all the executors.


AdrianOlosutean (approved):
Looks good, no test

scripts/bash/run_enceladus.sh
```diff
@@ -412,6 +412,12 @@ get_temp_log_file() {
     mktemp -p "$LOG_DIR" -t "$TEMPLATE"
 }
 
+add_keytab_to_files() {
+    MENAS_AUTH_KEYTAB_NAME = echo "${MENAS_AUTH_KEYTAB}" | grep -o '[^/]*$'
```
benedeki, line 416: What this `grep` supposed to do? Seems to me, it takes the tail of the string ignoring front slashes. 🤔

benedeki: Sorry mixed up star and dot.

benedeki (approved):

benedeki:
Running on AWS pas few days. Adding `PR: Tested

PR #1964, opened 2021-10-26
Partition sizing demo - potential library

benedeki:
This is just a demo, not to be merged. Also the class and method names are not given.


benedeki:
Served it's purpose, closing

Issue #1938, opened 2021-10-01
Make it possible to have mapping rule without a join condition


benedeki:
## Background
With the ability of filtering it is possible the mapping table is reduced enough, and there's no need to add a join condition.

## Feature
Make it possible to have no join condition on the Mapping Conformance Rule - support must exist both on back-end and front-end.

dk1844:
Release notes
Mapping Conformance Rule may now exist without a join conditions, but in such a case a non-empty filter must be defined instead.

PR #1970, opened 2021-11-03
1938 MappingCR without join a condition

dk1844:
As outlined in #1938, this PR makes a Mapping Conformance Rule exists without a join condition.
Originally, without mapping filters, this would have made little sense, but with the employment of mapping filters, there are viable use cases (see below in the _Testrun_)

## Affected layers
 - UI (validation) has been changed to allow mCR with 0 filter
 - Backend service no longer throws an exception on empty join-conitions, test suites updated

## Testrun
A superhero dataset has been used with a MT having zero join-on column:
```
+-------------+---------+
|      njValue|otherData|
+-------------+---------+
|      regular|      111|
|extraordinary|      222|
|         rare|      333|
+-------------+---------+
```

### NoJoin without filter = data explosion
Had there been no filter, the user would either have to add a preexisting literal column to join on, otherwise they would be facing data explosion:
no-join-no-filter-explosion

### NoJoin with filter = correct data enrichment
Using an appropriate filter with a no-join mapping table yields the expected result:
no-join-filtered

## Release notes suggestion
Mapping Conformance Rule may now exist without a join conditions, but in such a case a non-empty filter must be defined instead.


menas/ui/components/dataset/conformanceRule/ConformanceRuleForm.js
```diff
@@ -374,13 +374,8 @@ class MappingConformanceRuleForm extends ConformanceRuleForm {
...
 
   hasValidJoinConditions(fieldValue = []) {
-    let isValid = fieldValue.length >= 1;
-
-    if (!isValid) {
-      sap.m.MessageToast.show("At least 1 join condition is required.");
-    }
-
-    return isValid
+    // empty join conditions are allowed
```
Zejnilovic, line 377: How hard would it be to check either for >= 1 join conditions or >= 1 field equals filters?

AdrianOlosutean: Or why not remove the method and its uses?

dk1844: @Zejnilovic: I have done it on UI in https://github.com/AbsaOSS/enceladus/pull/1970/commits/c35ab4c15c93091d4b4dc6c10533553094a2da7a . It looks like this:

join-condition-or-filter

(@AdrianOlosutean, could be, but now it is used again)

AdrianOlosutean (approved):
Looks good, tested in Menas

PR #1971, opened 2021-11-03
1954 Fix mt conformance rule import validation ignoring structs

### LINKED ISSUE #1954: Import feature doesn't properly validate conformance rules if output is a struct
## Describe the bug
When importing a dataset having conformance rules that have structs as outputs, the validation does not add subfields to the evolving schema. This fails the validation if the output subfields are used in subsequent conformance rules

## To Reproduce
Steps to reproduce the behaviour OR commands run:
1. Have a dataset
2. Have Mapping conformance rule which output a struct
3. Have any other rule using a subfield from output struct of the 2nd step
4. Try importing this Dataset

## Expected behaviour
Import passes

## Additional context
Check Rules -> Mapping, Single Column, Coalesce

Zejnilovic:
I don't like how I wrote it. But I cannot right now come up with a better way. Please do be critical when code reviewing.

Closes #1954 

The same issue might be with DropConformance Rule when dropping a struct, its children will not get deleted. The use of Set[String] might be unfortunate and the Schema and SchemaField should maybe be enhanced so it is self-aware of the children. Would love a chat about this.

Release Notes: When importing a dataset a mapping table conformance rule can now output a struct and will be properly validated.


dk1844:
> I don't like how I wrote it. But I cannot right now come up with a better way. Please do be critical when code reviewing.

I tried to help in https://github.com/AbsaOSS/enceladus/pull/1973. Hope you find it useful. 👍

menas/src/main/scala/za/co/absa/enceladus/menas/services/DatasetService.scala
```diff
@@ -316,17 +316,32 @@ class DatasetService @Autowired()(datasetMongoRepository: DatasetMongoRepository
...
-    val inputValidation = mt.attributeMappings.values.map(validateInputColumn(fields, _))
+                           mtRule: MappingConformanceRule): RuleValidationsAndFields = {
+    val inputValidation = mtRule.attributeMappings.values.map(validateInputColumn(fields, _))
+    val allOutput = mtRule.allOutputColumns()
+    val outputColumns = mtRule.allOutputColumns().keySet
+
+    val mtFields = for {
+      someMappingTable <- datasetMongoRepository.getConnectedMappingTable(mtRule.mappingTable, mtRule.mappingTableVersion)
+      mtSchema <- someMappingTable.map(mt => datasetMongoRepository.getConnectedSchema(mt.schemaName, mt.schemaVersion)).get
+    } yield mtSchema.map(x => x.fields.flatMap(f => f.getAllChildrenBasePath :+ f.path).toSet ).get
```
AdrianOlosutean, line 327: Most I could do to make it a bit more readable
```suggestion
    val mtFields = for {
      someMappingTable <- datasetMongoRepository.getConnectedMappingTable(mtRule.mappingTable, mtRule.mappingTableVersion)
      mtSchema: Option[Schema] <- someMappingTable.map(mt => datasetMongoRepository.getConnectedSchema(mt.schemaName, mt.schemaVersion)).get
      result = mtSchema.map(_.
        fields.flatMap(f => f.getAllChildrenBasePath :+ f.path).toSet
      ).getOrElse(Set.empty)
    } yield result
```

AdrianOlosutean (commented):
Just some style things

menas/src/main/scala/za/co/absa/enceladus/menas/services/DatasetService.scala
```diff
@@ -316,17 +316,32 @@ class DatasetService @Autowired()(datasetMongoRepository: DatasetMongoRepository
...
       val updated: RuleValidationsAndFields = validateOutputColumn(acc.fields, outputCol)
       acc.updateWithFieldsReplace(updated)
     })
+
+    val outputColsFlat: Future[Set[String]] = for {
+      fieldsFromMT <- mtFields
+      oldFields <- validatedOutputCols.fields
+    } yield  oldFields ++ allOutput.flatMap { case (out, in) =>
+      DatasetService.replacePrefixIfFound(fieldsFromMT, out, in)
+    }
```
AdrianOlosutean, line 342:
```suggestion
   val outputColsFlat: Future[Set[String]] = for {
      fieldsFromMT <- mtFields
      oldFields <- validatedOutputCols.fields
      newFields = allOutput.flatMap { case (out, in) =>
        DatasetService.replacePrefixIfFound(fieldsFromMT, out, in)
      }
    } yield oldFields ++ newFields
```

AdrianOlosutean (commented):
Sorry for double review

Zejnilovic:
> Just some style things

Happy to apply them. I know I see things differently

dk1844 (approved):
LGTM 😄 (just read the code this time)

AdrianOlosutean (approved):
Tried locally, where a mapping table outputs two different structs from a mapping table. Validating dropping fields from those structs works, as well as those structs themselves. However, it also works to import a dataset where a subfield of an existing field is dropped, which I think it's wrong. I am not sure if this is related to this PR or not

PR #1973, opened 2021-11-05
Bugfix/1954 fix import validation mt rule suggestions

dk1844:
@Zejnilovic, let me humbly offer my suggestions on the PR #1971.


Zejnilovic (approved):
I love it. `stripPrefix` totally eluded me

PR #1980, opened 2021-11-11
#1867 filter from config removed, including tests.

### LINKED ISSUE #1867: Remove the ability to set a datset filter via the configuration file
## Background
When the UI will have been done, the ability to set the dataset filter via the configuration the more disruptive and confusing rather than helping.

## Feature
Remove the ability to set the dataset filter via the configuration file.

## Proposed Solution
Solution Ideas 
1. See the class `FilterFromConfig` and its usage.

dk1844:
Removal of `FilterFromConfig` (i.e. feature where a filter is allowed to be set from config instead of rule/mt) alongside its usage and test setup.

Closes #1867.

# RN suggestion
Enceladus no longer support loading filter from config (config tree `dataframefilter.$dataFrameName`).


AdrianOlosutean (approved):
Code reviewed

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/interpreter/rules/mapping/CommonMappingRuleInterpreter.scala
```diff
@@ -66,13 +65,9 @@ trait CommonMappingRuleInterpreter {
 
     val mappingTableDef = dao.getMappingTable(rule.mappingTable, rule.mappingTableVersion)
 
-    val ruleFilter = if (rule.mappingTableFilter.nonEmpty) {
-      rule.mappingTableFilter
-    } else {
-      // This is a workaround until UI supports filter definition. Until then, the filters can be set via configuration.
-      FilterFromConfig.loadFilter(rule.mappingTable)
-    }
+    val ruleFilter = rule.mappingTableFilter
```
benedeki, line 68: I think even this `val` can go.

dk1844: Well, it could go and one would just use `rule.mappingTableFilter`. However, my point here is that it improves the readability of the code. I deliberately left it as `ruleFilter` and `mappingTableFitler`, those two named filters that will get merged later:

```scala
    (ruleFilter, mappingTableFilter) match {
      case (Some(a), Some(b)) => Option(a and b)
      case (Some(a), None)    => Option(a)
      case (None, Some(a))    => Option(a)
      case (None, None)       => None
```

In my opinion, it conveys the meaning better than if the `val` gets removed. So here, I'd really like to keep it as-is if this perspective makes sense to you.

benedeki: Fair enough 👍

benedeki (approved):

benedeki:
Tested, also against an older version. Unlike in the older version the filter configuration is not taken into account.

Issue #1987, opened 2021-11-24
Repartitioning prevents processing of zero size files


benedeki:
## Describe the bug
After partition sizing has been introduced it's nor possible to process data sets with 0 rows. The application exits with `division by zero` exception.

## To Reproduce
Steps to reproduce the behavior OR commands run:
1. Have a Dataset prepared
2. In the row input have zero rows
3. Run _Standardization_

## Expected behavior
Even zero size data sets are processed.

## Screenshot
![image](https://user-images.githubusercontent.com/14905969/143242656-b4e739e9-7102-44e0-b727-dd08eb5d951d.png)

benedeki:
Release notes:
Fixed issue when trying to repartition an empty dataset with 0 partitions.

PR #1988, opened 2021-11-24
#1987: Repartitioning prevents processing of zero size files

benedeki:
* Handling 0 partition case in repartitioning
* .gitignore duplicate entry fixed

_Release notes:_
_Fixed issue when trying to repartition an empty dataset with 0 partitions._


dk1844 (approved):
LGTM (just read the code)

PR #2000, opened 2021-12-12
#1998: Ensure log4j v2 at least 2.15.0

### LINKED ISSUE #1998: Ensure log4j v2 at least 2.15.0
Some imported dependencies bring in log4j v 2.x Ensure that at least 2.15 is part of the project.

benedeki:
* Menas gets log4j via Spring, explicitly imported to ensure 2.15 is used
* Upgraded Hyperdrive in SparkJobs which already uses log4j 2.15

Fixes #1998

### DIFF
--- a/menas/pom.xml
+++ b/menas/pom.xml
@@ -124,6 +124,17 @@
             <version>${bson.codec.jsr310.version}</version>
         </dependency>
         <!-- SPRING -->
+        <!-- ensuring that log4j used us at least 2.15.0-->
+        <dependency>
+            <groupId>org.apache.logging.log4j</groupId>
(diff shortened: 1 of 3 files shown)


menas/pom.xml
```diff
@@ -124,6 +124,17 @@
             <version>${bson.codec.jsr310.version}</version>
         </dependency>
         <!-- SPRING -->
+        <!-- ensuring that log4j used us at least 2.15.0-->
```
Zejnilovic, line 127: When we merge to version-3 brach we should not forget about API vs web

Zejnilovic (approved):

PR #2006, opened 2022-01-05
Feature/1876 menas migration script

### LINKED ISSUE #1876: Migration application/script
## Background
As part of moving Enceladus to cloud, we need to get the existing configuration (dataset definitions) data to AWS.

## Feature
Create the application that will migrate - read, transform and write elsewhere (in cloud) the data of selected datasets

dk1844:
Migration script for Menas (mongo A -> mongo B)
 - `--help` explains a lot
```bash
./migrate_menas.py --help                                                                                             
usage: migrate_menas [-h] [-n] [-v] [-t TARGETDB] [-s SOURCEDB] (-d DATASET_NAME [DATASET_NAME ...] | -m MTABLE_NAME [MTABLE_NAME ...]) SOURCE TARGET

Menas MongoDB migration script.

positional arguments:
  SOURCE                connection string for source MongoDB
  TARGET                connection string for target MongoDB

options:
  -h, --help            show this help message and exit
  -n, --dryrun          if specified, skip the actual synchronization, just print what would be copied over. (default: False)
  -v, --verbose         prints extra information while running. (default: False)
  -t TARGETDB, --target-database TARGETDB
                        Name of db on target to migrate to. (default: menas)
  -s SOURCEDB, --source-database SOURCEDB
                        Name of db on source to migrate from. (default: menas)
  -d DATASET_NAME [DATASET_NAME ...], --datasets DATASET_NAME [DATASET_NAME ...]
                        list datasets to migrate (default: [])
  -m MTABLE_NAME [MTABLE_NAME ...], --mapping-tables MTABLE_NAME [MTABLE_NAME ...]
                        list datasets to migrate (default: [])

```

 - exactly one of `--datasets`/`--mapping-tables` is required
 - source DB is hardcoded to `menas`, target defaults to `menas`, but can be changed
 - migration possible "by DS names" (migrates DSs, DS schemas, MTs, MT schemas, and runs) and "by MT names" (migrates MTs and MT schemas)
 - the script is designed to be rerunnable again to fill in the blanks from previous runs
 - migration mechanism:
   1. discover migrated entities, per each entity type:
   2. lock (`locked=true`), assign migration tag, `userLocked=migration` and `dateLocked` structure
   3. migrate locked
   4. migrated entities are marked with `migrated=true`
 - duplicates on target are counted and warned about, but skipped (**no overwrite**)
 
Example of migrated tags on source:
locked-migrationHas-migrated_userLocked_dateLocked
BSON/JSON-example:
```json
"dateLocked" : {
        "dateTime" : {
            "date" : {
                "year" : 2022,
                "month" : 1,
                "day" : 10
            },
            "time" : {
                "hour" : 14,
                "minute" : 45,
                "second" : 5,
                "nano" : 281230000
            }
        },
        "offset" : 0,
        "zone" : "UTC"
    },
    "locked" : true,
    "migrationHash" : "13cdfa",
    "userLocked" : "migration",
    "migrated" : true
```

(target only has the `migrationHash`)

### By DS names migration testrun:
```bash
./migrate_menas.py mongodb://localhost:27017/admin mongodb://localhost:27017/admin -t mns2 -d mydataset1 test654       
Menas mongo migration
Running with settings: dryrun=False, verbose=False
Using migration #: '2916a3' and timestamp 2022-01-10 14:39:33.074195+00:00
  source connection-string: mongodb://localhost:27017/admin
  source DB: menas
  target connection-string: mongodb://localhost:27017/admin
  target DB: mns2
dataset names supplied: ['mydataset1', 'test654']
Dataset names to migrate: ['mydataset1', 'test654']
DS schemas to migrate: ['another_schema1', 'myschema1']
MTs to migrate: ['bogusMt']
MT schemas to migrate: ['testSchema3']
Runs to migrate: []
Attachments of schemas to migrate: ['another_schema1', 'myschema1', 'testSchema3']


Migration of collection schema_v1 started
Migration of collection schema_v1 finished, migrated 104 schemas.

Migration of collection dataset_v1 started
Migration of collection dataset_v1 finished, migrated 8 datasets.

Migration of collection mapping_table_v1 started
Migration of collection mapping_table_v1 finished, migrated 20 mapping tables.

No runs to migrate in run_v1, skipping.
Migration of collection attachment_v1 started
Migration of collection attachment_v1 finished, migrated 99 attachments.

Done.
```

### By MT names migration testrun:
```bash
./migrate_menas.py mongodb://localhost:27017/admin mongodb://localhost:27017/admin -t mns2 -m MyAwesomeMappingTable1   
Menas mongo migration
Running with settings: dryrun=False, verbose=False
Using migration #: '2916a3' and timestamp 2022-01-10 14:39:33.074195+00:00
  source connection-string: mongodb://localhost:27017/admin
  source DB: menas
  target connection-string: mongodb://localhost:27017/admin
  target DB: mns2
mapping table names supplied: ['MyAwesomeMappingTable1']
MTs to migrate: ['MyAwesomeMappingTable1']
MT schemas to migrate: ['testSchema4', 'tzfix']
Attachments of schemas to migrate: ['testSchema4', 'tzfix']


Migration of collection schema_v1 started
Migration of collection schema_v1 finished, migrated 12 schemas.

Migration of collection mapping_table_v1 started
Migration of collection mapping_table_v1 finished, migrated 17 mapping tables.

Migration of collection attachment_v1 started
Migration of collection attachment_v1 finished, migrated 9 attachments.

Done.
```
### Test-help:
If you feel like testing this script, you might want to reset the state of source menas to the original, non-migrated state. This little command might help:
```javascript
db.getCollection('schema_v1').updateMany( // same for collections `dataset_v1`, `mapping_table_v1`, `run_v1`, `attachment_v1`
    {}, 
    { $unset : {
        "migrationHash" : "",
        "locked": true,
        "migrated": true,
        "dateLocked": "",
        "userLocked": ""
    }}
)
```

Things to consider (besides what you figure out)
 1. target DB name - configurable via `-t/--target-database`, defaults to `menas`. ✔️ 
 2. migration success marking - marked by `migrated=true` ✔️ 
 3. when migrating "by MT", only MTs and MT schemas are migrated now. Should this include the lookup of DSs that use this MT and also schemas of these DSs, too? ✔️ 
 4. are the options sufficient in terms of DSs/MTs input (`--datasets mydataset1 mydataset2` ... or do we need to implement input from a file, too? ✔️ 
 5. how can we test/assess correctness? I have tested this while developing, of course, but that should not be the end of it.
 6. If some entities were migrated using other means, the migration script reports it and skips those entities. ✔️ 
 7. attachment migration as a lower priority item? ✔️ 

Closes #1876


scripts/migration/migrate_menas.py
```diff
@@ -0,0 +1,326 @@
...
+
+# Constants
+NOT_LOCKED_MONGO_FILTER = {"$or": [
+    {"locked": False},  # is not locked, or
+    {"locked": {"$exists": False}}  # or: there is no locking info at all
+]}
+EMPTY_MONGO_FILTER = {}
+
+
+def parse_args():
```
Zejnilovic, line 33: Do we add return types?

AdrianOlosutean (commented):
I would break the file a bit into a migrate_menas with `parse_args` and main logic + 1 or more others according to preference

@dk1844 replying:
Honestly, I'd rather keep it in one file, at least that was my initial intent. How strongly do you feel about it?

scripts/migration/migrate_menas.py
dk1844: I added param types and return types in the method signatures. The only place that I am unsure of is how to type the function-type params.

AdrianOlosutean:
1. I think it's a better idea to have a param for the destination database. Supposing the migration was done successfully and you would like to change the default name to `menas`, that would not really be possible without very expensive operations
2.  not sure
3.  only MT and its schema
4.  Don't think input file is really necessary now
5. 
6. I would say log and allow

AdrianOlosutean:
No strong opinion about having multiple files, just that it's a bit more difficult to follow the code.

On another topic, where do you set locked to false after successfully migrating?

Zejnilovic:
```shell
$> ./migrate_menas.py --help
Traceback (most recent call last):
  File "./migrate_menas.py", line 19, in <module>
    from minydra.dict import MinyDict  # dictionary with dot access
ModuleNotFoundError: No module named 'minydra'
$> python --version
Python 3.9.1
```

Zejnilovic:
> On another topic, where do you set locked to false after successfully migrating?

@AdrianOlosutean as far as I envision this script. Locked should remain true after migration so we do not create 2 independent or contradicting branches of the same entity

dk1844:
> ModuleNotFoundError: No module named 'minydra'
>
> I presume I need to install something. Should we add some README to this migration explaining this?

Yes, thanks for this. You need extra packages - I have updated the script and the readme: https://github.com/AbsaOSS/enceladus/pull/2006/commits/45b8c4304d707940f8ab4a4e8855014e2299bfb0

AdrianOlosutean:
> There is another thing I have noticed, in #1950, besides `locked` field, there is also `dateLocked` and `userLocked, but this is only populated when locking via the API. For the purpose of the script, should it also populate these fields by e.g. `user=migration/menas-migration/migrate_menas.py` and fill the date, too?

I would say it should be populated. I'm not sure if the user should just be admin or the migration script, since locking would almost exclusively be done by admin

dk1844:
> > There is another thing I have noticed, in #1950, besides `locked` field, there is also `dateLocked` and `userLocked, but this is only populated when locking via the API. For the purpose of the script, should it also populate these fields by e.g. `user=migration/menas-migration/migrate_menas.py` and fill the date, too?
> 
> I would say it should be populated. I'm not sure if the user should just be admin or the migration script, since locking would almost exclusively be done by admin

Updated the script, it now populates `userLocked=migration` and `dateLocked` with the, hopefully expected, datetime structure (I have tried locking via the API and mimicked the structure 😄 )

README.md
```diff
@@ -342,3 +342,9 @@ A module containing [examples](examples/README.md) of the project usage.
 
 ## How to contribute
 Please see our [**Contribution Guidelines**](CONTRIBUTING.md).
+
+
+# Extras
+ - For Menas migration, there is a useful script available in
+[scripts/migration/migrate_menas.py](scripts/migration/migrate_menas.py)
+   (you need python packages `pymongo` and `minydra` to run)
```
Zejnilovic, line 350: I like this. Could we also commit requirements.txt? 
[popreqs](https://github.com/bndr/pipreqs) will generate them for you, so in the end, users just run `pip install -r requirements.txt`

scripts/migration/migrate_menas.py
```diff
@@ -0,0 +1,433 @@
...
+
+defaults = MinyDict({
+    'verbose': False,
+    'dryrun': False,
+    'target_db_name': "menas"
+})
+
+migration_hash = secrets.token_hex(3)  # e.g. 34d4e10f
+
+SOURCE_DB_NAME = "menas"
```
Zejnilovic, line 47: feels a bit weird for this to be not part of the input

scripts/migration/migrate_menas.py
```diff
@@ -0,0 +1,433 @@
...
+
+def assemble_schemas_from_ds_names(db: Database, ds_names: List[str], not_locked_only: bool) -> List[str]:
+    return assemble_schemas_from_x(db, ds_names, "dataset_v1", "schemaName", not_locked_only)
+
+
+def assemble_schemas_from_mt_names(db: Database, mt_names: List[str], not_locked_only: bool) -> List[str]:
+    return assemble_schemas_from_x(db, mt_names, "mapping_table_v1", "schemaName", not_locked_only)
+
+
+def assemble_schemas_from_x(db: Database, entity_names: List[str], collection_name: str,
```
Zejnilovic, line 137: Not sure what the X is supposed to be

scripts/migration/migrate_menas.py
```diff
@@ -0,0 +1,433 @@
...
+
+    print("Done.")
+
+
+if __name__ == '__main__':
+    args = parse_args()
+
+    # script-global flags
+    dryrun = args.dryrun
+    verbose = args.verbose
```
Zejnilovic, line 458: Are these like global values?

Zejnilovic (commented):
Very sorry it took me so long. This is a first pass through the script. I do agree that it might be better if we could split it into smaller ones.

scripts/migration/migrate_menas.py
dk1844: yes, they are. That is why the methods use it even thought their signature does not contain it.

scripts/migration/migrate_menas.py
dk1844: this is a generalization of `assemble_schemas_from_ds_names` and `assemble_schemas_from_mt_names`, because their logic is very similar and that's why this common method exists.

scripts/migration/migrate_menas.py
dk1844: My assumption here was that while the user might want to change the target_db (e.g. for a careful migration/tryout), the source db will always be `menas`. Is there a case where it may not be?

If so, it is relatively easy to add an option to specify it

README.md
dk1844: wow, did not know this, I will try it, nice! :D

scripts/migration/migrate_menas.py
```diff
@@ -0,0 +1,433 @@
+#!/usr/bin/python3
```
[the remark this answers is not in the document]
dk1844: Thanks, @Zejnilovic for pushing this

README.md
dk1844: Great idea, done.

Zejnilovic:
From --help, it looks like Source and Target are the last parameters, but they are the first ones. 
Or is it just me?

Zejnilovic:
If Dataset is not found there is no warning, no error. Is that correct?

```
Menas mongo migration
Running with settings: dryrun=True, verbose=True
Using migration #: '30c9d4' and locking timestamp 2022-01-11 10:57:12.381195+00:00 (UTC)
  source connection-string: mongodb://localhost:27017
  source DB: menas
  target connection-string: mongodb://localhost:27017
  target DB: migration_aws_1
dataset names supplied: ['Alfa']
Dataset names given: ['Alfa']
Dataset names to migrate: []
DS schemas to migrate: []
MTs to migrate: []
MT schemas to migrate: []
Runs to migrate: []
All schemas (DS & MT) to migrate: []
Attachments of schemas to migrate: []
*** Dryrun selected, no actual migration will take place.
Done.
```

Zejnilovic:
A DB needs to already be initialized with all the fields needed. If 1st migration is run and then Menas, Menas won't start up.

This could be described in texts + maybe checked?

Zejnilovic:
We will need a feature where we can "Mass unlock" or "Mass un-migrate" entities. 100% a new Issue/PR if we agree it is needed

Zejnilovic:
When trying to migrate already migrated, instead of nothing to migrate, it would be nice to see a message "Already migrated"

Zejnilovic:
The outputs are very clean. I think it is enough to pipe it into a file. No extra file creation function needed

scripts/migration/migrate_menas.py
dk1844: I have added the option to define the source db name, too, it is `-s`/`--source-database`. Defaults to `menas`.

scripts/migration/migrate_menas.py
dk1844: renamed to: `assemble_schemas`

dk1844:
> From --help, it looks like Source and Target are the last parameters, but they are the first ones. Or is it just me?

It does not matter. Source and Target are positional parameters, the only two. They must be given in this order, but can be freely interleaved by optional param pairs. (`myscript.py --some-optional-param valueABC firstPostitional --some-other-option valueXYZ --yet-another-option`). This is common with python argparse processing and perhaps sort of a nice general behavior for combination of dash-prefixed optional params and positional ones.



> If Dataset is not found there is no warning, no error. Is that correct?
> ...

Well, it's not wrong, given the laziness of Mongo db/collection creation (_ lazy creation process [...] the database and corresponding collection are actually only created when a document is inserted._, src: [here](https://www.mongodb.com/blog/post/pymongo-monday-pymongo-create)). So it is the default behavior of pymongo. But surely, we can check if the source is empty. Maybe even if it is created.


> A DB needs to already be initialized with all the fields needed. If 1st migration is run and then Menas, Menas won't start up.
> 
> This could be described in texts + maybe checked?

What? Can you elaborate, please?




> We will need a feature where we can "Mass unlock" or "Mass un-migrate" entities. 100% a new Issue/PR if we agree it is needed

Can be done. I currently use the script(s) outline in the **Test-help** section of the first comment.



> When trying to migrate already migrated, instead of nothing to migrate, it would be nice to see a message "Already migrated"
If you only want to change the message (and keep the current logic), I could do it. I assume you mean the part where it says
```
No schemas to migrate in schema_v1, skipping.
No datasets to migrate in dataset_v1, skipping.
No mapping tables to migrate in mapping_table_v1, skipping.
No runs to migrate in run_v1, skipping.
No attachments to migrate in attachment_v1, skipping.
```
However, if you wanted to distinguish between nothing to migrate vs already migrated, then note, that the migration parts are oblivious to the difference.
First, there is a _discovery phase_ that, given the input, finds matching entities to migrate (in source only - so it know what should be migrated, but does not look in the target). Second, _migration phase_ is run with the lists of entities from the first phase -- this deals with possible already-migrated-but-not-marked entities on target, but does not look back how that list came to be (there was nothing at all in the source vs. nothing matched vs. nothing non-locked matched).
So let's be more specific here, please.

scripts/migration/migrate_menas.py
dk1844: I have checked sources like https://www.python.org/dev/peps/pep-0008/#global-variable-names and https://pythonguides.com/python-naming-conventions/ None of them suggest a prefix.
There is a `global` keyword, but it is for write access to global vars, which we don't need https://www.programiz.com/python-programming/global-keyword

Nevertheless, I have at least put them all in the same place (`__main__`) which should make it pretty readable.

Also, to minimize exposure, I have made the `dryrun` non-global`, as it does not make a lot of mess with bloating of the signatures (as opposed to `verbose` might). Also, it is fair for `dryrun` to be in the method signatures as it has significant effect on the script run. Verbose only results in more prints, so no significant behavior change.

dk1844:
> Very sorry it took me so long. This is a first pass through the script. I do agree that it might be better if we could split it into smaller ones. Might be me not being accustomed to python so it is hard for me to read.

I have put the constants in a separate file (not sure if this is a great idea after all, but anyway it perhaps cleaned the start a bit), but mainly, I have moved the `argparse` processing to the start. This gives a good overview of the script. `def main()` and `__main__` are customarily last.
Hope this helps.

Zejnilovic:
> It does not matter. Source and Target are positional parameters, the only two. They must be given in this order, but can be freely interleaved by optional params / param pairs. (myscript.py --some-optional-param valueABC firstPostitional --some-other-option valueXYZ --yet-another-option). This is common with python argparse processing and perhaps sort of a nice general behavior for combination of dash-prefixed optional params and positional ones.

But if I put them last it does not work. What am I missing?

```shell
╰─$ python migrate_menas.py -n -v -t migration_aws_1 -d Alfa mongodb://localhost:27017 mongodb://localhost:27017
usage: migrate_menas [-h] [-n] [-v] [-t TARGETDB] [-s SOURCEDB] (-d DATASET_NAME [DATASET_NAME ...] | -m MTABLE_NAME [MTABLE_NAME ...]) SOURCE TARGET
migrate_menas: error: the following arguments are required: SOURCE, TARGET
```

Zejnilovic:
> Well, it's not wrong, given the laziness of Mongo db/collection creation (_ lazy creation process [...] the database and corresponding collection are actually only created when a document is inserted._, src: here). So it is the default behavior of pymongo. But surely, we can check if the source is empty. Maybe even if it is created.

I would like this. Do you prefer another ticket?

> What? Can you elaborate, please?

If I do not have the target DB created by means beforehand and migration goes first, Menas will not work. Because it creates entities but not in full and as expected by Menas.

> Can be done. I currently use the script(s) outline in the Test-help section of the first comment.

I will create a ticket

> However, if you wanted to distinguish between nothing to migrate vs already migrated, then note, that the migration parts are oblivious to the difference.
First, there is a discovery phase that, given the input, finds matching entities to migrate (in source only - so it know what should be migrated, but does not look in the target). Second, migration phase is run with the lists of entities from the first phase -- this deals with possible already-migrated-but-not-marked entities on target, but does not look back how that list came to be (there was nothing at all in the source vs. nothing matched vs. nothing non-locked matched).
So let's be more specific here, please.

Did I misunderstand that in the discovery we are "discarding" any locked entities? What I tried to say is, that maybe we should print out that we discarded these X already locked entities.

Do stop me if I am taking this issue/PR too far.

AdrianOlosutean (dismissed):
Looks good for now, code reviewed

Zejnilovic:
#2012 Migration Script - Print out already locked/migrated entities that are skipped 
#2011 Migration Script - Add rollback of locking
#2010 Migration script - Check if target DB is initialized before running migration
#2009 Migration script - Check if the source returns empty

Zejnilovic:
> Done ✔️ , you can check it out, @Zejnilovic. Source DB is now check for having expected collections (only those we plan to migrate). For target, I did not do this, because we lack the means to prepare the fresh db first, yet.

Can I recommend doing it once using `db_version` collection? This would give a once-off answer to the whole question.

dk1844:
> #2012 Migration Script - Print out already locked/migrated entities that are skipped 
> #2011 Migration Script - Add rollback of locking
> #2010 Migration script - Check if target DB is initialized before running migration
> #2009 Migration script - Check if the source returns empty

Thank you hugely, @Zejnilovic .

> > Done ✔️ , you can check it out, @Zejnilovic. Source DB is now checked for having expected collections (only those we plan to migrate). For target, I did not do this, because we lack the means to prepare the fresh db first, yet.
> 
> Can I recommend doing it once using `db_version` collection? This would give a once-off answer to the whole question.

Doing what? Checking that a specified db contains `db_version` collection, perhaps even with the expected content of a record having `version=1`? Do you mean on source or on target? or both?

AdrianOlosutean (approved):
Code reviewed

AdrianOlosutean (approved):

Zejnilovic (approved):
Code reviewed

Zejnilovic (approved):
Ran tests. Works well.
Will add tickets for:
- "skip runs" as they are platform-specific (#2014)
- "don't lock runs" as they are not modifiable (#2015)

PR #2016, opened 2022-01-19
Feature/2013 menas init

### LINKED ISSUE #2013: Migration Script - initial menas db provisioning
## Background
With the latest updates to the migration script (namely checking (source and) target db validity (version collection, perhaps in the future also target collections), we need a way to initialize a blank DB to a proper state suitable for migration.

## Feature
Starting with an empty database, create necessary collections for migration (`schema_v1`, `dataset_v1`, `mapping_table_v1`, `run_v1`, `attachment_v1`) with indices and "db_version" with "version=1" record.

dk1844:
Menas DB initialization script
 - having a blank db, this script assert that:
   - migrating collections exist (`attachment_v1`, `dataset_v1`, `mapping_table_v1`, `schema_v1`, and `run_v1`, `propertydef_v1`, `landing_page_statistics_v1`)
   - appropriate indices are added to the collections above ⬆️ 
   - collection `db_version` exists and record version=1 is set.

A side-effect of this development is the creation of `menas_db` module with generalized `MenasDb` object, that is used in migration script, too. Also, the prior-to-migration check for migration is fortified by now requiring the target db properly initialized (version, collections) - exactly what this initialization script can enforce.

## Release notes suggestion
`initialize_menas` can be used to fully prepare MongoDB database for `migrate_menas`. The script creates necessary collections, their indices and `db_version` record.

Closes #2013


scripts/migration/initialize_menas.py
```diff
@@ -0,0 +1,167 @@
...
+    print('  target connection-string: {}'.format(target))
+    print('  target DB: {}'.format(target_db_name))
+    print("")
+
+    from menas_db import MenasDbError, MenasDb
+    target_db: MenasDb = MenasDb.from_connection_string(target, target_db_name, "target db", verbose)
+
+    try:
+        target_db.check_db_version()
+        target_db.check_menas_collections_exist()  # todo, check indexing, too?
```
Zejnilovic, line 142: I would not be so nice. You implemented create or repair basically. I do not think that should be the job of this script.

This script should check db_version collection. And if it is:
- none -> Start init
- < 1 -> Tell user to run means for upgrade
- == 1 -> Tell user all is fine

scripts/migration/initialize_menas.py
```diff
@@ -0,0 +1,167 @@
...
+                        help="Name of db on target to migrate to.")
+
+    return parser.parse_args()
+
+
+def initialize_menas_db(menas_db: MenasDb) -> None:
+
+    # create db_version record
+    print("'db_version' initialization:")
+    create_db_version(menas_db.mongodb)
```
Zejnilovic, line 49: Please move this to be the last creation of the init script.

scripts/migration/initialize_menas.py
dk1844: Redone in this fashion, thanks for the real-world-inspired opinion.

scripts/migration/initialize_menas.py
dk1844: Done.

scripts/migration/initialize_menas.py
```diff
@@ -0,0 +1,177 @@
...
+        insert_result = collection.insert_one({"version": 1})
+        print(f"  Created version=1 record with id {insert_result.inserted_id}")
+    print("")
+
+
+def create_collections(db: Database, collection_names: List[str]):
+    print("Initialization of migration-related collections")
+    existing_collections = db.list_collection_names()
+    if len(existing_collections) != 0:
+        print(f"Warning, there are unexpected existing collections in the DB already: {existing_collections}")
```
Zejnilovic, line 92: When I run init on a "bad" already existing DB. I get a warning that I am missing collection and then it goes on about indexes and then it fails. 

I need it to fail the second it sees that collections with our names are already in use

Zejnilovic:
```shell
Menas mongo initialization
Running with settings: dryrun=False, verbose=True
  target connection-string: mongodb://localhost:27017
  target DB: migration_aws_1

DB 'migration_aws_1' does not contain record version (DB migration_aws_1 (target db) does not contain collection db_version!)

Initializing DB 'migration_aws_1' for Menas...

Initialization of migration-related collections
Warning, there are unexpected existing collections in the DB already: ['dataset_v1', 'mapping_table_v1', 'schema_v1', 'attachment_v1']
  BEFORE: Aiming to create ['schema_v1', 'dataset_v1', 'mapping_table_v1', 'run_v1', 'attachment_v1'], found exising ['dataset_v1', 'mapping_table_v1', 'schema_v1', 'attachment_v1'].
  Creating missing collections ['run_v1'] ...
  AFTER: existing collections: ['dataset_v1', 'mapping_table_v1', 'schema_v1', 'run_v1', 'attachment_v1']

Initialization indices (1) on Menas collection 'schema_v1':
  BEFORE: index info: {'_id_': {'v': 2, 'key': [('_id', 1)]}}
  creating index for key [('name', 1), ('version', 1)], unique=True
  AFTER: index info: {'_id_': {'v': 2, 'key': [('_id', 1)]}, 'name_1_version_1': {'v': 2, 'unique': True, 'key': [('name', 1), ('version', 1)]}}

Initialization indices (1) on Menas collection 'dataset_v1':
  BEFORE: index info: {'_id_': {'v': 2, 'key': [('_id', 1)]}}
  creating index for key [('name', 1), ('version', 1)], unique=True
Traceback (most recent call last):
  File "/Users/.../Development/enceladus/scripts/migration/initialize_menas.py", line 141, in run
    target_db.check_db_version()
  File "/Users/.../Development/enceladus/scripts/migration/menas_db.py", line 84, in check_db_version
    raise MenasNoDbVersionRecordError(f"DB {self.mongodb.name}{self.hint} does not contain collection {DB_VERSION_COLLECTION}!")
menas_db.MenasNoDbVersionRecordError: DB migration_aws_1 (target db) does not contain collection db_version!

During handling of the above exception, another exception occurred:
```

Zejnilovic (commented):
Missing collections:
- `landing_page_statistics_v1`
- `propertydef_v1`

Indexes are not "full"
- `attachment_v1` misses `refName_1_refVersion_1`
- `run_v1` misses `uniqueId_1`

Tested: 
- basic flow (verified indexes)
- existing DB db_version version set to 1
- existing DB with no db_version
- existing DB with db_version version set to 2

dk1844:
Thank you, @Zejnilovic, for such an extensive test. Individual comments below

> Missing collections:
> 
> * `landing_page_statistics_v1`
> * `propertydef_v1`

Added, thanks. Migration of propertydef_v1 added as a separate issue: https://github.com/AbsaOSS/enceladus/issues/2017

 
> Indexes are not "full"
> 
> * `attachment_v1` misses `refName_1_refVersion_1`
> * `run_v1` misses `uniqueId_1`

You are right, I naively based the code on what I see in my test db, which apparently did not have all the expected indices for some reason. Fixed
For reference, the complete list should be based on [MigrationToV1.scala#L39](https://github.com/AbsaOSS/enceladus/blob/develop/migrations/src/main/scala/za/co/absa/enceladus/migrations/migrations/MigrationToV1.scala#L39), I suppose.

> 
> Tested:
> 
> * basic flow (verified indexes)
> * existing DB db_version version set to 1
> * existing DB with no db_version
> * existing DB with db_version version set to 2

AdrianOlosutean (approved):
Looks ok, code review

Zejnilovic:
@dk1844 pushed the chmod +x commit

Zejnilovic:
> For reference, the complete list should be based on MigrationToV1.scala#L39, I suppose.

I think we did not update this one with the latest and greatest, or have we? We might need to enhance it.

Zejnilovic (approved):

PR #2018, opened 2022-01-21
Feature/2015 run & attachments not locked

### LINKED ISSUE #2015: Migration Script - Do not lock runs, attachments
## Background
Runs & attachments are not editable and should not be locked during migration.

## Feature
Do not lock Attachments & Runs when migrating. But keep migrated=true.

dk1844:
Migration internals has been changed, so that the in-progress migration mark is now just the `migrationHash={generatedValue}`, not locking of the entities.

This allows making the locking feature optional. Thus, only schemas, datasets, and MTs are getting locked, while runs and attachments do not have any lock-related fields present (`locked`, `userLocked`, `dateLocked`).

I have test-ran it with example data and the result seemed satisfactory.

## Release notes suggestion
`migrate_menas` only locks schemas, datasets and MTs on source; runs and attachments no longer get locked when migrated.

Merge after #2016 (based on it).

Closes #2015.


AdrianOlosutean:
Can you base this PR on #2016 ?

dk1844:
> Can you base this PR on #2016 ?

I believe it is based on that already, as I am stating in the introductory comment. ( I know that there are a lot of commits there as well, but after squash and merge, it should be fine. Unfortunately, I created started this one with #2016 not being merged yet

scripts/migration/menas_db.py
```diff
@@ -98,37 +98,37 @@ def ensure_collections_exist(collection_names: List[str]) -> None:
         # deliberately only checking migrating collections, not all (landing_page_statistics may not be present)
         return ensure_collections_exist(DATA_MIGRATING_COLLECTIONS)
 
-    def get_distinct_ds_names_from_ds_names(self, ds_names: List[str], not_locked_only: bool) -> List[str]:
-        return self.get_distinct_entities_ids(ds_names, DATASET_COLLECTION, not_locked_only)
+    def get_distinct_ds_names_from_ds_names(self, ds_names: List[str], migration_free_only: bool) -> List[str]:
```
Zejnilovic, line 101: `migration_free_only` sounds, for some reason, very funny to me. `not_migrated_only` or `non_migrated_only` sounds more standard. Not pushing it.

scripts/migration/initialize_menas.py
```diff
@@ -159,6 +159,7 @@ def run(parsed_args: argparse.Namespace):
             print(f"DB '{target_db_name}' contains version record, but it is invalid: {version_err}.\n")
             exit(10)
         else:
+            assert found_version != 1  # this would not be a valid MenasDbVersionError
```
Zejnilovic, line 162: This line makes the bellow if a bit obsolete, doesn't it?

Zejnilovic (commented):
Have not run yet

scripts/migration/initialize_menas.py
dk1844: What? if the version is not 1, it can be lower or higher (in theory). Or are you suggesting to rely on the fact that version=0 should never appear, because the pre-1 Menas DBs did not have `db_version` at all?

Zejnilovic: So, AFAIK, assert will throw an exception if `found_version not equal 1`. On line 163 there is if `found_version lower than 1` (also true if not equal). So the line on 163 is never reached thanks to the assertion on line 162. Right?

dk1844: > AFAIK, assert will throw an exception if `found_version not equal 1`

No, assert will throw an exception if `found_version` **is equal to 1**. So we have a guarantee it is not equal to one and we process further to throw the appropriate error message by comparison to the actual `found_version`.

Zejnilovic: My bad

dk1844: No worries 👍

Zejnilovic (approved):
Tested:
- Attachments migration
  - of already migrated attachments is skipped
  - lock is not added
- Runs migration 
  - of already migrated attachments is skipped
  - lock is not added
- Other entities locked + migration hash added

PR #2032, opened 2022-03-21
Merge/develop ver 3 to develop spark 3

AdrianOlosutean:


AdrianOlosutean:
One unusual thing I noticed in this merge is that a bunch of tests started to fail due to the missing `scala-compiler`dependency which was supposed to be included by `spark-commons`, but it is marked as provided in Enceladus parent pom. However, this problem didn't occur in `develop-ver-3`. At the moment, I just added it to spark-commons and rest-api

PR #2033, opened 2022-03-21
2031 spark partition sizing

### LINKED ISSUE #2031: Use spark-partition-sizing library

AdrianOlosutean:
Closes #2031

### DIFF
--- a/spark-jobs/src/main/resources/reference.conf
+++ b/spark-jobs/src/main/resources/reference.conf
@@ -80,9 +80,18 @@ spline.mode=BEST_EFFORT
 #
 spline.producer.url="http://localhost:8085/producer"
 
-#Block size in bytes needed for repartition/coalesce
+#possible values: plan, recordCount, schema, dataFrame, schemaSummaries, sample
+
+partition.strategy="plan"
+#Block size in bytes needed for repartition/coalesce, needed for any strategy except for recordCount
 #min.processing.partition.size=31457280
 #max.processing.partition.size=134217728
+
+#if sample is selected
+#partition.sample.size=3
+
+#if recordCount is selected
+#max.record.partition.count=100
 
 # Control plugins
 # Several plugins can be used. In this case the last element of the key needs to be incremented for each plugin.
--- a/spark-jobs/src/main/resources/reference.conf
+++ b/spark-jobs/src/main/resources/reference.conf
@@ -80,7 +80,7 @@ spline.mode=BEST_EFFORT
 #
 spline.producer.url="http://localhost:8085/producer"
 
-#possible values: plan, recordCount, schema, dataFrame, schemaSummaries, sample
+#possible values: plan, dataFrame, sample
 
 partition.strategy="plan"
 #Block size in bytes needed for repartition/coalesce, needed for any strategy except for recordCount
@@ -88,10 +88,7 @@ partition.strategy="plan"
 #max.processing.partition.size=134217728
 
 #if sample is selected
-#partition.sample.size=3
-
-#if recordCount is selected
-#max.record.partition.count=100
+#partition.sample.size=100
 
 # Control plugins
 # Several plugins can be used. In this case the last element of the key needs to be incremented for each plugin.
--- a/spark-jobs/src/main/scala/za/co/absa/enceladus/common/Repartitioner.scala
+++ b/spark-jobs/src/main/scala/za/co/absa/enceladus/common/Repartitioner.scala
@@ -16,6 +16,7 @@
 package za.co.absa.enceladus.common
 
 import org.apache.spark.sql.DataFrame
+import org.slf4j.Logger
 import za.co.absa.enceladus.common.config.CommonConfConstants
 import za.co.absa.enceladus.utils.config.ConfigReader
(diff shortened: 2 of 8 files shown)


spark-jobs/src/main/scala/za/co/absa/enceladus/common/Repartitioner.scala
```diff
@@ -0,0 +1,79 @@
...
+
+import org.apache.spark.sql.DataFrame
+import za.co.absa.enceladus.common.config.CommonConfConstants
+import za.co.absa.enceladus.utils.config.ConfigReader
+import za.co.absa.spark.partition.sizing.DataFramePartitioner.DataFrameFunctions
+import za.co.absa.spark.partition.sizing.sizer._
+import za.co.absa.spark.partition.sizing.types.DataTypeSizes
+import za.co.absa.spark.partition.sizing.types.DataTypeSizes.DefaultDataTypeSizes
+
+class Repartitioner(configReader: ConfigReader) {
```
Zejnilovic, line 26: I know I have been hands-off on this, but can anybody tell me why this is part of Enceladus, but seems as it should/could be part of the spark-partition-sizing library.

AdrianOlosutean: Well, the library is supposed to provide the functions themselves. This class is supposed to accomodate the functions with the configs

benedeki: I like the class as an isolation level to gather and provide (re-)partitioning settings separated from the main flow.
But I think the current implementation is too ambitious. See my other comments.

spark-jobs/src/main/resources/reference.conf
```diff
@@ -80,10 +80,19 @@ spline.mode=BEST_EFFORT
 #
 spline.producer.url="http://localhost:8085/producer"
 
-#Block size in bytes needed for repartition/coalesce
+#possible values: plan, recordCount, schema, dataFrame, schemaSummaries, sample
```
benedeki, line 83: `recordCount` with it's setting doesn't make much sense this way. Record count is dataset specific to be useful for effective partitioning. I think we can eventually use it, but with some more development and interaction with the rest of the system.

benedeki: Also in the current state I think `schema` and `schemaSummaries` is premature to be used. I think the data type estimated are still to wild for practical usage.

spark-jobs/src/main/resources/reference.conf
```diff
@@ -80,10 +80,19 @@ spline.mode=BEST_EFFORT
...
+
+partition.strategy="plan"
+#Block size in bytes needed for repartition/coalesce, needed for any strategy except for recordCount
 #min.processing.partition.size=31457280
 #max.processing.partition.size=134217728
 
+#if sample is selected
+#partition.sample.size=3
```
benedeki, line 91: Considering in statistics the smallest reasonable sample size is 100, it would use at least that, and wouldn't be shy of 1000.

spark-jobs/src/main/scala/za/co/absa/enceladus/common/Repartitioner.scala
```diff
@@ -0,0 +1,79 @@
...
+    }
+  }
+
+  private def repartitionBySchemaWithSummaries(df: DataFrame): DataFrame = {
+    val sizer = new FromSchemaWithSummariesSizer()
+    df.repartitionByDesiredSize(sizer)(minPartition, maxPartition)
+  }
+
+  private def repartitionBySample(df: DataFrame): DataFrame = {
+    val maybeInt = configReader.getIntOption(CommonConfConstants.partitionSampleSizeKey)
```
benedeki, line 52: I would use some default here and file a warning.
Alternatively no repartitioning as it is, but still a warning.

spark-jobs/src/main/scala/za/co/absa/enceladus/common/Repartitioner.scala
```diff
@@ -0,0 +1,79 @@
...
+import za.co.absa.enceladus.common.config.CommonConfConstants
+import za.co.absa.enceladus.utils.config.ConfigReader
+import za.co.absa.spark.partition.sizing.DataFramePartitioner.DataFrameFunctions
+import za.co.absa.spark.partition.sizing.sizer._
+import za.co.absa.spark.partition.sizing.types.DataTypeSizes
+import za.co.absa.spark.partition.sizing.types.DataTypeSizes.DefaultDataTypeSizes
+
+class Repartitioner(configReader: ConfigReader) {
+
+  val minPartition: Option[Long] = configReader.getLongOption(CommonConfConstants.minPartitionSizeKey)
```
benedeki, line 28: IIRC when both values are `None` no repartitioning will happen, despite it's set-up. Would deserve a warning in that case, I would say.

benedeki (approved):

benedeki (approved):

benedeki:
Will be tested with Spark 3

Issue #1693, opened 2021-02-17
Experimental API 3 - Versioned Model


Zejnilovic:
## Background
3rd part of the API rewrite. 
Approach this as a PoC. 
API is enough. Do not touch UI.
Think about where Envelope makes sense

## Feature
All of this should be put under profile "api-experimental" in spring or a feature switch

- /{model} 
  - GET return list of {model} and maybe latest versions
    - Optional param searchQuery
  - POST create or update {model}
- /{model}/import imports a single entity. Updates or creates
- /{model}/{name}
  - GET Returns list of versions available
  - POST update - full object
  - DELETE disable
- /{model}/{name}/audit-trail gets audit trail
- /{model}/{name}/{version}
  - Or keyword latest
  - GET returns the specific {model}
  - **Extra?**  - POST?PUT to update with partial body?
  - DELETE disables - **on {name} might be enough**
- /{model}/{name}/{version}/validate gets a validation report
- /{model}/{name}/{version}/properties
  - Get the properties
  - /{propertyName} for specific property then GET and POST
- /{model}/{name}/{version}/used-in
  - Get the list of places this is used in
  **- Would something similar be useful on {name}?**
- /{model}/{name}/{version}/export
- /datasets/{name}/{version}/rules
  - GET get the rules defined
  - POST create rule ?? Allow position/order to be specified ??
  - /{number or order}
    - GET get the specific rule
    - POST to update the rule
    - DELETE to delete the rule
    - **Q:** /reorder POST
- /mapping-tables/{name}/{version}/defaults
  - GET all defaults
  - POST create a default
  - /{column name} to get the specific default
    - DELETE to delete the rule
- /schema/{name}/{version}/json
- /schema/{name}/{version}/original-file
- /schema/{name}((/{version}))/upload - this needs a **PoC**. Do we want 3 endpoints or one?

dk1844:
Release notes
 - `/api-v3/datasets/` Rest API V3 added. Includes checks on entities and is to be used externally. The API is now truly RESTful (endpoint naming, structure, methods, Location header on creating responses). Also, Swagger API doc reacts to Spring dev profile (JVM arg `-Dspring.profiles.active=dev`) to show legacy V2 API doc, too.

PR #2046, opened 2022-04-04
Feature/1693 api v3 versioned model - Dataset v3 implementation

### LINKED ISSUE #1611: VersionedModelController.create should yield Location header
## Describe the bug
All extending classes of `VersionedModelController` define a `/create` endpoint that persists an entity when called with POST. Correct response `201 Created` is returned. Customarily, this should be accompanied by the Location header with the URI pointing to the created item. 

## To Reproduce
1. Call e.g. `PUT /api/dataset/create` with a valid DS content.
2. Inspect the response.

## Expected behaviour
Location header to be generally present with all entity-creating calls.

## Additional context
An example of this is implemented in `PropertyDefinitionController.createDatasetProperty`

dk1844:
This PR currently implements Enceladus API v3 for
 - general versioned model v3
   - note that general `/{modelName}/{entityName}/{versions}/validate` exists
 - dataset model v3; for the dataset API, a few changes have been done in the design:
   - .../properties sub-endpoints has been moved from the general versioned model to datasets (dataset-specific feature)
   - .../properties/{propertyName} has not been implemented, since the `.../properties` endpoint has no additional logic of validation (because of the existence of a general validation endpoint instead)

- later ADDED: schema and MT existence checking (those referenced in datasets (for MTs - in MCR))

- schemas, mapping tables, and property definitions are to follow; probably in another PR.

Partially implements #1693
Closes #1611 

## Release notes suggestion
`/api-v3/datasets/` Rest API V3 added. Includes checks on entities and is to be used externally. The API is now truly restful (enpoint naming, structure, methods, `Location` header on creating responses). Also Swagger API doc reacts to Spring `dev` profile (JVM arg `-Dspring.profiles.active=dev`) to show legacy V2 API doc, too.


rest-api/src/main/scala/za/co/absa/enceladus/rest_api/services/DatasetService.scala
```diff
@@ -93,7 +94,15 @@ class DatasetService @Autowired()(datasetMongoRepository: DatasetMongoRepository
...
   override def getUsedIn(name: String, version: Option[Int]): Future[UsedIn] = {
-    Future.successful(UsedIn())
+    val existingEntityF = version match {
+      case Some(version) => getVersion(name, version)
+      case None          => getLatestVersion(name)
+    }
+
+    existingEntityF.flatMap {
+      case Some(_) => Future.successful(UsedIn()) // empty usedIn for existing datasets
+      case None => Future.failed(NotFoundException(s"Dataset '$name' in version ${version.getOrElse("any")}' nof found"))
```
AdrianOlosutean, line 104:
```suggestion
      case None => Future.failed(NotFoundException(s"Dataset '$name' in version ${version.getOrElse("any")}' not found"))
```

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/services/DatasetService.scala
```diff
@@ -93,7 +94,15 @@ class DatasetService @Autowired()(datasetMongoRepository: DatasetMongoRepository
   }
 
   override def getUsedIn(name: String, version: Option[Int]): Future[UsedIn] = {
-    Future.successful(UsedIn())
+    val existingEntityF = version match {
```
AdrianOlosutean, line 97: I prefer without the type initial in the name

```suggestion
    val existingEntity = version match {
```

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/services/DatasetService.scala
```diff
@@ -428,10 +454,20 @@ object DatasetService {
...
     }
   }
 
+  /**
+   * Removes properties having empty-string value. Effectively mapping such properties' values from Some("") to None.
+   * This is Backend-implementation related to DatasetService.replaceBlankProperties(dataset) on Frontend
+   * @param properties original properties
+   * @return properties without empty-string value entries
+   */
+  def removeBlankProperties(properties: Map[String, String]): Map[String, String]  = {
```
AdrianOlosutean, line 467: Good idea 👍

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/services/DatasetService.scala
```diff
@@ -114,8 +123,24 @@ class DatasetService @Autowired()(datasetMongoRepository: DatasetMongoRepository
     }
   }
 
-  def replaceProperties(username: String, datasetName: String,
-                        updatedProperties: Option[Map[String, String]]): Future[Option[Dataset]] = {
+  def updateProperties(username: String, datasetName: String, datasetVersion: Int,
+                       updatedProperties: Map[String, String]): Future[Option[Dataset]] = {
+    for {
+      s <- validateProperties(updatedProperties).flatMap {
+        case validation if !validation.isValid => Future.failed(ValidationException(validation)) // warnings are ok for update
```
AdrianOlosutean, line 130: Maybe warnings could be logged?

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/controllers/v3/VersionedModelControllerV3.scala
```diff
@@ -0,0 +1,217 @@
...
+  @GetMapping(Array("/{name}/{version}/export"))
+  @ResponseStatus(HttpStatus.OK)
+  def exportSingleEntity(@PathVariable name: String, @PathVariable version: String): CompletableFuture[String] = {
+    forVersionExpression(name, version)(versionedModelService.exportSingleItem)
+  }
+
+  @GetMapping(Array("/{name}/export"))
+  @ResponseStatus(HttpStatus.OK)
+  def exportLatestEntity(@PathVariable name: String): CompletableFuture[String] = {
+    versionedModelService.exportLatestItem(name) // todo: remove in favor of the above? (that supports /{name}/latest/export)
```
AdrianOlosutean, line 95: From my point of view, this can stay

AdrianOlosutean (commented):
I like it overall, just some small comments.

Also, do you plan to change the frontend to use this api?

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/services/DatasetService.scala
dk1844: Thanks for noticing, fixing it.

data-model/src/main/scala/za/co/absa/enceladus/model/versionedModel/VersionedSummary.scala
```diff
@@ -16,3 +16,5 @@
 package za.co.absa.enceladus.model.versionedModel
 
 case class VersionedSummary(_id: String, latestVersion: Int)
+
+case class VersionsList(_id: String, versions: Seq[Int])
```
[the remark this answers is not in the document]
dk1844: Thanks, makes more sense, `VersionList` it is.

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/services/DatasetService.scala
dk1844: Sure, that's better.

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/controllers/v3/VersionedModelControllerV3.scala
dk1844: Well, this is part of the v3 API outline. I consider this misleading, because since you have the option to export specific versions already ( at .../{model}/{name}/{versionEitherNumberOrLatestKeyword}/export, having the option for export here would have to have another meaning - e.g. export the whole entity with all the versions. But since this is designed as latest-export as well, I consider this unnecessary, and as I said, even misleading.

That is why I am proposing to omit this endpoint entirely.

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/services/DatasetService.scala
dk1844: I am considering doing even a better thing -> to communicate it back to the caller as payload.

dk1844:
> ...
> Also, do you plan to change the frontend to use this API?

In the end, I think the UI should get changed to use this API. But that is obviously outside of the scope of the API update PRs.

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/services/DatasetService.scala
dk1844: Done.

AdrianOlosutean (approved):

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/services/v3/DatasetServiceV3.scala
```diff
@@ -0,0 +1,78 @@
...
+    }
+  }
+
+  // general entity validation is extendable for V3 - here with properties validation
+  override def validate(item: Dataset): Future[Validation] = {
+    for {
+      originalValidation <- super.validate(item)
+      propertiesValidation <- validateProperties(item.propertiesAsMap)
+      schemaValidation <- validateSchemaExists(item.schemaName, item.schemaVersion)
+      rulesValidation <- validateRules(item)
```
benedeki, line 60: Ditto with concurrency:
```suggestion
    val originalValidationFuture = super.validate(item)
    val propertiesValidationFuture = validateProperties(item.propertiesAsMap)
    val schemaValidationFuture = validateSchemaExists(item.schemaName, item.schemaVersion)
    val rulesValidationFuture = validateRules(item)
    for {
      originalValidation <- originalValidationFuture 
      propertiesValidation <- propertiesValidationFuture 
      schemaValidation <- schemaValidationFuture 
      rulesValidation <- rulesValidationFuture 
```

data-model/src/main/scala/za/co/absa/enceladus/model/versionedModel/VersionedSummary.scala
```diff
@@ -16,3 +16,5 @@
 package za.co.absa.enceladus.model.versionedModel
 
 case class VersionedSummary(_id: String, latestVersion: Int)
+
+case class VersionList(_id: String, versions: Seq[Int])
```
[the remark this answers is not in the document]
dk1844: In theory, it could. Or even `maxVersion` would do (assuming that all entities start with v1).

Attachments are the exception, they take after the schema version and there may be holes, but attachments are not exposed directly.

Then there is the matter of not being able to prove that all versions are actually there in the DB, we assume they should be, but... how sure can you be. 

I would be hesitant to provide such guarantees right of the bat.

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/controllers/DatasetController.scala
```diff
@@ -113,7 +113,7 @@ class DatasetController @Autowired()(datasetService: DatasetService)
   def replaceProperties(@AuthenticationPrincipal principal: UserDetails,
                         @PathVariable datasetName: String,
                         @RequestBody newProperties: Optional[Map[String, String]]): CompletableFuture[ResponseEntity[Option[Dataset]]] = {
-    datasetService.replaceProperties(principal.getUsername, datasetName, newProperties.toScalaOption).map {
+    datasetService.updatePropertiesV2(principal.getUsername, datasetName, newProperties.toScalaOption).map {
```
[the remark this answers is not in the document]
dk1844: No, this has nothing to do with spring autowiring. This is just a way to have V2 and V3 implementation in some services so that the v2 version can be removed easily when only v3 is to stay.

If we want to keep v2 and v3 long-term, this would require to create separate services, e.g. `datasetService` and `datasetServiceV3`. One of these solutions is, I believe, necessary if you want to keep the original implementation working.

It then begs the question - what is the intended lifetime of V2 API? Knowing this better, it could help to design better: _copy-paste-and-change-a-few-things knowing that the V2 impl will be gone soon_ vs. keeping v3 extending v2 (no code duplication, but harder removal)

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/services/DatasetService.scala
```diff
@@ -196,7 +222,7 @@ class DatasetService @Autowired()(datasetMongoRepository: DatasetMongoRepository
     }
   }
 
-  def validateProperties(properties: Map[String, String], forRun: Boolean): Future[Validation] = {
+  def validateProperties(properties: Map[String, String], forRun: Boolean = false): Future[Validation] = {
```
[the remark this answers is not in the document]
dk1844: This is done because V3 needs to validate properties, but the concept of `forRun` validation is not considered there.

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/services/VersionedModelService.scala
```diff
@@ -189,9 +201,12 @@ abstract class VersionedModelService[C <: VersionedModel with Product with Audit
...
   }
 
-  private[rest_api] def create(item: C, username: String): Future[Option[C]] = {
+  private[rest_api] def create(item: C, username: String): Future[Option[(C, Validation)]] = {
     for {
-      validation <- validate(item)
+      validation <- for {
+        generalValidation <- validate(item)
+        creationValidation <- validateForCreation(item)
+      } yield generalValidation.merge(creationValidation)
```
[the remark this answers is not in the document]
dk1844: Yes, I don't see it as a problem (_faster multiple-at-once_ vs _possibly not needed_ - I don't even see a clear winner here)

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/services/v3/DatasetServiceV3.scala
dk1844: same as above.

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/services/DatasetService.scala
```diff
@@ -428,10 +455,20 @@ object DatasetService {
...
     }
   }
 
+  /**
+   * Removes properties having empty-string value. Effectively mapping such properties' values from Some("") to None.
+   * This is Backend-implementation related to DatasetService.replaceBlankProperties(dataset) on Frontend
+   * @param properties original properties
+   * @return properties without empty-string value entries
+   */
+  def removeBlankProperties(properties: Map[String, String]): Map[String, String]  = {
```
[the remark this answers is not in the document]
dk1844: Sure, applying.

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/SpringFoxConfig.scala
```diff
@@ -39,11 +39,17 @@ class SpringFoxConfig extends ProjectMetadata {
...
+      // api v2
+      regex("/api/dataset.*"), regex("/api/schema.*"),
       regex("/api/mappingTable.*"), regex("/api/properties.*"),
       regex("/api/monitoring.*"),regex("/api/runs.*"),
       regex("/api/user.*"), regex("/api/spark.*"),
-      regex("/api/configuration.*")
+      regex("/api/configuration.*"),
+
+      // api v3
+        regex("/api-v3/datasets.*")
```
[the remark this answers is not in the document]
dk1844: The paths are not the same otherwise anyway, I have added other paths that we need there and it seems more sensible to spell it out in this place.

data-model/src/main/scala/za/co/absa/enceladus/model/versionedModel/VersionedSummary.scala
```diff
@@ -16,3 +16,5 @@
 package za.co.absa.enceladus.model.versionedModel
 
 case class VersionedSummary(_id: String, latestVersion: Int)
+
```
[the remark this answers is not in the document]
dk1844: true, extracted.

AdrianOlosutean (dismissed):
Looks good to me

data-model/src/main/scala/za/co/absa/enceladus/model/versionedModel/VersionedSummary.scala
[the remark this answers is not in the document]
dk1844: Perhaps let's just use just VersionedSummary - with the latest version

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/controllers/DatasetController.scala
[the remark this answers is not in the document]
dk1844: TODO ... review again

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/SpringFoxConfig.scala
```diff
@@ -39,11 +39,18 @@ class SpringFoxConfig extends ProjectMetadata {
...
-    or[String](regex("/api/dataset.*"), regex("/api/schema.*"),
+    or[String](
+      // api v2
+      regex("/api/dataset.*"), regex("/api/schema.*"),
       regex("/api/mappingTable.*"), regex("/api/properties.*"),
-      regex("/api/monitoring.*"),regex("/api/runs.*"),
+      regex("/api/monitoring.*"), regex("/api/runs.*"),
       regex("/api/user.*"), regex("/api/spark.*"),
-      regex("/api/configuration.*")
+      regex("/api/configuration.*"),
```
Zejnilovic, line 48: I would not document v2 API as we do not want people to use it. Thoughts?

Maybe spring fox could have multiple profiles? Dev and non-dev

AdrianOlosutean: I agree with not documenting v2

data-model/src/main/scala/za/co/absa/enceladus/model/versionedModel/VersionedSummary.scala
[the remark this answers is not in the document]
dk1844: Done.

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/controllers/DatasetController.scala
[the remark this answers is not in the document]
dk1844: The v3 method moved to DatasetServiceV3, so now it is tidier. Thanks for the tip, @benedeki.

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/services/VersionedModelService.scala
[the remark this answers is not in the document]
dk1844: Explanatory comment added, thanks for the tip.

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/services/v3/DatasetServiceV3.scala
dk1844: Explanatory comment added, thanks for the tip.

data-model/src/main/scala/za/co/absa/enceladus/model/versionedModel/VersionedSummary.scala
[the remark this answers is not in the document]
dk1844: removed in the end -- reaction to the above.

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/controllers/v3/VersionedModelControllerV3.scala
```diff
@@ -0,0 +1,218 @@
...
+  }
+
+  @PostMapping(Array("/{name}/import"))
+  @ResponseStatus(HttpStatus.CREATED)
+  def importSingleEntity(@AuthenticationPrincipal principal: UserDetails,
+                         @PathVariable name: String,
+                         @RequestBody importObject: ExportableObject[C],
+                         request: HttpServletRequest): CompletableFuture[ResponseEntity[Validation]] = {
+    if (name != importObject.item.name) {
+      Future.failed(new IllegalArgumentException(s"URL and payload entity name mismatch: '$name' != '${importObject.item.name}'"))
```
Zejnilovic, line 105: I am not sure I like these plaintext methods. I have a feeling 
`{ "message": "XYZ" }` would fit better.

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/controllers/v3/VersionedModelControllerV3.scala
```diff
@@ -0,0 +1,218 @@
...
+  // todo maybe offset/limit?
+  @GetMapping(Array(""))
+  @ResponseStatus(HttpStatus.OK)
+  def getList(@RequestParam searchQuery: Optional[String]): CompletableFuture[Seq[VersionedSummary]] = {
+    versionedModelService.getLatestVersionsSummary(searchQuery.toScalaOption)
+  }
+
+  @GetMapping(Array("/{name}"))
+  @ResponseStatus(HttpStatus.OK)
+  def getVersionsList(@PathVariable name: String): CompletableFuture[VersionList] = {
```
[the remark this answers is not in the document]
Zejnilovic: Sorry already fixed

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/controllers/v3/VersionedModelControllerV3.scala
```diff
@@ -0,0 +1,218 @@
...
+  }
+
+
+  @GetMapping(Array("/{name}/audit-trail"))
+  @ResponseStatus(HttpStatus.OK)
+  def getAuditTrail(@PathVariable name: String): CompletableFuture[AuditTrail] = {
+    versionedModelService.getAuditTrail(name)
+  }
+
+  @GetMapping(Array("/{name}/{version}/used-in"))
```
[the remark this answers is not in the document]
dk1844: Here, I think it makes sense. 
Since the _disable_ is available for all entity types, so should _used-in_. The nice thing about it is that there can be a common approach to entities:
1. making sure that there is anything preventing from disabling (used-in)
2. dealing with linked references returned by used-in if any
3. disabling the entity knowing that it should succeed (without surprises)

It is quite reassuring for the API user to be able to call `used-in` even if currently for a certain type there cannot be any entity linked, it perhaps may be in the future?

All in all, I am a fan of having `used-in`s everywhere (where _disable_ is).

data-model/src/main/scala/za/co/absa/enceladus/model/versionedModel/VersionedSummary.scala
```diff
@@ -16,3 +16,5 @@
 package za.co.absa.enceladus.model.versionedModel
 
 case class VersionedSummary(_id: String, latestVersion: Int)
```
Zejnilovic, line 18: Since we are moving this towards ENC v 3 can we have a change that this is not `_id` but `name`?

Also, I am thinking how helpful a description would be at this point.

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/controllers/v3/VersionedModelControllerV3.scala
Zejnilovic: To further support my comment. 404 response is 
```json
{
	"timestamp": "2022-05-04T08:01:40.637+0000",
	"status": 404,
	"error": "Not Found",
	"message": "No message available",
	"path": "/menas/api-v3/mapping-tables/SourceSystemMappingTable/latest/default"
}
```

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/controllers/v3/VersionedModelControllerV3.scala
```diff
@@ -0,0 +1,218 @@
...
+
+  protected def createdWithNameVersionLocationBuilder(name: String, version: Int, request: HttpServletRequest,
+                                                      stripLastSegments: Int = 0, suffix: String = ""): ResponseEntity.BodyBuilder = {
+    val strippingPrefix = Range(0, stripLastSegments).map(_ => "/..").mkString
+
+    val location: URI = ServletUriComponentsBuilder.fromRequest(request)
+      .path(s"$strippingPrefix/{name}/{version}$suffix")
+      .buildAndExpand(name, version.toString)
+      .normalize() // will normalize `/one/two/../three` into `/one/tree`
+      .toUri() // will create location e.g. http:/domain.ext/api-v3/dataset/MyExampleDataset/1
```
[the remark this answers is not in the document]
dk1844: Suggestion applied. 👍

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/controllers/v3/VersionedModelControllerV3.scala
```diff
@@ -0,0 +1,218 @@
...
+
+    // hint on "/.." + normalize https://github.com/spring-projects/spring-framework/issues/14905#issuecomment-453400918
+
+    ResponseEntity.created(location)
+  }
+
+}
+
+object VersionedModelControllerV3 {
```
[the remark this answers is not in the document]
dk1844: Sure, why not. Done.

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/controllers/v3/VersionedModelControllerV3.scala
dk1844: After even further discussion, I have decided to remove the endpoint based on the following logic:
The structure of API endpoints representing a collection should, in general, behave in the following fashion:
 - `GET /entityNamePlural` -> list items or provide information about all/more items
 - `GET /entityNamePlural/individualEntityKeyEgId` -> give information about a specific entity

By that logic, `GET /{name}/export` would be appropriate if all entities or information about all/more was returned -- perhaps all versions, while `GET /{name}/{version}/export` would export just a single version.

But since we mapped just the single version (the latest one) export to a collection's endpoint, the logic does not fit. Moreover, given the fact that we also accept `latest` as a valid version identifier, it seem much more appropriate to use `GET /{name}/{version}/export` here (specifically `GET /{name}/latest/export`)

data-model/src/main/scala/za/co/absa/enceladus/model/versionedModel/VersionedSummary.scala
dk1844: I have created another case class that carries out the change (`_id`-> `name`). But only for API v3, because for V2 the original implementation. had to stay. And actually, the original case class is designed specifically as the mongo response would look like, so it is still useful

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/controllers/v3/VersionedModelControllerV3.scala
dk1844: @AdrianOlosutean @benedeki @Zejnilovic

dk1844: `RestResponse` or `RestError` application come in mind, perhaps

Zejnilovic: Will be part of the next PR. Created issue #2062

data-model/src/main/scala/za/co/absa/enceladus/model/versionedModel/VersionedSummary.scala
Zejnilovic: I think I might've posted the comment under the wrong line of code but what I am getting is 

image

And I expect the `_id` to be `name`. Am I doing something wrongly?

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/SpringFoxConfig.scala
dk1844: So, I have changed the implementation to the following: 
 - JVM being supplied `-Dspring.profiles.active=dev` will result in full V2 + V3 API as before, 
 - by default (no special profile setup) will only allow Swagger to show API V3 + import/export methods of API v2 (these we consider public API in V2 already).

Hope this is satisfactory.

### Showcase DEV `-Dspring.profiles.active=dev`
menas-api-dev


### Showcase Non-DEV
menas-api-nondev

data-model/src/main/scala/za/co/absa/enceladus/model/versionedModel/VersionedSummary.scala
Zejnilovic: Have it.

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/SpringFoxConfig.scala
```diff
@@ -28,27 +28,55 @@ import za.co.absa.enceladus.utils.general.ProjectMetadata
...
+    )
+    val v3paths = Seq(
+      regex("/api-v3/datasets.*"), regex("/api-v3/schemas.*"),
+      regex("/api-v3/mapping-tables.*"), regex("/api-v3/property-definitions.*")
+    )
+
+    val paths: Seq[Predicate[String]] = if (isDev) {
+      v2devPaths ++ v3paths
+    } else {
+      v2prodPaths ++ v3paths
```
Zejnilovic, line 69:
```suggestion
      v3paths
```

So prod APIs should be only v3 paths as the import-export being "prod" is only an ABSA thing. 

This would result in dev having all v2 and v3 paths and prod only v3.

dk1844: Thanks for the clarification, suggestion applied (along with removing the unnecessary code)

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/SpringFoxConfig.scala
dk1844: Redone due to https://github.com/AbsaOSS/enceladus/pull/2046#discussion_r867700135.
Menas DEV API remains unchanged, for Menas non-DEV API now only contains V3 API endpoints (i.e. V2 import/export removed)

Zejnilovic (approved):
Code reviewed
Api tested

benedeki (approved):

PR #2055, opened 2022-04-28
Feature/1693 api v3 delete recreate

dk1844:
- `usedIn` functionality both for `.../{name}/used-in` and `.../{name}/{version}/used-in` exists
   - for depencies of: schema->dataset, schema->mapping tables, mapping tables -> dataset, propdef->dataset (all based on version-less used-in)
 - disable for entities via `DELETE .../{name}`, enable for entities via `PUT .../{name}`
   - enable/disable is done for the whole entity (all versions at once), changes only possible on enabled entities (`!isDisabled` check).
 - `UsedIn` normalization, empty - V2 had it mixed (sometimes `None`, sometimes `Some(Seq.empty)` -> now it all `None`s. + unittest
 - all of the above reflected in integTests.

Partially implements https://github.com/AbsaOSS/enceladus/issues/1693

# Release notes suggestion
Both `/api-v3/{datasets|schemas|mapping-tables|property-defintions}/{name}/used-in` and `/api-v3/{datasets|schemas|mapping-tables|property-defintions}/{name}/{version}/used-in` REST API V3 added, the former is used for disable requests to check the dependencies. `UsedIn` normalization introduced.


data-model/src/main/scala/za/co/absa/enceladus/model/UsedIn.scala
```diff
@@ -15,15 +15,42 @@
...
 import za.co.absa.enceladus.model.menas.MenasReference
 
-case class UsedIn(
-  datasets: Option[Seq[MenasReference]] = None,
-  mappingTables: Option[Seq[MenasReference]] = None
-) {
+case class UsedIn(datasets: Option[Seq[MenasReference]] = None,
+                  mappingTables: Option[Seq[MenasReference]] = None) {
 
-  def nonEmpty: Boolean = {
```
dk1844, line 25: In order to assess correctness of the rewrite, the unit test for nonEmpty was written first and the implementation was changed after that.

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/repositories/DatasetMongoRepository.scala
```diff
@@ -77,13 +76,24 @@ class DatasetMongoRepository @Autowired()(mongoDb: MongoDatabase)
    * @return List of Menas references to Datasets, which contain the relevant conformance rules
    */
   def containsMappingRuleRefEqual(refColVal: (String, Any)*): Future[Seq[MenasReference]] = {
```
dk1844, line 78: Originally written by @lokm01, but some thing were missing (enabled check, sorting), but it it nicely general

rest-api/src/test/scala/za/co/absa/enceladus/rest_api/integration/controllers/DatasetApiIntegrationSuite.scala
```diff
@@ -437,4 +436,48 @@ class DatasetApiIntegrationSuite extends BaseRestApiTestV2 with BeforeAndAfterAl
     }
   }
 
+  s"DELETE $apiUrl/disable/{name}/{version}" can {
```
dk1844, line 439: Note, that this is the original V2 API - adding test cases here as a regression guard

AdrianOlosutean (dismissed):

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/controllers/v3/VersionedModelControllerV3.scala
```diff
@@ -148,25 +156,42 @@ abstract class VersionedModelControllerV3[C <: VersionedModel with Product
...
+        DisabledPayload(disabled = false)
+      } else {
+        throw NotFoundException(s"No versions for entity $name found to be enabled.")
+      }
+    }
+  }
+
+  @DeleteMapping(Array("/{name}"))
+  @ResponseStatus(HttpStatus.OK)
+  def disable(@PathVariable name: String): CompletableFuture[DisabledPayload] = {
```
Zejnilovic, line 187: I get that we block disabling if there is something using it. And I get that we need to show user what is wrong.
 But getting back a list `used-in` in this call might be weird and unexpected. Can we have something like 

```json
{
  "disabled": "true/false",
  "errors": ["whatever you need (like used-in) in here"]
}
```

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/controllers/v3/VersionedModelControllerV3.scala
```diff
@@ -148,25 +156,42 @@ abstract class VersionedModelControllerV3[C <: VersionedModel with Product
...
+  @PutMapping(Array("/{name}"))
   @ResponseStatus(HttpStatus.OK)
-  def disable(@PathVariable name: String,
-              @PathVariable version: Optional[String]): CompletableFuture[UpdateResult] = {
-    val v = if (version.isPresent) {
-      // For some reason Spring reads the Optional[Int] param as a Optional[String] and then throws ClassCastException
-      Some(version.get.toInt)
-    } else {
-      None
+  def enable(@PathVariable name: String): CompletableFuture[DisabledPayload] = {
```
Zejnilovic, line 175: Since we block the disable if the entity is used. Why are we not blocking re-enables if the entity is missing schema.

Step by step:
1. Disable datasetX that uses schemaX
2. Disable schemaX 
3. Enable datasetX passes. 

I would expect it to fail

Zejnilovic (commented):
Tested the disabling and enabling + used-in on name and on name+version.

I am curious. When entity is disable and somebody knows its name he can still "use it" which seems a tad weird. Maybe we should come up with a definition of disabled. Like
- disallow all PUT/POST calls except for PUT on /{name} to enable.
- validate should return `error disabled` so it cannot be run in spark-jobs.
- maybe also:
  - disallow export
  - disallow audit-trail
  - on GET /{name} print name, version and disabled true/false (in general not just here)
  - on GET /{name}/{version} just print basic info and say it is disabled

Happy to discuss more

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/controllers/v3/VersionedModelControllerV3.scala
```diff
@@ -148,25 +156,42 @@ abstract class VersionedModelControllerV3[C <: VersionedModel with Product
     } else if (version != item.version) {
       Future.failed(new IllegalArgumentException(s"URL and payload version mismatch: ${version} != ${item.version}"))
     } else {
-      versionedModelService.update(user.getUsername, item).map {
-        case Some((updatedEntity, validation)) =>
-          createdWithNameVersionLocationBuilder(updatedEntity.name, updatedEntity.version, request, stripLastSegments = 2).body(validation)
-        case None => throw notFound()
+      versionedModelService.isDisabled(item.name).flatMap { isDisabled =>
+        if (isDisabled) {
+          throw EntityDisabledException(s"Entity ${item.name} is disabled. Enable it first to create new versions.")
```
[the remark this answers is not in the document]
dk1844: Practically, there is no difference. `Future.failed` may be a bit more efficient approach, but on the other hand, in that case you need to make sure that you are mapping/flatMapping correctly.

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/controllers/v3/VersionedModelControllerV3.scala
dk1844: Well, I think it ok, because you can only expect to receive the expected object/message type on success - the `DisabledPayload` here. 
On error, you may get different response types and based on what is wrong, you get different things - e.g. 404 (no payload or message), 400 - a `UsedIn`, a `Validation` etc.

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/controllers/v3/VersionedModelControllerV3.scala
dk1844: Nice catch! Yes, this should be probably checked, too. Thank you.

Zejnilovic:
> Tested the disabling and enabling + used-in on name and on name+version.
> 
> I am curious. When entity is disable and somebody knows its name he can still "use it" which seems a tad weird. Maybe we should come up with a definition of disabled. Like
> 
> * disallow all PUT/POST calls except for PUT on /{name} to enable.
> * validate should return `error disabled` so it cannot be run in spark-jobs.
> * maybe also:
>   
>   * disallow export
>   * disallow audit-trail
>   * on GET /{name} print name, version and disabled true/false (in general not just here)
>   * on GET /{name}/{version} just print basic info and say it is disabled
> 
> Happy to discuss more


We agreed on:
- disallow all PUT/POST calls except for PUT on /{name} to enable.
- validate should return error disabled so it cannot be run in spark-jobs.
- on GET /{name} print name, version and status

dk1844:
> We agreed on:
> 
> * disallow all PUT/POST calls except for PUT on /{name} to enable.
// todo, will edit

> * validate should return error disabled so it cannot be run in spark-jobs.
// todo, will edit

> * on GET /{name} print name, version and status
 - implemented for V3 API by adding the `disabled` field to that "summary"

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/controllers/v3/VersionedModelControllerV3.scala
```diff
@@ -87,6 +89,12 @@ abstract class VersionedModelControllerV3[C <: VersionedModel with Product
     forVersionExpression(name, version) { case (name, versionInt) => versionedModelService.getUsedIn(name, Some(versionInt)) }
   }
 
+  @GetMapping(Array("/{name}/used-in"))
+  @ResponseStatus(HttpStatus.OK)
+  def usedIn(@PathVariable name: String): CompletableFuture[UsedIn] = {
```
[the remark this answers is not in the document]
dk1844: No, those should be probably `@JsonIgnore`d. Thanks for noticing and done.

Btw. On that account, I have also noticed that should there be any actual existing `MenasReference`s, their `collection` is always empty - I have searched this to be the case since 2019 when it appeared and it is also recorded exactly like this in the integration tests. I wonder, however, why is this so...?

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/controllers/v3/VersionedModelControllerV3.scala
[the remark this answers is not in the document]
dk1844: Got me wondering about the efficiency and found [this](https://stackoverflow.com/a/31195611/1773349).

It, nevertheless, supports the original claims: the same outcome with `Future.failed` being the more efficient way.

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/controllers/v3/VersionedModelControllerV3.scala
Zejnilovic: image

I will be honest. It took me a long time to get what is happening. What am I seeing and why I only understood it from the code. 

Purely logical it might make sense but it is not user friendly and I vote we do it a tad differently. That is add a message and then payload or something.

data-model/src/main/scala/za/co/absa/enceladus/model/versionedModel/NamedVersion.scala
```diff
@@ -15,4 +15,7 @@
 
 package za.co.absa.enceladus.model.versionedModel
 
-case class NamedVersion(name: String, version: Int)
+/**
+ * V3 Wrapper for [[za.co.absa.enceladus.model.versionedModel.VersionedSummary]]
+ */
+case class NamedVersion(name: String, version: Int, disabled: Boolean)
```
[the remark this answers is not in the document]
dk1844: Sure, we could. This was the most straightforward solution, clear to users. I find this adequate, but I am open to suggestions.

Perhaps consider suggesting the `def fn(disabledStatuses: Set[Boolean]): String` what you have in mind.

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/controllers/v3/VersionedModelControllerV3.scala
dk1844: This will not be trivial -> https://github.com/AbsaOSS/enceladus/issues/2065

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/controllers/v3/VersionedModelControllerV3.scala
dk1844: Wrapper with error message added.

AdrianOlosutean (approved):
Nice job, looks good to me. Code reviewed

data-model/src/main/scala/za/co/absa/enceladus/model/versionedModel/NamedVersion.scala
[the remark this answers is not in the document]
dk1844: Keeping as is due to offline discussions suggesting this is sufficient.

Zejnilovic (approved):
Reviewed, built, run.

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/controllers/v3/DatasetControllerV3.scala
```diff
@@ -57,14 +58,12 @@ class DatasetControllerV3 @Autowired()(datasetService: DatasetServiceV3)
         case Some((entity, validation)) =>
```
AdrianOlosutean, line 58: Shouln't this endpoint have the same validation?

Issue #2065, opened 2022-05-16
V3 API enable should check dependencies


dk1844:
## Background & Feature
Originating in https://github.com/AbsaOSS/enceladus/pull/2055#discussion_r870068092, the idea is that when enabling an entity, all its dependencies should be checked to be enabled as well.

## Feature
Since V3 API considers entity disable/enable as a single state (going forward), dependencies of all entity versions must be checked.

## Example
`DatasetA` v1 has `schemaA` v2 and mapping CR that is tied to `mappingTableA` v6 and also defines `propertyDefinionA`
`DatasetA v2 has `schemaB` v3, no mapping CR and defines `propertyDefintionB`.

`DatasetA` is disabled. 
When trying to (re)enable it, `schemaA` in v2, `schemaB` v3, `mappingTableA` v6, `propertyDefintionA|B` (all versions) must be check to be enabled in order for DatasetA to be correctly enabled.

## Proposed Solution [Optional]
Solution Ideas:
1. Utilize the fact that disabled entities fail validation? (but dependent entities failing validation for other reasons might block enabling entities)
2. Specifically create checks for the `disabled` state (not tied to validation itself) based on just unique entity names
3. point 2, but also checking versions because historically, there can be entities in mixed disabled/enabled state

PR #2066, opened 2022-05-16
Feature/1693 api v3 disabled fails validation

dk1844:
Implementation of PR review https://github.com/AbsaOSS/enceladus/pull/2055#issuecomment-1124699091, specifically:
> - disallow all PUT/POST calls except for PUT on /{name} to enable.
> - validate should return error disabled so it cannot be run in spark-jobs.

It has its own PR, because there are a lot of changes in the end (introducing `VersionedModelServiceV3` - and changing some abstract classes to traits in order to multi-inherit in specific services both from `VersionedModelServiceV3` (API V3 backing) and from V2 services that already hold a lot of reusable implementation.

Partial for #1693


rest-api/src/main/scala/za/co/absa/enceladus/rest_api/services/RunService.scala
```diff
@@ -31,8 +31,10 @@ import scala.concurrent.Future
 import scala.util.{Failure, Success, Try}
 
 @Service
-class RunService @Autowired()(runMongoRepository: RunMongoRepository)
-  extends ModelService(runMongoRepository) {
+class RunService @Autowired()(val mongoRepository: RunMongoRepository)
+  extends ModelService[Run] {
+
+  protected val runMongoRepository: RunMongoRepository = mongoRepository // alias
```
Zejnilovic, line 37: Is this just so you do not change the methods too much?

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/services/v3/VersionedModelServiceV3.scala
```diff
@@ -0,0 +1,105 @@
...
+  import scala.concurrent.ExecutionContext.Implicits.global
+
+  override private[services] val logger = LoggerFactory.getLogger(this.getClass)
+
+  def getLatestVersionSummary(name: String): Future[Option[VersionedSummary]] = {
+    mongoRepository.getLatestVersionSummary(name)
+  }
+
+  // v3 has internal validation on importItem (because it is based on update), v2 only had it on import
+  override def importSingleItem(item: C, username: String, metadata: Map[String, String]): Future[Option[(C, Validation)]] = {
```
AdrianOlosutean, line 42: Why is the metadata passed here?

dk1844: Thanks for noticing and making me realizing the problem! 🙇‍♂️ 
The metadata is passed there, because it should have been checked, too, but this implementation failed to do so. 

Should be fixed now, I have updated the code and added test cases for both V2 and V3 import API to hold this together.

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/services/v3/VersionedModelServiceV3.scala
```diff
@@ -0,0 +1,111 @@
...
+
+  /**
+   * Enables all versions of the entity by name.
+   * @param name
+   */
+  def enableEntity(name: String): Future[UpdateResult] = {
+    val auth = SecurityContextHolder.getContext.getAuthentication
+    val principal = auth.getPrincipal.asInstanceOf[UserDetails]
+
+    // todo check validation of used-in dependees? #2065
```
AdrianOlosutean, line 72: I think this comment can be deleted

AdrianOlosutean (dismissed):
Looks good, code review

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/services/v3/VersionedModelServiceV3.scala
dk1844: No, we cannot, because feature #2065 is still missing and it should be implemented here. But currently, it has not been given priority.

AdrianOlosutean (approved):
LGTM

AdrianOlosutean (commented):
I was testing the API and I think DatasetControllerV3 should also have this validation on updating the properties: PUT on"/{name}/{version}/properties"

dk1844:
> I was testing the API and I think DatasetControllerV3 should also have this validation on updating the properties: PUT on"/{name}/{version}/properties"

What do you mean? `DatasetServiceV3.updateProperties()` does validate properties being updated. Which part of validation do you consider missing, @AdrianOlosutean ?

AdrianOlosutean:
When I tested this endpoint on a disabled dataset, I didn't see the disabled validation

> > I was testing the API and I think DatasetControllerV3 should also have this validation on updating the properties: PUT on"/{name}/{version}/properties"
> 
> What do you mean? `DatasetServiceV3.updateProperties()` does validate properties being updated. Which part of validation do you consider missing, @AdrianOlosutean ?

dk1844:
> When I tested this endpoint on a disabled dataset, I didn't see the disabled validation
> 
> > > I was testing the API and I think DatasetControllerV3 should also have this validation on updating the properties: PUT on"/{name}/{version}/properties"
> > 
> > 
> > What do you mean? `DatasetServiceV3.updateProperties()` does validate properties being updated. Which part of validation do you consider missing, @AdrianOlosutean ?

I can see this covered in the (passing) integTests:
[direct file](https://github.com/AbsaOSS/enceladus/blob/b3ffe439847d2434a034581cb0fe32943ca4da82/rest-api/src/test/scala/za/co/absa/enceladus/rest_api/integration/controllers/v3/DatasetControllerV3IntegrationSuite.scala#L737)
[diff view](https://github.com/AbsaOSS/enceladus/pull/2066/files#diff-968796a28b59cae88f5bf58aa45e1fec0899728819f99eee92cf97356e6fbb98R737)

Zejnilovic (approved):
Build, run, code review

Issue #1692, opened 2021-02-17
Experimental API 2 - Runs


Zejnilovic:
## Background
2nd part of the API rewrite.
Approach this as a PoC.
API is enough. Do not touch UI.
Think about where Envelope makes sense

## Feature
All of this should be put under profile "api-experimental" in spring or a feature switch

- /runs now has the role of /runs/grouped 
  - Allows optional params of start date so you get only the last X days 
- /runs/{datasetName} now has role of  /runs/grouped/{datasetName}
  - Allows optional param start date so you get only the last X days.
- /runs/{datasetName}/{datasetVersion} 
  - GET gets you summaries for of runIDs
  - POST create new one
- /runs/summaries should not exist
- /runs/{datasetName}/{datasetVersion}/{runID} more PoCing
  - Latest instead of runID
  - GET your run info
  - POST To update run status 
  - **Q:** to have a separate /update-run-status, /udate-times
- /runs/{datasetName}/{datasetVersion}/{runID}/checkpoints
  - GET to get all of them
  - POST to add a checkpoint
  - /{name} to get only that one checkpoint
- /runs/{datasetName}/{datasetVersion}/{runID}/metadata
  - Metadata object from _INFO file
  - /{name} to get the value

dk1844:
## Reassurance of some of the details:
### `/runs[?startDate=dd-MM-yyyy|sparkAppId=...|uniqueId=...]`
 - returns (wrapped) list of `RunSummari`es (`Run` without `splineRef` and `controlMeasure`)
 - form of "last of each" (having runs grouped by `datasetName` and `versionName` - the `Run` with the latest `runId` is only returned

 ### `/runs/{datasetName}[?startDate=dd-MM-yyyy]`
 - same as above
 
 ### `/runs/{datasetName}/{datasetVersion}`
 - returns (wrapped) list of `RunSummari`es (`Run` without `splineRef` and `controlMeasure`)
 - no filtering, objects with all `runId`s should be returned

 ### `/runs/{datasetName}/{datasetVersion}/{runId}`
 - returns the `Run` object (contains `splineRef` and `controlMeasure`)

AdrianOlosutean:
I remember that `splineRef` was supposed to be removed, but I do not remember the details

Issue #2062, opened 2022-05-05
Consolidate REST API Error handling in api-v3 


Zejnilovic:
## Background
We need to consolidate responses and mainly error responses to the user from the REST API v3.

## Feature
Consolidate how we use return errors

## Notes
- This began in PR #2046 in [this comment](https://github.com/AbsaOSS/enceladus/pull/2046#discussion_r863756626)
- there are already `RestResponse` or `RestError` classes 
- we should keep some kind of ID field for better tracking through logs and connecting the issues

PR #2073, opened 2022-05-27
Feature/1692 api v3 runs

dk1844:
This PR implements Runs v3 API #1692. 
  - Pagination functionality is not included.
  - Endpoints description below
  - Endpoints are covered by an integTest, the `RunRepositoryV3` is covered by a separate integTest, too.
  - Special attention was given to efficiently aggregate the queries of the `run` collection

### `GET /runs[?startDate=yyyy-MM-dd|sparkAppId=...|uniqueId=...]`
 - returns (wrapped) list of `RunSummari`es (`Run` without `splineRef` and `controlMeasure`)
 - form of "last of each" (having runs grouped by `datasetName` and `versionName` - the `Run` with the latest `runId` is only returned

 ### `GET /runs/{datasetName}[?startDate=yyyy-MM-dd]`
 - same as above
 
 ### ` GET /runs/{datasetName}/{datasetVersion}[?startDate=yyyy-MM-dd]`
 - returns (wrapped) list of `RunSummari`es (`Run` without `splineRef` and `controlMeasure`)
 - no filtering, objects with all `runId`s should be returned

 ### `PUT /runs/{datasetName}/{datasetVersion}`
 - payload = `Run`
 - response contains the `Location` header with url to the newly created entity 

 ### `GET /runs/{datasetName}/{datasetVersion}/{runId}`
 - returns the `Run` object (contains `splineRef` and `controlMeasure`)
 - allows `latest` for a `runId` instead of #.
 

### `PUT /runs/{datasetName}/{datasetVersion}/{runId}`
 - payload = `RunStatus` (logically updated state, optional error)
 - there is currently no restrictions on the state transitions (details [here](https://github.com/AbsaOSS/enceladus/issues/1692#issuecomment-1144546563))

### `GET /runs/{datasetName}/{datasetVersion}/{runId}/checkpoints`
 - allows `latest` for a `runId` instead of #.

### `POST /runs/{datasetName}/{datasetVersion}/{runId}/checkpoints`
 - payload = `Checkpoint`
 - the new checkpoint is check for name duplicate and refused with a failed validation in such a case
 - response contains the `Location` header with url to the newly created entity 

### `GET /runs/{datasetName}/{datasetVersion}/{runId}/checkpoints/{checkpointName}`
 - allows `latest` for a `runId` instead of #

### `GET /runs/{datasetName}/{datasetVersion}/{runId}/metadata`
 - allows `latest` for a `runId` instead of #.
 - there is no endpoint for `GET .../metadata/{metadataName}`, because it does not seem to make a lot of sence (but I am open to suggstions in this part)

## Consideration for latest-of-each RunSummaries in MongoDB
 The latest-of-each retrieval poses an approach conundrum. I have basically thought about these options to implement it:
 1. grouping with max(runId) and joining on the original documents data - using `lookup` just as the V2 repository does it
 2. using Mongo's `aggregate`, while storing the documents in the with the group and then filtering by the `max(runId)`
 3. leverage a sort-of-all before grouping and then use grouping with `first` aggregate accumulator (guaranteed to be the max runId because of the pre-sorting)

The first option seems like overkill, but I admit I have not tried it. With options 2 and 3, I have written basic implementations of such aggregate queries and rudimentarily measured it on our UAT environment holding ~450k run records.

**Option 3 appeared to take about two times (~30s) more time than option 2 (~15s), provided that there is an appropriate index present in the collection for the pre-sorting.** Surprisingly, this index `{ $sort: { "runId": -1 } }` was present on our UAT, but must have been added extra, because our default setup does not introduce this index to the `runs_v1` collection. Thus, in the code, I have added it to be present for future MongoDB provisionings.

### Option 2 mongo script used:
```javascript
db.getCollection('run_v1').aggregate([
{ $sort: { "runId": -1 } }, // because of this sorting, the $first below yields max runId in the groups
{
    $group: {
      _id: {
           "dataset": "$dataset",
           "datasetVersion": "$datasetVersion"
      },
      datasetName: { $first : "$dataset"},
      datasetVersion: { $first : "$datasetVersion"},
      runId: { $first: "$runId" },
      status: { $first: "$runStatus.status" },
      startDateTime: { $first : "$startDateTime"},
      uniqueId: { $first: "$uniqueId" }

    },
},
{
    $project: {
        "_id" : 0 
    }
}
])
```

### Option 3 mongo script used:
```javascript
db.getCollection('run_v1').aggregate([
{
    $group: {
      _id: {
           "dataset": "$dataset",
           "datasetVersion": "$datasetVersion"
      },
      maxRunId: { $max: "$runId"},
      docs: {
          $push: '$$ROOT' // saving the documents
      }
    }
},
{
  $project: {
    _id: 0,
    doc: {
      "$setDifference": [{
          "$map": {
            "input": "$docs",
            "as": "mapped",
            "in": {
              "$cond": [{
                  "$eq": ["$maxRunId", "$$mapped.runId"] // filtering those with maxRunId here
                },
                "$$mapped",
                false
              ]
            }
          }
        },
        [false]
      ]
    }

  }
}, {
  $unwind: '$doc'
}, {
    $project: {
      datasetName: "$doc.dataset",
      datasetVersion: "$doc.datasetVersion",
      runId: "$doc.runId" ,
      status: "$doc.runStatus.status" ,
      startDateTime: "$doc.startDateTime",
      uniqueId:  "$doc.uniqueId" 
    }
}
  
], { allowDiskUse: true })
```


rest-api/src/test/scala/za/co/absa/enceladus/rest_api/integration/controllers/v3/DatasetControllerV3IntegrationSuite.scala
```diff
@@ -73,7 +73,7 @@ class DatasetControllerV3IntegrationSuite extends BaseRestApiTestV3 with BeforeA
         val response = sendPost[Dataset, Validation](apiUrl, bodyOpt = Some(dataset))
         assertCreated(response)
         response.getBody shouldBe Validation.empty.withWarning("keyD", "Property 'keyD' is recommended to be present, but was not found!")
-        val locationHeader = response.getHeaders.getFirst("location")
+        val locationHeader = response.getHeaders.getFirst("Location")
```
benedeki, line 76: How did this work before? Or is the function case insensitive and just convention dictates using capital `L`?

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/services/RunService.scala
```diff
@@ -16,32 +16,31 @@
...
-import za.co.absa.enceladus.rest_api.repositories.RunMongoRepository
+import za.co.absa.enceladus.rest_api.repositories.{MongoRepository, RunMongoRepository}
 import za.co.absa.enceladus.model.{Run, SplineReference, Validation}
 
 import scala.concurrent.Future
 import scala.util.{Failure, Success, Try}
 
-@Service
-class RunService @Autowired()(val mongoRepository: RunMongoRepository)
+@Service("runService") // by-name qualifier: making V2 autowiring un-ambiguous
```
benedeki, line 32: If you would have time to explain this to me (educating 😉 )

rest-api/src/test/scala/za/co/absa/enceladus/rest_api/integration/repositories/RunRepositoryIntegrationSuite.scala
```diff
@@ -42,6 +44,7 @@ class RunRepositoryIntegrationSuite extends BaseRepositoryTest {
   private val runFixture: RunFixtureService = null
 
   @Autowired
+  @Qualifier("runMongoRepository") // to correctly wire V2 runMongoRepository
```
benedeki, line 47: Another education request, please.

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/controllers/v3/RunControllerV3.scala
```diff
@@ -0,0 +1,215 @@
...
+
+  import scala.concurrent.ExecutionContext.Implicits.global
+
+  @GetMapping()
+  @ResponseStatus(HttpStatus.OK)
+  def list(@RequestParam startDate: Optional[String],
+           @RequestParam sparkAppId: Optional[String],
+           @RequestParam uniqueId: Optional[String]
+          ): CompletableFuture[Seq[RunSummary]] = {
+    require(Seq(startDate, sparkAppId, uniqueId).filter(_.isPresent).length <= 1,
```
benedeki, line 57:
```suggestion
    require(Seq(startDate, sparkAppId, uniqueId).count(_.isPresent) <= 1,
```

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/controllers/v3/RunControllerV3.scala
```diff
@@ -0,0 +1,215 @@
...
+  def getRunMetadata(@PathVariable datasetName: String,
+                     @PathVariable datasetVersion: Int,
+                     @PathVariable runId: String): CompletableFuture[ControlMeasureMetadata] = {
+    getRunForRunIdExpression(datasetName, datasetVersion, runId).map(_.controlMeasure.metadata)
+  }
+
+  /**
+   * For run's dataset name, version and runId (either a number of 'latest'), the `forVersionFn` is called.
+   */
+  protected def forRunIdExpression[T](datasetName: String, datasetVersion: Int, runIdStr: String)
```
benedeki, line 188: Seems unused?

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/controllers/v3/RunControllerV3.scala
```diff
@@ -0,0 +1,215 @@
...
+
+  /**
+   * Retrieves a Run by dataset name, version and runId (either a number of 'latest')
+   *
+   * @param datasetName    dataset name
+   * @param datasetVersion dateset version
+   * @param runIdStr       runId (either a number of 'latest')
+   * @return Run object
+   */
+  protected def getRunForRunIdExpression(datasetName: String, datasetVersion: Int, runIdStr: String): Future[Run] = {
```
benedeki, line 203: Why `protected` and not `private`?

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/repositories/v3/RunMongoRepositoryV3.scala
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    val datasetFilter: Option[Bson] = datasetNameVersionOptFilter(datasetName, datasetVersion)
+
+    val combinedFilter: Bson = (exclusiveOptsFilter, datasetFilter) match {
+      case (None, None) => BsonDocument() // empty filter
+      case (Some(filter1), None) => filter1
+      case (None, Some(filter2)) => filter2
+      case (Some(filter1), Some(filter2)) => Filters.and(filter1, filter2)
+    }
```
benedeki, line 67: What about to make this a routine? Used more times. (Here, lines 98-103,...)

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/controllers/v3/RunControllerV3.scala
```diff
@@ -0,0 +1,215 @@
...
+
+@RestController
+@RequestMapping(path = Array("/api-v3/runs"), produces = Array("application/json"))
+class RunControllerV3 @Autowired()(runService: RunServiceV3) extends BaseController {
+
+  import za.co.absa.enceladus.rest_api.utils.implicits._
+
+  import scala.concurrent.ExecutionContext.Implicits.global
+
+  @GetMapping()
```
Zejnilovic, line 51: I have no idea where this happens, but the on `/api-v3/runs` using `startDate` filter I get an extra result that is not supposed to be there.

image

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/services/RunService.scala
dk1844: `RunServiceV3` extends `RunService`. This has a few expected effects:
 - Wherever a field of type `RunServiceV3` is needed, `@Autowired` only has 1 option to use - `RunServiceV3`.
 - Wherever a field of type `RunService` is needed, `@Autowired` only has 2 options to use - `RunServiceV3` or `RunServiceV3` and thus fails with ambiguity exception, therefore it needs to be disambigued somehow (some details [here](https://www.baeldung.com/spring-autowire#disambiguation), but basically this is summed up by _By default, Spring resolves @Autowired entries by type._)

There are multiple ways how to hint the desired wiring to be used, e.g. mark one of the implementations by `@Primary` etc. The option I have used here is that I mark the bean with a custom qualifier that matches the field name where requested. So, I have:

```scala
@Service("runService") // by-name qualifier: making V2 autowiring un-ambiguous
class RunService()

@Service
class RunServiceV3()

class LandingPageController @Autowired() (
  ...,
  runService: RunService, // matches the qualifier of RunService => no ambiguity, RunService is wired.
)
```

rest-api/src/test/scala/za/co/absa/enceladus/rest_api/integration/repositories/RunRepositoryIntegrationSuite.scala
dk1844: This is the same case as https://github.com/AbsaOSS/enceladus/pull/2073#discussion_r897672416
A field of type `RunMongoRepository` can be wired with more than one class, so the qualifier here helps to resolve that -- when the field of this type is named `runMongoRepository`

Zejnilovic:
## Question 1

On non-existing URLs I do not get a 404 but rather an empty array.
`/api-v3/runs/DeleteMeXYZ` -> the dataset does not exist
`/api-v3/runs/DeleteMe/1223` -> the version doesn't exist
These return 200 and an empty array. 

I can see why return an empty array if the dataset exists but has no runs for the Name or the Version, but if the dataset name or combination of name and version don't even exist in Dataset collection, then it is misleading.

Thoughts?

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/controllers/v3/RunControllerV3.scala
dk1844: Should the RunControllerV3 ever get extended, it could get reused in the ancestor class, too.

rest-api/src/test/scala/za/co/absa/enceladus/rest_api/integration/controllers/v3/DatasetControllerV3IntegrationSuite.scala
dk1844: The latter exactly:
> function case insensitive and just convention dictates using capital `L`

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/controllers/v3/RunControllerV3.scala
dk1844: Thanks, suggestion applied.

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/repositories/RunMongoRepository.scala
```diff
@@ -325,6 +340,20 @@ class RunMongoRepository @Autowired()(mongoDb: MongoDatabase)
     ).headOption().map(_.map(bson => SerializationUtils.fromJson[Run](bson.toJson)))
   }
 
+  def appendCheckpoint(datasetName: String, datasetVersion: Int, runId: Int, newCheckpoint: Checkpoint): Future[Option[Run]] = {
```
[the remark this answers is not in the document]
dk1844: I have actually renamed the older one, it seemed to make more sense.

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/repositories/v3/RunMongoRepositoryV3.scala
dk1844: Done, thanks for the tip.

benedeki:
> ## Question 1
> 
> On non-existing URLs I do not get a 404 but rather an empty array. `/api-v3/runs/DeleteMeXYZ` -> the dataset does not exist `/api-v3/runs/DeleteMe/1223` -> the version doesn't exist These return 200 and an empty array.
> 
> I can see why return an empty array if the dataset exists but has no runs for the Name or the Version, but if the dataset name or combination of name and version don't even exist in Dataset collection, then it is misleading.
> 
> Thoughts?

I agree. 404 in case Dataset & Version does not exists.
Empty array if no runs for existing Dataset & Version

Zejnilovic:
## Suggestion 2

on Update status, that is `PUT /runs/{datasetName}/{datasetVersion}/{runId}`, we return plaintext message. I would rather go with 

```json
{
  "status": "updated",
  "message": "message"
}
```

But this already smells of the JSON envelope. So maybe we keep it for the next PR

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/controllers/v3/RunControllerV3.scala
dk1844: This issue has been addressed, the param now uses different date format `startDate=YYYY-MM-DD`

Zejnilovic (approved):
code reviewed
built
run

AdrianOlosutean (approved):
Code review

dk1844:
Hi guys, 

> > ## Question 1
> > On non-existing URLs I do not get a 404 but rather an empty array. `/api-v3/runs/DeleteMeXYZ` -> the dataset does not exist `/api-v3/runs/DeleteMe/1223` -> the version doesn't exist These return 200 and an empty array.
> > I can see why return an empty array if the dataset exists but has no runs for the Name or the Version, but if the dataset name or combination of name and version don't even exist in Dataset collection, then it is misleading.
> > Thoughts?
> 
> I agree. 404 in case Dataset & Version does not exists. Empty array if no runs for existing Dataset & Version

This I have fixed as suggested. For `/api-v3/runs/justDatasetNameWithoutVersion`, I have added check for latest-version, so this option is covered as well to produce 404 if no version is found for this dataset.
Tests added.

> ## Suggestion 2
> on Update status, that is `PUT /runs/{datasetName}/{datasetVersion}/{runId}`, we return plaintext message. I would rather go with
> 
> ```json
> {
>   "status": "updated",
>   "message": "message"
> }
> ```
> 
> But this already smells of the JSON envelope. So maybe we keep it for the next PR

Yes, it does smell a bit as to be part of #2062, thus, so far, I have just wrapped the response in an object with a message field (so it is a valid JSON with the message, but no other fields like `status`), but for the rest of the effort, we should solve this in #2062

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/controllers/v3/RunControllerV3.scala
```diff
@@ -212,4 +208,13 @@ class RunControllerV3 @Autowired()(runService: RunServiceV3) extends BaseControl
     }
   }
 
+  protected def parseYmdDate(dateOptStr: Optional[String]): Option[LocalDate] = {
+    dateOptStr.toScalaOption.map { dateStr =>
+      Try(LocalDate.parse(dateStr)) match {
+        case Success(parsed) => parsed
+        case Failure(e) => throw new IllegalArgumentException(s"Could not parse YYYY-MM-DD date from $dateStr: ${e.getMessage}")
```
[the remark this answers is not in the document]
dk1844: The service should not fail, it is just the response that will communicate the problem; with `None`, the invalid date would behave as not given at all.

I would understand this would be preferred if we were changing an implementation already used by users, but since this is V3 API that nobody uses yet, I feel we can afford to be stricter here. But I am willing to be persuaded.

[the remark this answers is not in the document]
benedeki: Sorry, misunderstood where it's used (not when querying data but when parsing request). This is totally OK of course.

benedeki (approved):

PR #2080, opened 2022-06-02
Feature/1692 api v3 runs to `develop`

dk1844:
(tryout PR to asses the incompatibilities etc) @AdrianOlosutean


dk1844:
Will be resolved in #2073

PR #2083, opened 2022-06-06
spline spark 3

AdrianOlosutean:


spark-jobs/pom.xml
```diff
@@ -181,9 +179,9 @@ spark-submit -(remove this)-packages org.apache.spark:spark-sql-kafka-0-10_2.11:
 -->
                             <exclude>org.apache.hadoop:*</exclude>
                             <exclude>org.scala-lang:scala-actors</exclude>
-                            <exclude>org.scala-lang:scala-compiler</exclude>
-                            <exclude>org.scala-lang:scala-reflect</exclude>
-                            <exclude>org.scala-lang:scala-library</exclude>
+<!--                            <exclude>org.scala-lang:scala-compiler</exclude>-->
```
benedeki, line 182: Do you know why this is present as commented out. And not removed fully?

benedeki (commented):
It's more a merge PR isn't it? The Spline change seems trivial, just a minor version change.
No UI changes needed?

spark-jobs/pom.xml
AdrianOlosutean: I think it's leftover code from before. It shouldn't be commented out

PR #2086, opened 2022-06-10
Feature/2060 pagination

### LINKED ISSUE #2060: Add pagination to the REST API v3
## Background
There are calls which pull a lot of data. `/datasets` and `/runs`. These calls should be paginated, maybe others as well.

## Feature
Add pagination to the REST API v3

dk1844:
_(In progress)_

This PR adds pagination functionality to listing entities, specifically to (with both optional params `offset` (default: `0`) and `limit` (default: `20`) - default applied if param is not specified or is invalid):
 - `GET /api-v3/{datasets|schemas|mapping-tables|property-defintions}[?offset=X][&limit=Y]`
 - `GET /api-v3/runs[?offset=X][&limit=Y]`
 - `GET /runs/{datasetName}[?offset=X][&limit=Y]`
 - `GET /runs/{datasetName}/{datasetVersion}[?offset=X][&limit=Y]`

with the responses being wrapped in the `Paginated` case class (yields the data `page` , used `offset`, used `limit` and `truncated` boolean flag to signify whether there is more data to be retrieved).

### Datasets example
`GET api-v3/datasets/?limit=3` ->

```json
{
  "page": [
    {"name": "dsA", "version": 1, "disabled": false},
    {"name": "dsB", "version": 1, "disabled": false},
    {"name": "dsC", "version": 1, "disabled": false}
  ],
  "offset": 0,
  "limit": 3,
  "truncated": true
}
```
(JSON was manually prettified, originally it is compacted)

### Runs example:
`GET api-v3/runs/dataset1?offset=3&limit=2` ->
```json
{
  "page": [
    {
      "datasetName": "dataset1",
      "datasetVersion": 4,
      "runId": 2,
      "status": "allSucceeded",
      "startDateTime": "04-12-2017 16:19:17 +0200",
      "runUniqueId": "20aa6300-f0e2-4432-a3f3-28c36da20a0d"
    },
    {
      "datasetName": "dataset1",
      "datasetVersion": 5,
      "runId": 1,
      "status": "allSucceeded",
      "startDateTime": "04-12-2017 16:19:17 +0200",
      "runUniqueId": "0559f3c4-907d-4cbf-981f-85d2b7e7b283"
    }
  ],
  "offset": 3,
  "limit": 2,
  "truncated": true
}
```
(JSON was manually prettified, originally it is compacted)

Closes #2060.

### DIFF
--- a/data-model/src/main/scala/za/co/absa/enceladus/model/versionedModel/NamedVersion.scala
+++ b/data-model/src/main/scala/za/co/absa/enceladus/model/versionedModel/NamedVersion.scala
@@ -18,4 +18,4 @@ package za.co.absa.enceladus.model.versionedModel
 /**
  * V3 Wrapper for [[za.co.absa.enceladus.model.versionedModel.VersionedSummary]]
  */
-case class NamedVersion(name: String, version: Int, disabled: Boolean)
+case class NamedVersion(name: String, version: Int, disabled: Boolean = false)
--- a/rest-api/src/main/scala/za/co/absa/enceladus/rest_api/models/rest/Paginated.scala
+++ b/rest-api/src/main/scala/za/co/absa/enceladus/rest_api/models/rest/Paginated.scala
@@ -15,15 +15,16 @@
 
 package za.co.absa.enceladus.rest_api.models.rest
 
-case class Paginated[T](page: Seq[T], offset:Int, limit: Int, truncated: Boolean)
+case class Paginated[T](page: Seq[T], offset: Int, limit: Int, truncated: Boolean)
 
 object Paginated {
 
   /**
    * Wraps data in Paginated and possibly truncates the data to fit the limit. `truncated` flag is set based on that.
-   * @param data data to be embedded and truncated to length of `limit`
+   *
+   * @param data   data to be embedded and truncated to length of `limit`
    * @param offset is only passed to the wrapper object, has no effect to data
-   * @param limit passed to the wrapper object and it used for truncating the data
+   * @param limit  passed to the wrapper object and it used for truncating the data
    * @tparam T
    * @return truncated data wrapped in Paginated - `truncated` is true if data truncated, false otherwise.
    */
--- a/rest-api/src/main/scala/za/co/absa/enceladus/rest_api/models/rest/Paginated.scala
+++ b/rest-api/src/main/scala/za/co/absa/enceladus/rest_api/models/rest/Paginated.scala
@@ -1,0 +1,38 @@
+/*
+ * Copyright 2018 ABSA Group Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.enceladus.rest_api.models.rest
+
+case class Paginated[T](page: Seq[T], offset:Int, limit: Int, truncated: Boolean)
+
+object Paginated {
+
+  /**
+   * Wraps data in Paginated and possibly truncates the data to fit the limit. `truncated` flag is set based on that.
+   * @param data data to be embedded and truncated to length of `limit`
+   * @param offset is only passed to the wrapper object, has no effect to data
+   * @param limit passed to the wrapper object and it used for truncating the data
+   * @tparam T
+   * @return truncated data wrapped in Paginated - `truncated` is true if data truncated, false otherwise.
+   */
+  def truncateToPaginated[T](data: Seq[T], offset: Int, limit: Int): Paginated[T] = {
+    if (data.length <= limit) {
+      Paginated(data, offset, limit, truncated = false)
+    } else {
+      Paginated(data.take(limit), offset, limit, truncated = true)
+    }
+
+  }
+}
--- a/rest-api/src/main/scala/za/co/absa/enceladus/rest_api/repositories/VersionedMongoRepository.scala
+++ b/rest-api/src/main/scala/za/co/absa/enceladus/rest_api/repositories/VersionedMongoRepository.scala
@@ -61,7 +61,7 @@ abstract class VersionedMongoRepository[C <: VersionedModel](mongoDb: MongoDatab
     collection.distinct[String]("name", getNotDisabledFilter).toFuture().map(_.sorted)
   }
 
-  def getLatestVersionsSummarySearch(searchQuery: Option[String] = None): Future[Seq[VersionedSummary]] = {
+  def getLatestVersionsSummarySearch(searchQuery: Option[String] = None, offset: Option[Int] = None, limit: Option[Int] = None): Future[Seq[VersionedSummary]] = {
     val searchFilter = searchQuery match {
       case Some(search) => Filters.regex("name", search, "i")
       case None => Filters.expr(true)
@@ -72,7 +72,10 @@ abstract class VersionedMongoRepository[C <: VersionedModel](mongoDb: MongoDatab
         Accumulators.max("latestVersion", "$version")
       ),
       sort(Sorts.ascending("_id"))
-    )
+    ) ++
+      offset.map(skipVal => Seq(Aggregates.skip(skipVal))).getOrElse(Seq.empty) ++
+      limit.map(limitVal => Seq(Aggregates.limit(limitVal))).getOrElse(Seq.empty)
+
     collection.aggregate[VersionedSummaryV2](pipeline).toFuture()
       .map(_.map(summaryV2 => VersionedSummary(summaryV2._id, summaryV2.latestVersion, Set(false)))) // because of the notDisabled filter
   }
--- a/rest-api/src/main/scala/za/co/absa/enceladus/rest_api/repositories/VersionedMongoRepository.scala
+++ b/rest-api/src/main/scala/za/co/absa/enceladus/rest_api/repositories/VersionedMongoRepository.scala
@@ -73,11 +73,17 @@ abstract class VersionedMongoRepository[C <: VersionedModel](mongoDb: MongoDatab
       ),
       sort(Sorts.ascending("_id"))
     ) ++
-      offset.map(skipVal => Seq(Aggregates.skip(skipVal))).getOrElse(Seq.empty) ++
+      offset.map(skipVal => Seq(Aggregates.skip(skipVal))).getOrElse(Seq.empty) ++ // todo reconsider using skip for performance?
       limit.map(limitVal => Seq(Aggregates.limit(limitVal))).getOrElse(Seq.empty)
 
     collection.aggregate[VersionedSummaryV2](pipeline).toFuture()
       .map(_.map(summaryV2 => VersionedSummary(summaryV2._id, summaryV2.latestVersion, Set(false)))) // because of the notDisabled filter
+
+    // todo, consider setting limit to batchsize, e.g. ?
+//    val aggregate =  collection.aggregate[VersionedSummaryV2](pipeline)
+//    limit.map(aggregate.batchSize).getOrElse(aggregate) // for defined limit, set batchsize to match
+//      .toFuture()
+//      .map(_.map(summaryV2 => VersionedSummary(summaryV2._id, summaryV2.latestVersion, Set(false)))) // because of the notDisabled filter
   }
 
   def getLatestVersions(missingProperty: Option[String]): Future[Seq[C]] = {
--- a/data-model/src/main/scala/za/co/absa/enceladus/model/test/factories/DatasetFactory.scala
+++ b/data-model/src/main/scala/za/co/absa/enceladus/model/test/factories/DatasetFactory.scala
@@ -18,7 +18,7 @@ package za.co.absa.enceladus.model.test.factories
 import za.co.absa.enceladus.model.Dataset
 import za.co.absa.enceladus.model.conformanceRule._
 import za.co.absa.enceladus.model.menas.MenasReference
-import za.co.absa.enceladus.model.versionedModel.VersionedSummary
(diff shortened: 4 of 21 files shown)


rest-api/src/main/scala/za/co/absa/enceladus/rest_api/controllers/v3/ControllerPagination.scala
```diff
@@ -0,0 +1,87 @@
...
+   * @param optField     value to attempt to extract from
+   * @param defaultValue value to use if extraction fails
+   * @return On extraction success, `extractedIntValue` is returned, otherwise (empty or invalid) `defaultValue` is returned.
+   */
+  private def extractDefinedValueOrDefault(optField: Option[String], defaultValue: Int): Int = {
+    optField match {
+      case None => defaultValue
+      case Some(intAsString) => Try(intAsString.toInt) match {
+        case Success(value) => value
+        case Failure(_) => defaultValue
```
lsulak, line 82: Shouldn't there be a warning log message about this? Something like "falling back to a default value X because Y couldn't be used/converted to int"?

otherwise, a potential problem with `optField` can go unnoticed (I know almost nothing about Sprint framework, perhaps this is a common practice)

dk1844: Hi, thanks for noticing, I agree, that it would be preferred to let the user know (similarly as @Zejnilovic suggests below in his Q#2).

Given the options, I have opted to throw an exception on this account resulting in HTTP 400 on the side of the REST API.

dk1844:
Thank you, @Zejnilovic. I has been cleared elsewhere, but for completeness sake, I will briefly comment here, too (below)

> ### Question 1
> did we talk about the size/length/count in the envelope or headers? Can we have it?

The `truncated` boolean flag is used instead, signifying the exhaustiveness of the result (when the result got truncated, something else is there yet to be retrieved).

> ### Question 2
> Can we have validation for negative values so it doesn't throw "random" 500, but something better? These 500s expose our mongo URLs, maybe more.

Yes, agreed. The param values are now validated and we now yield HTTP 400 with cause. Also covered in https://github.com/AbsaOSS/enceladus/pull/2086#discussion_r918833185.

 
> ### Question 3
> If I have 20 items in DB. The limit is 20 and I set the offset to 3. I am missing the first 3 items, but truncated is still false. Is this correct? Just checking.

Yes, exactly

Some examples:
| total items  |  offset  | limit  | truncated   | note   |
|---|---|---|---|---|
|  20  |  3  |  20  | false |  page contains 17 items out of possible 20. A complete result, no truncation - case from Q3  |
| 30  | 20  | 10  | false  |  last 10 items, nothing is left to retrieve - no truncation |
|  25 | 0  | 20   | true   |  first page with 20 items. There are still 5 items to retrieve - truncation has happened |

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/controllers/v3/ControllerPagination.scala
```diff
@@ -74,12 +74,14 @@ object ControllerPagination {
...
+  private def extractDefinedValueOrDefault(optField: Option[String], defaultValue: Int, paramNameHint: String): Int = {
     optField match {
       case None => defaultValue
       case Some(intAsString) => Try(intAsString.toInt) match {
-        case Success(value) => value
-        case Failure(_) => defaultValue
+        case Success(value) if value >= 0 => value
+        case Success(value) => throw new IllegalArgumentException(s"Value '$value' must be > 0 for the '$paramNameHint' param.")
+        case Failure(_) =>
+          throw new IllegalArgumentException(s"'$intAsString' could not be interpreted as int for the '$paramNameHint' param.")
```
lsulak, line 84: Can you please help me understand how this exception gets translated into 400?

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/controllers/v3/ControllerPagination.scala
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  /**
+   * For the `optField` we try to extract int value
+   *
+   * @param optField     value to attempt to extract from
+   * @param defaultValue value to use if extraction fails
+   * @return On extraction success, `extractedIntValue` is returned, otherwise (empty or invalid) `defaultValue` is returned.
+   */
+  private def extractDefinedValueOrDefault(optField: Option[String], defaultValue: Int, paramNameHint: String): Int = {
```
lsulak, line 77: (Edu, for me) I saw that some modules are adding exception annotation to methods, like `RestClient.scala` is using, e.g.: `@throws[DaoException]`. If I understood it correctly, this is more like documentation / for the compiler as an optional 'metadata' about a given method. Do you have a rule of thumb on when to use it?

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/controllers/v3/ControllerPagination.scala
dk1844: Certainly.

In "normal" scala/java code, it is passed along as a regular exception, with stack trace and all, but for Spring's REST, we define a [RestExcepstionHandler](https://github.com/AbsaOSS/enceladus/blob/develop/rest-api/src/main/scala/za/co/absa/enceladus/rest_api/controllers/RestExceptionHandler.scala#L47) that transforms known exceptions into specific REST responses (unmatched ones are passed as 500s with a stack trace).

Since our handler is defined for the `IllegalArgumentException` type, we can instill this behavior in integTests, too.

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/controllers/v3/ControllerPagination.scala
lsulak: Thank you, it makes sense. 

JFI I found some sources on this topic as well: 
* https://softwareengineering.stackexchange.com/questions/252378/when-to-declare-throws-with-a-scala-annotation-abstract-parent-or-concrete-im
* https://alvinalexander.com/scala/how-to-declare-scala-methods-throws-exceptions/

Zejnilovic (dismissed):
Built and Run. Did not do a full code review.

(Hopefully tonight I go over the code as well)

data-model/src/main/scala/za/co/absa/enceladus/model/versionedModel/NamedVersion.scala
```diff
@@ -18,4 +18,4 @@ package za.co.absa.enceladus.model.versionedModel
 /**
  * V3 Wrapper for [[za.co.absa.enceladus.model.versionedModel.VersionedSummary]]
  */
-case class NamedVersion(name: String, version: Int, disabled: Boolean)
+case class NamedVersion(name: String, version: Int, disabled: Boolean = false)
```
benedeki, line 21: Why this here&now?

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/models/rest/Paginated.scala
```diff
@@ -0,0 +1,39 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.enceladus.rest_api.models.rest
+
+case class Paginated[T](page: Seq[T], offset: Int, limit: Int, truncated: Boolean) //, message: Option[String] = None)
```
benedeki, line 18: Is `truncated` the standard name for the hint there are more data?
Probably just me, but gives me the impression, so of the **returned** data were truncated. 🤷

.sonarcloud.properties
```diff
@@ -14,3 +14,4 @@
 #
 
 sonar.scala.scalastyle.reportPaths=${project.basedir}/target/scalastyle-output.xml
+sonar.cpd.exclusions=${project.basedir}/*/src/test/scala/**Suite.scala,${project.basedir}/*/src/test/scala/**Test.scala,${project.basedir}/*/src/test/scala/**Spec.scala
```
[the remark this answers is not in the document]
dk1844: That is the way I understand it as well.

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/repositories/VersionedMongoRepository.scala
```diff
@@ -61,7 +61,7 @@ abstract class VersionedMongoRepository[C <: VersionedModel](mongoDb: MongoDatab
     collection.distinct[String]("name", getNotDisabledFilter).toFuture().map(_.sorted)
   }
 
-  def getLatestVersionsSummarySearch(searchQuery: Option[String] = None): Future[Seq[VersionedSummary]] = {
+  def getLatestVersionsSummarySearch(searchQuery: Option[String] = None, offset: Option[Int] = None, limit: Option[Int] = None): Future[Seq[VersionedSummary]] = {
```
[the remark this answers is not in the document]
dk1844: No, but `VersionedMongoRepository` is common for both versions via `VersionedModelService.getLatestVersionsSummarySearch`.

I consider this change to this common layer to be acceptable considering the alternative rearchitecting with V2 possibly going away at a certain point.

data-model/src/main/scala/za/co/absa/enceladus/model/versionedModel/NamedVersion.scala
dk1844: This was added as a sensible default. Now, because NamedVersion has been used in pagination-tests more extensively, where this default coming handy was more obvious (repetitive usage).

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/models/rest/Paginated.scala
dk1844: Not sure about _standard_, but this is [inspired by how AWS S3 API signifies that paged listing has more data to be retrieved](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjects.html#AmazonS3-ListObjects-response-IsTruncated).

The impression is correct to a certain degree - you request to obtain items of a collection and either you get all the items (when `collection.itemsCount <= page.size`) or the **returned data does get truncated to fit the page** (`collection.itemsCount > page.size`).

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/controllers/v3/ControllerPagination.scala
```diff
@@ -0,0 +1,89 @@
...
+    extractOptionOffsetOrDefault(optField.toScalaOption, defaultValue)
+  }
+
+  /** Offset value is extracted from `optField`, otherwise `defaultValue` returned
+   *
+   * @param optField
+   * @param defaultValue when extraction fails, this value is used. Default = [[ControllerPagination#DefaultOffset()]]
+   * @return
+   */
+  def extractOptionOffsetOrDefault(optField: Option[String], defaultValue: Int = DefaultOffset): Int = {
```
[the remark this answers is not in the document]
dk1844: I deliberately wanted to overload that public method so that the pagination helper object (`ControllerPagination`) is versatile enough not to care if `Optional[String]` or `Option[String]` is passed - this way, both works without the caller needing to think about this. (Because here, `Optional` (from Java) is what we need while `Option` (from Scala) is the signature we want there to be - because we are writing primarily Scala code.

Does it make sense? How strongly do you feel about this?

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/models/rest/Paginated.scala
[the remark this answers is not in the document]
dk1844: Thanks for noticing, yes, exactly, something I have included in a later-removed implementation. Removed.

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/controllers/v3/ControllerPagination.scala
```diff
@@ -0,0 +1,89 @@
...
+  private def extractDefinedValueOrDefault(optField: Option[String], defaultValue: Int, paramNameHint: String): Int = {
+    optField match {
+      case None => defaultValue
+      case Some(intAsString) => Try(intAsString.toInt) match {
+        case Success(value) if value >= 0 => value
+        case Success(value) => throw new IllegalArgumentException(s"Value '$value' must be > 0 for the '$paramNameHint' param.")
+        case Failure(_) =>
+          throw new IllegalArgumentException(s"'$intAsString' could not be interpreted as int for the '$paramNameHint' param.")
+      }
+    }
```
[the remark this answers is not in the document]
dk1844: I don't mind two matches inside each other (I do not consider that hard to read), but sure, why not. Accepted your suggestion

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/controllers/v3/ControllerPagination.scala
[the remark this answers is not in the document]
benedeki: It does make sense. Cool keep it as it is.
Cool it's deliberate and thought through 👍

data-model/src/main/scala/za/co/absa/enceladus/model/versionedModel/NamedVersion.scala
benedeki: Fair enough, good reason!

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/models/rest/Paginated.scala
benedeki: OK, if AWS uses this keyword, then it's probably good enough and widely accepted 😉

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/repositories/VersionedMongoRepository.scala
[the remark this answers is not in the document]
benedeki: OK, that is a good assumption. 👍

benedeki (approved):
- [X] code reviewed
- [X] pulled
- [X] built
- [ ] run

Issue #2072, opened 2022-05-24
Make 404 an optionally recoverable error when calling rest api endpoint


benedeki:
## Background
When call from _Spark Job_ to _Menas REST API_ fails, if the error is considered recoverable another of the configured servers is tried.

## Feature
`404` is not considered recoverable. Make it configurable that `404` can be considered recoverable to try a different server.

## Bonus [Optional]
Maybe it could be done generally, that a list of error status codes can be configured, that will be considered recoverable.

lsulak:
## Release notes
A user can specify a list of HTTP error status codes that are allowed to be retried when a REST API call yields such an HTTP error. Currently supported HTTP error status codes: 401, 403, and 404.

PR #2087, opened 2022-06-16
#2072 make 404 recoverable error

lsulak:
Feel free to review this PR. I'm far from satisfied with it at the moment - some comments:
* `CrossHostApiCaller.scala` - I don't like the current functionality around errors that are recoverable by default: ResourceAccessException and RestClientException - in my opinion, this should be changed somehow. I used the variable `canBeRecovered` and this approach might be an antipattern to Scala/Functional programming. This is more like a PoC though. **EDIT**: This was changed recently, it should be much better
* `CustomException.scala` - in the original design, `DaoException` was retryable by default. I couldn't find an easy way how to keep this approach and implement what was needed since 404 should be optionally recoverable (if it's recoverable always, then the solution would be super easy - we'd just need to change the parent of `NotFoundException`, but such solution wouldn't be correct). I checked the codebase and it looks like changing parent of `DaoException` shouldn't affect anything else (except changes in this PR, related to DAO module).
* Perhaps more unit tests are needed, in `spark-jobs` module.
* I wanted to do more testing against a real instance of Menas/restApi/spark job on my laptop, I'm hoping to do it next week.


dao/src/main/scala/za/co/absa/enceladus/dao/CustomException.scala
```diff
@@ -15,17 +15,16 @@
 
 package za.co.absa.enceladus.dao
 
-abstract class RetryableException(message: String, cause: Throwable) extends Exception(message, cause)
-
+abstract class CustomException(message: String, cause: Throwable) extends Exception(message, cause)
```
AdrianOlosutean, line 18: Does it make sense to rename this? Especially since fields where it's used are called retryableExceptions?

dao/src/main/scala/za/co/absa/enceladus/dao/rest/CrossHostApiCaller.scala
```diff
@@ -60,6 +69,8 @@ protected class CrossHostApiCaller private(apiBaseUrls: Vector[String], maxTryCo
 
 
   def call[T](fn: String => T): T = {
+    var canBeRecovered = false
```
AdrianOlosutean, line 72: Let's try to avoid using this vars

dao/src/test/scala/za/co/absa/enceladus/dao/rest/MenasRetryableExceptionsStringParserSuite.scala
```diff
@@ -0,0 +1,80 @@
...
+      "it is 400;404" in {
+        val inputExceptionsString = "400;404"
+        val expectedExceptionsSet = Set(
+          new CustomException("Optionally retryable exception - 400", None.orNull) {},
+          new CustomException("Optionally retryable exception - 404", None.orNull) {}
+        )
+
+        val result = MenasRetryableExceptionsStringParser.parse(inputExceptionsString)
+
+        result.size should be(expectedExceptionsSet.size)
```
AdrianOlosutean, line 55:
```suggestion
        result.size shouldEqual expectedExceptionsSet.size
```

lsulak: actually I changed this completely, please see the newest commits. Now it should handle all edge cases

dao/src/main/scala/za/co/absa/enceladus/dao/CustomException.scala
lsulak: The logic & reasoning for this is that before, if an exception was inherited from `RetryableException`, then such error was recoverable. Now this mechanism is completely different and it's not based on checking the exception type. So not just the name was changed, but also the logic (and thus, the name should reflect that).

dao/src/main/scala/za/co/absa/enceladus/dao/rest/CrossHostApiCaller.scala
lsulak: I know, I hate it :( it's because I couldn't find an easy solution for handling two errors that are recoverable by default: `ResourceAccessException` and `RestClientException` so now the logic is mixed. Probably I/we had to think about different design of code related to this.

dao/src/test/scala/za/co/absa/enceladus/dao/rest/CrossHostApiCallerSuite.scala
```diff
@@ -58,7 +58,11 @@ class CrossHostApiCallerSuite extends BaseTestSuite {
           .thenThrow(DaoException("Something went wrong B"))
           .thenReturn("success")
 
-        val result = CrossHostApiCaller(Vector("a", "b", "c"), 2, Some(0)).call { str =>
+        val retryableExceptions: Set[CustomException] = Set(
```
AdrianOlosutean, line 61: I would then rename these fields `customExceptios`

dao/src/main/scala/za/co/absa/enceladus/dao/rest/CrossHostApiCaller.scala
```diff
@@ -68,20 +78,33 @@ protected class CrossHostApiCaller private(apiBaseUrls: Vector[String], maxTryCo
...
+
+        case Failure(e: DaoException) if attemptNumber < maxTryCount =>
+          logFailure(e, url, attemptNumber, None)
+          attempt(url, attemptNumber + 1, urlsTried)
+
+        case Failure(e: MenasException) if retryableExceptions.contains(e) && attemptNumber < maxTryCount =>
+          logFailure(e, url, attemptNumber, None)
+          attempt(url, attemptNumber + 1, urlsTried)
+
         case _ => result
```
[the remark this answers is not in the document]
lsulak: Also, originally `DaoException` was considered to be retryable. Perhaps we should rename `DaoException` -> `RetryAbleException` and then rename `MenasException` -> `DaoException`. Please tell me what you think about it, I'm quite inclined towards it.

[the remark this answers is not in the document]
lsulak: Also the name of the module, `MenasException.scala` placed in dao module could be renamed to `DaoException.scala` or so. I don't want to refactor the whole module though (at least not without any prior discussion with the team 😄  )

dao/src/main/scala/za/co/absa/enceladus/dao/rest/CrossHostApiCaller.scala
```diff
@@ -68,20 +78,34 @@ protected class CrossHostApiCaller private(apiBaseUrls: Vector[String], maxTryCo
...
+          val nextUrl = nextBaseUrl()
+          logFailure(e, url, attemptNumber, Option(nextUrl))
+          attempt(nextUrl, 1, urlsTried + 1)
+
+        case Failure(e: MenasException) if retryableExceptions.contains(e) && attemptNumber < maxTryCount =>
           logFailure(e, url, attemptNumber, None)
           attempt(url, attemptNumber + 1, urlsTried)
-        case Failure(e: RetryableException) if urlsTried < baseUrlsCount =>
+
+        case Failure(e: MenasException) if retryableExceptions.contains(e) && urlsTried < baseUrlsCount =>
```
benedeki, line 104: This can be done nicer than with repeated code. `retryableExceptions` (let's rename it to `optionallyRetryableExceptions`) a set of exception **classes** can be extended by exceptions that are always retried. This can be actually a private `val` of the class (let's call it `retryableExceptions`). Then the `match` can exists with the original three `case`s.
How to organize the exception class hierarchy is up to debate then. 
* First of all, the retyping of the `ResourceAccessException` and `RestClientException`seems OK.
* It can go either to `DaoException` or the new `AutoRecoverableException`
* The exceptions can all be direct descendants of the `MenasException` (or if renamed). And the `DaoException` and eventually `AutoRecoverableException` will be always added to the internal set of `retryableExceptions` (regardless if the were in the original `optionallyRetryableExceptions`)
* Or to be cleaner the hierarchy can be:
`MenasException` (still would prefer a better name) 
->`RetryableException` -> (`DaoException`, `AutoRecoverableException`)
->`OptionallyRetriableException` -> (`UnauthorizedException`, `NotFoundException`)

The later has the advantage, that the companion object of `OptionallyRetriableException` can hold the map of HTTP status codes.

benedeki (changes requested):
Tried to explain the requested code changes as clear as I could. If it's not clear though I am happy to do a pair programming session.
Also the unit tests need to be reworked not to give false positive.

lsulak:
I've made a lot of changes but this is still a draft PR - I might refactor some things soon. I also need to understand the Scala typing system a bit more.

dao/src/main/scala/za/co/absa/enceladus/dao/rest/RestClient.scala
```diff
@@ -85,7 +85,7 @@ protected class RestClient(authClient: AuthClient,
         log.warn(s"Response - $statusCode : ${Option(response.getBody).getOrElse("None")}")
         log.warn(s"Unauthorized $method request for Menas URL: $url")
         if (retriesLeft <= 0) {
-          throw UnauthorizedException("Unable to reauthenticate, no retries left")
+          throw OptionallyRetryableException.UnauthorizedException("Unable to reauthenticate, no retries left")
```
lsulak, line 88: I still haven't split this case `case HttpStatus.UNAUTHORIZED | HttpStatus.FORBIDDEN =>` here into two cases. I started but then I realized that if this is split into two exceptions and two case branches, then exactly the same code would be executed (doesn't matter if this is encapsulated into a sub-function or not, that's just a cosmetic thing). 

`UnauthorizedException` is used in a couple of places and the usage is quite straightforward - this will be raised/caught whenever there are problems with authorization. So I've decided to keep it as is for now. But I'm happy to be persuaded.

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/HyperConformance.scala
```diff
@@ -167,7 +170,13 @@ object HyperConformance extends StreamTransformerFactory with HyperConformanceAt
     } else {
       None
     }
-    new HyperConformance(menasBaseUrls, menasUrlsRetryCount, menasSetup)
+    val optionallyRetryableExceptions: Set[OptionallyRetryableException.OptRetryableExceptionsType] =
+      conf.getList(classOf[Int], menasOptionallyRetryableExceptions)
+        .asScala
+        .map(OptionallyRetryableException.mapIntToOptionallyRetryableException(_))
+        .toSet
```
benedeki, line 177: **Minor:** would do the `.toSet` first and then `.map`. 😉

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/HyperConformanceAttributes.scala
```diff
@@ -48,6 +49,11 @@ trait HyperConformanceAttributes extends HasComponentAttributes {
       PropertyMetadata("Menas API URL retry count",
         Some("How many times a call to Menas API URL should be retried after failure before proceeding to the next URL. E.g. 2"),
         required = false),
+    menasOptionallyRetryableExceptions ->
+      PropertyMetadata("Menas API optionally retryable HTTP exceptions",
+        Some("""List of HTTP status codes that are allowed to be retryable, e.g. [404,402] (retry count is """ +
```
benedeki, line 54: Would list all supported ones, so the user doesn't need to investigate which ones are allowed.

spark-jobs/src/main/resources/reference.conf
```diff
@@ -25,10 +25,11 @@
 menas.rest.uri="http://localhost:8080,host2:9000/menas;https://localhost:8080,host2:9000/menas"
 # Each of the above uris can be tried multiple times for fault-tolerance (optional)
 menas.rest.retryCount=0
+menas.rest.optionallyRetryableExceptions=[401,403,404]
+
```
benedeki, line 29: This will change the behavior of Enceladus deployment. Keep it empty and only in comment list the supported ones.

benedeki: Btw, what's the behavior if the key is not present at all? Ideally should default to empty list (not keep compatibility. But if in `reference.conf` still OK.

dao/src/main/scala/za/co/absa/enceladus/dao/MenasException.scala
```diff
@@ -0,0 +1,54 @@
...
+
+package za.co.absa.enceladus.dao
+
+abstract class MenasException(message: String, cause: Throwable) extends Exception(message, cause)
+
+abstract class RetryableException(message: String, cause: Throwable) extends MenasException(message, cause)
+
+abstract class OptionallyRetryableException(message: String, cause: Throwable) extends MenasException(message, cause)
+
+object RetryableException {
```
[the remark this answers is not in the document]
lsulak: Okay, I have no objections to that. 

I usually preferred to implement abstract classes and interfaces on the top, and then specific implementations on the bottom of a file, so that a collection of all abstract things can be seen easily. But I guess that there are no good arguments pro or against this so I'm happy to follow whichever style you/team prefer.

dao/src/main/scala/za/co/absa/enceladus/dao/rest/CrossHostApiCaller.scala
```diff
@@ -30,25 +30,47 @@ object CrossHostApiCaller {
...
+  private def createInstance(
+      apiBaseUrls: Seq[String],
+      urlsRetryCount: Int,
+      startWith: Option[Int],
+      optionallyRetryableExceptions: Set[OptionallyRetryableException.OptRetryableExceptionsType]
+  ): CrossHostApiCaller = {
     val maxTryCount: Int = (if (urlsRetryCount < 0) {
-      logger.warn(s"Urls retry count cannot be negative ($urlsRetryCount). Using default number of retries instead ($DefaultUrlsRetryCount).") //scalastyle:ignore maxLineLength
+      logger.warn(
+        s"Urls retry count cannot be negative ($urlsRetryCount). " +
```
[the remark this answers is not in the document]
lsulak: I'm slightly more inclined to strictly maintain the max char limit per line, (which is I believe 120 and originally it had 143 characters), but perhaps we can talk about this on the next CQC - whether to be strict or not in these situations. What do you think?

spark-jobs/src/main/resources/reference.conf
lsulak: Very good point. I think that we have to make a decision about whether this config should be mandatory or not.

I can see that `menas.rest.retryCount` is optional, which might indicate that the whole 'retry mechanism' - driven by the configuration - is 'optional'. Can I change this in a way that if this is empty/not existing it will return an empty List?

(I didn't want to make this decision alone since I just joined the team/project.)

dao/src/main/scala/za/co/absa/enceladus/dao/MenasException.scala
```diff
@@ -0,0 +1,54 @@
...
+
+  type RetryableExceptionsType = Class[_ <: MenasException]
```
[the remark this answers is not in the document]
lsulak: Type removed, see my next commit. 

--
Regarding adding Set: I wouldn't be able to use it in `mapIntToOptionallyRetryableException` if it's a set.

--
Why `MenasException` and not `RetryableException` - I wasn't able to make the `union` (in the old version of `CrossHostApiCaller.scala` in this PR) work with `RetryablException` because of the type system - but it was fixed by using `++` and not `union`. I think that the reason is a type broadening trick that's used in `++` method:

`def ++[B >: A, That](that: GenTraversableOnce[B])(implicit bf: CanBuildFrom[Repr, B, That]): That = {`

(jfyi definition of the traversable collection that's used is: `trait TraversableLike[+A, +Repr]`)

dao/src/main/scala/za/co/absa/enceladus/dao/rest/AuthClient.scala
```diff
@@ -21,15 +21,16 @@ import org.springframework.security.kerberos.client.KerberosRestTemplate
 import org.springframework.util.LinkedMultiValueMap
 import org.springframework.web.client.RestTemplate
 import za.co.absa.enceladus.dao.auth._
-import za.co.absa.enceladus.dao.UnauthorizedException
+import za.co.absa.enceladus.dao.OptionallyRetryableException
```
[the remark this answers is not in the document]
lsulak: I just recently realized that this is possible, nice one, thank you!

dao/src/main/scala/za/co/absa/enceladus/dao/rest/RestClient.scala
benedeki: This is the last remaining open item, I see. Let's bring it up at tomorrow stand-up.

dao/src/main/scala/za/co/absa/enceladus/dao/rest/RestClient.scala
```diff
@@ -76,23 +83,37 @@ protected class RestClient(authClient: AuthClient,
...
     val statusCode = response.getStatusCode
 
+    /**
+     * This function handles unauthorized response by trying to authenticate
+     * (if there are still some retries attempt left) - this might be due to an expired session.
+     *
+     * @param exceptionToThrow Instance of an exception that should be thrown if there are no retries left.
+     */
+    def handleUnauthorizedResponse(exceptionToThrow: OptionallyRetryableException): Unit = {
+      log.warn(s"Response - ${statusCode} : ${Option(response.getBody).getOrElse("None")}")
```
dk1844, line 93:
```suggestion
      log.warn(s"Response - ${statusCode} : ${Option(response.getBody).getOrElse("<empty body>")}")
```
or just 
```suggestion
      log.warn(s"Response - ${statusCode} : ${Option(response.getBody).getOrElse("")}")
```
or something similar - `None` makes little sense when expecting a body from the log-reader's perspective.

dk1844 (dismissed):
LGTM (just read the code).

I have suggested some small things (optional - pun intended), but nicely done!

dao/src/main/scala/za/co/absa/enceladus/dao/rest/RestClient.scala
lsulak: I like it :)

spark-jobs/src/main/scala/za/co/absa/enceladus/common/CommonJobExecution.scala
```diff
@@ -56,8 +57,16 @@ trait CommonJobExecution extends ProjectMetadata {
 
   protected val log: Logger = LoggerFactory.getLogger(this.getClass)
   protected val configReader: ConfigReader = new ConfigReader()
+
   protected val menasBaseUrls: List[String] = MenasConnectionStringParser.parse(configReader.getString("menas.rest.uri"))
   protected val menasUrlsRetryCount: Option[Int] = configReader.getIntOption("menas.rest.retryCount")
+  protected val optionallyRetryableExceptions: Set[OptionallyRetryableException.OptRetryableExceptions] =
```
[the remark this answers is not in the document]
lsulak: I thought about it again and I agree that it's redundant information so I'll drop it.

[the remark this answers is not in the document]
dk1844: OK, I did not notice `-Type` suffix was already there. What you did is an improvement, fine by me.
People have different tastes in naming stuff, that's natural :)

dk1844 (dismissed):
LGTM (just read the code)

dao/src/main/scala/za/co/absa/enceladus/dao/rest/RestClient.scala
```diff
@@ -76,23 +83,37 @@ protected class RestClient(authClient: AuthClient,
 
     val statusCode = response.getStatusCode
 
+    /**
+     * This function handles unauthorized response by trying to authenticate
+     * (if there are still some retries attempt left) - this might be due to an expired session.
+     *
+     * @param exceptionToThrow Instance of an exception that should be thrown if there are no retries left.
+     */
+    def handleUnauthorizedResponse(exceptionToThrow: OptionallyRetryableException): Unit = {
```
benedeki, line 92:
```suggestion
    def handleUnauthorizedResponse(exceptionToThrow: =>OptionallyRetryableException): Unit = {
```
In this situation it's nice to use _by-name parameter_
Explanation can be found here for example:
https://www.tutorialspoint.com/scala/functions_call_by_name.htm
https://alvinalexander.com/scala/fp-book/how-to-use-by-name-parameters-scala-functions/

dao/src/main/scala/za/co/absa/enceladus/dao/rest/RestClient.scala
```diff
@@ -76,23 +83,37 @@ protected class RestClient(authClient: AuthClient,
...
-        log.warn(s"Unauthorized $method request for Menas URL: $url")
-        if (retriesLeft <= 0) {
-          throw UnauthorizedException("Unable to reauthenticate, no retries left")
-        }
-
-        log.warn(s"Expired session, reauthenticating")
-        authenticate()
+      case HttpStatus.UNAUTHORIZED =>
+        handleUnauthorizedResponse(UnauthorizedException("Unauthorized, unable to reauthenticate, no retries left"))
+        send[B, T](method, url, headers, bodyOpt, retriesLeft - 1)
```
benedeki, line 113: Why is the `send` not part of the `handleUnauthorizedResponse`?

benedeki (approved):
Approving as the two comments are totally not blocking, just educational.
(Still wouldn't mind them being addressed ;-) )

dao/src/main/scala/za/co/absa/enceladus/dao/rest/RestClient.scala
lsulak: I tried but it wouldn't be a tail recursive function - at least I couldn't make it work

benedeki: Indeed, I missed the fact that the function is recursive. 
It's good to comment such things - why some **obvious* thing was **not done*, similarly why some **not obvious** things **were done** 😉

benedeki (approved):

utils/src/main/scala/za/co/absa/enceladus/utils/config/ConfigReader.scala
```diff
@@ -60,6 +60,14 @@ class ConfigReader(val config: Config = ConfigFactory.load()) {
     config.getBoolean(path)
   }
 
+  def getIntList(path: String): List[Int] = {
```
benedeki, line 63: **Less important:**
The `config.getIntList` is sensitive, that the entry has to be JSON style list of whole number. That makes it impossible to override the value with `=Dkey=value` on command line, as it seems it turns the value into string always.
If not too much code move, would be nice to use the implementation from  https://github.com/AbsaOSS/enceladus/blob/6d1eb17a328fe6d3a4a7fa4b8e4936fad26490d4/utils/src/main/scala/za/co/absa/enceladus/utils/config/ConfigReader.scala#L63 
It even has a good amount of unit tests at https://github.com/AbsaOSS/enceladus/blob/6d1eb17a328fe6d3a4a7fa4b8e4936fad26490d4/utils/src/test/scala/za/co/absa/enceladus/utils/config/ConfigReaderSuite.scala#L116

benedeki (changes requested):

utils/src/main/scala/za/co/absa/enceladus/utils/config/ConfigReader.scala
lsulak: I was thinking about it for some time, but the reason why `getIntList` from the links above works is, I believe, due to `getListFromString` which is a fallback mechanism. Now, in `getListFromString` the whole module `za.co.absa.enceladus.utils.implicits.StringImplicits._` is being loaded, so it's actually a really big piece of code that I would need to copy. I'll still consider it because it might be useful sooner or later anyway - and we already have the code & unit tests, so it's not that I'm gonna actually write a tons of stuff from scratch. Thanks for the comment.

spark-jobs/src/main/scala/za/co/absa/enceladus/common/CommonJobExecution.scala
```diff
@@ -56,8 +58,16 @@ trait CommonJobExecution extends ProjectMetadata {
...
   protected val configReader: ConfigReader = new ConfigReader()
+
   protected val menasBaseUrls: List[String] = MenasConnectionStringParser.parse(configReader.getString("menas.rest.uri"))
   protected val menasUrlsRetryCount: Option[Int] = configReader.getIntOption("menas.rest.retryCount")
+  protected val optionallyRetryableExceptions: Set[OptRetryableExceptions] =
+    configReader
+      .getIntListOption("menas.rest.optionallyRetryableExceptions")
+      .getOrElse(Set.empty)
+      .toSet
+      .map(mapIntToOptionallyRetryableException(_: Int))
```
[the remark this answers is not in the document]
lsulak: I know, I just tried to be explicit. But since Scala compiler is pretty powerful, and should be able to check this during compile time, I can remove this.

spark-jobs/src/main/scala/za/co/absa/enceladus/common/CommonJobExecution.scala
```diff
@@ -56,8 +58,16 @@ trait CommonJobExecution extends ProjectMetadata {
 
   protected val log: Logger = LoggerFactory.getLogger(this.getClass)
   protected val configReader: ConfigReader = new ConfigReader()
+
   protected val menasBaseUrls: List[String] = MenasConnectionStringParser.parse(configReader.getString("menas.rest.uri"))
   protected val menasUrlsRetryCount: Option[Int] = configReader.getIntOption("menas.rest.retryCount")
+  protected val optionallyRetryableExceptions: Set[OptRetryableExceptions] =
+    configReader
```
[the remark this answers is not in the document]
lsulak: okay, I like the idea :)

dao/src/main/scala/za/co/absa/enceladus/dao/rest/AuthClient.scala
```diff
@@ -62,7 +63,7 @@ sealed abstract class AuthClient(username: String, restTemplate: RestTemplate, a
           log.info(s"Authentication successful: $username")
           getAuthHeaders(response)
         case _             =>
-          throw UnauthorizedException(s"Authentication failure ($statusCode): $username")
+          throw AuthenticationException(s"Authentication failure ($statusCode): $username")
```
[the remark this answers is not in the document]
lsulak: fixed; I'll think about the refactoring again, there might be a more intelligent way how to do this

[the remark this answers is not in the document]
lsulak: I was trying to add more test into RestClientSuite.scala, but then I realised that this scenario is very difficult to simulate with unit tests without overwriting a lot of code (or writing a new unit test module(s)). 

I think that an integration test would be a much much better candidate for covering this scenario with tests, because it just uses/calls too many modules (RestClient, AuthClient, CrossHostApiCaller, and a few helper modules).

jakipatryk (commented):
Just small notes. Besides them, LGTM, I did:
- build
- run unit tests
- run integration tests

spark-jobs/src/main/resources/reference.conf
```diff
@@ -25,10 +25,12 @@
 menas.rest.uri="http://localhost:8080,host2:9000/menas;https://localhost:8080,host2:9000/menas"
 # Each of the above uris can be tried multiple times for fault-tolerance (optional)
 menas.rest.retryCount=0
+# Currently supported ones are these: [401,403,404]
+menas.rest.optionallyRetryableExceptions=[]
```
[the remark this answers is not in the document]
lsulak: I was thinking about it, but I wanted to make it consistent with the rest of the repo and your change is not yet merged. But I can see that the README was already changed, so if you insist I'll change that

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/HyperConformanceAttributes.scala
```diff
@@ -48,6 +50,13 @@ trait HyperConformanceAttributes extends HasComponentAttributes {
       PropertyMetadata("Menas API URL retry count",
         Some("How many times a call to Menas API URL should be retried after failure before proceeding to the next URL. E.g. 2"),
         required = false),
+    menasOptionallyRetryableExceptions ->
+      PropertyMetadata("Menas API optionally retryable HTTP exceptions",
```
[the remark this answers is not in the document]
lsulak: I've changed all of these, please check the newest commit

jakipatryk (approved):

jakipatryk (dismissed):

dao/src/main/scala/za/co/absa/enceladus/dao/rest/AuthClient.scala
```diff
@@ -61,7 +61,7 @@ sealed abstract class AuthClient(username: String, restTemplate: RestTemplate, a
       val response = requestAuthentication(url(baseUrl))
       val statusCode = response.getStatusCode
 
-      lazy val errMessage = s"Authentication failure ($statusCode): $username"
+      def errMessage = s"Authentication failure ($statusCode): $username"
```
benedeki, line 64: Customary the embedded/nested function are defined right under the signature of the parent function (yes that would require this one to have at least one input parameter of `statusCode`

lsulak: All right, thank you. In that case I feel like it's not worth the effort, the benefit is too small since it's really just a string, that doesn't involve any demanding computation.

jakipatryk (approved):

benedeki (commented):
- [X] code reviewed
- [X] pulled
- [X] built
- [x] run

benedeki (approved):

PR #2098, opened 2022-07-28
#2090: Integrate Spark-Commons 0.3.x

### LINKED ISSUE #2090: Integrate Spark-Commons 0.3.x
## Background
_Spark-Commons 0.3.0_ has been released.

## Feature
Upgrade to and integrate the new version of _Spark-Commons_

benedeki:
* integration of Spark-commons and Spark-commons-test 0.3.1
* function `cache` replaced with `cacheIfNotCahcedyet`
* removed filter for test logs of `DataFrame already being cached
* `TZNormalizedSparkTestBase` take into account eventual timezone config

Closes #2090

### DIFF
--- a/spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/interpreter/DynamicInterpreter.scala
+++ b/spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/interpreter/DynamicInterpreter.scala
@@ -27,9 +27,10 @@ import za.co.absa.enceladus.conformance.config.ConformanceConfigParser
 import za.co.absa.enceladus.conformance.datasource.PartitioningUtils
 import za.co.absa.enceladus.conformance.interpreter.rules._
 import za.co.absa.enceladus.conformance.interpreter.rules.custom.CustomConformanceRule
-import za.co.absa.enceladus.conformance.interpreter.rules.mapping.{MappingRuleInterpreter, MappingRuleInterpreterBroadcast, MappingRuleInterpreterGroupExplode}
+import za.co.absa.enceladus.conformance.interpreter.rules.mapping.{MappingRuleInterpreter,
+  MappingRuleInterpreterBroadcast, MappingRuleInterpreterGroupExplode}
 import za.co.absa.enceladus.dao.MenasDAO
-import za.co.absa.enceladus.model.conformanceRule.{ConformanceRule, _}
+import za.co.absa.enceladus.model.conformanceRule._
 import za.co.absa.enceladus.model.{Dataset => ConfDataset}
 import za.co.absa.enceladus.utils.config.PathWithFs
 import za.co.absa.enceladus.utils.error.ErrorMessage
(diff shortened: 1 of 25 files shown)


utils/src/main/scala/za/co/absa/enceladus/utils/testUtils/TZNormalizedSparkTestBase.scala
```diff
@@ -24,7 +24,9 @@ trait TZNormalizedSparkTestBase extends SparkTestBase {
     val result = super.initSpark
 
     //TODO make conditional on empty SparkTestBase.timezone, once SparkCommons 0.3.0 will have been released
-    TimeZoneNormalizer.normalizeAll(result)
+    if (sparkConfig.timezone.isEmpty) {
```
lsulak, line 27: the comment above seems obsolete?

benedeki: Of course it is. Thank you for noticing. 👍

benedeki:
It's just simple refactoring, no feature code changed. Adding `Testing not needed` label.

utils/pom.xml
```diff
@@ -59,7 +59,12 @@
...
             <groupId>za.co.absa</groupId>
-            <artifactId>spark-commons_${scala.compat.version}</artifactId>
+            <artifactId>spark-commons-spark${spark.compat.version}_${scala.compat.version}</artifactId>
+            <version>${absa.spark.commons.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>za.co.absa</groupId>
+            <artifactId>spark-commons-test_${scala.compat.version}</artifactId>
             <version>${absa.spark.commons.version}</version>
         </dependency>
```
dk1844, line 69: At first glance, I would expect this dependency to be scoped to `test` only, but I understand that compile-scoped `testUtils` currently prevent you from doing so.

Perhaps some other time in the future, when we drop mvn support as it would get easier.

dk1844 (approved):
LGTM (read the code, checked out, ran tests)

dk1844:
> It's just simple refactoring, no feature code changed. Adding `Testing not needed` label.

Hey, I am sorry, but I beg to differ a bit - `cache` has changed to `chacheIfNotCachedYet()` and that _is_ a behavior change. Albeit I agree, this should only reflect in warnings. Whether we can afford not to test it is another question - and perhaps we can, since it is covered in the tests quite nicely.

benedeki:
> > It's just simple refactoring, no feature code changed. Adding `Testing not needed` label.
> 
> Hey, I am sorry, but I beg to differ a bit - `cache` has changed to `chacheIfNotCachedYet()` and that _is_ a behavior change. Albeit I agree, this should only reflect in warnings. Whether we can afford not to test it is another question - and perhaps we can, since it is covered in the tests quite nicely.

Well, even that should be "invisible" as the warnings were filtere out previously. The filter now removed.
So you are actually right, maybe it should/can be checked... 😉  (removing lable)

utils/pom.xml
benedeki: Indeed, on both accounts.

spark-jobs/src/main/scala/za/co/absa/enceladus/conformance/interpreter/DynamicInterpreter.scala
```diff
@@ -358,7 +360,7 @@ case class DynamicInterpreter(implicit inputFs: FileSystem) {
       // Cache the data first since Atum will execute an action for each control metric
       val cachedDf = persistStorageLevel match {
         case Some(level) => df.persist(level)
-        case None        => df.cache
+        case None        => df.cacheIfNotCachedYet()
```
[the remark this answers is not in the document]
benedeki: 🙇

PR #2103, opened 2022-08-02
GHP: Rest API v3 desc

### LINKED ISSUE #2092: Document the new REST API of Enceladus 3
## Background
There's a new REST API in Enceladus 3

## Feature
Add a description of reference to the REST API into https://absaoss.github.io/enceladus/

dk1844:
- Enceladus v3 REST API info has been added to GH pages (basic info, linking to built-in Swagger mini-site). 
 - V2-like Menas API doc page has been removed.

Closes #2092


Gemfile.lock
```diff
@@ -25,10 +25,12 @@ GEM
     ethon (0.12.0)
       ffi (>= 1.3.0)
     eventmachine (1.2.7)
+    eventmachine (1.2.7-x64-mingw32)
```
Zejnilovic, line 28: This is weird. Why are there 2 eventmachine dependencies?

Zejnilovic (approved):

Gemfile.lock
Zejnilovic: So this would be windows one? It should be fine, but if the build/site breaks, I would look at this as the first culprit.

benedeki: I have Gemfile.lock exempt from commit, I believe

benedeki (approved):

Gemfile.lock
dk1844: Hmm, maybe we should exempt it here, too, later on.

PR #2122, opened 2022-08-30
#2121 Implement retries quadratic+randomised backoff mechanism

### LINKED ISSUE #2121: On failed calling Rest-API wait for random moment
## Background
When a _Rest-API_ calls fails from _SparkJobs_ it is retried immediately, possibly creating an ingestion on the Rest endpoint.

## Feature
In case of a failure & retry on Rest API call add a random length ~2 seconds sleep.

jakipatryk:
- Added `retryBackoffStrategy` parameter to CrossHostApiCaller constructor
- Used `retryBackoffStrategy` as a step in retrying a call
- Implemented quadratic `retryBackoffStrategy` with randomised addition of 0-1s to delay

Closes #2121

### DIFF
--- a/dao/src/main/scala/za/co/absa/enceladus/dao/rest/CrossHostApiCaller.scala
+++ b/dao/src/main/scala/za/co/absa/enceladus/dao/rest/CrossHostApiCaller.scala
@@ -65,7 +65,7 @@ object CrossHostApiCaller {
   }
 
   private def quadraticRandomizedRetryBackoffStrategy(retryNumber: Int): Unit = {
-    val inSeconds: Double = retryNumber * retryNumber + Random.nextDouble()
+    val inSeconds: Double = retryNumber * retryNumber + retryNumber * Random.nextDouble()
     val inMillis = inSeconds * 1000
     Thread.sleep(inMillis.toLong)
   }
--- a/dao/src/main/scala/za/co/absa/enceladus/dao/rest/CrossHostApiCaller.scala
+++ b/dao/src/main/scala/za/co/absa/enceladus/dao/rest/CrossHostApiCaller.scala
@@ -37,7 +37,7 @@ object CrossHostApiCaller {
       urlsRetryCount: Int,
       startWith: Option[Int],
       optionallyRetryableExceptions: Set[OptRetryableExceptions],
-      retryBackoffStrategy: Int => Unit
+      retryBackoffStrategy: Int => Int
   ): CrossHostApiCaller = {
     val maxTryCount: Int = (if (urlsRetryCount < 0) {
       logger.warn(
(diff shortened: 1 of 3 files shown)


dao/src/main/scala/za/co/absa/enceladus/dao/rest/CrossHostApiCaller.scala
```diff
@@ -48,24 +49,34 @@ object CrossHostApiCaller {
...
-      optionallyRetryableExceptions: Set[OptRetryableExceptions] = Set.empty
+      optionallyRetryableExceptions: Set[OptRetryableExceptions] = Set.empty,
+      retryBackoffStrategy: Int => Unit = quadraticRandomizedRetryBackoffStrategy
    ): CrossHostApiCaller = {
-    createInstance(apiBaseUrls, urlsRetryCount, startWith, optionallyRetryableExceptions)
+    createInstance(apiBaseUrls, urlsRetryCount, startWith, optionallyRetryableExceptions, retryBackoffStrategy)
+  }
+
+  private def quadraticRandomizedRetryBackoffStrategy(retryNumber: Int): Unit = {
+    val inSeconds: Double = retryNumber * retryNumber + Random.nextDouble()
```
lsulak, line 68: I like how you added this functionality into the module, but I'm not sure if this is random enough. @benedeki what do you think, based on your experiences?

Imagine that 5 parallel processes would execute the same REST API call at the same time, then the timings would always differ only by 0-1 second. Let's say that the retryNumber is 4:
1. Fail, wait 1+(0 up to 1) seconds
2. Fail, wait 4+(0 up to 1) seconds
3. Fail, wait 9+(0 up to 1) seconds
4. Fail, wait 16+(0 up to 1) seconds.

The random part will be more and more insignificant over time. Something like this is more 'random' (=randomness will be slightly more impactful): `retryNumber * retryNumber + Random.nextDouble() * retryNumber`

But perhaps your solution is good enough as well for edge case scenarios.

jakipatryk: I'm open to any changes to the formula, I like `retryNumber * retryNumber + Random.nextDouble() * retryNumber` as well (now I'm thinking that this one is better than the current one), we can discuss it tomorrow/wait for @benedeki response.

lsulak:
Looks good, just a minor point / potential improvement about the algorithm.

- [X] code reviewed
- [X] pulled
- [X] built
- [x] run

lsulak (approved):

lsulak (dismissed):

dao/src/main/scala/za/co/absa/enceladus/dao/rest/CrossHostApiCaller.scala
```diff
@@ -48,24 +49,34 @@ object CrossHostApiCaller {
...
+    )
   }
 
   def apply(
       apiBaseUrls: Seq[String],
       urlsRetryCount: Int = DefaultUrlsRetryCount,
       startWith: Option[Int] = None,
-      optionallyRetryableExceptions: Set[OptRetryableExceptions] = Set.empty
+      optionallyRetryableExceptions: Set[OptRetryableExceptions] = Set.empty,
+      retryBackoffStrategy: Int => Unit = quadraticRandomizedRetryBackoffStrategy
```
dk1844, line 62: I might have missed that but it seems to me that `quadraticRandomizedRetryBackoffStrategy` is not used in tests at all. In such a case, wouldn't it make sense to cover it a bit, too?

jakipatryk: Changed signature from Int => Unit to Int => Int (now returns number of milliseconds to wait), and added tests.

dao/src/test/scala/za/co/absa/enceladus/dao/rest/CrossHostApiCallerSuite.scala
```diff
@@ -151,6 +151,89 @@ class CrossHostApiCallerSuite extends BaseTestSuite {
...
+          retryBackoffStrategy = mockedRetryBackoffStrategy
+        ).call { str =>
+          restClient.sendGet[String](str)
+        }
+
+        val inOrder = Mockito.inOrder(mockedRetryBackoffStrategy)
+        inOrder.verify(mockedRetryBackoffStrategy).apply(1)
+        inOrder.verify(mockedRetryBackoffStrategy).apply(2)
+        inOrder.verify(mockedRetryBackoffStrategy, Mockito.never()).apply(3)
+        // times 2 because one time in between URIs switch, and another after switched to new URI and failed
```
dk1844, line 232: It took me a while to understand why this is here, but spot on 👍

dk1844 (approved):
LGTM (read, checked out, built and fiddled with the tests)

It pretty cool that you managed bend the implementation to test the quadratic backoff without actually slowing down the tests run to so. 👍 

I wonder, however, if there should not be an integration test in the future that would actually show the true backoff sleeping (out of scope here, I assume).

Issue #2111, opened 2022-08-10
Rename entities in code that mention Menas but REST API is meant


jakipatryk:
Menas was API+UI, but now is just UI. However, there are still plenty of variables/classes/etc. with _menas_ in the name that should be somehow renamed, for example:
- `CommonJobExecution` has _vals_: `menasBaseUrls`, `menasUrlsRetryCount`, `menasSetup`
- `MenasConnectionStringParser`
- menas-credentials.conf (and other keytab/credentials related things)
- `MenasDAO` (?)
- data-model/menas (?)

## Note
While some of them are pretty obvious renames (like `menasBaseUrls` to `restApiBaseUrls`, or menas-credentials.conf to rest_api-credentials.conf), a few are non-obvious and probably need a bit of discussion (like `MenasDAO`, whether rename it at all, and if so into what).

jakipatryk:
Release notes
Renamed Menas API mentions to REST API.

PR #2127, opened 2022-09-08
#2126 Implement wrapper atop HDFS Service

### LINKED ISSUE #2126: Make HDFS work even without HADOOP_CONF_DIR
## Feature
Make HDFS work even without HADOOP_CONF_DIR

Zejnilovic:
1. Instead of ENV variable `HADOOP_CONF_DIR` use property `menas.fs.config.type` 
2. Add a wrapper around the FileSystem so we can implement `None` type and alter `S3` type after `HDFS` type

Closes #2126


rest-api/src/main/scala/za/co/absa/enceladus/rest_api/EnceladusFileSystem.scala
```diff
@@ -0,0 +1,35 @@
...
+
+  def isDirectory(path: Path): Boolean = fileSystem.getOrElse(throw exception).isDirectory(path)
+
+  def exists(path: Path): Boolean = fileSystem.getOrElse(throw exception).exists(path)
+}
+
+object EnceladusFileSystem {
+  def apply(fs: FileSystem): EnceladusFileSystem = EnceladusFileSystem(Some(fs))
+  def apply(): EnceladusFileSystem = EnceladusFileSystem(None)
+}
```
dk1844, line 35: Here, I'd rather use a sealed trait and distinct factory-provided implementation that delegate the logic.

```suggestion
sealed trait EnceladusFileSystem {
  def getStatus: FsStatus
  def listStatus(path: Path): Array[FileStatus]
  def isDirectory(path: Path): Boolean
  def exists(path: Path): Boolean
}

object EnceladusFileSystem {
  def apply(fs: FileSystem): EnceladusFileSystem = HdfsFileSystem(fs)
  def apply() : EnceladusFileSystem = empty
  val empty: EnceladusFileSystem = NoFileSystem
  
  private case class HdfsFileSystem(fileSystem: FileSystem) extends EnceladusFileSystem {
    def getStatus: FsStatus = fileSystem.getStatus
    def listStatus(path: Path): Array[FileStatus] = fileSystem.listStatus(path)
    def isDirectory(path: Path): Boolean = fileSystem.isDirectory(path)
    def exists(path: Path): Boolean = fileSystem.exists(path)
  }

  private case object NoFileSystem extends EnceladusFileSystem {
    val exception = new Exception("No FileSystem initialized")

    def getStatus: FsStatus = throw exception
    def listStatus(path: Path): Array[FileStatus] = throw exception
    def isDirectory(path: Path): Boolean = throw exception
    def exists(path: Path): Boolean = throw exception
  }
  
}
```

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/EnceladusFsConfig.scala
```diff
@@ -0,0 +1,48 @@
...
+    if (fsType.equalsIgnoreCase(HDFSOption)) {
+      logger.info(s"Using FS config for $HDFSOption")
+      EnceladusFileSystem(HDFSConfig.hadoopFS()(spark))
+    } else if (fsType.equalsIgnoreCase(NoneOption) || fsType.isEmpty) {
+      logger.info(s"Using FS config for $NoneOption")
+      EnceladusFileSystem()
+    } else {
+      throw new Exception("Unsupported FileSystem")
+    }
+  }
```
dk1844, line 42: Perhaps 

```suggestion
    fsType.toLowerCase match {
      case HDFSOption =>
        logger.info(s"Using FS config for $HDFSOption")
        EnceladusFileSystem(HDFSConfig.hadoopFS()(spark))
        
      case none if none == NoneOption || none.isEmpty =>
        logger.info(s"Using FS config for $NoneOption")
        EnceladusFileSystem.empty
        
      case _ =>
        throw new Exception("Unsupported FileSystem")
    }
```

would be more readable?

dk1844 (dismissed):
LGTM overall, thanks for explaining the usecase to me, @Zejnilovic (hint: perhaps the usecase may be written in the issue/pr desc?)

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/EnceladusFsConfig.scala
```diff
@@ -0,0 +1,48 @@
...
+}
+
+object EnceladusFsConfig {
+  final val HDFSOption = "hdfs"
+  final val NoneOption = "none"
+}
```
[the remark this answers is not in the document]
Zejnilovic: Thought about it. I am not sure, honestly. I use `@Value` to get the user-defined value. What would be the easy way of transforming that string into an Object?

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/EnceladusFileSystem.scala
Zejnilovic: This is a very nice/pretty solution.

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/EnceladusFsConfig.scala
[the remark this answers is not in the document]
dk1844: `Enums` have method `withName` for this purpose; with `case objects` you would need to do it yourself. With both you get exhaustivness checking.

One perspective is also code readability - simply put organizing set of options in a common structural type that can be referenced elsewhere. This does it too to a degree, but by convention, not by typing.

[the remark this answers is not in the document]
Zejnilovic: So I went with it. If you find it better, let's use it. If not, I can revert.

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/EnceladusFsConfig.scala
```diff
@@ -0,0 +1,54 @@
...
+        throw new Exception("Unsupported FileSystem")
+    }
+  }
+}
+
+object EnceladusFsType extends Enumeration {
+  val NoFs = Value("none")
+  val Empty = Value("")
+  val Hdfs = Value("hdfs")
+}
```
dk1844, line 54:
```suggestion
  def hadoopFS(): EnceladusFileSystem = {
    Try(EnceladusFsType.withName(fsType.toLowerCase)) match {
      case Success(EnceladusFsType.Hdfs) =>
        logger.info(s"Using FS config for HDFS.")
        EnceladusFileSystem(HDFSConfig.hadoopFS()(spark))

      case Success(EnceladusFsType.NoFs) =>
        logger.info(s"Not using any FS config.")
        EnceladusFileSystem.empty

      case _ =>
        throw new Exception(s"Unsupported FileSystem (found: '$fsType', supported: ${EnceladusFsType.values.mkString(", ")}, )")
    }
  }
}

object EnceladusFsType extends Enumeration {
  val NoFs = Value("none")
  val Hdfs = Value("hdfs")
}
```

And when you actually have only two options for `rest_api.fs.config.type`, perhaps you can afford to do it as suggested above - actually having just two types `none` and `hdfs` and not the empty third which seems to be there only for parsing purposes.

If you could provide a unit test for this (this would involve mocking the `rest_api.fs.config.type` value for each case), it would be even better.

dk1844 (commented):
It looks much better - just added one more thoughts mainly revolving aroung the existence of the third enum value `EnceladusFsType.Empty` which may not be needed, I believe.

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/EnceladusFsConfig.scala
dk1844: Sure, that seems like a suitable approach -- as long as we get it covered somewhere, I am ok.

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/HDFSConfig.scala
```diff
@@ -42,22 +41,25 @@ class HDFSConfig @Autowired() (spark: SparkSession) {
   val krb5username: String = ""
   @Value("${menas.hadoop.auth.krb5.keytab:}")
   val krb5keytab: String = ""
+  @Value("${menas.hadoop.conf.dir:}") // TODO change to rest-api
```
dk1844, line 44: Do you perhaps want to reference an issue here? E.g. #2111

dk1844 (approved):
LGTM

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/HDFSConfig.scala
Zejnilovic: I want to wait for @jakipatryk to merge his rename first.

Zejnilovic: Done

dk1844 (approved):
LGTM (just read the code)

miroslavpojer (approved):

PR #2150, opened 2022-11-22
Feature/2139 fix ASCII / Cobol unit test that uses the latest Cobrix

yruslan:
This is to be merged into Bart's branch


yruslan:
You're welcome!

Issue #1778, opened 2021-04-30
Speeding up Enceladus Build


Zejnilovic:
## Description
Our current build on Jenkins or CodeBuild takes 16+ minutes. This needs to be refactored a bit. Maybe we split out some parts of the project, maybe we move some tests around, like running integration only on integration.

This ticket is created to open up a discussion and not forget that this does pain us.

AdrianOlosutean:
I would like to propose the following ideas which could help:

- Have a bunch of profiles for building different parts of Enceladus together. Something like: Standardization/Conformance(building spark-jobs, utils, dao, plugins-builtin), rest-api (rest-api, data-model, migrations, utils), Menas(menas-ui) 
- Run integration tests only after having the Tested label(or another label) for a PR
- Have a label or a method of identifying a PR which doesn't need to run tests at all (such as updating README, scripts, examples, etc) which would skip tests

PR #2153, opened 2022-11-24
#1778 move to gh actions

Zejnilovic:
This does not completely fix the issue. But is a first step in bringing it closer.


Zejnilovic:
This will not be possible until we remove Integration tests from being part of unit tests or set up our own runners somewhere.

PR #2164, opened 2023-01-17
#2163 Loading of RecordIdGeneration settings fixed for Std/Std&Conf -…

### LINKED ISSUE #2163: IDGenerationStrategy setting is ignored for Standardization
## Describe the bug
IDGenerationStrategy setting `enceladus.recordId.generation.strategy` with values `uuid`, `stableHashId`, `none` (with `uuid` being the default) is available in Enceladus and expected to be used for both Standardizaton and Conformance.

The default (`uuid`) works as expected.
However, attempting to set non-default value (e.g. `stableHashId`) seems to result for this setting to be only respected in the Conformance phase, while Standardization running with the default `uuid` setting (effectively ignoring the setting).

## To Reproduce
1. set `enceladus.recordId.generation.strategy='stableHashId` in config or via `-D`
2. run std & conf
3. watch UUIDs being generated instead of stableHashId (integer values) being present for Std phase
4. (note that std/conf do not overwrite existing `enceladus_record_id` column)

## Expected behavior
non-default `enceladus.recordId.generation.strategy` setting should be respected in both std & conf phases.

## Additional context
Regression of this behavior may have appeared when the `spark-data-standardization` library was extracted. Thus consider:
1. remove any leftover code duplication (`RecordIdGeneration`)
2. add tests preventing from such a regression in the future

dk1844:
This PR attempts to fix #2163 

 - Loading of `RecordIdGeneration` settings fixed for Std/Std&Conf (extracted to a single method in StandardizationExecution)
 - RecordIdGeneration from spark-data-standardization library reused, local copy removed.

@miroslavpojer please test

Closes #2163


benedeki (approved):
- [x] code reviewed
- [ ] pulled
- [ ] built
- [ ] run

Not just seems to fix the issue, but improves the code too.

Would only check if the id is not generated twice if _Standardization_ and _Conformance_ are run separately. 🤔

miroslavpojer:
- [ ] code reviewed
- [ ] pulled
- [ ] built
- [ ] run

Problem tested on Enceladus's Hermes test. (Where problem was observed.) No Unit or Integration tests exist for this topic.
Periodical run of Hermes tests is required.

miroslavpojer (approved):

miroslavpojer:
> * [x]  code reviewed
> * [ ]  pulled
> * [ ]  built
> * [ ]  run
> 
> Not just seems to fix the issue, but improves the code too.
> 
> Would only check if the id is not generated twice if _Standardization_ and _Conformance_ are run separately. 🤔

I did test were Standardization and Conformance are done as separated spark-submit calls.

Issue #2159, opened 2023-01-10
Add "Run number" into additional info of info file


benedeki:
## Background
More info into the `INFO_FILE`.

## Feature
Put the _Run number_ into the `INFO_FILE`, as shown on the picture bellow

## Example
![image](https://user-images.githubusercontent.com/14905969/211532388-4a3972e1-1dfe-4f51-b243-9a8b27205314.png)

PR #2171, opened 2023-02-06
#2159: adding Run Number into Standardization spark job

lsulak:
Draft for now, I need to test a few things first & think about whether it should be part of Common Job Execution or only in the standardization. 

**Update**: tests passed and it works as expected in the Menas UI & INFO file & Mongo DB.

--

I can see how the standardization and conformance are logically intertwined. The current implementation should generally work but there might be a case where Run Number is not present at all in the INFO file - if Conformation Job runs separately against dataset that has INFO file produced by Standardization Job that didn't produce the Run Number (because it is older than the one in this PR). 

Q: Should the Run Number represent any standardization or conformance or combined job that ran against a dataset? If yes, then the current implementation is too naive and I'll add some check into the conformation job - something like: if the Run Number is not present in INFO file, add it.


lsulak:
I also implemented the sorting of Additional Info data in Menas, it's now sorted by ascending order.

Before:

Screenshot 2023-02-13 at 11 18 38

After:

Screenshot 2023-02-13 at 11 18 42

benedeki (approved):
- [x] code reviewed
- [ ] pulled
- [ ] built
- [ ] run

benedeki:
> 
> Q: Should the Run Number represent any standardization or conformance or combined job that ran against a dataset? If yes, then the current implementation is too naive and I'll add some check into the conformation job - something like: if the Run Number is not present in INFO file, add it.

I think it's good as it is now. No need to complicate it further.

miroslavpojer (approved):
- [x] code review
- [ ] pulled
- [ ] builts
- [ ] test run

dk1844 (approved):
LGTM (just read the code)

PR #2174, opened 2023-02-16
"Accept Dataset without conformance rules #2162"

### LINKED ISSUE #2162: Accept Dataset without conformance rules
## Background
When Dataset definition is submitted to Rest-API, it fails if list of conformance rules is not defined.
For easier operation (and eventual phase out of Conformance), it would be natural that if the Conformance rules are not provided it's considered and empty list.

## Feature
If `Dataset` (from `za.co.absa.enceladus.model`) has not defined property `conformance` during deserialization, consider it an empty list. 

## Proposed Solution
Solution Ideas:
1. Setting the property's default value to be `List.empty` should suffice
2. Verify then with Rest-APIs that insert a new Dataset or edit an existing one.

TebaleloS:
Defined the default value to be an empty list for `conformance` rule property in the `Dataset`
Added the test case for Dataset to consider `conformance` rules to be an `empty list` when they are not provided.
Closes #2162


data-model/pom.xml
```diff
@@ -87,6 +87,14 @@
                     </args>
                 </configuration>
             </plugin>
+            <plugin>
```
benedeki, line 90: Is this plugin needed in `data-model`? Try rolling it back if the whole thing builds.

data-model/pom.xml
```diff
@@ -87,6 +87,14 @@
...
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <source>8</source>
+                    <target>8</target>
+                </configuration>
```
benedeki, line 96: If **it is** needed then change it to:
```suggestion
                <version>${maven.compiler.plugin.version}</version>
                <configuration>
                    <source>${maven.compiler.source}</source>
                    <target>${maven.compiler.target}</target>
                </configuration>
```

menas/ui/npm-shrinkwrap.json
```diff
@@ -4,203 +4,288 @@
...
-      "version": "7.15.7",
-      "resolved": "https://registry.npmjs.org/@babel/helper-validator-identifier/-/helper-validator-identifier-7.15.7.tgz",
-      "integrity": "sha512-K4JvCtQqad9OY2+yTU8w+E82ywk/fe+ELNlt1G8z3bVGlZfn/hOcQQsUhGhW/N+tb3fxK800wLtKOE/aM0m72w==",
-      "dev": true
-    },
-    "@babel/highlight": {
-      "version": "7.14.5",
-      "resolved": "https://registry.npmjs.org/@babel/highlight/-/highlight-7.14.5.tgz",
-      "integrity": "sha512-qf9u2WFWVV0MppaL877j2dBtQIDgmidgjGk5VIMw3OadXvYaXn66U1BFlH2t4+t3i+8PhedppRv+i40ABzd+gg==",
+    "@ui5/cli": {
```
benedeki, line 7: I think it's an undesired artifact of build or merge. Please roll back,

menas/ui/package.json
```diff
@@ -9,7 +9,7 @@
     "@ui5/cli": "^2.8.1"
   },
   "config": {
-    "apiUrl": "${API_URL}",
+    "apiUrl": "http://localhost:8090/rest_api_war",
```
benedeki, line 12: Keep this change local, don't commit it please 😉

pom.xml
```diff
@@ -621,7 +621,7 @@
                         <configuration>
                             <!-- examples of class excluding -->
                             <!--
- excludes>
+                                excludes>
```
benedeki, line 624: Unnecessary change. Can stay, but if rolled back, even better. and best if new line remove too.

benedeki (changes requested):

menas/ui/package.json
TebaleloS: Okay, I actually forgot about that, will fix it

data-model/pom.xml
benedeki: * Was not rolled back (maybe it's needed, then add the comment) 😉 
* Please don't close a comment as `Resolved` when somebody else started it unless it's a _Suggestion_ you accepted

menas/ui/npm-shrinkwrap.json
benedeki: Still here.

benedeki (changes requested):

menas/ui/npm-shrinkwrap.json
TebaleloS: I think it came back after I rebased this branch from dev.

benedeki: _Rebase_ doesn't work that way. It keeps the changes you did. You need either to roll-back to a specific version of the file or easier just copy the old version's content and commit it. (The squash merge at the end will erase the fluctuation in the commit log)

Also, unless really needed, please refrain from _rebase_ once you presented your branch to others (= created a PR or cooperated with somebody on the branch). It avoids confusion and checkout issues.

data-model/src/main/scala/za/co/absa/enceladus/model/Dataset.scala
```diff
@@ -71,8 +71,7 @@ case class Dataset(name: String,
   def setSchemaVersion(newVersion: Int): Dataset = this.copy(schemaVersion = newVersion)
   def setHDFSPath(newPath: String): Dataset = this.copy(hdfsPath = newPath)
   def setHDFSPublishPath(newPublishPath: String): Dataset = this.copy(hdfsPublishPath = newPublishPath)
-  def setConformance(newConformance: Option[List[ConformanceRule]]): Dataset = this.copy(conformance = newConformance.getOrElse(Nil))
-  def setSchedule(newSchedule: Option[OozieSchedule]): Dataset = this.copy(schedule = newSchedule)
```
lsulak, line 75: (just a question for my own benefit)

I can see that you removed `setSchedule` method - okay, I can see that it's not used anywhere in the repo. However, what about the `schedule` attribute itself? I assume that for backward compatibility we want to keep it there - as well as `OozieSchedule` class and a few other usages across the codebase. If that's the case, can you just put a small comment above the attribute `schedule` about this?

What do you think, @benedeki?

lsulak (commented):
I didn't run the application on the localhost, but it looks good to me once these small comments will be resolved I will approve the PR.
- [x]  code reviewed
- [x]  pulled
- [x]  built
- [x]  ran tests

data-model/src/main/scala/za/co/absa/enceladus/model/Dataset.scala
benedeki: You are right @lsulak , the fields are kept in the model for compatibility (was planned to be removed with new data model).
Comment would be good. Please add it @TebaleloS

TebaleloS: Okay cool.

benedeki (commented):
- [x] code reviewed
- [ ] pulled
- [ ] built
- [ ] run

data-model/src/main/scala/za/co/absa/enceladus/model/Dataset.scala
```diff
@@ -49,7 +49,7 @@ case class Dataset(name: String,
                    userLocked: Option[String] = None,
                    conformance: List[ConformanceRule] = List.empty,
                    parent: Option[Reference] = None,
-                   schedule: Option[OozieSchedule] = None,
+                   schedule: Option[OozieSchedule] = None, //To be used for backward visioning compatibility
```
lsulak, line 52: visioning - typo?

benedeki (approved):

data-model/src/main/scala/za/co/absa/enceladus/model/Dataset.scala
TebaleloS: I have fixed it now.

lsulak (approved):
Didn't run it on localhost but I done everything else from the reviewer's perspective.

miroslavpojer (approved):
- [x] pulled
- [x] code review
- [x] built
- [x] test ran
- [x] manual API test done

PR #2178, opened 2023-03-02
Feature/2160 openapi3

### LINKED ISSUE #2160: Swagger API docs example value not showing proper values
## Describe the bug
When I run swagger API docs and look at the `example value` for both, create and returns, and I don't very well understand them, and they don't look like what I would expect.

## Screenshots
image

dk1844:
- OpenAPI 2 replaced by OpenAPI 3
 - supporting library switched from `springfox` (not maintianed anymore) to `springdoc`.
 - `MappingTable`'s dichtomy of both `defaultMappingValue: List` and `defaultMappingValues: Map` is now changed for json generation only to `defaultMappingValue` (unitTests updated)

## Swagger-reated
 - dev api vs prod api is now selectable in Swagger instead of depending on spring profile `dev`.
swagger-prod-api-dev-api-selector-labels

 - annotations added so that swagger site is populated better.
api-v3-swagger-openApi3

## Versions update notes
~Patch version update for `<spring.version>`: `2.0.0.RELEASE` -> `2.0.9.RELEASE`~ Reverted - just doing this breaks correct 401 reponses on JWT verification failing and yields 404 instead.

I have considered updating to
```xml
<spring.ldap.version>5.5.8</spring.ldap.version>
<spring.version>2.5.14</spring.version>
```
which seems to work ok (except for the 401->404 issue mentioned above), but the integration tests with embedded mongo do not work with this out of the box (or with light touchups at least - and there is no newer version of embedded mongo in the same major-version ballpark). However, a possible future consideration for a version update.
Note, for such a case, following config may be neede:
```properties
# https://stackoverflow.com/a/74430247
management.metrics.mongo.command.enabled=false
management.metrics.mongo.connectionpool.enabled=false
```

Closes #2160


data-model/pom.xml
```diff
@@ -58,6 +58,11 @@
             <artifactId>utils</artifactId>
             <version>${project.parent.version}</version>
         </dependency>
+        <dependency>
```
dk1844, line 61: This is added because in `data-model` we need annotations, while the API definition is done in `rest-api` module.

data-model/src/main/scala/za/co/absa/enceladus/model/MappingTable.scala
```diff
@@ -15,39 +15,78 @@
...
-                        hdfsPath: String,
+  @(AosSchema@field)(implementation = classOf[ZonedDateTime])
+  @BeanProperty dateDisabled: Option[ZonedDateTime] = None,
 
-                        schemaName: String,
-                        schemaVersion: Int,
+  @(AosSchema@field)(implementation = classOf[String], example = "user3")
+  @BeanProperty userDisabled: Option[String] = None,
 
-                        defaultMappingValue: List[DefaultValue] = List(),
```
Zejnilovic, line 34: Isn't this a change in the DB model? Thus quite a big change needed a Mongo update script as well.

dk1844: You are correct -- this is probably unwanted. I wanted to get rid of duplicate export fields `defaultMappingValues` and `defaultMappingValue`, that has stayed. Even though using the plural seemed better, it is probably not worth the hassle it would create on the persistence layer (I only imagined the change on the API level, which was bearable) - so I reverted the name to singular as it was.

Thanks for the idea.

Zejnilovic:
Have you thought about adding auth to it? Link to docs - https://swagger.io/docs/specification/authentication/bearer-authentication/

dk1844:
> ###EDIT
> Just tested. Now API returns 404 instead of 401 if I have a missing JWT header.

This part seems fixed now (I have used the original spring boot version from develop 🙄).

Now, the API should behave correctly and it's time to change auth that Swagger/openApi documents (I'm on it). Thanks big time for both of these realizations, @Zejnilovic!

dk1844:
> ###EDIT
> Just tested. Now API returns 404 instead of 401 if I have a missing JWT header.

This part seems fixed now (I have used the original spring boot version from develop 🙄).

Now, the API should behave correctly and it's time to change auth that Swagger/openApi documents (I'm on it). Thanks big time for both of these realizations, @Zejnilovic!

// edit: custom JWT auth option is now available in Swagger, too. Hint: try in Incognito, otherwise your browser might slip in SPNEGO token.

Zejnilovic:
All looks good now but I don't see a `login` documentation

dk1844:
> All looks good now but I don't see a `login` documentation

Correct observation. The endpoint is a special one introduced via the authorization config, so the usual annotation on the controller cannot be used. I added a manual ApiOperation description, heavily inspired by https://stackoverflow.com/a/74574834/1773349. So fixed in my 👀 .

Zejnilovic (dismissed):
Read code
Reviewed the swagger UI mostly

data-model/src/main/scala/za/co/absa/enceladus/model/MappingTable.scala
```diff
@@ -15,39 +15,78 @@
...
+import za.co.absa.enceladus.model.backend.audit._
+import za.co.absa.enceladus.model.dataFrameFilter.DataFrameFilter
+import za.co.absa.enceladus.model.versionedModel.VersionedModel
+
+import java.time.ZonedDateTime
+import scala.annotation.meta.field
+import scala.beans.BeanProperty
+
+case class MappingTable(
+  @(AosSchema@field)(example = "datasetA")
```
miroslavpojer, line 31: Is that copy&paste typo? "datasetA" ==> "mappingTableA"

data-model/src/main/scala/za/co/absa/enceladus/model/properties/PropertyDefinition.scala
```diff
@@ -15,39 +15,67 @@
...
+import scala.beans.BeanProperty
+
+case class PropertyDefinition(
+  @(AosSchema@field)(example = "propertyDefinitionA")
+  @BeanProperty name: String,
+
+  @(AosSchema@field)(example = "1")
+  @BeanProperty version: Int = 1,
+
+  @(AosSchema@field)(implementation = classOf[String], example = "dataset description")
```
miroslavpojer, line 38:
```suggestion
  @(AosSchema@field)(implementation = classOf[String], example = "property definition description")
```

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/Application.scala
```diff
@@ -25,13 +25,12 @@ import org.springframework.scheduling.annotation.EnableAsync
 import org.springframework.scheduling.concurrent.ThreadPoolTaskExecutor
 import org.springframework.security.core.context.SecurityContextHolder
 import org.springframework.scheduling.annotation.EnableScheduling
-import springfox.documentation.swagger2.annotations.EnableSwagger2
+
```
miroslavpojer, line 28: Additional empty row.

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/controllers/v3/VersionedModelControllerV3.scala
```diff
@@ -15,6 +15,8 @@
 
 package za.co.absa.enceladus.rest_api.controllers.v3
 
+import io.swagger.v3.oas.annotations.security.SecurityRequirement
+import io.swagger.v3.oas.annotations.{Operation, Parameter}
```
miroslavpojer, line 19:
```suggestion
import io.swagger.v3.oas.annotations.{Parameter}
```

miroslavpojer (commented):
- [x] Pulled
- [x] test ran
- [ ] code-review
- [ ] manual test

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/controllers/v3/VersionedModelControllerV3.scala
dk1844: Thanks, fixed.

rest-api/src/main/scala/za/co/absa/enceladus/rest_api/Application.scala
dk1844: Thanks, fixed.

data-model/src/main/scala/za/co/absa/enceladus/model/properties/PropertyDefinition.scala
dk1844: Yes, of course, thanks. Fixed.

data-model/src/main/scala/za/co/absa/enceladus/model/MappingTable.scala
dk1844: Yes, of course, thanks. Fixed.

miroslavpojer:
When using swagger switch from v2+3(dev) to v3, there are missing APIs for:
- user-info-controller
- spark-controller
- monitoring-controller

@benedeki what is you opinion?

miroslavpojer:
Missing documented return status 403 for all endpoint which require admin rights.
See endpoints with defined:
`  @PreAuthorize("@authConstants.hasAdminRole(authentication)")
`

miroslavpojer:
Is possible to hide the password here:

**
image
**

dk1844:
> Missing method names (next to end point paths).
> 
> image
> 
> See "/api/login Login" as expected example.

Hey, the closest thing I could find (without resorting to manual annotations for every method) is:
`springdoc.swagger-ui.displayOperationId=true`

Then, it looks like this:
swagger-op-id

(basically just `methodName` if globally unique, otherwise `methodName_#`)

Better than nothing or just too ugly?

dk1844:
> Is possible to hide the password here:
> 
> ** image **

The easiest option here is to define the password param's schema as
`new Schema().`type`("string").format("password")`
The result (masked entry field, but visible parameter):
passwordVisibleInParamSwagger

Using `x-www-form-urlencoded` (www form sending data in POST payload)
```
        .requestBody(new RequestBody()
          .required(true)
          .content(new Content()
            .addMediaType(
              "application/x-www-form-urlencoded",
              new MediaType().schema(new Schema()
                  .addProperty("username", new Schema().`type`("string"))
                .addRequiredItem("username")
                .addProperty("password", new Schema().`type`("string").format("password"))
                .addRequiredItem("password")
              )
            )
          )
        )
```
does not behave any better from the password-visibility-in-swagger-perspective:
www-form-swagger-passwordVisibleInSwagger

For simplicity, choosing the former, then.

miroslavpojer:
> > Missing method names (next to end point paths).
> > image
> > See "/api/login Login" as expected example.
> 
> Hey, the closest thing I could find (without resorting to manual annotations for every method) is: `springdoc.swagger-ui.displayOperationId=true`
> 
> Then, it looks like this: swagger-op-id
> 
> (basically just `methodName` if globally unique, otherwise `methodName_#`)
> 
> Better than nothing or just too ugly?

This solution is great.

miroslavpojer:
> > Is possible to hide the password here:
> > ** image **
> 
> The easiest option here is to define the password param's schema as `new Schema().`type`("string").format("password")` The result (masked entry field, but visible parameter): passwordVisibleInParamSwagger
> 
> Using `x-www-form-urlencoded` (www form sending data in POST payload)
> 
> ```
>         .requestBody(new RequestBody()
>           .required(true)
>           .content(new Content()
>             .addMediaType(
>               "application/x-www-form-urlencoded",
>               new MediaType().schema(new Schema()
>                   .addProperty("username", new Schema().`type`("string"))
>                 .addRequiredItem("username")
>                 .addProperty("password", new Schema().`type`("string").format("password"))
>                 .addRequiredItem("password")
>               )
>             )
>           )
>         )
> ```
> 
> does not behave any better from the password-visibility-in-swagger-perspective: www-form-swagger-passwordVisibleInSwagger
> 
> For simplicity, choosing the former, then.

Solution improved situation. Well done.

miroslavpojer:
> 

Missing information add.

miroslavpojer:
> Missing documented return status 403 for all endpoint which require admin rights. See endpoints with defined: ` @PreAuthorize("@authConstants.hasAdminRole(authentication)")`

Solved.

miroslavpojer (approved):

PR #2198, opened 2023-11-07
2195 Add s3a wrapper around run scripts

Zejnilovic:
A wrapper for any of the run scripts. Just prepend the run command with `./s3a_wrapper.sh` and execute as you usually would.

If you want to add jceks file through `--conf spark.hadoop.fs.s3a.bucket.SOME_NAME=jceks:/path/to/file.jceks \`

### DIFF
--- a/scripts/bash/enceladus_env.template.sh
+++ b/scripts/bash/enceladus_env.template.sh
@@ -120,3 +120,10 @@ ADDITIONAL_JVM_EXECUTOR_CONF_CLUSTER="$KRB5_CONF_CLUSTER $TRUST_STORE_CLUSTER $T
 # Switch that tells the script if it should exit if it encounters unrecognized.
 # On true it prints an Error and exits with 127, on false it only prints a warning
 EXIT_ON_UNRECOGNIZED_OPTIONS="true"
+
+# Variables for the s3a wrapper implementation
+MENAS_API="http://localhost:8080/menas/api"
+ECS_API_BASE="https://localhost"
+ECS_API_KK="$ecs_api_base/kk"
+ECS_API_MAP="$ecs_api_base/map"
+ECS_API_KEY="MY_SECRET_KEY"
--- a/scripts/bash/enceladus_env.template.sh
+++ b/scripts/bash/enceladus_env.template.sh
@@ -124,6 +124,6 @@ EXIT_ON_UNRECOGNIZED_OPTIONS="true"
 # Variables for the s3a wrapper implementation
 MENAS_API="http://localhost:8080/menas/api"
 ECS_API_BASE="https://localhost"
-ECS_API_KK="$ecs_api_base/kk"
-ECS_API_MAP="$ecs_api_base/map"
+ECS_API_KK="$ECS_API_BASE/kk"
+ECS_API_BUCKET="$ECS_API_BASE/bucket"
 ECS_API_KEY="MY_SECRET_KEY"
--- a/scripts/bash/s3a_wrapper.sh
+++ b/scripts/bash/s3a_wrapper.sh
@@ -135,7 +135,7 @@ fi
 exit_code=$?
 
 # Run versions cleanup for publish on s3a
-if [[ $$hdfsPublishPath == s3a://* ]]; then
+if [[ $hdfsPublishPath == s3a://* ]]; then
   echo "We have publish versions to clean:"
   curl -X GET \
     --header "x-api-key: $ECS_API_KEY" \
@@ -152,7 +152,7 @@ else
   echo "No publish versions to clean."
 fi
 
-if [[ $$STD_HDFS_PATH == s3a://* ]]; then
+if [[ $STD_HDFS_PATH == s3a://* ]]; then
   STD_HDFS_PATH_FILLED="${STD_HDFS_PATH//\{0\}/$dataset_name}"
   STD_HDFS_PATH_FILLED="${STD_HDFS_PATH_FILLED//\{1\}/$dataset_version}"
   STD_HDFS_PATH_FILLED="${STD_HDFS_PATH_FILLED//\{2\}/$report_date}"
--- a/scripts/bash/s3a_wrapper.sh
+++ b/scripts/bash/s3a_wrapper.sh
@@ -36,7 +36,7 @@ report_date=""
 report_version=""
 
 # The first argument is the name of the original script
-original_script="$(dirname "$0")/${1#./}"
+original_script="$(dirname "$0")/$(basename "$1")"
 
 # Shift the first argument so we can process the rest
 shift
@@ -53,7 +53,7 @@ other_args=()
 # Loop through arguments
 while [[ $# -gt 0 ]]; do
   case "$1" in
-    --keytab)
+    --menas-auth-keytab)
       keytab="$2"
       shift # past argument
       shift # past value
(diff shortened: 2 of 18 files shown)


scripts/bash/s3a_wrapper.sh
```diff
@@ -0,0 +1,174 @@
...
+if klist -s; then
+    echo "Kerberos ticket found."
+else
+    echo "No Kerberos ticket found or ticket is expired. Please run kinit."
+fi
+
+# Get Dataset info
+response=$(curl --negotiate -s -u : "$MENAS_API/dataset/$dataset_name/$dataset_version")
+if [ $? -ne 0 ]; then
+    echo "Curl command failed."
```
dk1844:
```suggestion
    echo "Could not load dataset info - $dataset_name v $dataset_version from Menas at $MENAS_API"
```

scripts/bash/s3a_wrapper.sh
```diff
@@ -0,0 +1,174 @@
...
+
+# Save the exit code
+exit_code=$?
+
+# Run versions cleanup for publish on s3a
+if [[ $hdfsPublishPath == s3a://* ]]; then
+  echo "We have publish versions to clean:"
+  curl -X GET \
+    --header "x-api-key: $ECS_API_KEY" \
+    -d "{\"ecs_path\":\"${hdfsPublishPath#s3a://}\"}" \
```
dk1844: what is `${hdfsPublishPath#s3a://}`?
It seems a bit similar to  [${parameter/#pattern/string}](https://unix.stackexchange.com/questions/126087/what-does-the-hash-in-parameter-pattern-string-do), but not quite.

scripts/bash/enceladus_env.template.sh
```diff
@@ -120,3 +120,10 @@ ADDITIONAL_JVM_EXECUTOR_CONF_CLUSTER="$KRB5_CONF_CLUSTER $TRUST_STORE_CLUSTER $T
 # Switch that tells the script if it should exit if it encounters unrecognized.
 # On true it prints an Error and exits with 127, on false it only prints a warning
 EXIT_ON_UNRECOGNIZED_OPTIONS="true"
+
+# Variables for the s3a wrapper implementation
+MENAS_API="http://localhost:8080/menas/api"
+ECS_API_BASE="https://localhost"
+ECS_API_KK="$ecs_api_base/kk"
+ECS_API_MAP="$ecs_api_base/map"
```
dk1844:
```suggestion
ECS_API_KK="$ecs_api_base/kk"
```
Why having both KK and MAP endpoint when the MAP one does not seem to be used?

dk1844 (commented):
Perhaps requirements to run the script could be mentioned somewhere: curl, jq,

scripts/bash/s3a_wrapper.sh
Zejnilovic:
```bash
$> hdfsPublishPath="s3a://some_bucket/my/path/there"
$> echo ${hdfsPublishPath#s3a://}
some_bucket/my/path/there
```

It strips the s3a prefix from the path saved in HDFS so it can be used by the mapping API

scripts/bash/enceladus_env.template.sh
Zejnilovic: Waiting for further instructions on doing an automatic jceks file resolution. 

TODO in code - https://github.com/AbsaOSS/enceladus/pull/2198/files#diff-82709faf8052f83cfd18429a9f6451da34687d8261923b52d065464293c11c14R24

The map endpoint was suggested, but it does not work both ways. I will remove depending on the on next steps.

scripts/bash/s3a_wrapper.sh
```diff
@@ -0,0 +1,174 @@
...
+
+# We have 2 states.
+#   Write Standardisation
+#   Write Conformance
+# The script will have to figure out which is being run and if it is being written to S3 or HDFS.
+#   In case of standardisation parsing the java properties supplied to figure out the standardisation path.
+#   In case of conformance calling Menas for info over REST API using supplied metadata
+# Then cleaning up the paths to conform to the path schema in query provided.
+
+# TODO - Auto load jceks files
```
miroslavpojer: When this is planned to be done? Exist Issue already?

miroslavpojer (commented):
- [x] pulled
- [x] code review
- [x] tests

miroslavpojer (commented):
When started script now action done. No hint provided.
There should be some hint how to used (place) script.

scripts/bash/s3a_wrapper.sh
```diff
@@ -0,0 +1,174 @@
...
+
+# Run klist to check for a current Kerberos ticket
+if klist -s; then
+    echo "Kerberos ticket found."
+else
+    echo "No Kerberos ticket found or ticket is expired. Please run kinit."
+fi
+
+# Get Dataset info
+response=$(curl --negotiate -s -u : "$MENAS_API/dataset/$dataset_name/$dataset_version")
```
miroslavpojer: If no MENAP_API available/running it simply ends without message. No error info. Debug output:`+response=` -- this is last line of script output (`bash -x`).

miroslavpojer: Problem with call of MENAS_API here.

- Postman: `http://localhost:8080/rest_api_war/api/dataset/literalConformanceRule/1` works
- same call inside of script does not work

```
++ curl --negotiate -s -u : http://localhost:8080/rest_api_war/api/dataset/literalConformanceRule/1
+ response=
+ '[' 0 -ne 0 ']'
+ echo
```

miroslavpojer (commented):
Documented outputs from review testing.

scripts/bash/s3a_wrapper.sh
Zejnilovic: Have you used `kinit` before running the curl with `--negotiate`?

Zejnilovic: So this is just a wrapper script for a transition period, not an application. It is not something I want to support, promote or figure out every edge case for.

scripts/bash/s3a_wrapper.sh
Zejnilovic: Now. 😄

scripts/bash/s3a_wrapper.sh
Zejnilovic: Sorry, you run it against localhost, so you don't have an SPNEGO setup, which is why it won't work.

scripts/bash/s3a_wrapper.sh
```diff
@@ -0,0 +1,152 @@
...
+#   - curl
+#   - klist
+
+set -e
+
+# Run klist to check for a current Kerberos ticket
+if klist -s; then
+    echo "Kerberos ticket found."
+else
+    echo "No Kerberos ticket found or ticket is expired. Please run kinit."
```
dk1844:
```suggestion
    error_exit "No Kerberos ticket found or ticket is expired. Please run kinit."
```

Shouldn't the script exit in this case? Does it make sense to go on?

dk1844 (approved):

scripts/bash/s3a_wrapper.sh
Zejnilovic: 100%. Thank you.

dk1844 (dismissed):

scripts/bash/s3a_wrapper.sh
miroslavpojer: Finally solved. Problem was in HTTPS_PROXY env value on env.

miroslavpojer (dismissed):
- [x] pull
- [x] code review
- [x] manual test

benedeki (approved):

PR #2215, opened 2024-03-22
Dell ECS cleanup now on partition

### LINKED ISSUE #2214: Dell ECS tmp cleanup should be done on the date partition
## Background
Dell ECS tmp cleanup should be done on the date partition not the whole root folder of the dataset

## Feature
Add report date to Dell ECS cleanup

Zejnilovic:
Closes #2214

### DIFF
--- a/scripts/bash/s3a_wrapper.sh
+++ b/scripts/bash/s3a_wrapper.sh
@@ -55,6 +55,7 @@ shift
 hdfsPublishPath=""
 hdfsPath=""
 jceks_flag=""
+std_path="$STD_HDFS_PATH"
 
 # Function to print error message and exit
 function error_exit() {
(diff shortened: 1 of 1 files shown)


scripts/bash/s3a_wrapper.sh
```diff
@@ -105,6 +106,18 @@ function cleanup_versions() {
   return 0
 }
 
+# Function to get evaluated tmp path for standardization
+# Replace placeholders if present
+function replace_std_path_placeholders() {
```
benedeki, line 111: Is this correct?
The function is more like a procedure and used as a function.
Also is the use of `$path` correct, not `$std_path`?

Zejnilovic: It is fine. Bash has no concept of procedures. And this way I can nicely chain it. 

Path fixed

benedeki (approved):
Looks legit now

Zejnilovic:
Tested by Paras 08/04 10:59

